PR #2, opened 2020-03-13
Synch endpoint

zloery:
Pull the work done on the synchronous CreateWorkspace endpoint into development branch.

This only contains commits from the week of 2/20 - 2/28, most of which were pair-programmed by zloery and MatthewBemis.

This is a somewhat messy commit (some files commented out, etc.) and some of this work is overwritten by the asynchronous endpoint changes included later. This version will never be released/used as-is, but breaking it out is helpful for organizing the past week's work into manageable chunks.


src/main/java/bio/terra/workspace/app/controller/TemplateApiController.java
```diff
@@ -0,0 +1,21 @@
+// package bio.terra.workspace.app.controller;
```
MatthewBemis: still needed?

src/test/java/bio/terra/workspace/service/ping/PingRestTest.java
```diff
@@ -0,0 +1,63 @@
+// package bio.terra.TEMPLATE.service.ping;
```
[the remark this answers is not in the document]
zloery: Deleted, nice catch

src/main/java/bio/terra/workspace/app/controller/TemplateApiController.java
zloery: Looks like it was only PingService, deleted it

PR #4, opened 2020-03-16
Add Spotless formatting plugin with google-java-format style, and for…

zloery:
…mat repo


src/main/java/bio/terra/workspace/service/migrate/MigrateService.java
```diff
@@ -11,72 +14,72 @@
...
+                changesetFile, lock.getId(), lock.getLockedBy(), lock.getLockGranted()));
 
-                 // We can get into this state where one of the APIs is running migrations and gets shut down so that
-                 // another API container can run. It will result in a lock that doesn't get released. This is similar
-                 // to the problems we will have from deploying multiple containers at once that try to run migrations.
-                logger.warn("Forcing lock release");
-                liquibase.forceReleaseLocks();
-            }
+        // We can get into this state where one of the APIs is running migrations and gets shut down
+        // so that
```
MatthewBemis: some of these comment format changes are odd, just fyi

zloery: Good catch - looks like the line wrapping adds newlines but doesn't like removing them. I'll try to manually fix these rq

PR #7, opened 2020-03-17
Move swagger assets to resources folder instead of pulling from webjar

zloery:
Warning - when testing/debugging this PR, the hosted .yaml file gets cached in chrome. You'll need to clear the cache between changes, otherwise the files that do/don't show up will be inconsistent.

This moves the Swagger UI dist files directly into the repo as resources, rather than relying on the swagger webjar to load resources. This is helpful because Istio does URL replacement: 

- Services are namespaced behind URL prefixes based on service name (e.g. terra-workspace-manager) and environment (e.g. zloery). For example, the CreateWorksapce endpoint of the workspace manager in the `zloery` environment is `[some IP]/zloery-terra-workspace-manager/api/v1/create`.
- Istio routes requests and removes the environment/service specific prefix - e.g. from the service's perspective, it's receiving requests at `[some IP]/api/v1/create`. Services do not know their own prefix.
- This makes serving a webpage that depends on css/javascript files hard: the webpage will request `[some IP]/swagger.js` when it should be requesting `[some IP]/zloery-terra-workspace-manager/swagger.js`

However, because the `server` field in the API specification also doesn't have the URL prefix, Swagger UI doesn't know where to send requests. This PR makes the UI render to show the API, but can't actually run requests.


zloery:
Re-evaluating this - I'll leave the branch up, but deleting this PR for now.

PR #8, opened 2020-03-17
(AS-197) Update Create endpoint to be asynchronous 

zloery:
This PR completes the asynchronous createWorkspace endpoint for the workspace manager.


src/main/java/bio/terra/workspace/service/create/CreateDAO.java
```diff
@@ -34,6 +34,14 @@ public String create(String workspaceId, JsonNullable<UUID> spendProfile) {
...
-    return workspaceId;
+    return workspaceId.toString();
+  }
+
+  public boolean deleteWorkspace(UUID workspaceId) {
+    Map<String, Object> paramMap = new HashMap<String, Object>();
+    paramMap.put("id", workspaceId.toString());
+    int rowsAffected =
+        jdbcTemplate.update("DELETE FROM workspace WHERE workspace_id = :id", paramMap);
+    return rowsAffected > 0;
```
davidangb: thinking out loud: we'll have to deal with foreign key constraints and other dependencies once the service supports them. Nothing to do right now.

src/main/java/bio/terra/workspace/service/create/flight/CreateWorkspaceAuthzStep.java
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    FlightMap inputMap = flightContext.getInputParameters();
+    UUID workspaceID = inputMap.get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
+    samService.deleteWorkspace(userReq.getRequiredToken(), workspaceID);
+    return StepResult.getStepResultSuccess();
+  }
+}
```
davidangb: Probably more of a question about stairway: does stairway provide all the logging we want/need to debug problems? IF not, add some logging - and potentially error-trapping in order to generate log messages - to these steps.

src/main/java/bio/terra/workspace/service/create/CreateDAO.java
```diff
@@ -34,6 +34,14 @@ public String create(String workspaceId, JsonNullable<UUID> spendProfile) {
     }
 
     jdbcTemplate.update(sql, paramMap);
-    return workspaceId;
+    return workspaceId.toString();
+  }
+
+  public boolean deleteWorkspace(UUID workspaceId) {
```
davidangb: seems strange to have a "delete" method in a "CreateDAO"

MatthewBemis: +1

I am wondering if we should move all of the DB actions to a centralized DAO that can be shared among the services. Given the current structure, I am writing a GetDAO at the moment which feels a bit strange especially since I could see it being used for other service classes within WS manager. I can imagine there being a lot of duplicated code unless we centralize the common DB code.

src/main/java/bio/terra/workspace/service/create/flight/CreateWorkspaceAuthzStep.java
zloery: Each stairway step (both forward and backwards) logs information when it starts or hits any exceptions - I think that's a reasonable level of logging for now, but I'm fine adding more logging if you think it'd be helpful

src/main/java/bio/terra/workspace/service/create/CreateDAO.java
zloery: Actually, that seems like a cleaner solution. It probably makes more sense to have one shared DAO (WorkspaceDAO? though it might also touch other DB tables like resources, data references, etc). I can add that here, but we should coordinate the change for `Get` too

zloery: Per discussion, we'll move to a shared DAO object for the workspace DB in a separate change.

PR #9, opened 2020-03-19
AS-227: get workspace

MatthewBemis:
~No tests yet. Will add shortly.~


src/main/resources/api/service_openapi.yaml
```diff
@@ -46,6 +46,27 @@ paths:
         500:
           description: Creation request error
           $ref: '#/components/responses/ErrorResponse'
+  '/api/v1/get':
+    post:
```
MatthewBemis: This is kinda funky. The endpoint accepts a body so it can't get a `GET`.... would be open to re-working this but given the current requirements it's what I've chosen.

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -10,11 +10,11 @@
 import org.springframework.stereotype.Component;
 
 @Component
-public class CreateDao {
+public class WorkspaceDao {
```
MatthewBemis: refactored to use centralized WorkspaceDAO for DB

src/main/java/bio/terra/workspace/service/get/GetService.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Autowired
+  public GetService(
+      ApplicationConfiguration configuration, WorkspaceDao workspaceDao, SamService samService) {
+    this.configuration = configuration;
+    this.workspaceDao = workspaceDao;
+    this.samService = samService;
+  }
+
+  public CreatedWorkspace getWorkspace(
```
MatthewBemis: Need to eventually return a more descriptive representation of the workspace but for now there's not much to a workspace. I could also wire through the spend profile and policies, I suppose...

src/main/resources/api/service_openapi.yaml
```diff
@@ -46,6 +46,27 @@ paths:
...
+        description: |
+          Workspace ID and user auth token.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/GetWorkspaceRequestBody'
+      responses:
+        200:
+          description: OK
+          $ref: '#/components/responses/CreateResponse'
```
MatthewBemis: We decided that this one wouldn't be async

src/main/java/bio/terra/workspace/service/get/GetService.java
```diff
@@ -0,0 +1,45 @@
...
+    this.workspaceDao = workspaceDao;
+    this.samService = samService;
+  }
+
+  public WorkspaceDescription getWorkspace(
+      GetWorkspaceRequestBody body, AuthenticatedUserRequest userReq) {
+
+    try {
+      samService.isAuthorized(
+          body.getAuthToken(),
```
zloery: If we decide to put tokens in headers, this should use the token from userReq instead of from the request body

src/main/resources/api/service_openapi.yaml
zloery: Can this be something like /api/v1/get/{id}? I think the auth tokens usually get passed via header, Create was just kind of a special case because it needs both user auth token + folder manager's auth token.

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -44,4 +45,25 @@ public boolean deleteWorkspace(UUID workspaceId) {
...
     return rowsAffected > 0;
   }
+
+  public WorkspaceDescription getWorkspace(UUID workspaceId) {
+    String sql = "SELECT * FROM workspace where workspace_id = (:id)";
+
+    Map<String, Object> paramMap = new HashMap<>();
+    paramMap.put("id", workspaceId.toString());
+
+    Map<String, Object> queryOutput = jdbcTemplate.queryForMap(sql, paramMap);
```
zloery: I realize I probably set a bad example on this, but could you create a new ErrorReport-based exception type to throw here, rather than returning JDBC's native EmptyResultDataAccessException?

src/test/java/bio/terra/workspace/service/get/GetServiceTest.java
```diff
@@ -0,0 +1,141 @@
...
+            .andReturn();
+
+    WorkspaceDescription desc =
+        objectMapper.readValue(
+            callResult.getResponse().getContentAsString(), WorkspaceDescription.class);
+
+    assertThat(desc.getId(), equalTo(workspaceId));
+  }
+
+  private CreatedWorkspace runCreateWorkspaceCall(CreateWorkspaceRequestBody request, String jobId)
```
MatthewBemis: I lifted this and the below code from `CreateServiceTest`. I can see this code being broadly useful so may pull it out into a common class a la `TestUtils.java`

src/main/java/bio/terra/workspace/service/get/GetService.java
```diff
@@ -0,0 +1,43 @@
...
+import bio.terra.workspace.generated.model.WorkspaceDescription;
+import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
+import bio.terra.workspace.service.iam.SamService;
+import org.broadinstitute.dsde.workbench.client.sam.ApiException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class GetService {
+  private ApplicationConfiguration configuration;
```
davidangb: why is this one not final?

src/main/resources/api/service_openapi.yaml
```diff
@@ -206,6 +222,26 @@ components:
...
+          type: string
+          nullable: true
+          format: uuid
+        policies:
+          description: Policies provided by the containing folder
+          type: array
+          nullable: true
+          items:
+            type: string
+            format: uuid
```
davidangb: does this current code actually return any policies?

src/main/java/bio/terra/workspace/service/get/GetService.java
MatthewBemis: it's actually unused so i'm just gonna remove it!

src/main/resources/api/service_openapi.yaml
MatthewBemis: no not yet. i added it to the model prematurely- happy to remove it to avoid any potential confusion but in my mind i just wanted to start wiring things through (as much as possible, at elast)

davidangb: I suggest removing it until it exists - so that any client stubs created against this openapi definition will be correct.

MatthewBemis: fair point! removed in latest commit

PR #10, opened 2020-04-03
use existing workspace resource type in sam

MatthewBemis:


zloery:
Does the workspace resource have the same action names (mostly "read" and "delete-job")? If so, lgtm

MatthewBemis:
@zloery yes `read` is an action on `workspace`. `delete-job` isn't, but none of the manager stuff exists either way so we still have to add that in at some point

MatthewBemis:
Merged here https://github.com/DataBiosphere/terra-workspace-manager/pull/14

PR #12, opened 2020-04-07
Fix local-dev and deployment to reflect new terra-config repo

zloery:


zloery:
Note that this change depends on updates to the terra-config repo. I'll hold off submitting this until those are complete.

zloery:
terra-config repo changes are commited, moving forward with this.
Also changing this to pull to master, now that we no longer need to use the development branch. Ignore the long list of commits here - most of those are already in master, but hidden because of commit squashing.

PR #14, opened 2020-04-13
switch to workspace sam resource type

MatthewBemis:


MatthewBemis:
This is the change that was in https://github.com/DataBiosphere/terra-workspace-manager/pull/10 but I'm just doing it as a new branch to avoid rebase and changing the base from development to master

PR #15, opened 2020-04-14
pin stairway to stable version

MatthewBemis:
Recently there were some stairway changes that overwrote the version that we were using in workspace manager. Some of the changes were breaking, so pinning to a self-published stable version for now until we can adopt the changes

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,7 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


zloery:
Just a heads up, there's a PR out for version 0.0.5 with some fixes, so this may need to be bumped again in the near future

PR #16, opened 2020-04-14
AS-213 & AS-270: Support Creating/Getting Data Repo Snapshots

MatthewBemis:


src/main/resources/api/service_openapi.yaml
```diff
@@ -62,6 +62,46 @@ paths:
         500:
           description: Get request error
           $ref: '#/components/responses/ErrorResponse'
+  '/api/v1/workspaces/{id}/references/data':
```
MatthewBemis, line 65: these paths were how i interpreted the design doc. "uncontrolled resource" seemed to be more of an internal concept, and not something to expose at the API level. references and resources seemed like the two key things to expose at this level.

i'm imagining once controlled resources are implemented, the paths might look something like:

```
/api/v1/workspaces/{workspaceId}/references/data/{name}             //uncontrolled
/api/v1/workspaces/{workspaceId}/resources/data/{name}              //controlled
/api/v1/workspaces/{workspaceId}/resources/compute/{name}           //controlled
etc...
```

let me know if you disagree with the thinking.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -59,6 +67,25 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     return new ResponseEntity<WorkspaceDescription>(desc, HttpStatus.OK);
   }
 
+  @Override
+  public ResponseEntity<JobModel> createDataReference(
+      @PathVariable("id") String id, @RequestBody CreateDataReferenceRequestBody body) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    createDataReferenceService.createDataReference(id, body, userReq);
+    // Look up the newly-created job
+    JobModel createJob = jobService.retrieveJob(body.getJobControl().getJobid(), userReq);
```
MatthewBemis, line 76: it seems like this might not need to be an async job anymore based on some discussion in appservices slack, but I had already implemented it as such so leaving it for now.

src/main/resources/db/changesets/20200227_initial_schema.yaml
```diff
@@ -103,6 +103,8 @@ databaseChangeLog:
               name: workspace_id
               type: text
               constraints:
+                references: workspace(workspace_id)
```
MatthewBemis, line 106: just some adjustments to the schema i noticed. i can break these out into a separate changeset if we're at a point where we want to not blow up dev DBs

src/main/resources/api/service_openapi.yaml
```diff
@@ -216,12 +263,90 @@ components:
...
+          format: uuid
+          nullable: true
+        referenceType:
+          description: The type of the reference, if an uncontrolled resource
+          type: string
+          enum: ["DataRepoSnapshot"] #eventually include GCS bucket, etc.
+          nullable: true
+        reference:
+          description: The description of the reference, if an uncontrolled resource
+          $ref: '#/components/schemas/DataRepoSnapshot' #TODO should be some generic reference, and then we can support multiple types
```
MatthewBemis, line 331: really what i want here is a DataReference type that things like DataRepoSnapshot extend. i'm still trying to figure out how to represent that in swagger..

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,120 @@
...
+        paramMap.put("reference_type", referenceType.get().getValue());
+        try {
+          paramMap.put("reference", objectMapper.writeValueAsString(reference));
+        } catch (JsonProcessingException e) {
+          throw new InvalidDataReferenceException("Could not write data reference to database");
+        }
+      } else
+        throw new InvalidDataReferenceException(
+            "DataReference must contain either a resource ID or a reference type and a reference");
+    }
```
zloery, line 69: You could use a ternary operator to simplify this a bit: something like
`paramMap.put("reference_type", resourceId.isPresent() ? null : referenceType.get());` and handling the exception from calling `get` when `referenceType` is missing

src/main/resources/api/service_openapi.yaml
zloery: I think there's definitely some confusing parts from the design doc - we have one table for data references and one table for workspace resources, but those aren't orthogonal. E.g.:

uncontrolled data references - in data_references, not in workspace_resources
controlled data references - in both data_references and workspace_resources (linked via resource_id)
controlled compute resources - not in data_references, in workspace_resources

From a user perspective though, it seems like we'd want one endpoint to return data references and one endpoint for returning resources (though controlled data references could show up in both). E.g. if I want to get info about a data reference, it seems strange to call different endpoints depending on whether it's controlled or uncontrolled.

I think an alternative path structure could be something like
```
/api/v1/workspaces/{workspaceId}/datareferences   // data resources, both controlled + uncontrolled
/api/v1/workspaces/{workspaceId}/resources // controlled resources, including both data + compute
```
with param-based filtering for get/list requests.

Definitely not 100% on this idea though, I'd love to hear thoughts on this

src/main/resources/api/service_openapi.yaml
```diff
@@ -216,12 +263,90 @@ components:
...
+          description: The ID of the resource
+          type: string
+          nullable: true
+        referenceType:
+          description: The type of the data reference (i.e. TDR snapshot, AWS bucket, etc)
+          type: string
+          nullable: true
+        reference:
+          description: The reference to the data (i.e. gs://...) #todo- describe as json instead
+          type: object
```
zloery, line 282: Since we're allowing arbitrary json here to make reference types extendible, can this just be a string?

src/main/resources/api/service_openapi.yaml
```diff
@@ -216,12 +263,90 @@ components:
...
+          type: string
+        workspaceId:
+          description: The ID of the workspace containing the reference
+          type: string
+          format: uuid
+        resourceId:
+          description: The ID of the resource, if it is controlled by Workspace Manager
+          type: string
+          format: uuid
+          nullable: true
```
zloery, line 323: Thoughts on making this something like a `ResourceDescription` object with fields from the workspace_resource table? It would still just be null for uncontrolled data resources, but it would have potentially useful information for controlled data resources.

src/main/resources/api/service_openapi.yaml
zloery: It seems like the only thing the references all share is that they have key-value maps. My first instinct would be to just make the `reference` field a JSON string and do parsing/validation against `referenceType` in the service itself, rather than OpenAPI.

src/main/resources/db/changesets/20200227_initial_schema.yaml
zloery: It should still be ok to edit these, but yeah it's getting close to the point where we'll need to be careful.

src/test/java/bio/terra/workspace/service/job/JobServiceTest.java
```diff
@@ -55,7 +55,6 @@ public void setup() {
     }
   }
 
-  @Test
```
zloery, line 58: Why remove this? It looks like a test to me

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,120 @@
...
+
+    DataReference reference = new DataReference();
+
+    reference.setWorkspaceId(UUID.fromString(queryOutput.get("workspace_id").toString()));
+    reference.setReferenceId(UUID.fromString(queryOutput.get("reference_id").toString()));
+    reference.setName(queryOutput.get("name").toString());
+    reference.setReferenceType(
+        JsonNullable.of(
+            DataReference.ReferenceTypeEnum.fromValue(
+                queryOutput.get("reference_type").toString())));
```
zloery, line 92: style nit: I just recently found out OpenApi generated objects have builder-style methods, so you can do something like
```
DataReference reference = new DataReference()
.workspaceId(...)
.referenceId(...)
...;
```
Mostly a style thing, so your call on doing that or not.

src/test/java/bio/terra/workspace/service/job/JobServiceTest.java
MatthewBemis: hah, didn't intend for that to slip through. I was just manually ignoring it during development because it's one of the slower tests. good catch

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -2,12 +2,14 @@
 
 import bio.terra.workspace.generated.controller.WorkspaceApi;
 import bio.terra.workspace.generated.model.*;
-import bio.terra.workspace.service.create.CreateService;
-import bio.terra.workspace.service.get.GetService;
+import bio.terra.workspace.service.datareference.create.CreateDataReferenceService;
```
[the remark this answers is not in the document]
MatthewBemis: +1 on DataReferenceService, WorkspaceService, etc. I think that's a sensible place to draw the line. I'll start the trend in this PR and then we can handle consolidating the WorkspaceService in a followup PR.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,120 @@
...
+        "INSERT INTO workspace_data_reference (workspace_id, reference_id, name, resource_id, credential_id, cloning_instructions, reference_type, reference) VALUES "
+            + "(:workspace_id, :reference_id, :name, :resource_id, :credential_id, :cloning_instructions, :reference_type, cast(:reference AS json))";
+
+    Map<String, Object> paramMap = new HashMap<>();
+    paramMap.put("workspace_id", workspaceId.toString());
+    paramMap.put("reference_id", referenceId.toString());
+    paramMap.put("name", name);
+    paramMap.put("cloning_instructions", cloningInstructions);
+
+    if (credentialId.isPresent()) {
```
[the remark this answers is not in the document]
MatthewBemis: TIL, thanks

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -0,0 +1,114 @@
...
+
+  public DataReferenceDescription createDataReference(
+      String workspaceId, CreateDataReferenceRequestBody body, AuthenticatedUserRequest userReq) {
+
+    // validate shape of request as soon as it comes in
+    if ((body.getReferenceType().isPresent() && body.getReference().isPresent())
+        == body.getResourceId().isPresent()) {
+      throw new InvalidDataReferenceException(
+          "Data reference must contain either a resource id or a reference type and a reference description");
+    }
```
davidangb, line 63: clever if clause!

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -0,0 +1,114 @@
...
+
+    try {
+      samService.isAuthorized(
+          userReq.getRequiredToken(),
+          SamUtils.SAM_WORKSPACE_RESOURCE,
+          workspaceId,
+          SamUtils.SAM_WORKSPACE_WRITE_ACTION);
+    } catch (ApiException samEx) {
+      throw new SamApiException(samEx);
+    }
```
davidangb, line 73: I know this only occurs twice in this class, but I hate the boilerplate. I found myself skimming over the boilerplate during code review, which is an opportunity for errors to slip through.

How would you feel about
```
  private void authz(AuthenticatedUserRequest userReq, String workspaceId, String action) {
    try {
      samService.isAuthorized(
              userReq.getRequiredToken(),
              SamUtils.SAM_WORKSPACE_RESOURCE,
              workspaceId,
              action);
    } catch (ApiException samEx) {
      throw new SamApiException(samEx);
    }
  }
```

and then here, you can call
```
authz(userReq, workspaceId, SamUtils.SAM_WORKSPACE_READ_ACTION);
```
which I find easier to read? Then also call `authz` from `getDataReference`.

src/main/java/bio/terra/workspace/service/datareference/flight/CreateDataReferenceStep.java
```diff
@@ -0,0 +1,57 @@
...
+    FlightMap inputMap = flightContext.getInputParameters();
+    UUID referenceId = inputMap.get(DataReferenceFlightMapKeys.REFERENCE_ID, UUID.class);
+    UUID workspaceId = inputMap.get(DataReferenceFlightMapKeys.WORKSPACE_ID, UUID.class);
+    String name = inputMap.get(DataReferenceFlightMapKeys.NAME, String.class);
+    UUID resourceId = inputMap.get(DataReferenceFlightMapKeys.RESOURCE_ID, UUID.class);
+    String credentialId = inputMap.get(DataReferenceFlightMapKeys.CREDENTIAL_ID, String.class);
+    String cloningInstructions =
+        inputMap.get(DataReferenceFlightMapKeys.CLONING_INSTRUCTIONS, String.class);
+    String referenceType = inputMap.get(DataReferenceFlightMapKeys.REFERENCE_TYPE, String.class);
+    String reference = inputMap.get(DataReferenceFlightMapKeys.REFERENCE, String.class);
```
davidangb, line 33: boy I wish `FlightMap` had `getString`, `getUUID`, etc.

src/main/resources/api/service_openapi.yaml
davidangb: If we *expect* json, let's leave it as an `object`.  Does this really allow any string, whether or not it is json?

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,87 @@
...
+            DataReferenceDescription.CloningInstructionsEnum.fromValue(
+                queryOutput.get("cloning_instructions").toString()))
+        .name(queryOutput.get("name").toString())
+        .referenceType(
+            DataReferenceDescription.ReferenceTypeEnum.fromValue(
+                queryOutput.get("reference_type").toString()))
+        .reference(queryOutput.get("reference").toString())
+        // TODO: query for resource once controlled resources are implemented
+        .resourceDescription(null);
+  }
```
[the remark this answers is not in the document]
MatthewBemis: I definitely prefer the RowMapper. I'll pull that in- thanks!

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,87 @@
...
+      JsonNullable<String> referenceType,
+      JsonNullable<String> reference) {
+    String sql =
+        "INSERT INTO workspace_data_reference (workspace_id, reference_id, name, resource_id, credential_id, cloning_instructions, reference_type, reference) VALUES "
+            + "(:workspace_id, :reference_id, :name, :resource_id, :credential_id, :cloning_instructions, :reference_type, cast(:reference AS json))";
+
+    Map<String, Object> paramMap = new HashMap<>();
+    paramMap.put("workspace_id", workspaceId.toString());
+    paramMap.put("reference_id", referenceId.toString());
+    paramMap.put("name", name);
```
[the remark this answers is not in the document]
MatthewBemis: it's required in `CreateDataReferenceRequestBody` and in the database schema. where were you thinking the null would come from?

src/main/resources/api/service_openapi.yaml
MatthewBemis: i don't think any string will be valid- it will only ever be json.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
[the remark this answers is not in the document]
davidangb: I did a pretty cursory reverse-engineering of the code, thought it would be possible. If you're confident, that's good enough for me, thanks!

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -44,9 +45,11 @@ public String createDataReference(
...
     paramMap.put("cloning_instructions", cloningInstructions);
     paramMap.put("credential_id", credentialId.orElse(null));
-    paramMap.put("resource_id", resourceId.isPresent() ? resourceId.get() : null);
-    paramMap.put("reference_type", referenceType.isPresent() ? referenceType.get() : null);
-    paramMap.put("reference", reference.isPresent() ? reference.get() : null);
+    paramMap.put("resource_id", resourceId.orElse(null));
+    paramMap.put("reference_type", referenceType.orElse(null));
+    paramMap.put("reference", reference.orElse(null));
+
+    System.out.println(reference);
```
davidangb, line 52: remove `println`

src/main/java/bio/terra/workspace/service/datareference/flight/CreateDataReferenceStep.java
```diff
@@ -0,0 +1,57 @@
...
+
+  private DataReferenceDao dataReferenceDao;
+
+  public CreateDataReferenceStep(DataReferenceDao dataReferenceDao) {
+    this.dataReferenceDao = dataReferenceDao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) throws RetryException {
+    FlightMap inputMap = flightContext.getInputParameters();
```
zloery, line 24: This FlightMap should have the actual request (a CreateDataReferenceRequestBody) stored under the key JobMapKeys.REQUEST, which gets populated by the JobBuilder. Instead of creating a new custom key for each field of the request, it might be easier to retrieve the request and use its fields directly. You'd still need a few custom fields for info that wasn't in the request though, like workspaceId or referenceId.

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -0,0 +1,107 @@
...
+
+    return dataReferenceDao.getDataReference(referenceId);
+  }
+
+  private void authz(AuthenticatedUserRequest userReq, String workspaceId, String action) {
+    try {
+      samService.isAuthorized(
+          userReq.getRequiredToken(), SamUtils.SAM_WORKSPACE_RESOURCE, workspaceId, action);
+    } catch (ApiException samEx) {
+      throw new SamApiException(samEx);
```
zloery, line 104: ! This was something I just realized recently -  `isAuthorized` returns authorization status as boolean that's being ignored here. 
The ApiException gets thrown if Sam fails to call the API, but if the user is unauthorized it returns `false` without throwing an error.

PR #17, opened 2020-04-17
Add DataReferenceDescription and ResourceDescription models to OpenAPI doc

zloery:
This adds new model objects to the OpenAPI document, but does not add any endpoints that use them. This doesn't add any useful functionality, but allows us to build multiple DataReference-based endpoints at the same time without merge conflicts.

The actual references are currently raw JSON strings with no schema specified, which we'll need to validate in the service itself. I'd like to use the `oneOf` operator (https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/) to validate them against one of several different schemas, but those don't work in the openapi-generator for spring-boot yet (see https://github.com/OpenAPITools/openapi-generator/issues/5381)

### DIFF
--- a/src/main/resources/api/service_openapi.yaml
+++ b/src/main/resources/api/service_openapi.yaml
@@ -234,6 +234,72 @@ components:
           type: string
           nullable: true
           format: uuid
+    DataReferenceDescription:
+      type: object
+      properties:
+        referenceId:
+          description: The ID of the data reference
+          type: string
+          format: uuid
+        name:
+          description: The name of the data reference; used to refer to the reference
+          type: string
+        workspaceId:
+          description: The ID of the workspace containing the reference
+          type: string
+          format: uuid
+        resourceDescription:
+          description: Description of the workspace resource, if this is a controlled reference.
+          nullable: true
+          $ref: '#/components/schemas/ResourceDescription'
+        referenceType:
+          description: The type of the reference, if an uncontrolled resource
+          type: string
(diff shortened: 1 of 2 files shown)


src/main/resources/api/service_openapi.yaml
```diff
@@ -234,6 +234,72 @@ components:
...
+        name:
+          description: The name of the data reference; used to refer to the reference
+          type: string
+        workspaceId:
+          description: The ID of the workspace containing the reference
+          type: string
+          format: uuid
+        resourceDescription:
+          description: Description of the workspace resource, if this is a controlled reference.
+          nullable: true
```
MatthewBemis, line 253: actually after pulling this bit into my changes, it looks like `resourceDescription` needs to be defined as:

```
resourceDescription:
  description: Description of the workspace resource, if this is a controlled reference.
  nullable: true
  allOf:
  - $ref: '#/components/schemas/ResourceDescription'
```

apparently the `$ref` keyword obliterates sibling keywords, so the nullable property was getting ignored. this is the solution/workaround

per https://stackoverflow.com/questions/40920441/how-to-specify-a-property-can-be-null-or-a-reference-with-swagger

zloery: Good catch! That's also good to know about nullable in general, I don't see other places in this API that it would cause issues (except maybe error code descriptions), but it'll probably come up again in the future

PR #19, opened 2020-04-17
AS271 Enumerate DataReferences

zloery:
This PR adds a paginated `enumerateDataReferences` endpoint to the WorkspaceManager. It manages pagination through options for `limit` and `offset`. It lists both controlled and uncontrolled resources, and currently has no way to filter or sort the responses.


src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,142 @@
...
+    List<String> whereClauses = new ArrayList<>();
+    whereClauses.add("(ref.workspace_id = :id)");
+    whereClauses.add(filterControlledWhereClause(filterControlled, "ref"));
+    whereClauses.add(uncontrolledOrVisibleResourcesClause(owner, "resource", "ref"));
+    String filterSql = combineWhereClauses(whereClauses);
+    String sql =
+        "SELECT ref.workspace_id, ref.reference_id, ref.name, ref.resource_id, ref.credential_id, ref.cloning_instructions, ref.reference_type, ref.reference, "
+            + "resource.resource_id, resource.associated_app, resource.is_visible, resource.owner, resource.attributes  "
+            + "FROM workspace_data_reference AS ref "
+            + "LEFT JOIN workspace_resource AS resource ON ref.resource_id = resource.resource_id "
```
ddietterich, line 43: Won't there always be a workspace_resource for every workspace_data_reference?
I don't think this should be a LEFT JOIN.

I think you want this the other way around: scan all workspace_resource and LEFT JOIN any data_reference if there is one.

src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -0,0 +1,36 @@
...
+import bio.terra.workspace.common.exception.ValidationException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import org.apache.commons.lang3.StringUtils;
+
+public final class ControllerValidationUtils {
+
+  private static final List<String> VALID_FILTER_CONTROLLED_OPTIONS =
+      Arrays.asList("controlled", "uncontrolled", "all");
```
ddietterich, line 12: I think these should be in an ENUM. Then you can use the same strings here as in the logic later.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -78,4 +81,19 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
         jobService.retrieveJobResult(id, Object.class, userReq);
     return new ResponseEntity<>(jobResultHolder.getResult(), jobResultHolder.getStatusCode());
   }
+
+  @Override
+  public ResponseEntity<DataReferenceList> enumerateReferences(
+      @PathVariable("id") String id,
+      @Valid @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
+      @Valid @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit,
+      @Valid @RequestParam(value = "filterControlled", required = false, defaultValue = "all")
```
ddietterich, line 90: Do you expect to have other kinds of filters eventually? This one seems narrow.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,142 @@
...
+            + filterSql
+            + " ORDER BY ref.reference_id "
+            + "OFFSET :offset "
+            + "LIMIT :limit";
```
ddietterich, line 47: nit: I'd suggest being consistent about where the space separators are in the string. I put them at the start after a + because it is easier for me to forget to put them at the end.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,142 @@
...
+public class DataReferenceDao {
+
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+
+  @Autowired
+  public DataReferenceDao(WorkspaceManagerJdbcConfiguration jdbcConfiguration) {
+    this.jdbcTemplate = new NamedParameterJdbcTemplate(jdbcConfiguration.getDataSource());
+  }
+
+  public DataReferenceList enumerateDataReferences(
```
ddietterich, line 32: I'm confused about why you are enumerating data references as opposed to enumerating resources.
Data reference is a specialization - one kind of resource. I would have expected an `enumerateResource` that had a filter to only see data references.

There should always be a resource row. If that resource is a reference, then there should also be a data reference row. The general scheme here is fine, but I think it is backwards.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,142 @@
...
+          .referenceType(ReferenceTypeEnum.fromValue(rs.getString("reference_type")))
+          .reference(rs.getString("reference"));
+    }
+  }
+
+  // Returns a SQL condition as a string that selects controlled resources, uncontrolled resources,
+  // or both, depending on the value of filterControlled.
+  public static String filterControlledWhereClause(
+      String filterControlled, String referenceTableAlias) {
+    // filterControlled is one of "controlled", "uncontrolled", or "all" (enum in OpenAPI yaml).
```
ddietterich, line 91: I think this is backwards. There would never be null resource_ids. There are just some resources that have associated data references.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,142 @@
...
+        + resourceTableAlias
+        + ".owner = \'"
+        + owner
+        + "\')";
+  }
+
+  // Combines a list of String SQL conditions with the delimiter `" AND "` to create a single
+  // SQL `WHERE` clause. Ignores null and empty strings.
+  public static String combineWhereClauses(List<String> clauses) {
+    return "WHERE ("
```
ddietterich, line 136: nit: nothing wrong with it, but you don't need the parens.

ddietterich:
Well, there is a misunderstanding, but it is mine... You have it right.

src/main/resources/api/service_openapi.yaml
```diff
@@ -68,6 +68,55 @@ paths:
...
+          default: 10
+      - in: query
+        name: filterControlled
+        description: |
+          Filter for selecting controlled vs. uncontrolled resources.
+        schema:
+          type: string
+          default: all
+          enum:
+          - controlled
```
ddietterich, line 101: This should make an Enum that you can use elsewhere.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
zloery: I was planning to add other filters as separate params in the future. I'm not sure which fields we need to filter on yet, or if there's a cleaner way to filter on multiple settings than just using multiple params.

src/main/resources/api/service_openapi.yaml
zloery: TIL OpenAPI doesn't generate the enum class if you define it in-line with parameters. pulled this out into the schema section and wired it through.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
zloery: Agreed it's unnecessary, but it was a little easier for me to parse this way when I was debugging. If that's weird style for SQL I can take them out

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
zloery: Done, also added a space to the start of the WHERE clause to maintain the pattern

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
davidangb: for forward compatibility, how about a generic "filter" parameter, which accepts a JSON key-value map? Or some other form of generic/expandable payload like a query DSL?

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -78,4 +81,18 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+          FilterControlledEnum filterControlled) {
+    ControllerValidationUtils.ValidatePaginationParams(offset, limit);
+    DataReferenceList enumerateResult =
+        dataReferenceService.enumerateDataReferences(
+            id, offset, limit, filterControlled, getAuthenticatedInfo());
+    return ResponseEntity.ok(enumerateResult);
+  }
```
davidangb, line 97: honest question, I don't know the answer: is there any way to centrally define the offset and limit (and maybe "filter") `@RequestParam`s, and then have multiple methods inherit those definitions?

As we build out more APIs, I want to make sure that we are consistent in our pagination, sorting, and filtering parameters. Defining them centrally and inheriting them would be a good way to do that.

Also - probably want to add sorting parameters to this API.

src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -0,0 +1,21 @@
+package bio.terra.workspace.common.utils;
+
+import bio.terra.workspace.common.exception.ValidationException;
+import java.util.ArrayList;
+import java.util.List;
+
+public final class ControllerValidationUtils {
+
+  public static void ValidatePaginationParams(int offset, int limit) {
```
davidangb, line 9: nit: don't start method with a capital (is this a convention I don't know about?)

src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -0,0 +1,21 @@
...
+    List<String> errors = new ArrayList<>();
+    if (offset < 0) {
+      errors.add("offset must be greater than or equal to 0.");
+    }
+    if (limit < 1) {
+      errors.add("limit must be greater than or equal to 1.");
+    }
```
davidangb, line 16: what's your opinion: should we add a sanity-check upper bound for page size?  Else, users could request limit=999999999, which could be bad for the system.

Conversely, should there be a way for a user to specify they truly want all results?

If we start building any extensive validation, let's consider pulling in a library like apache commons validator, or something.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -0,0 +1,152 @@
...
+  // to all workspace readers.
+  public static String visibleResourcesClause(String owner, String resourceTableAlias) {
+    return "("
+        + resourceTableAlias
+        + ".is_visible = true OR "
+        + resourceTableAlias
+        + ".owner = \'"
+        + owner
+        + "\')";
+  }
```
davidangb, line 143: this should use `:owner` syntax and SQL parameter replacement, else it is vulnerable to injection

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -0,0 +1,54 @@
...
+        SamUtils.SAM_WORKSPACE_READ_ACTION)) {
+      throw new SamUnauthorizedException(
+          "User " + userReq.getEmail() + " is not authorized to read workspace " + workspaceId);
+    }
+    // TODO: do we want to give users a different error when the specified workspace doesn't exist?
+    //  It requires an additional DB lookup (as an empty enumerate doesn't tell you whether
+    //  or not a workspace exists), but it's a nicer user experience for catching small mistakes.
+    if (!workspaceDao.workspaceExists(workspaceId)) {
+      throw new WorkspaceNotFoundException("No workspace exists with ID " + workspaceId);
+    }
```
davidangb, line 49: if the workspace does not exist, then won't the previous Sam call return an error? I may be wrong about that.

We should not allow users to fish for the existence/nonexistence of a workspace.
* if the workspace exists and user is allowed to see it, all good
* if and only if the user has permission to list all workspaces (in a given context, such as if they are a folder owner), but the workspace does not exist, then we can/should return a 404
* if the user does not have permission to list all workspaces, or the user is unauthorized for this workspace, then we return 403 with a generic message of "user is not authorized or workspace does not exist"

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
ddietterich: OpenAPI v3 has support for shared definitions. I have not used it yet, but it is one of the big benefits of moving from v2 to v3.

zloery: I can move the parameters to a shared place in the OpenAPI doc, so we could generate all the parameters in methods of the WorkspaceApi interface with the same settings (names, types, defaults, etc). 
To be honest, I'm not totally sure how that affects our implementation of the interface - I don't really understand how annotations mix with interfaces well enough to say if that changes anything, since we're re-specifying the annotations manually.

src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
zloery: For some reason I sometimes think of static methods starting with a capital. I don't think it's a convention anywhere, fixed.

src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
zloery: I'm a little wary of adding more specific validation logic to shared utils like this, but I'm also not a fan of leaving limit unbounded. 
I'd lean towards leaving it unbounded for now and changing it if we expect users to call this directly, but I'm willing to change that.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
zloery: Good catch!

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
zloery: Checking the documentation, it looks like Sam returns false if a user is unauthorized or the resource doesn't exist so yep, it'll throw an error. Took out the "workspace exists" check and updated the error message.

Since Sam doesn't tell us whether the authorization failed or the resource just doesn't exist, I'm not sure how to check for your second case here without adding more business logic, and I don't think the convenience is worth that.

zloery: Also removed the missing workspace test, since that's now not distinct from the "unauthorized" case.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
zloery: A key-value map seems like a nice, generic approach and also lets us move the filter parameter to a shared definition. The only drawback seems like we'd be validating in the service itself instead of right at the API, but that should be manageable with documentation.

zloery:
> This looks good to me. My PR will probably also be in a mergeable state today so let's coordinate how we want to merge our PRs. This one looks like it might have a tricky rebase.
> 
> In the future I think we should try doing feature branches. I've generally had pretty good experiences with them when working on rapid/coordinated development like this.

For this PR, the easiest way would probably just be: 1. you merge your PR as-is, then 2. I'll pull those changes into my branch. That second merge might be a little messy, but I don't think it'll be too bad.

Definitely agree we should coordinate more carefully if we're both working on the same area. I'm still pretty new to version control with Git and I'm not entirely familiar with feature branches, but if it's worked before I'd try it out!

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
zloery: Thinking about this more: a JSON kv map seems like a generally good solution, but I don't have a great sense of all the filtering use cases we'll want to cover in the future, and guessing about those before they show up kind of feels like premature optimization. Given that filtering is explicitly out of scope for AS271, I'm leaning towards removing it from this PR entirely. 
Does that seem reasonable?

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -86,22 +85,25 @@ public DataReferenceDescription createDataReference(
...
-      throw new SamApiException(samEx);
-    }
+    boolean isAuthorized =
+        samService.isAuthorized(
+            userReq.getRequiredToken(), SamUtils.SAM_WORKSPACE_RESOURCE, workspaceId, action);
+    if (!isAuthorized)
+      throw new SamUnauthorizedException(
+          "User "
+              + userReq.getEmail()
+              + " is not authorized to"
```
davidangb, line 103: nit: need space between "to" and `action`

PR #21, opened 2020-04-23
AS-272: Add delete data reference endpoint

MatthewBemis:


src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -80,6 +80,15 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     return new ResponseEntity<DataReferenceDescription>(ref, HttpStatus.OK);
   }
 
+  @Override
+  public ResponseEntity<Void> deleteDataReference(
+      @PathVariable("id") String workspaceId, @PathVariable("referenceId") String referenceId) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    dataReferenceService.deleteDataReference(workspaceId, referenceId, userReq);
```
zloery, line 87: This is good for uncontrolled references, but the current `deleteDataReference` method doesn't handle controlled resources. It deletes the row in `workspace_data_reference`, but not the corresponding row in `workspace_resource` and doesn't delete cloud resources. 
I'm not sure how we want to structure creating + deleting controlled resources atm, but in the meantime we should make sure this method isn't being called on controlled resources (checking that resourceId is null probably works fine)

PR #22, opened 2020-04-24
AS-291

ichengchang:
Prototype Create Workspace Gatling test.
I would not merge to master until the wider team has a chance to review and comment on the strategy in this document

[https://docs.google.com/document/d/1MgXncMiXOz39vUP88CRcwgKBy0xi-VRjkyZrvH0KPtI/edit#heading=h.9tik61r6ax2](url)

### DIFF
--- a/src/gatling/resources/logback.xml
+++ b/src/gatling/resources/logback.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+
+    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender" immediateFlush="false">
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%-5level] %logger{15} - %msg%n%rEx</pattern>
+        </encoder>
+    </appender>
+
+    <!-- uncomment and set to DEBUG to log all failing HTTP requests -->
+    <!-- uncomment and set to TRACE to log all HTTP requests -->
+    <logger name="io.gatling.http.engine.response" level="TRACE"/>
+
+    <!-- uncomment for logging around wblibs google API calls -->
+    <!-- <logger name="org.broadinstitute.dsde.workbench.google" level="DEBUG" />-->
+
+    <!-- useful information about service account creation -->
+    <!-- <logger name="simulations.TerraUserHelper" level="INFO" />-->
+
+    <root level="WARN">
+        <appender-ref ref="CONSOLE"/>
+    </root>
+
+</configuration>
--- a/src/gatling/simulations/BasicSimulation.scala
+++ b/src/gatling/simulations/BasicSimulation.scala
@@ -1,0 +1,87 @@
+import com.google.auth.oauth2.GoogleCredentials
+import java.io.ByteArrayInputStream
+import java.util.UUID
+
+import com.typesafe.config.{Config, ConfigFactory}
+import io.gatling.http.Predef._
+import io.gatling.core.Predef._
+import scala.collection.JavaConverters._
+import sys.process._
+
+class BasicSimulation extends Simulation {
+  def vaultToken: String = {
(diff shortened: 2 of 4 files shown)


src/gatling/resources/logback.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ansingh7115, line 1: The `gatling` folder should be either inside `src/test` or (if java vs scala gives you trouble) at the root. 

Other services like rawls and leo have an `automation` folder at the root for automated integration tests.

src/gatling/simulations/BasicSimulation.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+  /**
+   * Get the data JSON string from the given vault path
+   * @param path path in vault
+   * @return JSON string at path
+   */
+  def getVaultJson(path: String) : String = {
+    val vaultAddr = "https://clotho.broadinstitute.org:8200"
+    val vaultCmd = s"docker run --cap-add IPC_LOCK --rm -e VAULT_TOKEN=${vaultToken} -e VAULT_ADDR=${vaultAddr} vault:1.1.0 vault read -format json -field data ${path}"
+    vaultCmd.!!.stripLineEnd
```
ansingh7115, line 29: Unsure if you're already planning on removing this - but running a CLI command here is not a good idea. We have ways to handle getting secrets from vault and loading them into configs. We can talk about how to do this face to face if you would like

src/gatling/simulations/BasicSimulation.scala
```diff
@@ -0,0 +1,87 @@
...
+      "https://www.googleapis.com/auth/devstorage.full_control",
+      "https://www.googleapis.com/auth/cloud-platform"
+    )
+    val serviceAccountJson = getVaultJson( serviceAccountVaultPath )
+    val credentials = GoogleCredentials.fromStream(
+      new ByteArrayInputStream(serviceAccountJson.getBytes())).createScoped(scopes.asJava)
+      .createDelegated(onBehalfOfUser)
+    credentials.refreshIfExpired()
+    val newAccessToken = credentials.getAccessToken
+    newAccessToken.getTokenValue
```
ansingh7115, line 51: You could also do `credentials.refreshAccessToken().getTokenValue`  instead of `refreshIfExpired`

src/gatling/simulations/BasicSimulation.scala
```diff
@@ -0,0 +1,87 @@
...
+import java.io.ByteArrayInputStream
+import java.util.UUID
+
+import com.typesafe.config.{Config, ConfigFactory}
+import io.gatling.http.Predef._
+import io.gatling.core.Predef._
+import scala.collection.JavaConverters._
+import sys.process._
+
+class BasicSimulation extends Simulation {
```
ansingh7115, line 11: We talked about keeping as much of the code in java as possible and importing it into the Simulation file, making the Simulation file as light and bare bones as possible. Can we attempt to do that here?

src/gatling/resources/logback.xml
ichengchang: This can be done within ```build.gradle```. I think we should have an ```automation``` folder and place the ```gatling``` folder underneath it so that ```gatling``` tests can be separately deployed from unit tests.

src/gatling/simulations/BasicSimulation.scala
ichengchang: You are right, this will be replaced by whatever your team choose to obtain secrets from vault. This is definitely not in its final form yet but ok for POC to get gatling test running.

src/gatling/simulations/BasicSimulation.scala
ichengchang: Yes, let's use the current POC to discuss what needs to be refactored into Java.

PR #23, opened 2020-04-27
AS228 Delete Workspace

zloery:
Adds an endpoint to delete a workspace.

This uses Stairway, but `delete` seems like a weird operation for the Saga pattern - generally, it's really hard to undo partial deletes. It's also currently missing steps for planned but unimplemented features: deleting controlled resources with the Resource Manager library, notifying registered applications, and deleting policy objects in the Policy Manager.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
+++ b/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
@@ -61,7 +61,7 @@ public class WorkspaceService {
   public void deleteWorkspace(String id, String userToken) {
 
     AuthenticatedUserRequest userReq = new AuthenticatedUserRequest().token(Optional.of(userToken));
-    samService.workspaceAuthz(userReq, id, SamUtils.SAM_WORKSPACE_WRITE_ACTION);
+    samService.workspaceAuthz(userReq, id, SamUtils.SAM_WORKSPACE_DELETE_ACTION);
 
     String description = "Delete workspace " + id;
     JobBuilder deleteJob =
--- a/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
+++ b/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
@@ -1,6 +1,5 @@
 package bio.terra.workspace.service.workspace;
 
-import bio.terra.workspace.common.exception.SamUnauthorizedException;
 import bio.terra.workspace.common.utils.SamUtils;
 import bio.terra.workspace.db.WorkspaceDao;
 import bio.terra.workspace.generated.model.CreateWorkspaceRequestBody;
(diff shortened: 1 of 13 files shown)


src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -52,16 +53,26 @@ public CreatedWorkspace createWorkspace(
...
-
+    samService.workspaceAuthz(userReq, id, SamUtils.SAM_WORKSPACE_READ_ACTION);
     WorkspaceDescription result = workspaceDao.getWorkspace(id);
     return result;
   }
+
+  public void deleteWorkspace(String id, String userToken) {
+
+    AuthenticatedUserRequest userReq = new AuthenticatedUserRequest().token(Optional.of(userToken));
+    samService.workspaceAuthz(userReq, id, SamUtils.SAM_WORKSPACE_WRITE_ACTION);
```
MatthewBemis, line 64: There's a `delete` action in Sam for workspaces- we should probably use that here instead.

src/main/java/bio/terra/workspace/service/workspace/flight/DeleteWorkspaceAuthzStep.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  // Note: This doesn't fully undo the do-step, as the newly-created IAM resource will not have
+  // the same access policies as the old one. However, it prevents strange half-deleted states
+  // from occuring.
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    FlightMap inputMap = flightContext.getInputParameters();
+    UUID workspaceID = inputMap.get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
+    samService.createWorkspaceWithDefaults(userReq.getRequiredToken(), workspaceID);
```
MatthewBemis, line 37: I don't think that this will work. Sam prevents ID reuse for workspaces, so we won't actually be able to re-create the workspace resource in Sam if this fails.

I am trying to think of another approach for dealing with this failure but I haven't landed on anything yet. If the workspace resource is deleted, then it will appear deleted for the user even if it's still in the WSM db... but that doesn't mean that all of the contents of the workspace were deleted- so users could falsely be led to believe that they're not going to be accruing costs associated with the workspace anymore. It's a tricky case to handle. It adds a lot of complexity but maybe we could periodically sweep for orphaned parts of the workspace state?

Does stairway allow for us to retry? We could put our best effort in to retry the step until it's eventually deleted and log it if we can't delete it.

A lot of this was just me thinking out loud...

zloery: Good to know on the ID reuse. It already felt a little weird to try re-creating a deleted resource, but if we can't do that at all it seems better to just do the same thing as the `DeleteWorkspaceStateStep` and tell Stairway there's nothing we can do.
I added a retry on Sam errors in the `do` step which should handle transient errors, but I think ultimately if we hit the undo step the best we can do is log this as a dismal failure.
That sweep might make sense at some point, but for now we can probably log/monitor dismal failures from Stairway and take manual action from there.

PR #27, opened 2020-05-01
Use new Data Repo swagger client instead of hand-rolled REST calls

MatthewBemis:


build.gradle
```diff
@@ -30,6 +30,8 @@ repositories {
 
 dependencies {
 	compile(group: 'bio.terra', name: 'stairway', version: '0.0.4.1-SNAPSHOT')
+	compile(group: 'bio.terra', name: 'datarepo-client', version: '1.0.0-SNAPSHOT')
+	compile(group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: '2.29.1')
```
MatthewBemis, line 34: i pulled this in because without it i was getting `java.lang.IllegalStateException: InjectionManagerFactory not found.` every time the DR client was called. not sure if there was a dependency collision (w/ the Sam client?) or if it was just missing entirely but this fixed it.

src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
```diff
@@ -1,47 +1,32 @@
...
-      HttpEntity<String> entity = new HttpEntity<>(headers);
-      ResponseEntity response =
-          restTemplate.exchange(
-              getSnapshotUrl(instance, snapshotId), HttpMethod.GET, entity, Object.class);
+    RepositoryApi repositoryApi = repositoryApi(instance, userReq);
 
-      return response.getStatusCode().is2xxSuccessful();
-    } catch (RestClientException e) {
+    try {
+      repositoryApi.retrieveSnapshot(snapshotId);
```
davidangb, line 27: this throws an `ApiException` if the snapshot is not found/user does not have permission, correct?  I believe that's true but want to confirm.

build.gradle
davidangb: hrm this is a little troubling, I'd prefer not to clutter our build.gradle if we don't have to - the provenance gets lost. At minimum, add a comment to explain why the dependency is here. At best - contact the data repo folks and ask if the published artifact is missing a dependency?

MatthewBemis: @ddietterich could you please confirm/refute my understanding of this issue?

src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
MatthewBemis: yes it should throw for all non-successful status codes, which would cover this case

build.gradle
ddietterich: Interesting. Let me do a little digging. I'm not sure why swagger-codegen needs jersey anything to generate the **client** code.

ddietterich: Yes, there was a missing dependency in the datarepo-client. I have fixed this and tested it with a jadecli build that doesn't have the jersey-hk2 thing.

Try client 1.0.1-SNAPSHOT and let me know if you can successfully remove that dependency.

MatthewBemis: That fixed it. Thanks Dan!

PR #30, opened 2020-05-05
AS-317 Fix CreateWorkspace error serialization and auth before making SQL calls

zloery:
This PR fixes the issue raised directly in AS-317 (WM not validating tokens before performing a createWorkspace SQL call) and also several issues that this uncovered. The full list of changes is:

-WM now authenticates with Sam to create an IAM object before creating a workspace in its own DB. This prevents unauthorized users from fishing for the existence of workspaces.
-WM now returns a user-friendly DuplicateWorkspaceException when an authorized user attempts to create a new Workspace with an ID that's already used, rather than directly returning an SQL error.
-WM now correctly uses the authentication token provided in the request body for create calls, rather than the token from the header. The token in the header is intended to validate that a request comes from the Folder Manager, not to represent a particular user.
-ErrorReportExceptions raised by steps in Stairway are now serialized (and de-serialized) with their error code, rather than only using the `message` and `causes` fields of the ErrorReportException.
-API documentation updated to reflect that the "permission denied" response to a `create` call has a 401 status, not a 403.

### DIFF
--- a/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
+++ b/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
@@ -1,6 +1,7 @@
 package bio.terra.workspace.db;
 
 import bio.terra.workspace.app.configuration.WorkspaceManagerJdbcConfiguration;
+import bio.terra.workspace.common.exception.DuplicateWorkspaceException;
 import bio.terra.workspace.common.exception.WorkspaceNotFoundException;
 import bio.terra.workspace.generated.model.WorkspaceDescription;
 import java.util.HashMap;
@@ -8,6 +9,7 @@ import java.util.Map;
 import java.util.UUID;
 import org.openapitools.jackson.nullable.JsonNullable;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.dao.DuplicateKeyException;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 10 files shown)


src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -35,7 +37,12 @@ public String createWorkspace(UUID workspaceId, JsonNullable<UUID> spendProfile)
     paramMap.put("spend_profile", spendProfile.orElse(null));
     paramMap.put("spend_profile_settable", !spendProfile.isPresent());
 
-    jdbcTemplate.update(sql, paramMap);
+    try {
+      jdbcTemplate.update(sql, paramMap);
+    } catch (DuplicateKeyException e) {
+      throw new DuplicateWorkspaceException(
+          "Workspace " + workspaceId.toString() + " already exists.", e);
+    }
```
davidangb, line 45: could you add a unit test for this? I know it's a bunch more code. It would be good to see that we return a nicely-worded error message (without raw SQL!) in this case.

zloery: Done, added one at both DAO and service-level.

Also, the alternating messages was caused by Stairway failing to undo steps. You'd get something like this:

Run 1:
1. Create workspace with ID in the WM's DB. (succeeds)
2. Create workspace IAM object in Sam (fails because there's no token)
3. Try to undo the create-in-Sam step by deleting in Sam (also fails because there's still no token). This gives the "no token" error.
4. Because there was a failed undo step, Stairway can't undo any more, so it throws up its hands and calls the flight done. The workspace is never cleaned up from the DB.

Run 2:
1. Create workspace with ID in the WM's DB (fails, since it already exists from Run 1). This gives the "workspace already exists" error.
2. Undo the create-workspace step by deleting the workspace (succeeds). Because the workspace has been cleaned up, the next call will be Run 1 again.

Side note, this meant that unauthenticated users had the ability to delete any workspace by calling "create" on it when it already exists. That's... a little terrifying, and I want to think about our other Flights to make sure this kind of bug (accidentally changing state by "undoing" differently than "doing") doesn't happen elsewhere.

PR #32, opened 2020-05-13
AS-305: Add endpoint to look up data reference by name, add unique index

MatthewBemis:


src/main/resources/db/changesets/20200512_reference_unique_index.yaml
```diff
@@ -0,0 +1,17 @@
+databaseChangeLog:
```
MatthewBemis, line 1: figured since there's a dev instance it's time to start adding new changesets

zloery: Yeah, I think that's the right call

src/test/java/bio/terra/workspace/db/DataReferenceDaoTest.java
```diff
@@ -177,7 +230,7 @@ public void enumerateWorkspaceReferences() throws Exception {
     dataReferenceDao.createDataReference(
         secondReferenceId,
         workspaceId,
-        name,
+        name + "2",
```
zloery, line 233: Why the "2"?

MatthewBemis: Since uniqueness between `(workspaceId, referenceType, name)` is now enforced, this one clashed with the reference created above it

PR #34, opened 2020-05-15
AS-340: Adhere to API versioning standard

MatthewBemis:


src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -146,7 +146,7 @@ public void duplicateWorkspaceRejected() throws Exception {
 
     MvcResult failureResult =
         mvc.perform(
-                post("/api/v1/workspaces")
```
ansingh7115, line 149: Might be good to define this string once somewhere instead of repeating it everywhere - but possibly outside of the direct scope of this ticket

PR #36, opened 2020-05-18
AS-324 Add minnie-kenny config to Workspace Manager

zloery:


zloery:
This may disrupt integration and performance tests if the machines running those don't have git-secrets installed. I'll hold off merging this one in particular until I've checked with UC and Ivan

zloery:
AppSec is working on an alternate solution to this problem - we'll use that instead of minnie-kenny when it's ready soon.

PR #38, opened 2020-05-20
Move k8s definitions to Helm

gmalkov:
This implements the changes described in this doc:
https://docs.google.com/document/d/103d2H6816feTyHDYSQXBVft4pOssdW6dFz-4KSzxs9M/edit?usp=sharing

Basically, this pulls out all of the k8s resources and puts them into a Helm chart that is packaged and versioned in a separate repo.


local-dev/skaffold.yaml.template
```diff
@@ -7,6 +7,16 @@ build:
     context: ../
     jib: {}
 deploy:
-  kustomize:
-    paths:
-    - terra-config/NAMESPACE/terra-workspace-manager
+  helm:
+    releases:
+      - name: workspacemanager-ENV
+        namespace: terra-ENV
```
zloery, line 13: I might be missing something, but aren't our namespaces currently just ENV instead of "terra-ENV"?

gmalkov: Yep, this is a change we are consciously making. Also it allows a smooth transition between the kustomize-managed and helm-managed environments. @choover-broad proposed adding the terra- prefix to easily identify namespaces that correspond to Terra environments vs other k8s deployments and I agree.

PR #41, opened 2020-05-22
AS-350: Reduce the amount of times that we process the reference json string

MatthewBemis:


src/main/resources/api/service_openapi.yaml
```diff
@@ -428,8 +428,8 @@ components:
           type: string
           nullable: true
         reference:
-          description: The reference to the data (i.e. gs://...) #todo- describe as json instead
-          type: object
+          description: The reference to the data (i.e. gs://...)
```
davidangb, line 431: how about something like "serialized object containing details specific to the referenceType. For instance, TDR snapshots must be serialized json."

PR #43, opened 2020-06-03
update gitignore

ungwudik:
- gitignore properties file for integration tests
- update todo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,4 +44,4 @@ pom.xml
 
 ### Integration configs
 *account.json
-
+application-integration-test.properties
--- a/render_config.sh
+++ b/render_config.sh
@@ -3,7 +3,7 @@
 VAULT_TOKEN=${1:-$(cat $HOME/.vault-token)}
 TARGET_ENV=${2:-dev}
 
-//todo: Make config-rendering reusable for testing other services AS-366
+#todo: Make config-rendering reusable for testing other services AS-366
 
 DOCKER_MOUNT_DIRECTORY=terra-workspace-manager
 TEMPLATE_INPUT_PATH=${DOCKER_MOUNT_DIRECTORY}/config-templates


zarsky-broad:
Does it make sense to create e.g. `src/test/resources/rendered`, and we can put this properties file there, along with the other things that are currently rendered to `/tmp/` when running the setup script, and then ignore the folder's contents?

ungwudik:
> Does it make sense to create e.g. `src/test/resources/rendered`, and we can put this properties file there, along with the other things that are currently rendered to `/tmp/` when running the setup script, and then ignore the folder's contents?

ungwudik:
Mistakenly closed this PR, but reopened now!

ungwudik:
> Does it make sense to create e.g. `src/test/resources/rendered`, and we can put this properties file there, along with the other things that are currently rendered to `/tmp/` when running the setup script, and then ignore the folder's contents?

Yes, that's similar to what David An just suggested in Slack. I can add the 'rendered' directory or whatever we'd like to name it.
I'm a little hesitant on moving the service-account json secret file in `/tmp` though. I like the idea that the file in /tmp is cleared by a local system job every now and then.
Also, I know we can gitignore that file when it's under `src` directory, but does that presents an opportunity (although not likely) for the json secret file to accidentally end up in source control due to human error or something else? Especially given that the human user never interacts with json secret file directly.

zarsky-broad:
I think I mainly don't like the idea of a script writing a file to a location that isn't within my user purview. That said, I think you can probably merge this as-is, and we can think about refactoring later.

ungwudik:
> I think I mainly don't like the idea of a script writing a file to a location that isn't within my user purview. That said, I think you can probably merge this as-is, and we can think about refactoring later.

Created AS-397 for the refactor.

PR #44, opened 2020-06-04
Whitelist datarepo instances [AS-303]

zarsky-broad:
Adds an whitelist of data repo instances that can be communicated with and enforces it, with test.

It's desirable to have the whitelist be different per-environment; this will be follow-on work that I'll create a ticket for, if this looks okay as-is.

with https://github.com/broadinstitute/firecloud-develop/pull/2159 and https://github.com/broadinstitute/rawls/pull/1232


src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
```diff
@@ -3,22 +3,39 @@
...
     client.setAccessToken(accessToken);
     return client;
   }
 
   private RepositoryApi repositoryApi(String instance, AuthenticatedUserRequest userReq) {
+    validateInstance(instance);
     return new RepositoryApi(getApiClient(userReq.getRequiredToken()).setBasePath(instance));
   }
 
+  public void validateInstance(String instance) {
```
zloery, line 33: I think this should be private - I don't think we expect anything outside this class to call `validateInstance`.

That means we can't test it directly, but testing calls to `snapshotExists` with a bad instance would work.

src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
```diff
@@ -3,22 +3,39 @@
...
+    if (!dataRepoConfig.getInstances().contains(instance)) {
+      throw new ValidationException("Data repository instance " + instance + " is not allowed.");
```
zloery, line 35: Since instances aren't a secret, could you also print the valid instances?

src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
zarsky-broad: Mentioned to you on Slack, but commenting here to solicit suggestions:
I'm not sure how to test this behavior if this particular method becomes private, because I haven't yet figured out how to mock the dependency on `bio.terra.datarepo.api.RepositoryApi`. Any ideas are welcome!

zloery: After more slack discussion, we have 4 options:
1. leave the method public + separate
2. move validateInstance to snapshotExists, mock out repositoryApi in the test
3. Add the Powermock library and use that to mock a private method
4. Break validateInstance out into a static method that takes an instances list + instance
(plus probably more we haven't thought of)
I'd lean towards 1 or 2. For 3 I don't like introducing another heavyweight mock library unless we need to, and for 4 we don't expect anything to call this outside of DataRepoService, so a static method feels like the wrong choice.
Open to more thoughts!

src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
```diff
@@ -3,22 +3,43 @@
...
+  public void validateInstance(String instance) {
+    if (!dataRepoConfig.getInstances().contains(instance)) {
+      throw new ValidationException(
+          "Data repository instance "
+              + instance
+              + " is not allowed. Valid instances are: "
+              + String.join(", ", dataRepoConfig.getInstances()));
+    }
+  }
+
```
davidangb, line 42: for user-friendliness, this should be lenient to differences in case (since urls are generally not case-sensitive) ... and potentially trim whitespace from beginning/end?

src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
davidangb: Option 1 - leaving it public - works for me ... we might even find reason to call this method from elsewhere, in the future

src/main/resources/application.properties
```diff
@@ -1,15 +1,21 @@
...
+
 db.stairway.uri=jdbc:postgresql://127.0.0.1:5432/${STAIRWAY_DATABASE_NAME}
 db.stairway.username=${STAIRWAY_DATABASE_USER}
 db.stairway.password=${STAIRWAY_DATABASE_USER_PASSWORD}
 db.stairway.migrateUpgrade=false
 db.stairway.forceClean=true
+
 sam.basePath=${SAM_ADDRESS}
+
+datarepo.instances=https://jade.datarepo-dev.broadinstitute.org,https://jade-terra.datarepo-prod.broadinstitute.org,https://data.terra.bio
```
davidangb, line 21: I really think these should be key-value pairs, instead of raw values:
* terra-prod = "https://jade-terra.datarepo-prod.broadinstitute.org"
* terra-dev = "https://jade.datarepo-dev.broadinstitute.org"

When we persist data references to the db, we should persist them by key; e.g. the data reference would contain "terra-prod" as its instance value.

When Terra's prod data repo moves from "https://jade-terra.datarepo-prod.broadinstitute.org" to "https://data.terra.bio", we can simply update the value inside this config, and pre-existing references will use the new hostname. Else, we'd have to run a migration against all pre-existing references - or leave them broken - or ask data repo to maintain the old hostname forever.

As an additional benefit, the user-facing message inside the validation exception "Valid instances are ... " need only output the valid keys, so we keep our inner workings of hostnames abstracted from end users.

I realize that to do this easily in config, we might want to switch to [YAML instead of properties](https://docs.spring.io/spring-boot/docs/1.2.0.M1/reference/html/boot-features-external-config.html#boot-features-external-config-yaml) syntax.

I mentioned this in person (in passing) - feel free to push back and say it should be a follow-on PR.

zarsky-broad: I suspect it's worth it, avoiding a migration is reason enough for me, but I agree with your other points, too. Thanks for the detailed explanation of your thoughts, it's really helpful!

zarsky-broad: For the moment, I encoded the current prod and the future prod under different keys - this is because in my experience from the UI, there's an overlap period when the new DNS name is pointed and some people/configs still have the old url, so we can have both active at once and remove the current one at our leisure, or never. Your thoughts, @davidangb?

src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
zarsky-broad: Going to leave it as-is for now, then.

src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
```diff
@@ -31,12 +31,12 @@ private RepositoryApi repositoryApi(String instance, AuthenticatedUserRequest us
   }
 
   public void validateInstance(String instance) {
-    if (!dataRepoConfig.getInstances().contains(instance)) {
+    if (!dataRepoConfig.getInstances().containsValue(instance.toLowerCase().trim())) {
```
davidangb, line 34: this should be `containsKey` - the data references should be storing the string `terra-dev` or `terra-prod`

src/main/resources/application.properties
zarsky-broad: Ah, just caught the second part of this that I didn't parse the first time, that we store the keys in our db. Let me take a pass at that.

src/test/java/bio/terra/workspace/service/datareference/DataRepoServiceTest.java
```diff
@@ -37,7 +37,8 @@ public void testValidateInvalidDataRepoInstance() throws Exception {
   public void testValidateValidDataRepoInstance() throws Exception {
     try {
       // the valid url is set in test/resources/application.properties
-      dataRepoService.validateInstance("https://fake-valid-data-repo-url.broadinstitute.org");
+      // we trim and toLowerCase the string, so this verifies that too
+      dataRepoService.validateInstance(" https://FaKe-VaLiD-data-repo-url.broadinstitute.org  ");
```
davidangb, line 41: this should be `fake-test` or `FaKe-tEsT`

zarsky-broad:
**NB**: with the change from ingesting the raw URL of the data repo instance to taking a key representing it, there will need to be a matching PR to Rawls. I'll get on that now.

davidangb:
@zarsky-broad fyi I think this will need an accompanying firecloud-develop PR, so that Rawls sends a key instead of a url: https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/rawls/rawls.conf.ctmpl#L25 (again, if you want to do the key-value change as a follow-on PR, that's ok with me)

zarsky-broad:
@davidangb yeah, @MatthewBemis and I were just starting to realize the scope of this, I'm planning to bring it up at standup

terra-workspace-manager-client/build.gradle
```diff
@@ -66,35 +66,35 @@ dependencies {
...
-            'interfaceOnly' : 'true',                  // only generate the model and API interfaces
-            'library'       : 'jersey2',               // do jersey code generation so we use Jackson JSON
-            'dateLibrary'   : 'java8',                 // do java8 code generation
-            'apiPackage'    : 'bio.terra.workspace.api',
-            'modelPackage'  : 'bio.terra.workspace.model',
-            'invokerPackage': 'bio.terra.workspace.client',
-            'sourceFolder'  : openapiTargetFolder
-        ])
-        new DefaultGenerator().opts(config.toClientOptInput()).generate()
+openApiGenerate {
```
MatthewBemis, line 70: does the "Publishing" portion of the README need to be updated with this change?

PR #46, opened 2020-06-11
Run unit tests on every push

zarsky-broad:
This runs unit tests (with `--scan`, so you can see the results) on every push. It caches Gradle dependencies (tied to the Gradle version, so the cache doesn't get too huge), along with the wrapper bin itself, so they don't all have to be re-downloaded every time.

I've also created a `local-postgres-init.sql` file to be used to set up the databases for the unit tests, and I've pointed people to use it for local setup in DEVELOPMENT.md.


.gitignore
```diff
@@ -1,7 +1,6 @@
 HELP.md
 .gradle/
 build/
-local-dev/
```
zloery, line 4: I don't think we want to remove this - local-dev usually holds clones of helm + config repos that shouldn't be tracked in git.
If it's to track the postgres script from this PR, I'd recommend explicitly including the script with `!`

PR #48, opened 2020-06-15
AS-318: Run integration tests against dev on merge to master

MatthewBemis:


zarsky-broad:
(I helped write this, I think it is good.)

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,70 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Run Integration Tests
+
+on:
+  push:
+    branches:
+      - master
```
ungwudik, line 9: I think it's also worth running the tests on non-master branch commits. We'd want to detect a code issue (if there's one) **before** it gets to master, IMO.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,70 @@
...
+jobs:
+  integration-test:
+
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:9.6
+        env:
+          POSTGRES_PASSWORD: postgres
```
ungwudik, line 20: Could there a potential concern (now or in the future) having passwords here in plain text?

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Grant execute permission for render_config
+        run: chmod +x render_config.sh
```
davidangb, line 58: could we remove this step if we committed `render_config.sh` as executable (see `git update-index --chmod`)?

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,70 @@
...
+          VAULT_ADDR: https://clotho.broadinstitute.org:8200
+        run: |
+          VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
+            -e "VAULT_ADDR=${VAULT_ADDR}" \
+            vault:1.1.0 \
+            vault write -field token \
+              auth/approle/login role_id=${{ secrets.VAULT_APPROLE_ROLE_ID }} \
+              secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+          echo ::set-output name=vault-token::$VAULT_TOKEN
+          echo ::add-mask::$VAULT_TOKEN
```
davidangb, line 44: potentially for a follow-on PR: this should probably be extracted into a reusable action so we're not copy-pasting it everywhere

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,70 @@
...
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Grant execute permission for render_config
+        run: chmod +x render_config.sh
+      - name: Render configuration for tests
+        run: ./render_config.sh ${{ steps.vault-token-step.outputs.vault-token }} dev
+      - name: Ensure read permissions for rendered files
+        run: sudo chmod +r /tmp/wsm-firecloud-account.json src/test/resources/application-integration-test.properties
```
davidangb, line 62: is there something that should change in `render_config.sh` to make this step obsolete?

.github/workflows/integration-test.yml
MatthewBemis: I agree, there will be a big effort to refactor this workflow once we have on-demand MC-Terra/FiaB integration working. Then we can run this on every PR and prevent merge if the tests fail.

I think the original motivation for just running against master is to not have multiple PRs running tests against the same environment at the same time. Also, are the tests ever going to conflict with each other if run concurrently from multiple PRs? (I haven't audited the existing test but that's something we'd need to keep in mind for future ones until we have on-demand MC-Terras)

.github/workflows/integration-test.yml
MatthewBemis: I am inclined to say that in this case it's okay, we've sort of set a standard for doing it with throwaway DBs like this one. The nature of this one isn't really "secret"

.github/workflows/integration-test.yml
MatthewBemis: Agreed as follow-on! If I recall from a meeting I had with DevOps, Greg said that extracting this was planned especially since it would set a standard for how we're supposed to mask tokens and such

.github/workflows/integration-test.yml
ungwudik: The existing test does create a workspace in a self-contained manner that shouldn't conflict with say, another CreateWorkspace request, in theory. This has not been test out.

We currently know of the CreateWorkspace quota limitation (about 70 requests/10 seconds) that will be tackled in AS-299, but it's probably unlikely that we'll have enough concurrent PR's to run into that issue for integration tests.
That said, I don't know whether there are other limitations for concurrent runs. But in any case - we should always aim to build the tests out in a way that allows concurrent runs.

But yes, we can start with a workflow for only master and refactor later to include PRs.

.github/workflows/integration-test.yml
MatthewBemis: It's something that only has to happen because of how the GitHub workflow is run, so doing it in the render_config.sh seemed out of place. Eventually we will not do this through an explicit Docker run and this step won't be needed.

.github/workflows/integration-test.yml
MatthewBemis: I looked into this a bit, oddly enough the permissions on the file are _already_ `755`, so it should be executable. 

I'm not sure what exactly is going on here. I'll dig into it a bit more to understand the discrepancy, but it did seem to require this step.

PR #49, opened 2020-06-17
AS-373 /status endpoint for WM

zloery:
This PR adds a /status endpoint for workspace manager. 

This adds `BaseStatusService` which should be extensible to other Spring-based MC-Terra components as a replacement for the health-checking functionality in `workbench-libs`. This allows us to check the status of downstream dependencies asynchronously on a schedule, while only returning cached results for the actual endpoint call. These dependencies can be critical (non-ok status in dependency leads to overall non-ok status) or not.

Currently I'm checking Sam (critical dependency), all allowable Data Repo instances (non-critical dependencies), and Postgres (critical dependency). These were pretty much judgement calls, so I'm happy to add more dependencies or change criticality on any of these.

It's an unauthenticated api endpoint (located at `/status` rather than `/api/...`). You can try it out at https://workspace.zloery.integ.envs.broadinstitute.org/swagger-ui.html


src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -72,4 +78,24 @@ public void workspaceAuthz(AuthenticatedUserRequest userReq, String workspaceId,
...
+
+  public SystemStatusSystems status() {
+    // No access token needed since this is an unauthenticated API.
+    StatusApi statusApi = new StatusApi(getApiClient(null));
+
+    try {
+      // Note the SystemStatus class here is from the Sam client library, not generated by WM's
+      // OpenAPI, so this requires some amount of reshaping objects.
+      SystemStatus samStatus = statusApi.getSystemStatus();
+      Map<String, Map<String, String>> subsystemStatusMap =
```
zloery, line 90: I'd like other people's thoughts on this one - samStatus.getSystems() returns an Object because that's all that's guaranteed by their API, but it feels really fragile to be casting an Object like this. 
An alternative is including the `getSystems().toString()` as a single message, but that means we lose machine-readability of Sam subsystems in this message. Thoughts on this tradeoff?

src/main/java/bio/terra/workspace/service/status/WorkspaceManagerStatusService.java
```diff
@@ -0,0 +1,55 @@
...
+      @Value("${status-check-staleness-threshold.in.milliseconds}") long staleThresholdMillis) {
+    super(staleThresholdMillis);
+    this.jdbcTemplate = new NamedParameterJdbcTemplate(jdbcConfiguration.getDataSource());
+    Supplier<SystemStatusSystems> dbHealthFn =
+        () ->
+            new SystemStatusSystems()
+                .ok(jdbcTemplate.getJdbcTemplate().execute(this::isConnectionValid));
+    registerSubsystem("Postgres", new StatusSubsystem(dbHealthFn, /*isCritical=*/ true));
+
+    // TODO: should we really be checking every instance? Are any of them critical?
```
zloery, line 39: TODO from the ticket that I'd appreciate opinions on.

davidangb: I think that once the config files are properly environment-specific, each Workspace Manager will only have 1-2 data repo instances in its config ... so yes, let's keep checking all of them.

src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
```diff
@@ -0,0 +1,81 @@
+package bio.terra.workspace.common.utils;
+
+import bio.terra.workspace.generated.model.SystemStatus;
+import bio.terra.workspace.generated.model.SystemStatusSystems;
```
davidangb, line 4: bleh, I know it's generated code, the name `SystemStatusSystems` is gross. Not a blocker for this PR!

thinking out loud ... if we want this to be a basis for other services to extend for their own status endpoints, should it use generated code?

src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
```diff
@@ -0,0 +1,81 @@
...
+          "Subsystem status has not been checked since "
+              + lastCheckDate.toString()
+              + ", exceeding deadline of "
+              + staleThresholdMillis
+              + " ms.";
+      return new SystemStatus()
+          .ok(false)
+          .putSystemsItem(
+              "Staleness", new SystemStatusSystems().ok(false).addMessagesItem(timeoutMessage));
+    }
```
davidangb, line 78: the staleness check is a good addition, nice job!

src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
```diff
@@ -0,0 +1,81 @@
...
+    AtomicBoolean systemOk = new AtomicBoolean(true);
+    subsystems.forEach(
+        /*parallelismThreshold=*/ 1,
+        (name, subsystem) -> {
+          SystemStatusSystems subsystemStatus = subsystem.getStatusCheckFn().get();
+          tmpSubsystemStatusMap.put(name, subsystemStatus);
+          if (subsystem.isCritical() && !subsystemStatus.getOk()) {
+            systemOk.set(false);
+          }
+        });
```
davidangb, line 54: this should probably have error-catching around each subsystems' `getStatusCheckFn` call - if our attempt to check status results in an Exception, let's mark that subsystem as unhealthy.

src/main/java/bio/terra/workspace/service/iam/SamService.java
davidangb: hmm. `samStatus.getSystems()` should return a `Map<String, SubsystemStatus>` and the fact that it doesn't is a problem in the Sam client library (which is a result of awkward swagger definitions, and I don't believe is trivial to fix).

I somehow doubt that casting directly to `Map<String, SubsystemStatus>` will work. What if you serialized the output of `samStatus.getSystems()` to JSON, then deserialized from JSON to `Map<String, SubsystemStatus>`, with error-trapping? (is that too heavyweight?)

src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
zloery: Yeah, I'm really not a fan of that name.
We definitely could use this generated code for other services - the `SystemStatus` was one of the original objects in the POC service so other services have the same object. But, that means we're redefining the same thing in a bunch of places, which makes it hard to change later...

src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: I like that solution better than casting directly, especially since failures give us JSON exceptions instead of null pointer exceptions.

src/main/java/bio/terra/workspace/service/status/WorkspaceManagerStatusService.java
zloery: SGTM

PR #51, opened 2020-06-19
AS-368: Add Request-Id to every log message associated with an API call

MatthewBemis:
Log messages will look something like:

```
2020-06-19 09:40:41.535 pn3N7J3M ERROR 17147 --- [nio-8080-exec-1] b.t.w.a.c.GlobalExceptionHandler         :    cause: bio.terra.workspace.common.exception.DuplicateWorkspaceException@2696eb2[causes=<null>,statusCode=400 BAD_REQUEST]
```

This does _not_ add the ID to anything that happens asynchronously within stairway. It will be added to the message that says the flight was launched, but no subsequent messages from within the flight.

### DIFF
--- a/src/main/java/bio/terra/workspace/mdc/MdcLogEnhancerFilter.java
+++ b/src/main/java/bio/terra/workspace/mdc/MdcLogEnhancerFilter.java
@@ -11,6 +11,8 @@ import org.springframework.stereotype.Component;
 @Component
 public class MdcLogEnhancerFilter implements Filter {
 
+  private Hashids hashids = new Hashids("requestIdSalt", 8);
+
   @Override
   public void doFilter(
       ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain)
@@ -34,7 +36,6 @@ public class MdcLogEnhancerFilter implements Filter {
   }
 
   private String generateRequestId() {
-    Hashids hashids = new Hashids("requestIdSalt", 8); // min length will be 8
     long generatedLong = ThreadLocalRandom.current().nextLong(0, Integer.MAX_VALUE);
 
     return hashids.encode(generatedLong);
--- a/src/main/java/bio/terra/workspace/mdc/MdcLogEnhancerFilter.java
+++ b/src/main/java/bio/terra/workspace/mdc/MdcLogEnhancerFilter.java
@@ -17,12 +17,19 @@ public class MdcLogEnhancerFilter implements Filter {
       throws IOException, ServletException {
     HttpServletRequest httpRequest = (HttpServletRequest) servletRequest;
 
-    String requestId =
-        ((requestId = httpRequest.getHeader("Request-Id")) != null)
-            ? requestId
-            : generateRequestId();
+    /* Services calling Workspace Manager can supply an X-Request-ID or X-Correlation-ID
+       HTTP header to trace requests through the system. X-Request-ID will be preferred over
+       X-Correlation-ID. If neither header is present, Workspace Manager will generate one.
+    */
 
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/workspace/mdc/MdcLogEnhancerFilter.java
```diff
@@ -0,0 +1,35 @@
...
+public class MdcLogEnhancerFilter implements Filter {
+
+  @Override
+  public void doFilter(
+      ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain)
+      throws IOException, ServletException {
+    HttpServletRequest httpRequest = (HttpServletRequest) servletRequest;
+
+    String requestId =
+        ((requestId = httpRequest.getHeader("Request-Id")) != null)
```
MatthewBemis, line 21: `getHeader` is case-insensitive, fortunately

src/main/java/bio/terra/workspace/mdc/MdcLogEnhancerFilter.java
```diff
@@ -0,0 +1,35 @@
...
+            ? requestId
+            : generateRequestId();
+
+    MDC.put("requestId", requestId + " ");
```
MatthewBemis, line 25: I added a space after the `requestId` as extra padding because when testing locally it was bumping up against the log-level in the message i.e. it was producing messages that looked like `pn3N7J3MERROR`. That may have just been formatting in my terminal not properly showing a tab or something. If there are objects I can remove.

src/main/java/bio/terra/workspace/mdc/MdcLogEnhancerFilter.java
davidangb: could this be `X-Request-ID` or `X-Correlation-ID`, maybe even look for both before generating one ourselves? https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

src/main/resources/application.properties
```diff
@@ -20,3 +20,5 @@ sam.basePath=${SAM_ADDRESS}
 
 datarepo.instances.terra-dev=https://jade.datarepo-dev.broadinstitute.org
 datarepo.instances.terra-prod=https://jade-terra.datarepo-prod.broadinstitute.org
+
+logging.pattern.level = %X{requestId}%5p
```
davidangb, line 24: I think if you change this to `%X{requestId} %5p` (note the space) then you won't have to append a space to the requestid in `MdcLogEnhancerFilter`. See https://docs.spring.io/autorepo/docs/spring-boot/1.4.0.RELEASE/reference/html/boot-features-logging.html (search for "You can add MDC and other ad-hoc content")

src/main/java/bio/terra/workspace/mdc/MdcLogEnhancerFilter.java
```diff
@@ -0,0 +1,35 @@
...
+    filterChain.doFilter(servletRequest, servletResponse);
+  }
+
+  private String generateRequestId() {
+    Hashids hashids = new Hashids("requestIdSalt", 8); // min length will be 8
```
davidangb, line 30: is it worth creating the hashids instance at the class level? That would avoid calling the constructor for every request. The call to `encode` *looks* threadsafe.

MatthewBemis:
@davidangb I've addressed your comments. +1 for creating a followup story for stairway transactions. I'll create that.

I also added another filter to add `X-Request-ID` to the HTTP response headers for each request. Seems like it could be useful. Let me know if you think it's superfluous!

PR #52, opened 2020-06-24
Fix integration test & simplify config [AS-397]

zarsky-broad:
I think there's a ticket for this, I'll find it and hook it up if we're ok with this approach.


render_config.sh
```diff
@@ -18,23 +13,17 @@ fi
...
+      vault read -format=json ${FIRECLOUD_ACCOUNT_VAULT_PATH} | \
+      jq .data`
+fi
 
 if [[ -z "$SERVICE_ACCOUNT_CREDS" ]]; then
     printf "\033[0;31m Could not fetch service account creds. Check your vault token. \n\033[0m"
     exit 1
 fi
 
 echo "$SERVICE_ACCOUNT_CREDS" > ${SERVICE_ACCOUNT_OUTPUT_FILE_PATH}
```
ungwudik, line 29: This line currently fails, since the `rendered` directory doesn't yet exist.
`./render_config.sh: line 29: ./src/test/resources/rendered/wsm-firecloud-account.json: No such file or directory`

render_config.sh
```diff
@@ -18,23 +13,17 @@ fi
 
 DSDE_TOOLBOX_DOCKER_IMAGE=broadinstitute/dsde-toolbox:consul-0.20.0
 
-SERVICE_ACCOUNT_CREDS=`docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN ${DSDE_TOOLBOX_DOCKER_IMAGE} \
-    vault read -format=json ${FIRECLOUD_ACCOUNT_VAULT_PATH} | \
-    jq .data`
+if type vault > /dev/null; then
```
ungwudik, line 16: I don't have `vault` installed currently, and this line fails.
`./render_config.sh: line 16: type: vault: not found`

src/test/java/bio/terra/workspace/integration/common/configuration/TestConfiguration.java
```diff
@@ -9,16 +10,21 @@
...
+  }
+
+  public void setWsmUrls(HashMap<String, String> wsmUrls) {
+    this.wsmUrls = wsmUrls;
   }
 
-  public void setWsmCreateWorkspaceUrl(String wsmCreateWorkspaceUrl) {
-    this.wsmCreateWorkspaceUrl = wsmCreateWorkspaceUrl;
+  public String getWsmCreateWorkspaceUrl() {
+    return this.wsmUrls.get("dev") + this.wsmEndpoints.get("createWorkspace");
```
ungwudik, line 27: For clarity, can this be updated to accept/use the environment dynamically?

render_config.sh
davidangb: yeah, I say stick with the docker image, to prevent issues with different versions of the `vault` binary, or aliases, or anything like that.

config-templates/application-integration-test.properties.ctmpl
```diff
@@ -1,11 +0,0 @@
...
-{{- with $wsmBaseUrl := (( printf "%s/dev-terra-workspace-manager" $host )) -}}
-{{- with $wsmCreateWorkspaceUrl := (( printf "%s/api/workspaces/v1" $wsmBaseUrl )) -}}
-{{- with $serviceAccountEmail := (( "william.thunderlord@test.firecloud.org" )) -}}
-{{- with $serviceAccountFilePath := env "SERVICE_ACCOUNT_FILE_PATH" -}}
-
-it.wsmCreateWorkspaceUrl={{$wsmCreateWorkspaceUrl}}
-it.serviceAccountEmail={{$serviceAccountEmail}}
-it.serviceAccountFilePath={{$serviceAccountFilePath}}
-
-{{end}}{{end}}{{end}}{{end}}{{end}}
```
davidangb, line 11: we'll have to put all the environment-specific code back in at some point (see https://github.com/DataBiosphere/terra-workspace-manager/pull/31/files#r431286413), I'm fine with whatever approach we take until we get to that point.

src/test/resources/application-integration-test.properties
```diff
@@ -0,0 +1,6 @@
+it.wsmUrls.dev=https://workspace.dsde-dev.broadinstitute.org/
+
+it.wsmEndpoints.createWorkspace=api/workspaces/v1
+
+it.serviceAccountEmail=william.thunderlord@test.firecloud.org
+it.serviceAccountFilePath=src/test/resources/rendered/wsm-firecloud-account.json
```
davidangb, line 6: how about moving to YAML? https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-yaml

config-templates/application-integration-test.properties.ctmpl
zarsky-broad: Yeah; we thought that for the test resources, the environment selection should be able to happen at test run time rather than at render time.

src/test/resources/application-integration-test.properties
zarsky-broad: I discussed this at some point with (I think?) @zloery; for the time being, I don't want to have a mix of languages, but I'd be quite interested to discuss this more.

davidangb: we could adopt yaml everywhere and avoid a mix of languages ...

PR #53, opened 2020-06-29
add version endpoint [AS-376]

zarsky-broad:


src/main/java/bio/terra/workspace/app/configuration/VersionConfiguration.java
```diff
@@ -0,0 +1,31 @@
+package bio.terra.workspace.app.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
+
+@Configuration
+@EnableConfigurationProperties
```
zloery, line 9: I don't think you need the `@EnableConfigurationProperties` annotation - from what I understand that should be used in property/context classes elsewhere to pull in configurations that aren't in beans. There's more detail + examples at https://stackoverflow.com/questions/49880453/what-difference-does-enableconfigurationproperties-make-if-a-bean-is-already-an/49888642
Since the `@Configuration` annotation already makes this a bean, we don't need it (confirmed running locally). Out of scope for this PR, but I also need to go fix that for some of our other config classes.

src/main/java/bio/terra/workspace/app/controller/UnauthenticatedApiController.java
```diff
@@ -24,4 +29,16 @@ public UnauthenticatedApiController(WorkspaceManagerStatusService statusService)
...
+
+  @Override
+  public ResponseEntity<SystemVersion> serviceVersion() {
+    SystemVersion currentVersion =
+        new SystemVersion()
+            .tag(versionConfiguration.getTag())
+            .hash(versionConfiguration.getHash())
+            .github(
+                "https://github.com/DataBiosphere/terra-workspace-manager/commit/"
+                    + versionConfiguration.getHash());
```
davidangb, line 41: this `currentVersion` object will not change for the life of the running java service. So, you could create `currentVersion` once at the class level instead of re-creating a new one each time this method runs.

PR #57, opened 2020-07-02
Add deleteWorkspace integration test and cleanup code

ungwudik:


src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
```diff
@@ -28,24 +29,91 @@
...
-
   @Autowired private TestConfiguration testConfig;
-
   @Autowired private AuthService authService;
+  private final String TAG_NEEDS_CLEANUP = "needs-cleanup";
+  private static final Logger logger = LoggerFactory.getLogger(WorkspaceIntegrationTest.class);
+
+  // Variable for workspace Id's to clean up in tearDown(TestInfo testInfo) method, after
+  // running applicable test method(s)
+  private UUID workspaceIdToCleanup;
```
davidangb, line 44: I'm worried about using a single mutable variable here for use across multiple tests. If we ever switched to using `@TestInstance(Lifecycle.PER_CLASS)`, this could cause consistency problems where multiple tests/teardown operations overwrite each other. I suggest this move to a threadsafe/concurrent collection - probably a `ConcurrentHashMap`, where the key can be a test name (maybe `TestInfo.getTestMethod()`?) and the value can be a list of workspace Ids to clean up. (We should also support a list of workspaces that need cleaning up, not just a single one; tests may need to create/clean up multiple workspaces).

Is this worth it? If this is overkill, at minimum add a comment along the lines of "if we switch TestInstance modes this variable will be a problem".

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
```diff
@@ -28,24 +29,91 @@
...
 
   @BeforeEach
   public void setup() {}
 
+  /*
+   * TODO: Cloud Recourse Library (CRL) is planning to broadly address resource cleanup for integration tests in MC Terra
```
davidangb, line 50: typo: "Resource" ... though "Cloud Recourse" is interesting in its own right ...

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
```diff
@@ -28,24 +29,91 @@
...
+   *  applications.
+   *  The doc below proposes a "Resource Tracking" approach for cleaning up. We should follow the developments in CRL for
+   *  potential utilization here in Workspace Manager.
+   *  Doc: https://docs.google.com/document/d/13mYVJML_fOLsX1gxQxRJgECUNT27dAMKzEJxUEvtQqM/edit#heading=h.x06ofvfgp7wt
+   */
+  @AfterEach
+  public void tearDown(TestInfo testInfo) throws Exception {
+    Set<String> tags = testInfo.getTags();
+    if (tags != null && tags.contains(TAG_NEEDS_CLEANUP)) {
+      cleanUpWorkspace();
```
davidangb, line 60: assuming we move to tracking all workspaces-to-be-cleaned in a `ConcurrentHashMap`, this should calculate test-specific info and pass it into `cleanUpWorkspace()` - either pass the test name or pass the list of ids to clean up.

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
```diff
@@ -28,24 +29,91 @@
...
+    String jsonBody = testUtils.mapToJson(body);
+
+    WorkspaceResponse<?> deleteWorkspaceResponse =
+        workspaceManagerTestClient.delete(userEmail, path, jsonBody);
+
+    Assertions.assertEquals(deleteWorkspaceResponse.getStatusCode(), HttpStatus.NO_CONTENT);
+  }
+
+  @Test
+  @Tag(TAG_NEEDS_CLEANUP)
```
davidangb, line 95: if you wanted to, you could create a custom `@NeedsCleanup` annotation, which could be reusable across many test classes: https://junit.org/junit5/docs/current/user-guide/#writing-tests-meta-annotations

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
ungwudik: I think it's definitely worth it. This variable is workable but not ideal, and I should have noted that.
I've implemented a `ConcurrentHashMap`.
`TestInfo.getDisplayName()` returns a simpler value (i.e. strips additional metadata that may not be needed).

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
ungwudik: Indeed! :)

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
ungwudik: Created AS-431 for this, as it looks like we may need to annotate a few other things.

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
ungwudik: One more thing: I don't think it's necessarily to remove the Id's from the map post-cleanup, but let me know if you disagree.

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
```diff
@@ -28,24 +29,94 @@
...
+    String token = authService.getAuthToken(userEmail);
+    String path = testConfig.getWsmWorkspacesBaseUrl() + "/" + workspaceId;
+    DeleteWorkspaceRequestBody body = new DeleteWorkspaceRequestBody().authToken(token);
+    String jsonBody = testUtils.mapToJson(body);
+
+    WorkspaceResponse<?> deleteWorkspaceResponse =
+        workspaceManagerTestClient.delete(userEmail, path, jsonBody);
+
+    Assertions.assertEquals(deleteWorkspaceResponse.getStatusCode(), HttpStatus.NO_CONTENT);
+  }
```
davidangb, line 95: should probably mark this one as needing cleanup too. If this test creates a workspace, but then fails to delete it, the workspace will be orphaned and need cleanup. Potentially, if the final assertion succeeds (i.e. the test passes), you could remove the entry in the cleanup map so the tearDown doesn't try to delete it a second time.

ungwudik: My initial thought was: The 'cleanup' does the same thing as the deleteWorkspace step in this test (i.e. they both call the same http endpoint with the same user, path, etc). So if the deleteWorkspace step fails, the cleanup is likely to fail (unless it's a non-persistent issue such as temporary unavailability, timeout, etc).

I can remove the entry and proceed to clean up for this test. However, I'm wondering if the things a test will need to know/do are beginning to add up. Currently, a test (code/human) that needs cleanup will need to know to (1) use the `@Tag(TAG_NEEDS_CLEANUP)` annotation, and (2) add the the Id's to the map. Maybe there's a way we can make the preparation for cleanup a little more seamless. I'll think of ideas, but don't have any right now. Also, the development of Janitor service _should_ be able to help with this.

Let me know what you think.

PR #58, opened 2020-07-08
Some initial OAS cleanup [pre- AS-391]

zarsky-broad:
Hopefully non-controversial. Nothing here requires any java code changes, and I don't believe it would affect what's published.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -152,9 +152,5 @@ spotless {
 	java {
 		googleJavaFormat()
 	}
-	format 'yaml', {
-		target '/src/main/resources/api/service_openapi.yaml'
-		prettier().config(['parser': 'yaml', 'printWidth': 100, 'singleQuote': true, 'trailingCommas': 'none'])
-	}
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -152,5 +152,9 @@ spotless {
 	java {
 		googleJavaFormat()
 	}
+	format 'yaml', {
+		target '/src/main/resources/api/service_openapi.yaml'
+		prettier().config(['parser': 'yaml', 'printWidth': 100, 'singleQuote': true, 'trailingCommas': 'none'])
+	}
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -46,7 +46,6 @@ dependencies {
 	implementation group: 'org.webjars', name: 'swagger-ui', version: '3.28.0'
 	implementation group: 'bio.terra', name: 'terra-workspace-manager-client', version: '0.0.3-SNAPSHOT'
 	implementation group: 'org.hashids', name: 'hashids', version: '1.0.3'
-	developmentOnly 'org.springframework.boot:spring-boot-devtools'
 
 	// -- OpenAPI CodeGen dependencies --
 	// TODO: this version of swagger-annotations is old, but the code gen is still relying on it
--- a/build.gradle
+++ b/build.gradle
@@ -44,8 +44,9 @@ dependencies {
 	implementation group: 'org.liquibase' , name: 'liquibase-core', version: '3.9.0'
 	implementation group: 'org.webjars', name: 'webjars-locator-core', version: '0.46'
 	implementation group: 'org.webjars', name: 'swagger-ui', version: '3.28.0'
-	implementation(group: 'bio.terra', name: 'terra-workspace-manager-client', version: '0.0.3-SNAPSHOT')
-	implementation(group: 'org.hashids', name: 'hashids', version: '1.0.3')
+	implementation group: 'bio.terra', name: 'terra-workspace-manager-client', version: '0.0.3-SNAPSHOT'
+	implementation group: 'org.hashids', name: 'hashids', version: '1.0.3'
+	developmentOnly 'org.springframework.boot:spring-boot-devtools'
 
 	// -- OpenAPI CodeGen dependencies --
 	// TODO: this version of swagger-annotations is old, but the code gen is still relying on it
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -42,9 +42,10 @@ dependencies {
 	implementation group: 'org.apache.commons', name: 'commons-pool2', version: '2.8.0'
 	implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.10'
 	implementation group: 'org.liquibase' , name: 'liquibase-core', version: '3.9.0'
-	implementation group: 'org.webjars', name: 'swagger-ui', version: '3.28.0' // if updating, change path in ApiResourceConfig.java
-	implementation(group: 'bio.terra', name: 'terra-workspace-manager-client', version: '0.0.3-SNAPSHOT')
-	implementation(group: 'org.hashids', name: 'hashids', version: '1.0.3')
+	implementation group: 'org.webjars', name: 'webjars-locator-core', version: '0.46'
```
zarsky-broad, line 45: Allows us to use version-agnostic links to swagger-ui static resources

src/main/resources/api/service_openapi.yaml
```diff
@@ -1,264 +1,231 @@
...
+  description: Workspace Manager REST API to manage workspaces.
   version: 0.0.1
 
-servers:
-- url: /
-
 paths:
-  '/status':
+  # Unauthorized Paths
+  /status:
```
zarsky-broad, line 9: All the swagger examples have these unquoted, so I unquoted them here

src/main/resources/api/service_openapi.yaml
```diff
@@ -1,264 +1,231 @@
...
-      description: |
-        Returns the operational status of the service
+      summary: Returns the operational status of the service
       operationId: serviceStatus
-      tags:
-      - unauthenticated
+      tags: [Unauthenticated]
       responses:
-        200:
+        '200':
```
zarsky-broad, line 15: The spec _technically_ requires that these have quotes, so I normalized

src/main/resources/api/service_openapi.yaml
```diff
@@ -1,264 +1,231 @@
...
-- url: /
-
 paths:
-  '/status':
+  # Unauthorized Paths
+  /status:
     get:
-      description: |
-        Returns the operational status of the service
+      summary: Returns the operational status of the service
```
zarsky-broad, line 11: Summaries can be seen when the list is collapsed, descriptions cannot.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1,264 +1,231 @@
...
       operationId: serviceStatus
-      tags:
-      - unauthenticated
+      tags: [Unauthenticated]
```
zarsky-broad, line 13: These are equivalent list formats, so I'm using brackets whenever it's just a list of primitives. The capitalization doesn't change anything in the generated code because it's automatically camel-cased, so it only affects the swagger ui page.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1,264 +1,231 @@
...
       responses:
-        200:
+        '200':
           description: Service is functional
-          $ref: '#/components/responses/StatusResponse'
-        500:
+          content:
```
zarsky-broad, line 17: This way the description is preserved.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1,264 +1,231 @@
...
+      summary: Create a new Workspace. This should only be called by Folder Manager.
       operationId: createWorkspace
-      tags:
-      - workspace
+      tags: [Workspace]
       requestBody:
-        description: |
-          Auth token, spend profile, and policy information from
-          Folder Manager, as well as async job information.
+        required: true
```
zarsky-broad, line 43: openapi-generator doesn't support the `requestBodies` field for refs, for some reason, but if we switch to swagger codegen, the schemas for these can move to that.

src/main/resources/api/service_openapi.yaml
```diff
@@ -267,41 +234,49 @@ components:
       required: true
       schema:
         type: string
+        # format: uuid
```
zarsky-broad, line 237: This is correct, but turning it on requires some actual changes to the java code, so I've left it commented out for now.

zarsky-broad:
> LGTM, though heads up that using the `prettier` yaml formatter requires devs to have npm installed

Oh, nards. I'll remove it, then.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1,264 +1,237 @@
...
+                $ref: '#/components/schemas/ErrorReport'
     delete:
-      description: |
-        Delete a Workspace. This should only be called by the Folder Manager.
+      summary: Delete a Workspace. This should only be called by the Folder Manager.
       operationId: deleteWorkspace
-      tags:
-      - workspace
+      tags: [Workspace]
       requestBody:
```
davidangb, line 88: this line still throws a validation error: `DELETE operations cannot have a requestBody.` I realize that we're documenting what the code is actually doing, so we shouldn't change it - and that error seems to be against the spec. I wonder if there's anything we can do about it.

No changes necessary for this PR, I'm just thinking out loud.

src/main/resources/api/swagger-ui.html
```diff
@@ -48,14 +81,18 @@
...
         plugins: [
-          SwaggerUIBundle.plugins.DownloadUrl
+          SwaggerUIBundle.plugins.DownloadUrl,
+          cleanupPlugin
         ],
-        layout: "StandaloneLayout"
+        layout: 'StandaloneLayout',
+        defaultModelsExpandDepth: -1, // hide the huge list of schemas under the routes
+        defaultModelRendering: 'model', // schema has the descriptions, unlike the example value
+        defaultModelExpandDepth: 2, // affects the schema shown for a request or response
```
davidangb, line 90: I like these changes ... AND I would like to think about ways to make this reusable, so our swagger UIs are consistent across services once we have more MCTerra services. Same as previous comment: nothing to change in this PR, this is for future thinking.

PR #59, opened 2020-07-14
AS-367: Add Stackdriver Tracing

ansingh7115:
Terra helm: https://github.com/broadinstitute/terra-helm/pull/72/files
Terra helmfile: https://github.com/broadinstitute/terra-helmfile/pull/51/files


src/main/java/bio/terra/workspace/service/trace/StackdriverTrace.java
```diff
@@ -0,0 +1,63 @@
...
+@Component
+public class StackdriverTrace {
+  private final StackdriverConfiguration stackdriverConfiguration;
+
+
+  @Autowired
+  public StackdriverTrace(StackdriverConfiguration stackdriverConfiguration) {
+    this.stackdriverConfiguration = stackdriverConfiguration;
+    // Creates and registers the exporter. This must be happen each time WSM is instantiated.
+    // createAndRegister();
```
ansingh7115, line 29: When I call createAndRegister() here, I get a bunch of unit test failures of this type:

`DataReferenceServiceTest > enumerateFailsUnauthorized() STARTED

DataReferenceServiceTest > enumerateFailsUnauthorized() FAILED
    java.lang.IllegalStateException at DefaultCacheAwareContextLoaderDelegate.java:132
        Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException at ConstructorResolver.java:797
            Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException at ConstructorResolver.java:797
                Caused by: org.springframework.beans.factory.BeanCreationException at ConstructorResolver.java:313
                    Caused by: org.springframework.beans.BeanInstantiationException at BeanUtils.java:217
                        Caused by: java.lang.IllegalStateException at Preconditions.java:511
`

ansingh7115: I imagine it's having some issues starting the exporter for tests? The quick fix for this would be to start the exporter in StartUpInitializer. But, I don't know if you have some java magic that will help here :)

src/test/java/bio/terra/workspace/service/trace/StackdriverTraceTest.java
```diff
@@ -0,0 +1,36 @@
...
+  @Test
+  void createScope() {
+    String span1Name = "span1";
+    Scope scope = trace.scope(span1Name);
+
+  }
+
+  @Test
+  void annotate() {}
+}
```
ansingh7115, line 36: Yeah...this file is sad. I'm not sure how to test the StackdriverTrace class without somehow creating a mock Stackdriver. It may just be too thin for tests to actually create any value here

src/test/java/bio/terra/workspace/service/trace/StackdriverTraceTest.java
```diff
@@ -0,0 +1,36 @@
...
+class StackdriverTraceTest {
+
+  @Autowired
+  private StackdriverTrace trace;
+
+  @Test
+  void createScope() {
+    String span1Name = "span1";
+    Scope scope = trace.scope(span1Name);
+
```
ansingh7115, line 31: The only thing you can do with a Scope after you've created it is to close it. So, there not really anything to check.

src/main/java/bio/terra/workspace/service/trace/StackdriverTrace.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  private final Logger logger = LoggerFactory.getLogger(StackdriverTrace.class);
```
davidangb, line 32: logger is not used anywhere

src/main/java/bio/terra/workspace/service/trace/StackdriverTrace.java
```diff
@@ -0,0 +1,63 @@
...
+    } catch (IOException e) {
+      throw new StackdriverRegistrationException(
+          "Failed to create and register OpenCensus Stackdriver Trace exporter", e);
+    }
+  }
+
+  public Scope scope(String traceName) {
+    return tracer
+        .spanBuilder(traceName)
+        .setSampler(Samplers.probabilitySampler(stackdriverConfiguration.getSamplingProbability()))
```
davidangb, line 56: could you create the `Sampler` instance at the class level, and reuse it for each span? (is there any reason to not reuse the Sampler?) It feels wasteful to init a new one each time.

src/test/java/bio/terra/workspace/service/trace/StackdriverTraceTest.java
davidangb: I agree that that it's pretty thin. `annotate()` doesn't do anything except wrap the call; I don't think that's worth fighting against.

For the future, or for this PR if you feel like tackling it: you *could* write a custom exporter for unit tests, and use that instead of the stackdriver exporter: https://opencensus.io/exporters/custom-exporter/java/trace/. That would allow you to write a test that says "when I call a given API, do I properly capture the traces I expected to capture?"

src/main/resources/application.properties
```diff
@@ -4,6 +4,11 @@ workspace.maxStairwayThreads=4
 workspace.resourceId=mc-terra-workspace-manager
 workspace.stairwayTimeoutSeconds=1800
 
+stackdriver.serviceAccountFilePath=src/test/resources/rendered/wsm-firecloud-account.json
```
zloery, line 7: I think we probably want a way to enable/disable tracing via configuration. As is, you can't run WM at all without access to a firecloud account, which is a problem for HCLS/anyone else external.

Relying on stuff in the test directory for our main workflow worries me a bit. Could this be rendered to the main resources directory?

src/main/java/bio/terra/workspace/service/trace/StackdriverTrace.java
zloery: `org.springframework.beans.factory.UnsatisfiedDependencyException` means some component isn't getting all of its dependencies wired in correctly - when I ran it, I was getting errors pointing to StackdriverConfiguration not being able to find the SA credentials in tests, but that was because I hadn't rendered the file

Does this work when called from StartUpInitializer?

src/main/java/bio/terra/workspace/service/trace/StackdriverTrace.java
```diff
@@ -0,0 +1,63 @@
...
+        .startScopedSpan();
```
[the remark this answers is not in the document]
ansingh7115: It seems to nest the spans correctly without calling the parent spans. I did a quick nested trace for getWorkspace just to see what it did: 

```
  public WorkspaceDescription getWorkspace(String id, AuthenticatedUserRequest userReq) {
    try (Scope s = trace.scope("getWorkspaceSpan")) {
      samService.workspaceAuthz(userReq, id, SamUtils.SAM_WORKSPACE_READ_ACTION);
      trace.annotate("authed!");
      try (Scope ss = trace.scope("getWorkspaceDescription")) {
        WorkspaceDescription result = workspaceDao.getWorkspace(id);
        return result;
      }
    }
  }
```
Which ended up with: 
Screen Shot 2020-07-16 at 7 09 14 PM

Here's a [link](https://console.cloud.google.com/traces/traces?project=broad-dsde-dev&start=1594335983607&end=1594940783607&q=getWorkspaceSpan&tid=cc81bfa91316a80f145dfec71ff129fb) to the trace

src/main/resources/application.properties
ansingh7115: I changed it so that this file renders in main. Also created [AS-455](https://broadworkbench.atlassian.net/browse/AS-455) to re-think WSM service accounts. The current SA we're using isn't dedicated to WSM. 

As for enabling/disabling tracing - I can add that to config and do something like:

```
if (traceEnabled) {
    createAndRegister();
}
```

This will mean that traces are created, but will not be exported. There might be better ways to do this though. The alternative way I found (https://github.com/DataBiosphere/terra-workspace-manager/pull/61) has an enable/disable feature

PR #60, opened 2020-07-17
AS-412: POC Orchestrating CreateWorkspaceManager performance tests using GitHubActions and Helm workflows.

ichengchang:
This is fundamental work related to how we can quickly develop and orchestrate perf tests using

1) GitHub Actions
2) Skaffold / Helm

All the setups and performance tests layouts are within the scope of this ticket.

### DIFF
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -54,6 +54,11 @@ jobs:
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
       - name: Run Gatling performance tests
+        env:
+          WSM_BASE_URL: https://workspace.ichang.integ.envs.broadinstitute.org/
+          GATLING_CONCURRENCY: 10
+          CICD: GHA
+          FIRECLOUD_SERVICE_ACCOUNT_PATH: ../src/test/resources/rendered/wsm-firecloud-account.json
         run: |
           cd terra-workspace-manager-performance
           ../gradlew gatlingRun
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -20,10 +20,6 @@ jobs:
         id: vault-token-step
         env:
           VAULT_ADDR: https://clotho.broadinstitute.org:8200
-          WSM_BASE_URL: https://workspace.ichang.integ.envs.broadinstitute.org/
-          GATLING_CONCURRENCY: 10
-          CICD: GHA
-          FIRECLOUD_SERVICE_ACCOUNT_PATH: ../src/test/resources/rendered/wsm-firecloud-account.json
         run: |
           VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
             -e "VAULT_ADDR=${VAULT_ADDR}" \
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -55,5 +55,11 @@ jobs:
         run: chmod +x gradlew
       - name: Change directory to performance project
         run: cd terra-workspace-manager-performance
+      - name: Debug Step 1
+        run: ls -l
+      - name: Debug Step 2
+        run: pwd
+      - name: Debug Step 3
+        run: ls -l ..
       - name: Run Gatling performance tests
         run: ../gradlew gatlingRun
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -53,13 +53,7 @@ jobs:
         run: sudo chmod +r src/test/resources/rendered/wsm-firecloud-account.json
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
-      - name: Change directory to performance project
-        run: cd terra-workspace-manager-performance
-      - name: Debug Step 1
-        run: ls -l
-      - name: Debug Step 2
-        run: pwd
-      - name: Debug Step 3
-        run: ls -l ..
       - name: Run Gatling performance tests
-        run: ../gradlew gatlingRun
+        run: |
+          cd terra-workspace-manager-performance
+          ../gradlew gatlingRun
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -17,7 +17,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      # See master_push.yml
+      # See dev_push.yml
       - name: Get Vault token
         id: vault-token-step
         env:
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -1,0 +1,59 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Run Gatling Performance Tests
+
+on:
+  push:
+    branches:
+      - iv-perf-devbase
+
+jobs:
+  performance-test:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      # See master_push.yml
+      - name: Get Vault token
+        id: vault-token-step
+        env:
+          VAULT_ADDR: https://clotho.broadinstitute.org:8200
+          WSM_BASE_URL: https://workspace.ichang.integ.envs.broadinstitute.org/
+          GATLING_CONCURRENCY: 10
+          CICD: GHA
+          FIRECLOUD_SERVICE_ACCOUNT_PATH: /tmp/wsm-firecloud-account.json
+        run: |
+          VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
+            -e "VAULT_ADDR=${VAULT_ADDR}" \
+            vault:1.1.0 \
+            vault write -field token \
+              auth/approle/login role_id=${{ secrets.VAULT_APPROLE_ROLE_ID }} \
+              secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+          echo ::set-output name=vault-token::$VAULT_TOKEN
+          echo ::add-mask::$VAULT_TOKEN
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Grant execute permission for render_config
+        run: chmod +x render_config.sh
+      - name: Render configuration for tests
+        run: ./render_config.sh ${{ steps.vault-token-step.outputs.vault-token }} dev
+      - name: Ensure read permissions for rendered files
+        run: sudo chmod +r /tmp/wsm-firecloud-account.json src/test/resources/application-integration-test.properties
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      - name: Change directory to performance project
+        run: cd terra-workspace-manager-performance
+      - name: Run Gatling performance tests
+        run: ../gradlew gatlingRun
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -23,7 +23,7 @@ jobs:
           WSM_BASE_URL: https://workspace.ichang.integ.envs.broadinstitute.org/
           GATLING_CONCURRENCY: 10
           CICD: GHA
-          FIRECLOUD_SERVICE_ACCOUNT_PATH: /tmp/wsm-firecloud-account.json
+          FIRECLOUD_SERVICE_ACCOUNT_PATH: ../src/test/resources/rendered/wsm-firecloud-account.json
         run: |
           VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
             -e "VAULT_ADDR=${VAULT_ADDR}" \
@@ -50,7 +50,7 @@ jobs:
       - name: Render configuration for tests
         run: ./render_config.sh ${{ steps.vault-token-step.outputs.vault-token }} dev
       - name: Ensure read permissions for rendered files
-        run: sudo chmod +r /tmp/wsm-firecloud-account.json src/test/resources/application-integration-test.properties
+        run: sudo chmod +r src/test/resources/rendered/wsm-firecloud-account.json
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
       - name: Change directory to performance project
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -6,7 +6,9 @@ name: Run Gatling Performance Tests
 on:
   push:
     branches:
+      - dev
       - iv-perf-devbase
+    tags:
 
 jobs:
   performance-test:
--- a/local-perf/README.md
+++ b/local-perf/README.md
@@ -1,0 +1,35 @@
+This directory contains scripts for running continuous development of performance test builds.
+This is not necessary for deployment, but it can be helpful for developing faster.
+The provided setup script clones the terra-helm and terra-helmfile git repos,
+and templates in the desired Terra environment/k8s namespace to target.
+If you need to pull changes to either terra-helm or terra-helmfile, rerun this script.
+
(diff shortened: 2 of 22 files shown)


local-perf/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+The local-perf provides an example on **Distributed load testing using GKE** as described in this article:
+https://cloud.google.com/solutions/distributed-load-testing-using-gke
+
+Using a containerzied solution for load testing gives us the following advantages
+1) The load test runner is completely ephemeral
+2) Complete ease to scale and run tests reliably and cost effectively
+
+To use this, first ensure Skaffold i
+(available at https://skaffold.dev
```
zloery, line 18: It looks like text is missing here

local-perf/setup_local_env.sh
```diff
@@ -0,0 +1,29 @@
...
+# Optional input
+TERRA_HELM_BRANCH=${2:-master}
+TERRA_HELMFILE_BRANCH=${3:-master}
+
+# Clone Helm chart and helmfile repos
+rm -rf terra-helm
+rm -rf terra-helmfile
+#git clone -b "$TERRA_HELM_BRANCH" --single-branch https://github.com/broadinstitute/terra-helm
+#git clone -b "$TERRA_HELMFILE_BRANCH" --single-branch https://github.com/broadinstitute/terra-helmfile
+git clone -b "$TERRA_HELMFILE_BRANCH" --single-branch ssh://git@github.com/broadinstitute/terra-helm
```
zloery, line 20: Should this be -b "$TERRA_HELM_BRANCH"?

local-perf/setup_local_env.sh
```diff
@@ -0,0 +1,29 @@
...
+# Required input
+ENV=$1
+# Optional input
+TERRA_HELM_BRANCH=${2:-master}
+TERRA_HELMFILE_BRANCH=${3:-master}
+
+# Clone Helm chart and helmfile repos
+rm -rf terra-helm
+rm -rf terra-helmfile
+#git clone -b "$TERRA_HELM_BRANCH" --single-branch https://github.com/broadinstitute/terra-helm
```
zloery, line 18: Can these comments be deleted?

local-perf/terra-helm-local/charts/workspacemanager-performance/Chart.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v2
```
zloery, line 1: Having these charts in local-perf/terra-helm-local/* kind of confuses me. The setup_local_env scripts clones in the terra-helm and -helmfile repos, but it looks like you're actually using local charts instead of the ones from the cloned repo.

I think you probably want put the charts in the helm repo (https://github.com/broadinstitute/terra-helm) and reference the locally cloned repo

terra-workspace-manager-performance/build.gradle
```diff
@@ -0,0 +1,83 @@
...
+}
+
+def jibPluginId = 'com.google.cloud.tools.jib'
+final hasPlugin = project.getPlugins().hasPlugin(jibPluginId);
+if (hasPlugin) {
+    final Plugin plugin = project.getPlugins().getPlugin(jibPluginId)
+    println 'Plugin already applied - version ' + plugin.properties['jibPluginId']
+} else {
+    apply plugin: "com.google.cloud.tools.jib"
+}
```
zloery, line 25: Why is this section needed? Does applying the jib plugin to this project conflict with the main project?

ichengchang:
> The following is regarding the deployment/running of the tests, I am not really versed in Gradle/Gatling so I will not comment on the framework or tests themselves.
> 
> I don't see a case for running these tests in k8s. It adds an unnecessary level of complexity. It also yields this awkward dichotomy of how the tests run locally vs how they run when initiated by GH actions. Why not implement the tests in a Dockerfile which developers can run locally and that the GH actions workflow references as well? The added complexity of Skaffold/Helm for this seems like overkill to me.
> 
> Helm is a poor fit for running one-off jobs. It's meant to be a declarative spec for k8s deploys. The best practice is to have jobs defined in pre- or post-deploy [hooks](https://helm.sh/docs/topics/charts_hooks/) around the main, permanent k8s configuration that is being deployed. I have never seen a Helm chart where the main piece of configuration being deployed is a job.
> 
> If this was an appropriate use case for Helm, service repositories are not the place to check in Helm charts, we have a [dedicated repository](https://github.com/broadinstitute/terra-helm) for that, that comes with its own set of best practices.
> 
> I realize that this is something you have been working on without too much DevOps guidance/input. Part of the reason for that is that our team is pretty thinly stretched at the moment, and we did not have the bandwidth to take a look at what you were working on earlier. If we had, we would have pointed out the above much sooner. We were under the impression that the performance setup you were working on was very much an ad-hoc thing, and that we would have a chance to review the deployment tooling involved and tweak it to be more in line with our MC-Terra CI/CD best practices.
> 
> Let's schedule some time to discuss these changes, and maybe a way to keep our teams' efforts more in-sync.

Here's an article on Distributed load testing in GKE [https://cloud.google.com/solutions/distributed-load-testing-using-kubernetes](url).

Right now, the dichotomy (GHA vs Helm) arises because of the different ways of storing config data that is easy to reconcile. 

Running containerized load tests within GKE will enable us to easily stand up ephemeral tests, scale and run them in parallel. We can certainly use Docker but since Jib was used I am sticking to the Jib tool to build image for performance tests rather than using Docker.

terra-workspace-manager-performance/build.gradle
ichengchang: Yes that's correct, the main project already reference the jib class path which causes a conflict in the performance subproject. I haven't found a better way to fix this yet.

local-perf/terra-helm-local/charts/workspacemanager-performance/Chart.yaml
ichengchang: Correct! The reason we are using `terra-helm-local` here is because we haven't yet put the performance test chart in `terra-helm` repo. This is pending further sync up with DevOps in the coming QA / DevOps sync up.

local-perf/README.md
ichengchang: Fixed

local-perf/setup_local_env.sh
ichengchang: Fixed.

local-perf/setup_local_env.sh
ichengchang: Comments removed.

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,62 @@
...
+    tags:
+
+jobs:
+  performance-test:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      # See master_push.yml
```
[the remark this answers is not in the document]
ichengchang: Thanks for catching this! Comment has been updated.

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Ensure read permissions for rendered files
+        run: sudo chmod +r src/test/resources/rendered/wsm-firecloud-account.json
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      - name: Run Gatling performance tests
+        env:
+          WSM_BASE_URL: https://workspace.ichang.integ.envs.broadinstitute.org/
+          GATLING_CONCURRENCY: 10
+          CICD: GHA
+          FIRECLOUD_SERVICE_ACCOUNT_PATH: ../src/test/resources/rendered/wsm-firecloud-account.json
```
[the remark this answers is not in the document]
ichengchang: That's correct. WSM is still using SAM dev instance. Obviously this config will change when it's ready to do so.

local-perf/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+To use this, first ensure Skaffold is installed on your local machine 
+(available at https://skaffold.dev/). 
+
+> Older versions of Skaffold (v1.4.0 and earlier) do not have support for Helm 3 and will fail to deploy your 
+changes. If you're seeing errors like `UPGRADE FAILED: "(Release name)" has no 
+deployed releases`, try updating Skaffold.
+
+You may need to use gcloud to provide GCR
+ credentials with `gcloud auth configure-docker`. Finally, run local-run.sh with
```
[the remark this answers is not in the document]
ichengchang: Fixed

terra-workspace-manager-performance/src/gatling/resources/application.conf
```diff
@@ -0,0 +1,16 @@
+dev {
+  terra {
+    env {
+      wsmBaseUrl = "WSM_BASE_URL"
+    },
+    wsmBaseUrl = "https://workspace.ichang.integ.envs.broadinstitute.org/"
```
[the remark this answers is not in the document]
ichengchang: Removed the unused config

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,62 @@
...
+        run: |
+          cd terra-workspace-manager-performance
```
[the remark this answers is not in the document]
ichengchang: This is to make sure that the `build.gradle` of module / subproject `terra-workspace-manager-performance` is loaded rather than the root project's `build.gradle`. We structured the configuration so that the load test framework is defined in the subproject's `build.gradle` (in this case Gatling framework). In the future, when we develop perf tests using say Jmeter framework, we can update the subproject's `build.gradle` to Jmeter framework and configured the source sets to load tests from say src/test/jmeter within the subproject.

ichengchang:
> Thanks Ivan for flagging me on the review. As we'd discussed, I'm mostly interested in seeing how this work interacts with what TDR team is working on re: performance testing, and how whatever solution we align on might scale out across different MC Terra components. I'm starting to get a better sense of where a non-Gatling test executor might fit in here.
> 
> One thing that may still be interesting to explore is what it would look like to use Gatling to bootstrap the test runs, but have as much simulation code as possible use the standard WSM Java code style. I think there's a baseline expectation / goal that the majority of test code should be in the same language as the service code. That may easily be left for future effort, but it's something worth considering.

I have already started exploring Jmeter as a candidate of our next load test framework. There are couple Jmeter gradle plugins and there's a way to write load tests entirely in Java using Jmeter Java API, I think this will be appealing to teams who prefer a pure 100% Java codebase rather than Scala or some other scripting languages based on recording web sessions.

ichengchang:
> > The following is regarding the deployment/running of the tests, I am not really versed in Gradle/Gatling so I will not comment on the framework or tests themselves.
> > I don't see a case for running these tests in k8s. It adds an unnecessary level of complexity. It also yields this awkward dichotomy of how the tests run locally vs how they run when initiated by GH actions. Why not implement the tests in a Dockerfile which developers can run locally and that the GH actions workflow references as well? The added complexity of Skaffold/Helm for this seems like overkill to me.
> > Helm is a poor fit for running one-off jobs. It's meant to be a declarative spec for k8s deploys. The best practice is to have jobs defined in pre- or post-deploy [hooks](https://helm.sh/docs/topics/charts_hooks/) around the main, permanent k8s configuration that is being deployed. I have never seen a Helm chart where the main piece of configuration being deployed is a job.
> > If this was an appropriate use case for Helm, service repositories are not the place to check in Helm charts, we have a [dedicated repository](https://github.com/broadinstitute/terra-helm) for that, that comes with its own set of best practices.
> > I realize that this is something you have been working on without too much DevOps guidance/input. Part of the reason for that is that our team is pretty thinly stretched at the moment, and we did not have the bandwidth to take a look at what you were working on earlier. If we had, we would have pointed out the above much sooner. We were under the impression that the performance setup you were working on was very much an ad-hoc thing, and that we would have a chance to review the deployment tooling involved and tweak it to be more in line with our MC-Terra CI/CD best practices.
> > Let's schedule some time to discuss these changes, and maybe a way to keep our teams' efforts more in-sync.
> 
> Here's an article on Distributed load testing in GKE [https://cloud.google.com/solutions/distributed-load-testing-using-kubernetes](url).
> 
> Right now, the dichotomy (GHA vs Helm) arises because of the different ways of storing config data that is easy to reconcile.
> 
> Running containerized load tests within GKE will enable us to easily stand up ephemeral tests, scale and run them in parallel. We can certainly use Docker but since Jib was used I am sticking to the Jib tool to build image for performance tests rather than using Docker.

Here's another article on running perf tests in k8s:

https://dzone.com/articles/performance-testing-in-kubernetes

ansingh7115:
I agree with Devops, I don't think we should be running tests with Kubernetes/Helm. Not only (as Chelsea and Greg have said) are they not meant for that purpose generally, but it doesn't fit with our current infrastructure decisions. Our current agreement is to use Helm for defining deployments and GitHub Actions for running CI/CD pipelines, as we have done for integration tests. If we use Kubernetes Jobs, we now have two tools for running CI/CD pipelines, which is not necessary. It also adds to the burden on other engineers to now familiarize themselves with two different tools.

My strong recommendation is to use GitHub Actions for running performance tests, which is consistent with how we run both unit tests and integration tests. I have not yet really heard any reason not to do the same for performance tests.   

In the end, this is an infrastructure decision, and those belong to Devops. If Devops thinks we should be using GitHub Actions for this, then I think we should listen to them here.

terra-workspace-manager-performance/src/gatling/resources/application.conf
```diff
@@ -0,0 +1,15 @@
...
+      wsmBaseUrl = "WSM_BASE_URL"
+    }
+  },
+  sam {
+    basePath = "https://sam.dsde-dev.broadinstitute.org",
+    samBillingProject = "sam-perf-test-user",
+    email = "william.thunderlord@test.firecloud.org",
+    firecloudServiceAccount = "FIRECLOUD_SERVICE_ACCOUNT_CREDS",
+    firecloudServiceAccountPath = "FIRECLOUD_SERVICE_ACCOUNT_PATH",
+    serviceAccountFilePath = "SERVICE_ACCOUNT_FILE_PATH"
```
ungwudik, line 13: Not critical - but I was expecting to see actual "accounts" and "paths" as values of these variables. Maybe these variable names can be updated?

ungwudik: > Not critical - but I was expecting to see actual "accounts" and "paths" as values of these variables. Maybe these variable names can be updated?

Ahh this is pulling from an env variable. Nevermind.

terra-workspace-manager-performance/src/gatling/simulations/CreateWorkspaceSimulation.scala
```diff
@@ -0,0 +1,99 @@
...
+  val authToken = newAccessToken.getTokenValue
+
+  val workspaceIdFeeder = Iterator.continually(
+    Map("workspaceId" -> UUID.randomUUID)
+  )
+  val tokenFeeder = Iterator.continually(
+    Map("authToken" -> authToken)
+  )
+
+  val api : String = "api/workspaces/v1"
```
ungwudik, line 77: Can this endpoint be defined in `application.conf` file or somewhere similar? I imagine it's going to be reused by other tests.

davidangb:
I'm chiming in with a +1 to what's been said before, regarding the high-level architecture of moving test jobs out of Helm/k8s and into Github Actions. I'll hold off on lower-level code review until the architecture settles down.

terra-workspace-manager-performance/src/gatling/simulations/CreateWorkspaceSimulation.scala
```diff
@@ -0,0 +1,99 @@
+import com.google.auth.oauth2.GoogleCredentials
```
ungwudik, line 1: We may benefit from moving some of the test-agnostic 'config' stuff (auth, GHA, etc) out of this test file. I understand that this is a POC, but I think it's worth mentioning.

ungwudik:
Touched base with Ivan regarding the Helm-related idea that's yet to be hashed out here.
I believe the current plan is to discuss the issue in the inaugural "QA/DevOps Sync" meeting next Tuesday, so I tried to keep my review comments outside of that topic.

terra-workspace-manager-performance/src/gatling/simulations/CreateWorkspaceSimulation.scala
ichengchang: Yes good point, a lot of these stuff can either be defined in `application.conf` of env vars. I don't know which way teams prefer. I think it would make sense to simplify this to a single approach. Let's have a discussion.

terra-workspace-manager-performance/src/gatling/resources/application.conf
ichengchang: Cool

terra-workspace-manager-performance/src/gatling/simulations/CreateWorkspaceSimulation.scala
ichengchang: Yup definitely that's the plan. Integration tests also use the same library, maybe we should have a shared library that both performance and integration tests can use.

PR #61, opened 2020-07-20
AS-367: Alternative: Spring Cloud Trace

ansingh7115:
Uses Spring Cloud:
Example code: https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample
Docs: https://docs.spring.io/spring-cloud-sleuth/docs/current-SNAPSHOT/reference/html/

Here's what the example trace I added looks like:

Screen Shot 2020-07-20 at 4 34 33 AM


src/main/java/bio/terra/workspace/app/configuration/ApiResourceConfig.java
```diff
@@ -11,4 +17,22 @@
...
+          public boolean preHandle(
+              HttpServletRequest request, HttpServletResponse response, Object handler)
+              throws Exception {
+            spanCustomizer.tag("session-id", request.getSession().getId());
+            // add tags to Stackdriver traces here
+
+            return true;
+          }
+        });
+  }
```
ansingh7115, line 37: creates a parent span for each API call

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -51,9 +57,14 @@ public CreatedWorkspace createWorkspace(
...
+  @NewSpan
   public WorkspaceDescription getWorkspace(String id, AuthenticatedUserRequest userReq) {
-
-    samService.workspaceAuthz(userReq, id, SamUtils.SAM_WORKSPACE_READ_ACTION);
+    Span newSpan = tracer.nextSpan().name("workspaceAuthz");
+    try (Tracer.SpanInScope ws = tracer.withSpanInScope(newSpan.start())) {
+      samService.workspaceAuthz(userReq, id, SamUtils.SAM_WORKSPACE_READ_ACTION);
+    } finally {
+      newSpan.finish();
+    }
```
ansingh7115, line 67: Here's an example trace 

- We have @NewSpan, which creates a span with the same name as the method
- We are also creating a span explicitly, and then closing it. I tried to make this cleaner by attempting to make a Span class that implements Autocloseable, but it got a little hairy.

I imagine we wouldn't make too many spans explicitly, and use @NewSpan generally

render_config.sh
```diff
@@ -4,7 +4,8 @@ VAULT_TOKEN=${1:-$(cat $HOME/.vault-token)}
 TARGET_ENV=${2:-dev}
 
 FIRECLOUD_ACCOUNT_VAULT_PATH=secret/dsde/firecloud/${TARGET_ENV}/common/firecloud-account.json
-SERVICE_ACCOUNT_OUTPUT_FILE_PATH="$(dirname $0)"/src/test/resources/rendered/wsm-firecloud-account.json
+SERVICE_ACCOUNT_OUTPUT_FILE_PATH="$(dirname $0)"/src/main/resources/generated/wsm-firecloud-account.json
+SERVICE_ACCOUNT_TEST_OUTPUT_FILE_PATH="$(dirname $0)"/src/test/resources/rendered/wsm-firecloud-account.json
```
ansingh7115, line 8: Created https://broadworkbench.atlassian.net/browse/AS-455 to handle figuring out how SAs should work / where they should be rendered for WSM.

src/test/resources/application-integration-test.properties
```diff
@@ -3,4 +3,4 @@ it.wsmUrls.dev=https://workspace.dsde-dev.broadinstitute.org/
 it.wsmEndpoints.workspaces=api/workspaces/v1
 
 it.serviceAccountEmail=william.thunderlord@test.firecloud.org
-it.serviceAccountFilePath=src/test/resources/rendered/wsm-firecloud-account.json
+it.serviceAccountFilePath=src/main/resources/generated/wsm-firecloud-account.json
```
zloery, line 6: Should this one be reading from the test dir since you're rendering to both locations?

src/test/resources/application.properties
```diff
@@ -4,6 +4,11 @@ workspace.maxStairwayThreads=4
 workspace.resourceId=mc-terra-workspace-manager
 workspace.stairwayTimeoutSeconds=1800
 
+spring.cloud.gcp.trace.enabled=true
```
zloery, line 7: Do we want to enable tracing for all tests? It seems like unit tests would pollute the dev environments logs when we could just turn it off

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
zloery: Are these two different ways of opening spans, or do we need to use both in some places?

ansingh7115: Yeah, I'm just modeling two different ways of creating a span. I think we'll probably use @NewSpan generally (which, btw, you can also specify a name instead of automatically using the name of the method by doing @NewSpan("different name") ).

ansingh7115: For the explicitly created span, `SpanInScope` implements `Closeable`, not `AutoCloseable`, which is why we need to explicitly close the span in that `finally { ... }` block :( 
I attempted to try to finagle trying to make something that uses `Autocloseable` but it got messy and is probably not worth it

src/test/resources/application.properties
ansingh7115: true! I will disable for tests

src/main/java/bio/terra/workspace/app/configuration/ApiResourceConfig.java
davidangb: can this use the MDC id instead of/in addition to the session id?

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
davidangb: The explicitly-created spans are definitely more verbose/awkward than the opencensus implementation. I *do* think we need the level of detail of multiple inner spans, inside a complicated API. Is your thinking here that if we need inner spans, all the business logic for that span should be refactored into a standalone method, which could use @NewSpan?

ansingh7115: Yes - I was thinking we could generally break out methods if we need to create spans for blocks of code and use @NewSpan.

But, idk if that's good practice. It keeps the code cleaner. But then we're restructuring our code for the sake of a trace. Do you think it's worth it?

render_config.sh
ansingh7115: I am leaving the test SA in for now, AS-455 will handle figuring out SAs

src/main/java/bio/terra/workspace/app/configuration/ApiResourceConfig.java
```diff
@@ -28,7 +28,12 @@ public void addInterceptors(InterceptorRegistry registry) {
           public boolean preHandle(
               HttpServletRequest request, HttpServletResponse response, Object handler)
               throws Exception {
-            spanCustomizer.tag("session-id", request.getSession().getId());
+            String mdcRequestId = request.getHeader("X-Request-ID");
+            if (mdcRequestId != null) spanCustomizer.tag("mdc-request-id", mdcRequestId);
+
+            String mdcCorrelationId = request.getHeader("X-Correlation-ID");
+            if (mdcCorrelationId != null)
+              spanCustomizer.tag("mdc-correlation-id", mdcCorrelationId);
```
davidangb, line 36: hm, is there a way to reuse the implementation in `MdcLogEnhancerFilter`? That filter also generates a unique MDC id if none exists, it would be good to have the generated id if neither header is present.

ansingh7115: I believe `addinterceptor` is being called before `MdcLogEnhancerFilter`. I think the only way to get id found/generated in the filter to be available in the interceptor is to move the MDC log code from the filter to the interceptor. 

That way, we get the MDC id here, create it if it doesn't exist and then add the tag to the customizer, all in the interceptor. 

Unsure if there's a better way to do it than that.

src/main/java/bio/terra/workspace/app/configuration/ApiResourceConfig.java
```diff
@@ -11,4 +19,45 @@
...
+
+            // get an mdc id from the request (if not found, create one), and pass it along in the response
+            String requestId = getMDCRequestId(httpRequest);
+            httpResponse.addHeader("X-Request-ID", requestId);
+
+            // add tags to Stackdriver traces
+            spanCustomizer.tag("requestId", requestId);
+
+            return true;
+          }
```
ansingh7115, line 42: @MatthewBemis  Moved MDC stuff from `Filter` into `preHandle`. I removed `MDC.put(...)`. Was that being used for something I was not aware of (I assumed it was just so that the `MdcResponseFilter` had access to it)

davidangb: Isn't the `MDC.put(...)` the critical piece that adds the id into the log statements? On this branch, does the id still appear in logs?

ansingh7115: Ahhh. Okay - my bad - I didn't take the time to learn about MDC fully and assumed what put() was for something else, which was incorrect.

ansingh7115: Adding the put() back in, the log statements do again show the mdc id. 
Screen Shot 2020-07-23 at 10 40 59 AM

PR #64, opened 2020-07-27
Upgrade to Java 11 [AS-394]

zarsky-broad:
Still creating a Java 8 artifact of the client for Rawls. I think this is everything.


davidangb:
What criteria did you use to choose `joschi/setup-jdk` over any other github action? For instance, I see https://github.com/AdoptOpenJDK/install-jdk which feels more official. I haven't done any research into what's out there, so please consider this an honest question.

zarsky-broad:
@davidangb The reason I didn't pick that one is that it's not in the official actions marketplace, so I totally didn't see that it exists! I'll switch over, I like that it's official.

davidangb:
Also - could you add some notes to DEVELOPMENT.md or README.md, giving developers some pointers/links to using AdoptOpenJDK locally, i.e. in their IntelliJ setups? I don't think we should rewrite info that's already out there, but an overview/links would be nice

zarsky-broad:
After a bit more reading, `joschi/setup-jdk` uses a more standard pipeline and seems to be updated more regularly, plus it's actually published on the marketplace, which means it's widely used. Of the options on the marketplace, it's the only one that will let us install the specific openjdk that we want (without also installing scala). I'm going to stick with this. I'll add some documentation.

zarsky-broad:
Also, for simplifying dev experience outside IntelliJ, what would you think about encouraging people to use [jEnv](https://www.jenv.be/) so that we could add a `.java-version` file and people's shells would switch to the correct version automatically? The suggestion would be:

```sh
brew install jenv
# follow postinstall instructions to add jenv to shell...

# to add previously installed versions of Java to jEnv, list them:
# /usr/libexec/java_home -V
# and then add them:
# jenv add /Library/Java/JavaVirtualMachines/<JAVA VERSION GOES HERE>/Contents/Home

brew cask install adoptopenjdk11
jenv add /Library/Java/JavaVirtualMachines/
jenv add /Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home
```

davidangb:
@zarsky-broad "standard pipeline", "updated regularly", "published on the marketplace", and "widely used" are good criteria for choosing this action. Are there any other actions that do the same thing that are *more* widely used or standard or regularly updated?

Whatever action we choose here is likely to be used by other repos, so it is worthwhile to be able to defend the choice and have confidence in it.

re: jenv ... sure, I think it is fine to encourage but not require its use. I have no experience with it but it looks useful and if it's anything like `nvm`, it's useful. Devs would still need to configure their IntelliJ projects, I assume.

zarsky-broad:
The action that Github publishes on the marketplace doesn't let us specify AdoptOpenJDK, so that's out. The one more widely used option insists on also installing Scala, so I disqualified it as well.

I actually quite like jEnv over, say, nvm, because it doesn't do any installing, it just manages _which_ installed version is used _when_.

zarsky-broad:
@davidangb How's this look?

PR #67, opened 2020-07-28
[AS-430] Cleanup for findings in security report

zloery:
The enum value changes will break dev Rawls. This PR should be merged at the same time as a change to Rawls: https://github.com/broadinstitute/rawls/pull/1263

This includes a variety of small cleanup changes to address issues raised in Workspace Manager's security review. Generally, they fall into a few categories:

- Properly typing enum and UUID inputs, rather than accepting and manually validating raw strings.
- Removing user-provided input from Workspace Manager's output. Many validation error messages no longer include the invalid input, as it's considered best practice not to return potentially malicious strings.
- Changing request parsing errors from HTTP status 500 to 400 to reflect that these are really errors in the user's input.
- Changing reference type enum values to CAPITALS_WITH_UNDERSCORES for consistency with Java style and default Spring enum parsing for parameters.


src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -21,16 +21,15 @@ public DataReferenceValidationUtils(ObjectMapper objectMapper, DataRepoService d
   }
 
   public String validateReference(
-      DataReferenceDescription.ReferenceTypeEnum referenceType,
-      String reference,
-      AuthenticatedUserRequest userReq) {
+      ReferenceTypeEnum referenceType, String reference, AuthenticatedUserRequest userReq) {
 
-    if (referenceType.equals(DataReferenceDescription.ReferenceTypeEnum.DATAREPOSNAPSHOT)) {
+    if (referenceType != null && referenceType.equals(ReferenceTypeEnum.DATA_REPO_SNAPSHOT)) {
```
zarsky-broad, line 26: Can we invert this? `ReferenceTypeEnum.DATA_REPO_SNAPSHOT.equals(referenceType)` would seem to imply the not-null, right?

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -71,10 +71,13 @@ public boolean isAuthorized(
...
   }
 
-  public void workspaceAuthz(AuthenticatedUserRequest userReq, String workspaceId, String action) {
+  public void workspaceAuthz(AuthenticatedUserRequest userReq, UUID workspaceId, String action) {
     boolean isAuthorized =
         isAuthorized(
-            userReq.getRequiredToken(), SamUtils.SAM_WORKSPACE_RESOURCE, workspaceId, action);
+            userReq.getRequiredToken(),
+            SamUtils.SAM_WORKSPACE_RESOURCE,
+            workspaceId.toString(),
```
zarsky-broad, line 79: Ugh, annoying

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zloery: Nice, that works

zloery:
I think the actual valid strings actually have changed, we're going from `DataRepoSnapshot` to `DATA_REPO_SNAPSHOT`

src/main/java/bio/terra/workspace/app/configuration/exceptions/InputJsonParsingException.java
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  public InputJsonParsingException(String message, List<String> causes) {
+    super(message, causes);
+  }
+
+  public InputJsonParsingException(String message, Throwable cause, List<String> causes) {
+    super(message, cause, causes);
+  }
+}
```
davidangb, line 27: where is this class used?

src/main/java/bio/terra/workspace/app/controller/GlobalExceptionHandler.java
```diff
@@ -29,11 +31,25 @@
...
+    // Instead, we return a generic error message about input validation.
+    String validationErrorMessage =
+        "Request could not be parsed or was invalid: "
+            + ex.getClass().getSimpleName()
+            + ". Ensure that all types are correct and that enums have valid values.";
+    ErrorReport errorReport =
+        new ErrorReport()
+            .message(validationErrorMessage)
+            .statusCode(HttpStatus.BAD_REQUEST.value());
+    return new ResponseEntity<>(errorReport, HttpStatus.BAD_REQUEST);
```
davidangb, line 52: this is such a balance between security and user-friendliness. If possible, let's try to at least output the *name* of the variable that failed validation, e.g. "SnapshotId value could not be parsed or was invalid". I realize this won't always be possible, and would need to be captured at the point of validation and passed to this handler (as a field in the various Exceptions?). No need to hold this PR for that, but let's try as we continue to develop.

src/main/resources/api/service_openapi.yaml
```diff
@@ -424,6 +432,20 @@ components:
...
+
+    ReferenceTypeEnum:
+      type: string
+      description: The type of a reference, if an uncontrolled resource
+      # TODO: swagger-codegen always looks for and removes a common prefix for
+      # enum values, which breaks enums that only have a single value.
+      # PLACEHOLDER_VALUE should be removed when we have an actual second value
+      # to use.
+      enum: ['DATA_REPO_SNAPSHOT', 'PLACEHOLDER_VALUE'] #eventually include GCS bucket, etc.
+
```
davidangb, line 448: oof. Is there any other solution to this? (fine to do in a follow-on PR)

src/main/java/bio/terra/workspace/app/configuration/exceptions/InputJsonParsingException.java
zloery: Ah good catch! This was my first approach that I replaced with changes to the GlobalExceptionHandler

src/main/resources/api/service_openapi.yaml
zloery: Yeah... 

As far as I can tell, there's no option to disable this behavior in swagger-codegen (there is back in OpenApiGenerator, but we've already had one migration...). 
We could change the value to `DATAREPOSNAPSHOT`, but the lack of underscores is more difficult to read.
The old value of `DataRepoSnapshot` caused issues with enum parsing, since by default swagger-codegen enums and their string values don't always match

src/main/java/bio/terra/workspace/app/controller/GlobalExceptionHandler.java
zloery: Agreed, I'm a little concerned that we're packing too many types of errors into a general "your request is bad" message.

zloery:
Rawls PR: https://github.com/broadinstitute/rawls/pull/1263

src/main/resources/api/service_openapi.yaml
davidangb: This comment about using a custom generator is enticing: https://github.com/swagger-api/swagger-codegen-generators/issues/246#issuecomment-442629121 but it feels like overkill to use a custom generator just for this.

PR #74, opened 2020-07-30
Reorganize service accounts between tests and local instance

ansingh7115:
This PR addresses how service accounts get rendered and how/when the application accesses them.

Before, we were rendering the same service account twice - once in `src/main/resources/generated` (for running the app locally) and once in `src/test/resources/rendered/` (for running integration tests). Now, we could create a `/rendered` directory and render the service account there once. 

**Changes in this PR:**
- `render_configs.sh` now only saves the service account used for local runs and integration tests once, to a `/rendered` directory it creates. 
- Add this `/rendered` directory to `.gitignore`. 
- Change to `DEVELOPMENT.md` instructions on running locally: you must now run `render_configs` before running locally, since the app now relies on a service account key being present.
- Change GitHub Actions workflow that runs integration test to point to this new file location for the SA.
- Change the `application-integration-test.properties` to point to this new file location for the SA
- Change the app's `application.properties` to read the SA for Cloud Trace from the env variable `CLOUD_TRACE_SA_PATH`. When deployed in K8s, Helm creates this SA and sets the env variable to the correct file path ([PR for this](https://github.com/broadinstitute/terra-helm/pull/92)). When deploying locally, this is set by the developer.

The slightly confusing thing is that we use a special service account that only has Cloud Trace Admin access for when we deploy to the real environments. But, when we deploy locally or run integration tests, we are using the same all-purpose service account for everything.


DEVELOPMENT.md
```diff
@@ -85,8 +85,15 @@ export STAIRWAY_DATABASE_USER=stairwayuser
 export STAIRWAY_DATABASE_USER_PASSWORD=stairwaypwd
 export STAIRWAY_DATABASE_NAME=stairwaylib
 export SAMPLING_PROBABILITY=1.0
+export CLOUD_TRACE_SA_PATH=rendered/service-account.json
+export SERVICE_GOOGLE_PROJECT=broad-dsde-dev
+export SAM_ADDRESS=https://sam.dsde-dev.broadinstitute.org
+```
+
+and render configs if you haven't already 
+```sh
```
zarsky-broad, line 94: Nit: add newline before code block (technically correct, some markdown readers require it)

PR #79, opened 2020-08-04
AS-423: Non-Gatling Tool for Perf testing (JMeter)

ichengchang:
This PR is a POC for demonstrating the capability of JMeter as a possible replacement of Gatling.
Writing Java code for test development rather than using Scala.

The `CreateWorkspaceSimulation` class demonstrates the typical workflow of writing a JMeter test in Java. Although it seems like a lot of code, most are boilerplate code that can be refactored into the following sequence of steps. In addition, the code also include *Gui imports which supports the generation of _jmx_ script. The _jmx_ script can be loaded into the JMeter GUI tool for test developers who are more familiar with GUI tool than Java for local development and testing purposes.

- Create a top-level test plan node for all JMeter tasks
- Create a setUp thread group that run once for initialization (e.g. obtain a bearer token for the test)
- Create regular thread groups representing concurrent users and configure load profile
- Parameterize user parameters for each thread
- Build Http request for CreateWorkspace API including required headers
- Configure backend listener for the reporting framework (either Graphite or InfluxDb)

Our next step would be to use this PR as a step forward to investigate what kind of custom header does JMeter need to provide to support the tracking of trace or log information. The reporting framework will need to be extended to include these headers as well.

NB: The `Jmeter-director-worker` module has been left out from this PR to keep the PR small. There will be a separate PR for this which will include devOps as reviewers.

### DIFF
--- a/automation-jmeter-plugins/src/main/java/automation/plugins/jmeter/functions/OAuth2Token.java
+++ b/automation-jmeter-plugins/src/main/java/automation/plugins/jmeter/functions/OAuth2Token.java
@@ -1,0 +1,178 @@
+package automation.plugins.jmeter.functions;
+
+import com.google.auth.oauth2.GoogleCredentials;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.jmeter.engine.util.CompoundVariable;
+import org.apache.jmeter.functions.AbstractFunction;
+import org.apache.jmeter.functions.InvalidVariableException;
+import org.apache.jmeter.samplers.SampleResult;
+import org.apache.jmeter.samplers.Sampler;
+import org.apache.jmeter.threads.JMeterVariables;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.*;
+
+public class OAuth2Token extends AbstractFunction {
+    private static final Logger log = LoggerFactory.getLogger(OAuth2Token.class);
+    private static final List<String> desc = new LinkedList<>();
+
+    private static final String KEY = "__OAuth2Token"; //$NON-NLS-1$
+
+    // Number of parameters expected - used to reject invalid calls
+    private static final int MIN_PARAMETER_COUNT = 3;
+    private static final int MAX_PARAMETER_COUNT = 4;
+
+    static {
+        desc.add(getResString("oauth2_provider")); //$NON-NLS-1$
+        desc.add(getResString("oauth2_service_accountkey_env_or_configpath")); //$NON-NLS-1$
+        desc.add(getResString("oauth2_user_email")); //$NON-NLS-1$
+        desc.add(getResString("function_name_paropt")); //$NON-NLS-1$
+    }
+
+    private Object[] values;
+
+    public OAuth2Token() {
+
+    }
+
+    @Override
+    public List<String> getArgumentDesc() {
+        return desc;
+    }
+
+    @Override
+    public String execute(SampleResult previousResult, Sampler currentSampler) throws InvalidVariableException {
+        String authToken = null;
+
+        JMeterVariables vars = getVariables();
+
+        CloudProvider cloudProvider = CloudProvider.fromLabel(((CompoundVariable) values[0]).execute().trim());
+
+        String saKeyEnvVarOrConfigPath = ((CompoundVariable) values[1]).execute().trim();
+        String userEmail = System.getenv(((CompoundVariable) values[2]).execute().trim());
+        byte[] saKey = getBytesFromEnvVarOrConfigPath(saKeyEnvVarOrConfigPath);
+
+        switch (cloudProvider) {
+            case GCP:
+                GoogleCredentials credentials = null;
+                try {
+                    credentials = GoogleCredentials.fromStream(new ByteArrayInputStream(saKey))
+                            .createScoped(new ArrayList<String>() {
+                                private static final long serialVersionUID = 2644360116969700916L;
+                                {
+                                    add("profile");
+                                    add("email");
+                                    add("openid");
+                                }
+                            }).createDelegated(userEmail);
+                    credentials.refreshIfExpired();
+                    authToken = credentials.getAccessToken().getTokenValue();
+                } catch (IOException e) {
+                    log.warn(e.getMessage());
+                    throw new RuntimeException(e.getMessage());
+                }
+                break;
+            case AWS:
+            case AZURE:
+                log.warn(cloudProvider.getDescription() + " not implemented");
+                throw new RuntimeException(cloudProvider.getDescription() + " not implemented");
+            default:
+                log.warn(cloudProvider.getDescription());
+                throw new RuntimeException(cloudProvider.getDescription());
+        }
+
+        if (StringUtils.isEmpty(authToken)) {
+            log.warn("Unable to obtain oauth2 token for user '{}'", userEmail);
+            throw new RuntimeException(String.format("Unable to obtain oauth2 token for user '%s'", userEmail));
+        }
+
+        String varName = null;
+
+        if (values.length > 3)
+            varName = ((CompoundVariable) values[3]).execute().trim();
+
+        if (vars != null && varName != null) {
+            vars.put(varName, authToken);
+        }
+
+        return authToken;
+    }
+
+    @Override
+    public void setParameters(Collection<CompoundVariable> parameters) throws InvalidVariableException {
+        checkParameterCount(parameters, MIN_PARAMETER_COUNT, MAX_PARAMETER_COUNT);
+        values = parameters.toArray();
+    }
+
+    @Override
+    public String getReferenceKey() {
+        return KEY;
+    }
+
+    private static String getResString(String key) {
+        ResourceBundle bundle = ResourceBundle.getBundle("messages");
+        return bundle.containsKey(key) ? bundle.getString(key) : String.format("[res_key=\"%s\"]", key);
+    }
+
+    private static byte[] getBytesFromEnvVarOrConfigPath(String env) {
+        String envVarOrConfigPath = System.getenv(env);
+        File f = new File(envVarOrConfigPath);
+        if (f.exists()) {
+            try {
+                return Files.readAllBytes(f.toPath());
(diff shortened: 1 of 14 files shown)


automation-jmeter-plugins/src/main/java/automation/plugins/jmeter/functions/OAuth2Token.java
```diff
@@ -0,0 +1,178 @@
...
+                return Files.readAllBytes(f.toPath());
+            } catch (IOException e) {
+                throw new RuntimeException(e.getMessage());
+            }
+        }
+        return envVarOrConfigPath.getBytes();
+    }
+
+    // Cloud providers
+    enum CloudProvider {
```
ungwudik, line 136: Should this be moved to its own (separate) file?

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
```diff
@@ -0,0 +1,307 @@
...
+    private static JSR223PreProcessor jsr223PreProcessor;
+    private static ThreadGroup threadGroup;
+    private static LoopController loopController;
+    private static HTTPSamplerProxy httpSamplerProxy;
+    private static HashTree testPlanTree;
+    private static HeaderManager headerManager;
+    private static UserParameters userParameters;
+    private static BackendListener backendListener;
+
+    private static void initializeTestPlan(String testPlanName) {
```
ungwudik, line 53: Am I correct to assume that abstracting this function will be covered by the `Create a top-level test plan node for all JMeter tasks` bullet point in the PR description?
That being said, it'll be worth making notes/TODOs (possibly with JIRA tickets) within the codes that we'll need to refactor. This will help ensure that we won't lose track of anything, since this is all new.

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    private static void setHttpSampler(String requestType, String setPath) {
+        httpSamplerProxy = new HTTPSamplerProxy();
+        httpSamplerProxy.setDomain("${__env(WSM_SERVER)}");
+        httpSamplerProxy.setProperty("HTTPSampler.port", "${__env(WSM_PORT)}");
+        httpSamplerProxy.setProperty("HTTPSampler.protocol", "${__env(WSM_PROTOCOL)}");
+        httpSamplerProxy.setPath(setPath);
+        httpSamplerProxy.setMethod(requestType);
+        String body = "{" + "\"id\"" + ":" + " \"" + "${workspaceId}" + "\", " + "\"authToken\"" + ":" + " \""
```
ungwudik, line 129: We have a model for the `CreateWorkspaceRequestBody` that can be (mapped into a string and) reused here.

https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/resources/api/service_openapi.yaml#L390

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
```diff
@@ -0,0 +1,307 @@
...
+    public static void main(String[] argv) throws Exception {
+
+        // for capturing HTTP traffic
+        // System.setProperty("http.proxyHost", "127.0.0.1");
+        // System.setProperty("http.proxyPort", "8888");
+        // for capturing HTTPS traffic
+        // System.setProperty("https.proxyHost", "127.0.0.1");
+        // System.setProperty("https.proxyPort", "8888");
+
+        File jmeterHome = new File("/apache-jmeter-5.3");
```
ungwudik, line 260: The test(s) will need to be updated when the jmeter version changes in the gradle file, correct? Any way we can make it more seamless (maybe fetch via a gradle task or something)?

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
```diff
@@ -0,0 +1,307 @@
...
+        JMeterUtils.initLogging();// you can comment this line out to see extra log messages of i.e. DEBUG level
+        JMeterUtils.initLocale();
+
+        initializeTestPlan("Terra Workspace Manager Test Plan");
+        setUpLoopController();
+        setUpThreadGroup();
+        setJSR223PreProcessor();
+        setLoopController(1);
+        setThreadGroup(2, 2);
+        setHttpSampler("POST", "api/workspaces/v1");
```
ungwudik, line 277: Endpoint can be moved to a properties file accessible by all tests.

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
```diff
@@ -0,0 +1,307 @@
...
+
+        // Store execution results into a .jtl file
+        String logFile = jmeterHome + slash + "createworkspace.jtl";
+        ResultCollector logger = new ResultCollector(summer);
+        logger.setFilename(logFile);
+        testPlanTree.add(testPlanTree.getArray()[0], logger);
+
+        // Run Test Plan
+        //jMeterEngine.configure(testPlanTree);
+        //jMeterEngine.run();
```
ungwudik, line 302: Nit: did you mean to leave these lines commented out?

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
```diff
@@ -0,0 +1,307 @@
+package automation.jmeter;
```
ungwudik, line 1: I understand that there's refactor work coming in the future. However, I think it's worth separating the test itself from all the config stuff as part of this very PR.
Currently, I'm not able to fully evaluate this POC, as it's not clear to me how much work is involved when writing new tests.

terra-workspace-manager-jmeter/README.md
```diff
@@ -0,0 +1,33 @@
+# terra-workspace-manager-jmeter
```
ungwudik, line 1: Looks like this file was missed when updating the language from `master/slave` to `director/worker`

automation-jmeter-plugins/src/main/java/automation/plugins/jmeter/functions/OAuth2Token.java
ichengchang: Let's just keep the `enum` In place until one more use case needs it.

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
ichengchang: Added a TODO note as a reminder.

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
ichengchang: Is everyone using `io.swagger.codegen` for this? Is there an example from Unit or Integration tests I can reuse, otherwise I can figure this out.

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
ichengchang: The Jmeter location will be replaced by an envvar (JMETER_HOME), this applies to all environments. Added a TODO for this.

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
ichengchang: I think we can also parse the swagger.yaml (just like one of the above comments related to request body) to get the API Path using the name of the API as key (CreateWorkspace).

ichengchang: Added a TODO for this recommendatiuon.

terra-workspace-manager-jmeter/README.md
ichengchang: Renamed everything to director/worker language.

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
ichengchang: Yes these lines are commented out for now.

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
ichengchang: Let's have a separate discussion on this. I agreed, there needs to be refactoring to make writing JMeter test more code efficient.

terra-workspace-manager-jmeter/src/main/java/automation/jmeter/CreateWorkspaceSimulation.java
ungwudik: Here's an example https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java#L122-L127

I don't know that all teams are currently using Swagger codegen, but I believe the future plan is to align on it. @gjuggler may be able to speak more to that.

davidangb:
As a POC, I am generally positive on this. Since this is a POC I haven't done a detailed code review, as I expect a lot of code would change. My two big questions are:
1. What would a more complex test look like? This POC appears to only test a single API; what would it look like if we needed to test a user's path through the system (i.e. a CUJ)?
2. How could the boilerplate be reduced by moving into abstract classes, utility functions, etc.? As-is, it looks like a lot of work to create a single test.

davidangb:
@ichengchang last update on this PR was late August. Can this be closed?

PR #81, opened 2020-08-05
AS-423: Helm charts for deploying JMeter director/worker infrastructure

ichengchang:
This PR is related to PR [https://github.com/DataBiosphere/terra-workspace-manager/pull/79](https://github.com/DataBiosphere/terra-workspace-manager/pull/79). 

It concerns the set up of the JMeter director/worker infrastructure so that load tests can run from this infrastructure.

This PR is only a starting point. Specifics on cluster ownership and additional configs required to integrate with preview environments will be worked out with devOps as an ongoing process.


jmeter-director-worker/jmeter-director/Dockerfile
```diff
@@ -0,0 +1,42 @@
+# This is the Dockerfile for the JMeter Director.
+# TODO: Refactor common code from JMeter Director / Worker into a Base file.
+
+FROM openjdk:slim
```
davidangb, line 4: we're trying to standardize on adoptopenjdk; could you use that instead? https://hub.docker.com/_/adoptopenjdk

jmeter-director-worker/jmeter_demo_setup.sh
```diff
@@ -0,0 +1,85 @@
...
+    echo ""
+    echo "Below is the list of namespaces in the kubernetes cluster"
+
+    kubectl get namespaces
+    echo ""
+    echo "Please check and try again"
+    exit
+fi
+
+##Wait until Influxdb Deployment is up and running
```
davidangb, line 24: what is the purpose of Influxdb? Is it worth bringing in a new technology?

PR #82, opened 2020-08-05
[AS-406] Update stairway version to 0.0.20

zloery:
Made the following changes to be compatible with new Stairway version:

- Change Stairway constructor to use Builder pattern. Provided values are unchanged
- Places that caught StairwayExceptions now also catch InterruptedExceptions and treat them the same way
- Added a call to `recoverAndStart` as the third Stairway initialization step, as detailed in [Stairway code](https://github.com/DataBiosphere/stairway/blob/develop/src/main/java/bio/terra/stairway/Stairway.java#L244)
- Pulled `waitForFlight` into Workspace Manager instead of Stairway. The method was marked as deprecated in Stairway because it didn't work nicely with Stairway's new multi-instance functionality, but we don't use this new functionality.
- Changed polling time for stairway jobs from 10s to 1s, as all current WM Stairway jobs execute quickly.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/job/JobService.java
+++ b/src/main/java/bio/terra/workspace/service/job/JobService.java
@@ -136,7 +136,7 @@ public class JobService {
 
   void waitForJob(String jobId) {
     try {
-      int pollIntervalSeconds = 10;
+      int pollIntervalSeconds = 1;
       waitForFlight(
           jobId, pollIntervalSeconds, appConfig.getStairwayTimeoutSeconds() / pollIntervalSeconds);
     } catch (StairwayException | InterruptedException stairwayEx) {
--- a/src/main/java/bio/terra/workspace/service/job/JobService.java
+++ b/src/main/java/bio/terra/workspace/service/job/JobService.java
@@ -136,9 +136,10 @@ public class JobService {
 
   void waitForJob(String jobId) {
     try {
-      int pollIntervalSeconds = 1;
       waitForFlight(
-          jobId, pollIntervalSeconds, appConfig.getStairwayTimeoutSeconds() / pollIntervalSeconds);
+          jobId,
+          appConfig.getStairwayPollingIntervalSeconds(),
+          appConfig.getStairwayTimeoutSeconds() / appConfig.getStairwayPollingIntervalSeconds());
     } catch (StairwayException | InterruptedException stairwayEx) {
       throw new InternalStairwayException(stairwayEx);
     }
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -123,12 +136,27 @@ protected String submit(
...
+      throws DatabaseOperationException, FlightException, InterruptedException {
+    for (int pollCount = 0; pollCount < pollCycles; ++pollCount) {
+      TimeUnit.SECONDS.sleep(pollSeconds);
+      FlightState state = stairway.getFlightState(flightId);
+      if (!state.isActive()) {
+        return state;
+      }
+    }
+    throw new FlightException("Flight did not complete in the allowed wait time");
+  }
```
davidangb, line 158: could you rewrite this to use `ScheduledExecutorService` instead of `sleep`?

src/main/resources/application.properties
```diff
@@ -3,6 +3,7 @@ server.port=8080
 workspace.maxStairwayThreads=4
 workspace.resourceId=mc-terra-workspace-manager
 workspace.stairwayTimeoutSeconds=1800
+workspace.stairwayPollingIntervalSeconds=1
```
davidangb, line 6: was this previously 10? I'm wondering if polling every second is too fast.

zloery: It was previously 10s - it turns out 10s is actually too slow, and this is much closer to how long things actually take. It turns out that ~2/3 of our test runtime was waiting for this polling, and this change alone cuts tests from ~7 minutes to ~3 minutes.
That said, it may be too short for long-running operations, like creating stuff on clouds. It's probably worth revisiting then, but the performance boost now seems worth it.

src/main/java/bio/terra/workspace/service/job/JobService.java
zloery: Done. I'm not familiar with `ScheduledExecutorService` so this may not be what you had in mind - let me know if I've messed something up.

src/main/resources/application.properties
davidangb: Let's keep an eye on this. I think the best approach in the long run might be to introduce backoff into the polling strategy (maybe not exponential, maybe a linear backoff?) so we catch completion of fast flights quickly, but don't overwhelm slow flights with lots of poll requests.

And then ... polling for Stairway flight status with backoff is almost certainly reusable across services, so it should go into a shared library somewhere. Nothing necessary for this PR; let's follow up with another story.

src/main/java/bio/terra/workspace/service/job/JobService.java
davidangb: this looks right to me! I originally thought we could use `scheduleWithFixedDelay` but that doesn't return a value; I think you've done the right thing here.

src/main/resources/application.properties
zloery: Sounds good, created [AS-467] to track that. It's probably also worth adding that functionality to the template repo, which also has a JobService.

PR #84, opened 2020-08-10
[AS-371] Adopt Data Repo's snapshot naming standard for data references

zloery:
This adopts TDR's existing naming standard for data reference names in Workspace Manager. The standard is 1-63 alphanumeric characters and underscores, with the first character not an underscore. To quote TDR:

> Dataset, table, and column names follow this pattern. This should be used for the name of any object in the
> system. It enforces BigQuery naming rules except it disallows a leading underscore so we avoid collisions with
> any extra columns the DR adds. For table names, this is shorter than what BigQuery allows.

By aligning on a standard, we eliminates a source of potential bugs. I chose TDR's existing standard for snapshots because we expect that in an e2e user journey, the name WM stores for a data reference to a snapshot will match TDR's name for that snapshot and will be surfaced in a UI.

Some existing snapshots may already have invalid names. This PR will disallow accessing these references by name. We should keep that in mind until the next time we recreate DBs (which may not be long, given the pending postgres upgrade).

This shouldn't require any changes for Rawls. Because names may only include alphanumeric characters and underscores, we do not need to URL encode data reference names in requests to WM.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
+++ b/src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
@@ -8,6 +8,7 @@ import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.regex.Pattern;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.stereotype.Component;
 
 @Component
@@ -15,7 +16,7 @@ public class DataReferenceValidationUtils {
 
   private ObjectMapper objectMapper;
   private DataRepoService dataRepoService;
-  final Pattern nameValidationPattern = Pattern.compile("^[a-zA-Z0-9][_a-zA-Z0-9]*$");
+  final Pattern nameValidationPattern = Pattern.compile("^[a-zA-Z0-9][_a-zA-Z0-9]{0,62}$");
 
   public DataReferenceValidationUtils(ObjectMapper objectMapper, DataRepoService dataRepoService) {
     this.objectMapper = objectMapper;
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -7,19 +7,29 @@
...
+  final Pattern nameValidationPattern = Pattern.compile("^[a-zA-Z0-9][_a-zA-Z0-9]*$");
 
   public DataReferenceValidationUtils(ObjectMapper objectMapper, DataRepoService dataRepoService) {
     this.objectMapper = objectMapper;
     this.dataRepoService = dataRepoService;
   }
 
+  public void validateReferenceName(String name) {
+    if ((name.length() < 1 || name.length() > 63)
+        || !nameValidationPattern.matcher(name).matches()) {
```
davidangb, line 27: probably worth adding a check for null here. Even if it's not possible for name to be null in current usage, this method could be used elsewhere in the future.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -7,19 +7,29 @@
...
   private DataRepoService dataRepoService;
+  final Pattern nameValidationPattern = Pattern.compile("^[a-zA-Z0-9][_a-zA-Z0-9]*$");
 
   public DataReferenceValidationUtils(ObjectMapper objectMapper, DataRepoService dataRepoService) {
     this.objectMapper = objectMapper;
     this.dataRepoService = dataRepoService;
   }
 
+  public void validateReferenceName(String name) {
+    if ((name.length() < 1 || name.length() > 63)
```
MatthewBemis, line 26: can you also do the length check as part of the regex?

PR #85, opened 2020-08-13
Just stick with handling DRS for now [AS-391 part 2]

zarsky-broad:
Holding off on `oneOf`.

See [the doc](https://docs.google.com/document/d/1iG8tPDQ4JMld4FrQxCUYC1t4L2oRRZuoslNCtZnHTB8/edit) for context.

Goes with https://github.com/broadinstitute/rawls/pull/1271.


src/main/resources/api/service_openapi.yaml
```diff
@@ -466,8 +466,7 @@ components:
         referenceType:
           $ref: '#/components/schemas/ReferenceTypeEnum'
         reference:
-          description: The serialized object containing details specific to the referenceType. For instance, TDR snapshots must be serialized json.
-          type: string
+          $ref: '#/components/schemas/DataRepoSnapshot'
```
MatthewBemis, line 469: changing this to an actual object from a string representation of an object will impact services upstream, like rawls

zarsky-broad: Yes, but the actual transmitted JSON will be the same, so while we'll need to make that change in Rawls when we publish and are using the new client, it'll be fine on release.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -56,7 +63,13 @@ public String createDataReference(
     paramMap.put("credential_id", credentialId);
     paramMap.put("resource_id", resourceId == null ? null : resourceId.toString());
     paramMap.put("reference_type", referenceType == null ? null : referenceType.toString());
-    paramMap.put("reference", reference);
+    try {
+      paramMap.put("reference", objectMapper.writeValueAsString(reference));
+    } catch (JsonProcessingException e) {
+      // TODO: add logger and print out the reference
+      throw new InvalidDataReferenceException(
+          "Couldn't convert reference to JSON. This... shouldn't happen.");
```
davidangb, line 71: this could easily happen, since it is user-supplied input, right?

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -44,19 +45,17 @@ public String validateReference(
...
-      return ref;
-    } catch (JsonProcessingException e) {
+      DataRepoSnapshot ref, AuthenticatedUserRequest userReq) {
+    if (ref.getInstanceName() == null || ref.getSnapshot() == null) {
       throw new InvalidDataReferenceException(
-          "Input could not be parsed as a Data Repo snapshot"
-              + " reference. The only valid fields are instanceName and snapshot.");
+          "Invalid Data Repo Snapshot identifier: "
+              + "instanceName and snapshot must both be provided.");
     }
```
davidangb, line 53: maybe use `StringUtils.isEmpty()` here to also look for the empty string

src/main/resources/api/service_openapi.yaml
davidangb: is it true that the transmitted value is the same? I thought I saw some string-encoding going on somewhere. I am not certain of this.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
zarsky-broad: It shouldn't, we should already have validated it and checked that the reference exists by calling `validateReference`.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zarsky-broad: I'm gonna go with `isBlank` because it will also catch whitespace-only

src/main/resources/api/service_openapi.yaml
zarsky-broad: @zloery any insight?

zloery: This actually made me curious so I went and checked - DA is right, this does change the transmitted value (and it's mutually unintelligible between the two versions).

A call to the existing system looks like
`curl -X POST "https://workspace.dsde-dev.broadinstitute.org/api/workspaces/v1/3fa85f64-5717-4562-b3fc-2c963f66afa6/datareferences" -H  "accept: */*" -H  "Authorization: Bearer ___" -H  "Content-Type: application/json" -d "{\"name\":\"snapshotName\",\"referenceType\":\"DATA_REPO_SNAPSHOT\",\"reference\":\"{\\\"instanceName\\\": \\\"terra-dev\\\", \\\"snapshot\\\": \\\"93dc1e76-8f1c-4949-8f9b-07a087f3ce7b\\\"}\",\"cloningInstructions\":\"COPY_NOTHING\"}"`

A call to the new system looks like
`curl -X POST "http://localhost:8080/api/workspaces/v1/3fa85f64-5717-4562-b3fc-2c963f66afa6/datareferences" -H  "accept: */*" -H  "Authorization: Bearer ___" -H  "Content-Type: application/json" -d "{\"name\":\"snapshotName\",\"referenceType\":\"DATA_REPO_SNAPSHOT\",\"reference\":{\"instanceName\":\"terra-dev\",\"snapshot\":\"93dc1e76-8f1c-4949-8f9b-07a087f3ce7b\"},\"cloningInstructions\":\"COPY_NOTHING\"}"`

The string encoding on the snapshot actually does change, so we'll have to update Rawls after all

PR #86, opened 2020-08-14
update deps and address codacy findings

zarsky-broad:


zarsky-broad:
Yup. I turned off the markdown linter.

PR #87, opened 2020-08-14
[AS-381] Add integration test for creating a reference to a Data Repo Snapshot

zloery:
This adds an integration test which covers the flow of adding a reference to a Data Repo snapshot to a workspace. It does not create the referenced snapshot or dataset in Data Repo, and instead assumes that it persists as part of the dev environment. This pattern will need to be revisited if we ever test against non-dev environments of Data Repo.

### DIFF
--- a/src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
+++ b/src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
@@ -166,6 +166,13 @@ public class WorkspaceIntegrationTest {
   @Test
   @Tag(TAG_NEEDS_CLEANUP)
   public void createDataReference(TestInfo testInfo) throws Exception {
+    // This test relies on a persistent snapshot existing in Data Repo, currently in dev.
+    // This snapshot was created using our dev service account, which is a steward in dev Data Repo.
+    // First, I created a dataset "wm_integration_test_dataset" using TDR's
+    // snapshot-test-dataset.json. Then, I created the snapshot "workspace_integration_test_snapshot"
+    // using the "byFullView" mode. Finally, I added the integration test user as a reader of this snapshot.
+    // These steps should only need to be repeated if the dev DataRepo data is deleted, or to support
+    // this test in other DataRepo environments.
     UUID workspaceId = UUID.randomUUID();
     testToWorkspaceIdsMap.put(testInfo.getDisplayName(), Collections.singletonList(workspaceId));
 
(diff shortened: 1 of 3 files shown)


src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
```diff
@@ -158,6 +163,49 @@ public void deleteWorkspaceWithInvalidToken(TestInfo testInfo) throws Exception
     Assertions.assertTrue(deleteWorkspaceResponse.isErrorObject());
   }
 
+  @Test
+  @Tag(TAG_NEEDS_CLEANUP)
+  public void createDataReference(TestInfo testInfo) throws Exception {
+    // This test relies on a persistent snapshot existing in Data Repo, currently in dev.
+    // This snapshot was created using our dev service account, which is a steward in dev Data Repo.
```
ungwudik, line 170: Is there any blocker on creating the dataset/snapshot as part of this test code? I think tests should self-contain its resources whenever possible.
Also, I assume testing against non-dev environments of Data Repo is probably not a matter of `if`, but `when`?

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
```diff
@@ -158,6 +163,49 @@ public void deleteWorkspaceWithInvalidToken(TestInfo testInfo) throws Exception
...
+
+    createDefaultWorkspace(workspaceId);
+
+    String userEmail = testConfig.getServiceAccountEmail();
+    String path = testConfig.getWsmWorkspacesBaseUrl() + "/" + workspaceId + "/datareferences";
+
+    DataRepoSnapshot snapshotReference =
+        new DataRepoSnapshot()
+            .snapshot("97b5559a-2f8f-4df3-89ae-5a249173ee0c")
+            .instanceName("terra-dev");
```
ungwudik, line 187: This can be moved into the properties file https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/test/resources/application-integration-test.properties

 and consumed/configured here https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/test/java/bio/terra/workspace/integration/common/configuration/TestConfiguration.java

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
zloery: I don't think there's anything blocking us from creating the dataset as part of the test. That said, that's a lot of extra setup and teardown functionality that isn't really part of WM, and I'm a little concerned that there's a lot of room for test flakiness there. On the other hand, as implemented the test will start failing every time TDR clears their dev database (which should be rarely, but still). I'd need to think about this a bit

re: your second point, I'm not actually sure. At some point we'll definitely test against non-dev Data Repos, but will we be running these integration tests or something else?

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
zloery: Good point, that's a pretty clean way of handling the environment-specific bits. I'll move the instance name there now. The snapshot name would also go there if we use the persistent dev snapshot, but not if we create a dataset with each test run.

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
zloery: Coming back to this, I talked to Dan from TDR about this and he recommended using a persistent snapshot. Their dev environment is fairly stable and they make a reasonable effort to preserve the data, so failures due to deleted resources should be rare. Given that it removes a lot of setup/teardown complexity from this test (especially on functionality that isn't used elsewhere in WM), I'd prefer not to create a new snapshot each time we run the test.

ungwudik: 1. In that case, the current strategy works for me. I'd be interesting in learning what those extra setup/teardown functionality is for creating the dataset/snapshot, though.

2. Also, I'm wondering - after creating a data reference of a snapshot, are there any changes in the snapshot itself that we may need to undo/cleanup after the tests runs? Or does deleting the workspace which contains the data reference take care of any potential (_I don't know of any_) buildup in the snapshot?

3. To answer your question - Yes, I expect we'll be running these integration tests in other environments in the future.

ungwudik: Also, would you mind adding the rationale above (i.e. `…Their dev environment is fairly stable and they make a reasonable effort to preserve the data…`) to the existing comments at the beginning of this `createDataReference` test?

ungwudik:
I added a few follow-up comments, but the failing unit test is what's blocking me from approving.

zloery:
It looked like the failure was unrelated, and re-running the tests fixed it. I'm looking into the flaky test, but it shouldn't be a problem for this PR

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
zloery: 1. Creating the snapshot is three calls to TDR as the dev service account: creating a dataset (using [this schema](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/test/resources/snapshot-test-dataset.json)), creating a snapshot on that dataset with the right settings, and then granting access to william.thunderlord (since we're running the TDR calls as the dev service account, rather than as william.thunderlord).
Cleaning up would be two delete calls, one for the snapshot and one for the dataset.
It's not super complex, but it would slow the test down and introduce a lot more surface area for errors that we shouldn't really be testing for

2. Snapshots in TDR are totally immutable(*), and WM's calls are all read-only.
(*) I think there's an exception for when patients request their data be removed and possibly another case, but from WM's perspective we should consider them immutable.

3. Nice, thanks! Updated comment

PR #92, opened 2020-08-24
fix overlooked test...

zarsky-broad:


ungwudik:
Looks straight-forward enough, but I'm not sure what "overlooked" means here. Was it failing? Removing unnecessary code? Something else?

zarsky-broad:
It was failing; I somehow didn't catch that it needed to be updated with the last PR.

PR #93, opened 2020-08-27
AS-466: A reusable JMeter plugin for MDC log tracing

ichengchang:
This function can be used by any Java JMeter code for MDC log tracing.

### DIFF
--- a/jmeter-plugins/src/main/java/plugins/jmeter/functions/XCorrelationIDHeader.java
+++ b/jmeter-plugins/src/main/java/plugins/jmeter/functions/XCorrelationIDHeader.java
@@ -29,6 +29,7 @@ public class XCorrelationIDHeader extends AbstractFunction {
     private static final int MAX_PARAMETER_COUNT = 2;
 
     static {
+        desc.add(getResString("x_correlation_id_prefix")); //$NON-NLS-1$
         desc.add(getResString("function_name_paropt")); //$NON-NLS-1$
     }
 
--- a/jmeter-plugins/src/main/java/plugins/jmeter/functions/XCorrelationIDHeader.java
+++ b/jmeter-plugins/src/main/java/plugins/jmeter/functions/XCorrelationIDHeader.java
@@ -1,0 +1,95 @@
+package plugins.jmeter.functions;
+
+import java.util.Collection;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.ResourceBundle;
+import java.util.concurrent.ThreadLocalRandom;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.jmeter.engine.util.CompoundVariable;
+import org.apache.jmeter.functions.AbstractFunction;
+import org.apache.jmeter.functions.InvalidVariableException;
+import org.apache.jmeter.samplers.SampleResult;
+import org.apache.jmeter.samplers.Sampler;
+import org.apache.jmeter.threads.JMeterVariables;
+import org.hashids.Hashids;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class XCorrelationIDHeader extends AbstractFunction {
+    private Hashids hashids = new Hashids("XCorrelatiomnIDSalt", 8);
+    private static final String MDC_CORRELATION_ID_HEADER = "X-Correlation-ID";
+    private static final Logger log = LoggerFactory.getLogger(XCorrelationIDHeader.class);
+    private static final List<String> desc = new LinkedList<>();
+
+    private static final String KEY = "__XCorrelationIDHeader"; //$NON-NLS-1$
+
+    // Number of parameters expected - used to reject invalid calls
+    private static final int MIN_PARAMETER_COUNT = 1;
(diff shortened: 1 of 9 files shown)


jmeter-plugins/src/main/java/plugins/jmeter/functions/XCorrelationIDHeader.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.jmeter.functions.InvalidVariableException;
+import org.apache.jmeter.samplers.SampleResult;
+import org.apache.jmeter.samplers.Sampler;
+import org.apache.jmeter.threads.JMeterVariables;
+import org.hashids.Hashids;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class XCorrelationIDHeader extends AbstractFunction {
+    private Hashids hashids = new Hashids("XCorrelatiomnIDSalt", 8);
```
ungwudik, line 20: Nit: typo `XCorrelatiomnIDSalt `

ungwudik:
Looks good to me. I imagine Anu and/or Isaac are in a better position to approve.

jmeter-plugins/src/main/java/plugins/jmeter/functions/XCorrelationIDHeader.java
```diff
@@ -0,0 +1,96 @@
...
+        StringBuilder header = new StringBuilder();
+        header.append(MDC_CORRELATION_ID_HEADER);
+        header.append(": ");
+        String correlationId = generateCorrelationId();
+        String prefix = ((CompoundVariable) values[0]).execute().trim();
+        if (prefix.length() > 0) {
+            header.append(prefix);
+            header.append("-");
+        }
+        header.append(correlationId);
```
davidangb, line 58: it feels awkward that this code is only working with a delimited string. Instead, could you create a `org.apache.jmeter.protocol.http.control.Header` object, which keeps the header name/value in separate member variables? You could then use `vars.putObject` to store the `Header` as a variable, and finally you could return `header.toString()`. Otherwise, you are requiring any header managers to retrieve the variable as a String, parse the string (fragile!), and create their own `Header` objects.

Note that you'll have to include `ApacheJMeter_http` as a build dependency to have access to the `Header` class.

Alternately, just have this function be responsible for generating a unique id and optionally prefixing that id. It doesn't have to know anything about the `X-Correlation-ID` header, and in fact it doesn't care where the unique id is used. It just generates an optionally-prefixed unique id, and it is up to any callers to use that id wherever they please. This could make it useful outside of MDC, too.

jmeter-plugins/src/main/resources/messages.properties
```diff
@@ -0,0 +1,2 @@
+x_correlation_id_prefix=A custom prefix for X-Correlation-ID
+function_name_paropt=Name of variable in which to store the result (optional)
```
davidangb, line 2: what is a "function_name_paropt"? I think this would be clearer as "variable_name", or something else more descriptive.

davidangb:
One more question: what is big-picture goal of this code? Is the goal to be able to correlate the logs generated by the perf test clients with the logs generated by the services being tested? If so, the test code itself will need access to these IDs and will need to be modified to log these IDs.

jmeter-plugins/src/main/java/plugins/jmeter/functions/XCorrelationIDHeader.java
ichengchang: I like the alternative approach. We will just implement the random correlation id so users can just use this Jmeter function to get the id without worrying about the underlying implementation.

jmeter-plugins/src/main/resources/messages.properties
ichengchang: The "function_name_paropt" is used throughout the JMeter function implementation, since it is already defined as a resource. So I keep the name consistent with JMeter way of naming the var. It's am awkward name I agree but I prefer to stick with the convention.

jmeter-plugins/src/main/java/plugins/jmeter/functions/XCorrelationIDHeader.java
ichengchang: Fixed.

ichengchang:
> Some comments inline. In addition to those:
> 
> 1. If this is meant to be a reusable plugin, it should not live in the workspace manager repo; that's too tightly coupled.
> 2. Instead of building a jar to a bucket, and requiring users to download from that bucket, can this publish to an artifact repository? We have Artifactory available; I'm ok if we use Google or Github package repos if those are better.

The plugin will be moved to our new repo jmeter-tools. Isaac and I will work on the publish strategy, yes Artifactory will be more appropriate.

ichengchang:
> One more question: what is big-picture goal of this code? Is the goal to be able to correlate the logs generated by the perf test clients with the logs generated by the services being tested? If so, the test code itself will need access to these IDs and will need to be modified to log these IDs.

That's where the prefix comes in. The test client can prefix the ID with metadata that helps identify the test runs, e.g. prefix by timestamp or some consistent naming convention.

PR #95, opened 2020-09-16
[AS-470] Add alpha and staging data repos as valid instances

zloery:
Adds `terra-alpha` and `terra-staging` as valid Data Repo instances.

All WM requests to reference snapshots from data repo specify which instance they are calling. Because we currently have no tests or calls that run on the alpha or staging instances of Workspace Manager, this is the only change needed for AS-470.


davidangb (approved):
it hurts my brain how inconsistent the data repo urls are, but that is out of our control! :+1:

PR #97, opened 2020-09-21
Expose 'critical' field of subsystem status

zloery:
This change allows callers to see which subsystems of WM are considered "critical" and which are not. Workspace Manager's overall status is defined as ok if all critical subsystems are ok. This change was brought up at a sprint review several months back, but never made it to a ticket.

Because this is an unauthenticated endpoint, I'm also going to explore using the new preview environments for this.


zloery:
preview-create

MatthewBemis (approved):

zloery:
preview-delete

PR #99, opened 2020-09-23
Overhaul, clarify, and parameterize config [AS-396]

zarsky-broad:
Before you look at this, note that it makes a distinction between two classes of config:
- Static, which may be _env-specific_, but can be specified here in the repo
- Dynamic, which may either be a secret whose value could rotate or _instance-specific_ (not just env) and definitely need to be provided at runtime

With that in mind, here are the parts of this PR:
1. **YAML:** all .properties files have been converted to .yaml. The exception to this is the autogenerated `version.properties`, because it's not designed to be maintained by humans.
2. **Namespacing:** to match [terra-resource-janitor](https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/main/resources/application.yml), all config keys that are specific to the app have been namespaced under `workspace`.
3. ~**Profiles:** different environments are represented in the config as different profiles, and the active profile is selected based on the value of the `TERRA_ENV` env variable. The default is dev. The config files have been moved into a `config` subfolder in `resources`, since there are now multiple.~
4. **Local development:** the config sets defaults for all values based on what we use for local development, so exporting variables is no longer required when running locally.
    - Option: local could be its own profile, if we don't like setting defaults.
5. **More judicious/explicit use of env:** all env variables used by the config are now pulled in at one place, at the top of `application.yml`. There may be some more things that aren't secret per-env (like database name/username) that could be put into the profile config.

If we like this, I'll make PR's for the helm and helmfile repos adding `TERRA_ENV` that would merge before this, and then a second set of PR's deleting all the old values (and for fc-develop to change the Rawls key for the data repo), and finally we would delete the keyed data repo instances in `application.yml`.

Related PR's:
https://github.com/broadinstitute/terra-helm/pull/145
https://github.com/broadinstitute/terra-helmfile/pull/269


davidangb:
> Profiles: different environments are represented in the config as different profiles, and the active profile is selected based on the value of the TERRA_ENV env variable

@zarsky-broad does this mean that production secrets are rendered and deployed (but unused) in non-prod environments? If so, that's a non-starter for security reasons

zarsky-broad:
Ah, secrets are still provided by env variables.

src/main/resources/config/application.yml
```diff
@@ -0,0 +1,67 @@
...
+  resources:
+    cache:
+      cachecontrol:
+        max-age: 0
+        must-revalidate: true
+    static-locations: classpath:/api/
+  sleuth.sampler.probability: 0.1
+
+workspace:
+  datarepo.instances:
```
gjuggler, line 39: I think it's important for the main application.yml to have entries for all possible application-specific configs. This file doesn't have an entry for workspace.datarepo.instances.terra, but that value does show up in the env-specific overrides. Two questions / suggestions:

1) For this PR, can you add an entry for workspace.datarepo.instances.terra = 'jade.datarepo-dev'? This would follow your pattern of "the base application.yml has values we use for local testing against dev".

2) Would there be a longer-term plan to move entirely away from having all of these values hard-coded in the main application.yml, and have each WSM server only have a single TDR server it's configured to point to?

src/main/resources/config/application.yml
```diff
@@ -0,0 +1,67 @@
...
+      uri: ${env.db.host}/${env.db.stairway.name}
+      username: ${env.db.stairway.user}
+    workspace:
+      initializeOnStart: ${env.db.init}
+      password: ${env.db.ws.pass}
+      upgradeOnStart: true
+      uri: ${env.db.host}/${env.db.ws.name}
+      username: ${env.db.ws.user}
+
+  maxStairwayThreads: 4
```
gjuggler, line 59: In Janitor, the Stairway-specific stuff is nested beneath a 'janitor.stairway' namespace. May make sense here, where you've got 3 fields that have "stairway" in their name. Something like:

```
workspace:
  stairway:
    max-threads: 4
    polling-interval-seconds: 1
    timeout-seconds: 1800
```

src/main/resources/config/application.yml
```diff
@@ -0,0 +1,67 @@
...
+  resourceId: mc-terra-workspace-manager
+  stairwayPollingIntervalSeconds: 1
+  stairwayTimeoutSeconds: 1800
+
+  # Cache sub-system status results for 1m by default
+  status-check-frequency.in.milliseconds: 60000
```
gjuggler, line 65: nit: the "in" as a config namespace is entirely meaningless here. One convention we might aim for is a {service}.server namespace which holds service-specific configs like this. Something like:

```
workspace:
  server:
    status-check:
      frequency-in-ms: 60000
      staleness-threshold-ms: 60000
```

src/main/resources/config/application.yml
```diff
@@ -0,0 +1,67 @@
...
+workspace:
+  datarepo.instances:
+    terra-alpha: https://data.alpha.envs-terra.bio
+    terra-dev: https://jade.datarepo-dev.broadinstitute.org
+    terra-prod: https://jade-terra.datarepo-prod.broadinstitute.org
+    terra-staging: https://data.staging.envs-terra.bio
+
+  db:
+    stairway:
+      forceClean: true
```
gjuggler, line 47: nit: prefer using kebab-case everywhere in this file for self-consistency.

src/test/resources/config/application.yml
```diff
@@ -0,0 +1,35 @@
+# Because the @TestPropertySource annotation doesn't support loading yaml files, this
```
gjuggler, line 1: What I set up for Janitor was to have test-specific overrides to the default application.yaml, rather than specifying an entirely distinct application.yml. This reduced the duplication of config values across the main & test source directories.

See https://github.com/DataBiosphere/terra-resource-janitor/tree/master/src/test/resources for the concrete examples.

What this does require is that all unit & integration tests specify a known @profile annotation on the test class; this was made easier by creating a base unit and integration test class to unify the annotations, e.g. https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/test/java/bio/terra/janitor/common/BaseUnitTest.java

See also the section in my configs doc that tries to identify some patterns and conventions around test configs: https://docs.google.com/document/d/1VfgL8D6PpJg7irbL_gKnmyZ5_GH5OYGfK8GPN45odUs/edit#heading=h.hek1a2xrbqo8

I'd like for that document to eventually be a reference point for new MC Terra services (and likely get baked back into the POC repo). Let me know if you feel anything there is unwarranted or goes against the direction you're taking here!

src/main/java/bio/terra/workspace/app/configuration/DataRepoConfig.java
```diff
@@ -7,7 +7,7 @@
 
 @Configuration
 @EnableConfigurationProperties
-@ConfigurationProperties(prefix = "datarepo")
+@ConfigurationProperties(prefix = "workspace.datarepo")
 public class DataRepoConfig {
   private HashMap<String, String> instances;
```
gjuggler, line 12: [aside / nit] I don't love the idea that we have an untyped key-value hash being specified in config like this. It just generally goes against the grain of our pattern "configs are mapped to FooConfig classes with typed value bindings".

At the very least – one convention I'd like to align on is that the configuration Java class is the canonical place where config values are documented (see [this section](https://docs.google.com/document/d/1VfgL8D6PpJg7irbL_gKnmyZ5_GH5OYGfK8GPN45odUs/edit#heading=h.83dycur23l5f)). Maybe doesn't need to be in this PR, but it would be great to see a descriptive comment added to the field definition for every value that's getting bound to a Spring app config.

src/main/resources/config/application.yml
```diff
@@ -0,0 +1,67 @@
+# All env variables that are used in one place
+# Exception: TERRA_ENV (which sets the active profile), which is parsed before this is.
+env:
```
gjuggler, line 3: [aside] I'm still coming to terms with the idea of this "env" style of mapping env variables to Spring app config. It's attractive in some regards, but I think it also has some potential to create more confusion.

Concrete example: with this PR, there are now 3 totally different ways of referring to the "config value that tells WSM whether to initialize its database on startup":
1) `INIT_DB` env variable
2) `env.db.init` Spring config
3) `workspace.db.workspace.initializeOnStart` Spring config

IMO, we should eventually aim for there to be one canonical path for each config value, no matter where it's getting specified.

I don't think this is any reason to hold off on this PR. But I'd just note that I suspect we may eventually seek a solution that doesn't require a separate 'env' stanza like this.

src/main/resources/config/application.yml
```diff
@@ -0,0 +1,67 @@
...
+    terra-staging: https://data.staging.envs-terra.bio
+
+  db:
+    stairway:
+      forceClean: true
+      migrateUpgrade: ${env.db.init}
+      password: ${env.db.stairway.pass}
+      uri: ${env.db.host}/${env.db.stairway.name}
+      username: ${env.db.stairway.user}
+    workspace:
```
gjuggler, line 52: [nit / opt] The config path here, `workspace.db.workspace` feels a bit awkward due to repetition.

What I did for Janitor was to nest the Janitor-specific DB config directly under the Janitor namespace (e.g. `janitor.db`), and put the Stairway DB stuff under `janitor.stairway.db`. It flips around the hierarchy a bit but felt reasonable enough to me. ([code ref](https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/main/resources/application.yml#L2))

gjuggler (approved):
Thanks for pushing on this Isaac! I dropped a handful of notes / comments / nitpicks. Feel free to take or leave any of them as you see fit.

src/main/resources/config/application.yml
zarsky-broad: The goal is to remove the datarepo.instances stanza from the main config entirely, and have it in the profiles only; this is an in-between state that would be resolved by the end of the chain of PR's that this would be part of.

zarsky-broad: The end state for this is the same as the current state of `workspace.sam.basePath`

src/main/java/bio/terra/workspace/app/configuration/DataRepoConfig.java
davidangb: I'm curious what would be better than a `HashMap<String, String>` in this particular instance. We could potentially make this a little more strongly typed as `HashMap<String, URL>` and/or we could create a bespoke model class to represent the keys, something like `HashMap<DataRepoName, URL>`. Would that be better?

For this particular config value, we use a name to refer to each data repo instance, to allow the repo instance URLs to change without having to change any persisted data that references that instance. We *could* hardcode the keys (i.e. the names), but that also feels wrong.

Totally agreed that a code comment would help.

src/test/resources/config/application-integration-test.yml
```diff
@@ -0,0 +1,12 @@
...
+    dev: terra
+  dataRepoSnapshotId:
+    dev: 97b5559a-2f8f-4df3-89ae-5a249173ee0c
+  serviceAccountEmail: william.thunderlord@test.firecloud.org
+  serviceAccountFilePath: rendered/service-account.json
+  wsmEndpoints:
+    workspaces: api/workspaces/v1
+  wsmUrls:
+    dev: https://workspace.dsde-dev.broadinstitute.org/
+#    dev: http://localhost:8080/
```
davidangb, line 12: nit: delete if not used

davidangb (commented):
one nit inline. I'm generally in favor of this PR and will re-review once Greg's comments are addressed!

src/main/java/bio/terra/workspace/app/configuration/DataRepoConfig.java
zarsky-broad: What if we add metadata? https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-configuration-metadata.htm

zarsky-broad: Also, I'm going to adhere to that doc you linked Greg, very nice

gjuggler: I guess I was more wondering aloud about why this property needs to support arbitrary keys. I can think of ~two scenarios:

(1) Once we can specify each per-environment TDR target more easily (which this PR starts to enable via profiles), we no longer need to specify additional named TDR targets, e.g. there's only one. So this can become a singular property `workspace.datarepo.instance`.

(2) We still want to support differing numbers / sets of TDR targets. If this is the case, I would question whether a simple key-value will be flexible enough in the long term. For example, is it plausible that we might need to track more complex metadata about each target TDR? I might be inclined to treat this config property as a single JSON-serialized string that maps to a list of objects with a known POJO structure. For example:

```
--- 
workspace: 
  datarepo: 
    # Instances is a multiline string that serializes to a list of 
    instances: |-
        [
         {"name": "azure", "url": "https://data.terra-azure.bio", "apiVersion": 2}
        ]
    primary-instance: "https://data.terra.bio"
```

src/test/resources/config/application-integration-test.yml
```diff
@@ -0,0 +1,12 @@
+workspace.integration-test:
+  data-repo-instance-names:
+    dev: terra
```
ungwudik, line 3: I expected the value here to be `terra-dev, unless this is a welcome change?

zarsky-broad: Yeah, there will be some other PR's that land before this, the goal is that each environment's data repo instance is just called `terra`.

src/test/resources/config/application-integration-test.yml
zarsky-broad: Ah, we had this before. This is so that it's easy to toggle the comment on these two lines locally when you want to run integration tests against local, since intellij doesn't expose an easy way to override an environment variable.

davidangb: add a code comment to that effect then? e.g.
```
    dev: https://workspace.dsde-dev.broadinstitute.org/
    # toggle to the following value to run integration tests against the local environment
    # dev: http://localhost:8080/
```

src/main/resources/application.yml
```diff
@@ -0,0 +1,80 @@
...
+    username: ${env.db.stairway.user}
+
+  status-check:
+    frequency-ms: 60000
+    staleness-threshold-ms: 600000
+
+  workspace-database:
+    initialize-on-start: ${env.db.init}
+    password: ${env.db.ws.pass}
+    upgrade-on-start: true
```
davidangb, line 78: we may eventually want this to be a variable. No need to change as part of this PR.

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -35,26 +35,15 @@
...
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.MvcResult;
 
-@Tag("unit")
-@ExtendWith(SpringExtension.class)
-@ContextConfiguration(classes = Main.class)
-@SpringBootTest
-@AutoConfigureMockMvc
-public class WorkspaceServiceTest {
+public class WorkspaceServiceTest extends BaseUnitTest {
```
davidangb, line 46: was the `BaseUnitTest`  / `BaseIntegrationTest` refactor driven by any of the config changes in this PR?

davidangb (approved):
:+1: lgtm; I did not test in any way!

src/main/resources/application.yml
```diff
@@ -0,0 +1,80 @@
...
+    host: jdbc:postgresql://${DATABASE_HOSTNAME:127.0.0.1}:5432
+    init: ${INIT_DB:false}
+    stairway:
+      name: ${STAIRWAY_DATABASE_NAME:stairwaylib}
+      pass: ${STAIRWAY_DATABASE_USER_PASSWORD:stairwaypwd}
+      user: ${STAIRWAY_DATABASE_USER:stairwayuser}
+    ws:
+      name: ${STAIRWAY_DATABASE_NAME:wm}
+      pass: ${STAIRWAY_DATABASE_USER_PASSWORD:wmpwd}
+      user: ${STAIRWAY_DATABASE_USER:wmuser}
```
zloery, line 14: This looks like the wrong env vars are getting pulled in

zloery (approved):
LGTM pending 1 fix

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
zarsky-broad: Since the test now require profiles, this seemed like a good time to make this abstraction.

PR #101, opened 2020-09-30
Cleanup after config overhaul [post- AS-396]

zarsky-broad:
Depends on https://github.com/broadinstitute/firecloud-develop/pull/2295

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -48,10 +48,6 @@ spring:
 workspace:
   datarepo.instances:
     terra: ${env.urls.terra-datarepo}
-    terra-alpha: https://data.alpha.envs-terra.bio
-    terra-dev: https://jade.datarepo-dev.broadinstitute.org
-    terra-prod: https://jade-terra.datarepo-prod.broadinstitute.org
-    terra-staging: https://data.staging.envs-terra.bio
 
   job:
     max-threads: 4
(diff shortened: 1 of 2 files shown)


davidangb (approved):
:+1: code looks fine. Does this change overall mean that any pre-existing snapshot references - which probably used `terra-dev` - are now broken? If so, make sure to communicate that to any dev team that might have pre-existing references

zarsky-broad:
> 👍 code looks fine. Does this change overall mean that any pre-existing snapshot references - which probably used `terra-dev` - are now broken? If so, make sure to communicate that to any dev team that might have pre-existing references

I'm not sure! @zloery, any insight?

zloery:
> > 👍 code looks fine. Does this change overall mean that any pre-existing snapshot references - which probably used `terra-dev` - are now broken? If so, make sure to communicate that to any dev team that might have pre-existing references
> 
> I'm not sure! @zloery, any insight?

It won't break the DB since we store entire serialized references. I don't actually know if it breaks existing snapshots - does Rawls verify the data repo instance when it looks up snapshots in WM? 
If not, existing snapshots will just have a different value, and it won't cause problems as it's not used after reference creation.

zarsky-broad:
Looks like this will break things, yeah. @davidangb I'm going to go ahead and merge this, since the deed is already done, but any idea who I should give a heads up to?

PR #102, opened 2020-10-01
Disable integration tests that talk to Data Repo

ungwudik:
Data Repo dev environment is having stability issues. Jade team needs to about two weeks to look into fixing it, so disabling automated integration tests that involve Data Repo. 

I've created a small task AS-506 to follow up on re-enabling the tests.


zarsky-broad (commented):
Why not just use `@Disabled`? https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Disabled.html

ungwudik:
> Why not just use `@Disabled`? https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Disabled.html

I was too quick to think "Let's not start the tests at all." But yes, @Disabled is cleaner. Thanks!

zarsky-broad (approved):

PR #103, opened 2020-10-01
update deps

zarsky-broad:
Read changelogs and ran tests; nothing of particular interest changed, except that there's now an option in the swagger ui to persist authorization between page loads. There's a fix for a regression on how postgres numbers binary array indices, but we don't read/write any arrays from the database.


MatthewBemis (approved):
thanks for taking care of these updates as often as you do

PR #104, opened 2020-10-01
Update logger to use spring cloud json format

zloery:
Switches WM logs to use Spring's Stackdriver JSON log format, per Terra tech stack doc. This change makes the logs work with Stackdriver and GCP Error Reporting better.

The change to GlobelExceptionHandler isn't required, but makes log messages easier to understand by condensing multiple log entries into a single JSON blob.


build.gradle
```diff
@@ -68,6 +68,7 @@ dependencies {
 	implementation group: "org.hashids", name: "hashids", version: "1.0.3"
 	implementation group: "org.liquibase", name: "liquibase-core", version: "4.1.0"
 	implementation group: "org.springframework.cloud", name: "spring-cloud-gcp-starter-trace", version: "1.2.5.RELEASE"
+  implementation group: "org.springframework.cloud", name: "spring-cloud-gcp-starter-logging", version: "1.2.5.RELEASE"
```
zarsky-broad, line 71: Nit: indentation, alphabetize

zarsky-broad (commented):
So, sorry, I'm confused, how does this mix with opencensus in #111? Or it's log specific, nothing to do with traces...?

build.gradle
zloery: > So, sorry, I'm confused, how does this mix with opencensus in #111? Or it's log specific, nothing to do with traces...?

This is log specific and shouldn't affect traces at all (though yeah, the two libraries have pretty similar names...). This is updating WM's log format to let us use things like Error Reporting [[AS-449]](https://broadworkbench.atlassian.net/browse/AS-449) and generally to have more structured logs

zarsky-broad:
Could we do the config [in java](https://github.com/qos-ch/logback-extensions/wiki/Spring#using-spring-java-config)?

zloery:
The groovy file is an exact replacement for the `logback.xml` logback file that page mentions. Logback fully supports both XML and groovy configs, and I semi-arbitrarily picked groovy because TDR uses it and I don't like XML. 

I think that config still requires a logback file, it just allows further customization inside Spring.

build.gradle
```diff
@@ -93,6 +94,8 @@ dependencies {
 	testImplementation project(":workspace-manager-client")
 
 	annotationProcessor group: "org.springframework.boot", name: "spring-boot-configuration-processor"
+	// Used only for logback configuration file.
+	runtime group: 'org.codehaus.groovy', name: 'groovy', version: '2.5.7'
```
zarsky-broad, line 98: This should probably go with the normal deps, so above postgres. It's icky to me that we have to bring in something this big just to parse a config, but I don't really see an alternative, since we really don't like xml.

zarsky-broad (approved):

davidangb (approved):
:+1:, assuming you've looked in SD to make sure logs are landing there properly!

PR #107, opened 2020-10-08
Properly use the JdbcTemplate bean

zloery:
Inspired by comments in https://github.com/DataBiosphere/jade-data-repo/pull/641

Currently, we create a NamedParameterJdbcTemplate bean and never use it. Instead, our code re-creates the NamedParameterJdbcTemplate from the same input in several places in the code. This is unnecessary duplication, and we should use the common bean throughout the code.


zarsky-broad (approved):
Didn't test, but since the github tests passed this clearly works. Looks good to me!

PR #108, opened 2020-10-08
Remove unused Sam resource types/actions

zloery:
Will's investigation into Workspace creation authz found a few Sam constants that aren't used (and aren't valid in Sam), as well as a "steward" check we forgot to remove. The check is never executed as we've temporarily disabled the `deleteJob` endpoint, but we should still remove it.

### DIFF
--- a/src/main/java/bio/terra/workspace/common/utils/SamUtils.java
+++ b/src/main/java/bio/terra/workspace/common/utils/SamUtils.java
@@ -3,10 +3,6 @@ package bio.terra.workspace.common.utils;
 public class SamUtils {
 
   public static String SAM_WORKSPACE_RESOURCE = "workspace";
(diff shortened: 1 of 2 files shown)


zloery:
Related question for @MatthewBemis - back in February we created the `mc-workspace` resource type in Sam, but ultimately decided to use the existing workspace resource type for compatibility/migration reasons. Does anything else use the `mc-workspace` type, and if not can we delete it from Sam?

MatthewBemis:
@zloery nothing else was using the mc-workspace resource type so it should be all clear to delete it from Sam

zloery:
ty, created https://github.com/broadinstitute/sam/pull/463 for that

MatthewBemis (approved):

PR #110, opened 2020-10-13
PF-152 Add initial Flight & Steps for Project creation.

wchamber:
Add initial flight and steps to create a Google Project as the cloud context for a workspace.
Add new connected test type. Connected tests can talk to WM dependencies (e.g. Google Cloud), but are not exercised through only the WM public API (e.g. integration tests). Add gradle/github action/base test class for connected tests.

Still TODOs for deleting the project/context on workspace deletion, separate flight for deleting the project/context, as well as APIs and plumbing to make this flight used.


src/main/java/bio/terra/workspace/service/workspace/flight/GenerateProjectIdStep.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    return StepResult.getStepResultSuccess();
+  }
+
+  /** Generate a pseudo-random project id. */
+  public static String randomProjectId() {
+    return "wm-" + Long.valueOf(UUID.randomUUID().getMostSignificantBits()).toString();
+  }
```
davidangb, line 35: fwiw, we're using hashids in WSM for generating MDC ids: https://hashids.org/java/ ... you might adopt that or one of its alternatives for random-string creation

src/main/java/bio/terra/workspace/service/workspace/flight/CreateProjectStep.java
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * A {@link Step} for creating a Google Project for a workspace.
+ *
+ * <p>TODO(PF-156): Use RBS for project creation instead.
+ */
+public class CreateProjectStep implements Step {
+  @VisibleForTesting
+  static final ImmutableList<String> ENABLED_SERVICES =
+      ImmutableList.of("storage-api.googleapis.com");
```
davidangb, line 30: might this list - and the folder id - need to be passed in as arguments to the flight instead of hard-coded/in configuration?

wchamber: This project creation is all temporary since we're planning on migrating to RBS to buffer projects, so I'm inclined to make it a simple as possible in the short term.
I'm not sure if you're thinking that the folder and enabled services should be parameters to setting up a project, or exposed through the WM API. I'm going to assume the former.
Things like the folder id & enabled APIs (but medium term, which RBS pool to grab a project from) will be FlightMap inputParameters. It's not obvious how that will be reflected in the API yet. This relates to our "workspace profiles" to allow different project configuration for AoU vs CWB vs different Enterprise partners. (this is lightly touched in PF-158). I'm inclined to let us figure that out when we get further along and start simple for now. But yes, I expected there to be more parameters for how to set up a project.

src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    String projectId =
+        flightState
+            .getResultMap()
+            .get()
+            .get(WorkspaceFlightMapKeys.GOOGLE_PROJECT_ID, String.class);
+    assertEquals(
+        WorkspaceCloudContext.createGoogleContext(projectId),
+        workspaceDao.getCloudContext(workspaceId));
+    Project project = resourceManager.projects().get(projectId).execute();
```
zloery, line 63: Do we need to mark this test or do anything for Janitor, or will it just handle cleaning this up?

src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
```diff
@@ -0,0 +1,136 @@
...
+import com.google.api.services.serviceusage.v1.model.ListServicesResponse;
+import java.time.Duration;
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class CreateGoogleContextFlightTest extends BaseIntegrationTest {
```
zloery, line 28: This test is blurring the line between unit and integration differently than the other tests in the repo.

This is using integration configs and creating real GCP projects (which feels like an integration test), but it's executing against the local service instead of a remote one, whereas all our integration tests use the client library. 

I'd rather make this a full integration test that runs against a remote server (though ack that it wouldn't be possible to write yet given that this functionality isn't exposed).

src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
wchamber: As long as CRL is configured to talk to the Janitor, we're good. From previous PR, we have CRL configured to do so with testing profiles. So, as structured, these tests work.
For preview environments, or other non-persistent environments, we'll need to think more about how to start up WM with CRL in testing mode so that it does talk to Janitor.

src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
wchamber: I think there's a discussion to be had here. We need to be able to test cases that are we don't want to expose through an API. Case in point, the Error flight in this case - we don't really want an endpoint that creates flights that always fail. So, we need some notion of tests that don't run against a remote server. I'm open to creating new tests, or testing terminology - `locally connected` vs `remote`, but I think it's a mistake to try to make all integration tests run against remote servers.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateProjectStep.java
davidangb: 👍 , sounds like you've thought through it and I agree with the approach of keeping it simple for now and adding when necessary!

src/main/java/bio/terra/workspace/service/workspace/flight/CreateProjectStep.java
```diff
@@ -0,0 +1,151 @@
...
+                      .services()
+                      .batchEnable(
+                          projectName,
+                          new BatchEnableServicesRequest().setServiceIds(ENABLED_SERVICES))
+                      .execute());
+      pollUntilSuccess(operation, Duration.ofSeconds(30), Duration.ofMinutes(5));
+    } catch (IOException e) {
+      throw new RetryException("Error enabling services.", e);
+    }
+  }
```
davidangb, line 109: I believe this is the operation we've seen take the longest and be the most unreliable. No change necessary for this PR, but keep in mind for the future ... it may be easier to debug/retry/etc if create-project and enable-services are separate stairway steps.

davidangb (approved):
lgtm, modulo anything outstanding from Zach's comments

src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
zloery: I see the testing need here, and I don't think we can cover everything with just unit tests and remote integration tests like we have now. My main objection is to blurring the test types together, which makes it hard to tell what services are being mocked vs. run locally vs. run remotely. 

I'd like to get other people's input on this, but one possibility is splitting this new test type (`connected`?) out into its own directory + profile + spring configuration + base testing class + gradle task, similar to the way we currently split integration tests out from unit tests. That's pretty heavyweight, but it would make the test types distinct.

wchamber: Per offline talk, going forward with a new test type `connected`. Going to circulate a doc soon describing these test types.
I've added a new profile/spring/base testing class/gradle task/github action.
I did not separate this into its own directory. I think the connected tests are often closely tied to a single class under test, as this one is for `CreateGoogleContextFlight`. Keeping it in the same package, similar to unit test, drives the connection home. That does mean that unit tests and connected tests will sometimes be in the same directory, but I think having an easy way to find the relevant test outweighs possible confusion.
PTAL.

zloery (approved):

.github/workflows/connected-test.yml
```diff
@@ -0,0 +1,77 @@
...
+    branches: [ dev ]
+  pull_request:
+    branches: [ '**' ]
+jobs:
+  connected-test:
+
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
```
ungwudik, line 17: Just FYI - there's a PR (currently in review) that upgrades the postgres version from `9.6` to `12.3` https://github.com/DataBiosphere/terra-workspace-manager/pull/134

src/test/resources/application-connected-test.yml
```diff
@@ -0,0 +1,3 @@
+workspace:
```
ungwudik, line 1: How do you see this file evolving? I'm wondering how it's going to include `folder-id` values for other environments, in addition to that of dev.

.github/workflows/connected-test.yml
wchamber: Thanks for the heads up. @davidangb I think I might merge before you, so please update this file as well.

src/test/resources/application-connected-test.yml
wchamber: We don't need to run these tests against other environments. Just as we don't run unit tests "against" environments, we don't run connected tests against environments. Rather, we run the tests within a deployment environment where we have access to the dependencies and resources we need. For the test I'm adding here, we need a Service Account that has the privileges to create projects within a folder. There is such a credential and folder within the "dev" environment, and our `render_config.sh` already grabs dev SA credentials.
We don't  even necessarily need to run against "dev", we could run against a test only environment, or preview environments. Dev's just convenient at the moment as preview environments aren't ready.

I expect us to add more configuration values as needed to this file. In an ideal world, more of this test configuration that's based on the output of terraform would be rendered somewhere central, both for testing and for different helm deployments of environments. We're not here yet - we don't have a good way to pipe values from terraform to k8s helm/test config. So, for now, we're hard coding this value and hard coding the vault location of the service account credentials.
Happy to talk more offline too.

PR #111, opened 2020-10-14
use opencensus for tracing [AS-493]

zarsky-broad:
~I'm opening this as a draft for now just to solicit feedback.~


src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
```diff
@@ -0,0 +1,113 @@
...
+              .build());
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+
+    TraceConfig globalTraceConfig = Tracing.getTraceConfig();
+    globalTraceConfig.updateActiveTraceParams(
+        globalTraceConfig.getActiveTraceParams().toBuilder()
+            .setSampler(Samplers.probabilitySampler(tracingConfiguration.getProbability()))
+            .build());
```
wchamber, line 65: optional: This is generic StackdriverTraceExporter configuration set up. It's not specific to the TraceInterceptor any more than any other place we start use traces. Consider giving Stackdriver its own configuration class that just does this setup and keep this class as only the work needed to adding spans/attributes to each incoming request.
That's a natural division as OpenCensus encourages us to plugin different exporters behind the same tracing api.

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
```diff
@@ -0,0 +1,113 @@
...
+                .getCurrentSpan()
+                .putAttributes(
+                    Map.of(
+                        MDC_REQUEST_ID_KEY,
+                        AttributeValue.stringAttributeValue(requestId),
+                        "route",
+                        AttributeValue.stringAttributeValue(
+                            ((HandlerMethod) handler)
+                                .getMethodAnnotation(RequestMapping.class)
+                                .path()[0])));
```
davidangb, line 94: is there any chance that extraction of the route here will fail? Does this need try/catch or other failsafes?

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
```diff
@@ -0,0 +1,113 @@
...
+
+    return hashids.encode(generatedLong);
+  }
+
+  private String getMDCRequestId(HttpServletRequest httpRequest) {
+    return firstNonNull(
+        httpRequest.getHeader(MDC_REQUEST_ID_HEADER),
+        httpRequest.getHeader(MDC_CORRELATION_ID_HEADER),
+        generateRequestId());
+  }
```
davidangb, line 112: I _believe_ this is going to execute the relatively expensive `generateRequestId()` every time, then pass the resultant value into `firstNonNull` and discard it if the request has one of the desired headers. Could you rewrite this for efficiency so that it only calls `generateRequestId()` when it really needs to?

davidangb (commented):
a couple comments inline!

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
```diff
@@ -0,0 +1,119 @@
...
+
+    return hashids.encode(generatedLong);
+  }
+
+  private String getMDCRequestId(HttpServletRequest httpRequest) {
+    return getFirstNonNull(
+        () -> httpRequest.getHeader(MDC_REQUEST_ID_HEADER),
+        () -> httpRequest.getHeader(MDC_CORRELATION_ID_HEADER),
+        this::generateRequestId);
+  }
```
davidangb, line 118: good change here!

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
zarsky-broad: I think I did what you were describing here.

src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java
```diff
@@ -15,6 +19,21 @@
 
 @Configuration
 public class BeanConfig {
+  @Bean
+  public Tracer tracer() {
+    return Tracing.getTracer();
+  }
+
+  @Bean
+  public TracingAsyncClientHttpRequestInterceptor requestInterceptor() {
```
wchamber, line 28: did we ever figure out why we needed this? (/sanity check it's still needed?)
maybe a comment about why we have it would be helpful. probably ditto for the other beans.

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
wchamber: I meant split this into a separate class, but up to you.

src/main/resources/application.yml
```diff
@@ -68,6 +62,12 @@ workspace:
     frequency-ms: 60000
     staleness-threshold-ms: 600000
 
+  tracing:
+    enabled: ${env.tracing.enabled}
+    probability: ${env.tracing.probability}
+    project-id: ${env.tracing.project-id}
+    sa-path: ${env.tracing.sa-path}
```
wchamber, line 69: not really a comment on this PR, but why split these into the `env` declaration with shadow environment variables and the usage here? When should I have a `env` variable, and should I ever override them in other application property files.
We can override the values here from environment variables directly already, e.g. Spring maps `WORKSPACE_TRACING_ENABLED` to the workspace.tracing.enabled proprety. We don't need `CLOUD_TRACE_ENABLED` as well.

wchamber (approved):

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
zarsky-broad: Ah, that's what I did if you look in the files tab, sorry

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
```diff
@@ -0,0 +1,96 @@
...
+  private final Hashids hashids = new Hashids("requestIdSalt", 8);
+
+  @Autowired
+  public TraceInterceptorConfig(TracingConfiguration tracingConfiguration) {
+
+    TraceConfig globalTraceConfig = Tracing.getTraceConfig();
+    globalTraceConfig.updateActiveTraceParams(
+        globalTraceConfig.getActiveTraceParams().toBuilder()
+            .setSampler(Samplers.probabilitySampler(tracingConfiguration.getProbability()))
+            .build());
```
wchamber, line 43: move this into StackdriverTraceExporterConfig - it's the general setup of tracing as well.

wchamber (approved):

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
wchamber: oops, sorry!

src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java
zarsky-broad: We never did, it still is.
Added comments for the beans.

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
zarsky-broad: I'd been thinking of this as not being stackdriver specific, are we expecting different rules for different tracers?

davidangb (approved):

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
zarsky-broad: @wchamber is that 👍 makes sense, or 👍 it is stackdriver-specific and should be moved?

zarsky-broad:
@wchamber @davidangb Before I pull the trigger on this, the new version of the google opentelemetry exporter [was just published](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/41), thoughts on switching to opentelemetry instead of this? Or should we just go ahead here?

src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
wchamber: thumbs up that makes sense. Leave where is. Sorry for the ambiguity.

wchamber:
> @wchamber @davidangb Before I pull the trigger on this, the new version of the google opentelemetry exporter [was just published](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/41), thoughts on switching to opentelemetry instead of this? Or should we just go ahead here?

Pretty strongly feel we go ahead with OpenCensus. Even thought they've fixed the starting issue, it's till very much beta. Let's not be the bleeding edge.

davidangb:
agreed; let's stick with the stable library for now

PR #112, opened 2020-10-15
update deps & lock versions

zarsky-broad:
In addition to updating some deps (read changelogs and tested, nothing noteworthy), this adds version locking. While we use explicit versions for all of our direct dependencies, this will lock the versions of all transitive dependencies as well, so builds are actually reproducible.


gradle.lockfile
```diff
@@ -0,0 +1,239 @@
+# This is a Gradle generated file for dependency locking.
+# Manual edits can break the build and are not advised.
+# This file is expected to be part of source control.
```
wchamber, line 3: So does this just get created/modified automatically? Do we need to change anything about the developer workflow?

build.gradle
```diff
@@ -95,6 +95,10 @@ dependencies {
 	annotationProcessor group: "org.springframework.boot", name: "spring-boot-configuration-processor"
 }
 
+dependencyLocking {
+	lockAllConfigurations()
```
wchamber, line 99: If there's any really good reference for this, it might be nice to link somewhere.

gradle.lockfile
```diff
@@ -0,0 +1,239 @@
...
+com.jayway.jsonpath:json-path:2.4.0=testCompileClasspath,testRuntimeClasspath
+com.squareup.okhttp3:logging-interceptor:3.14.9=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+com.squareup.okhttp3:okhttp:3.14.9=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+com.squareup.okio:okio:1.17.2=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+com.vaadin.external.google:android-json:0.0.20131108.vaadin1=testCompileClasspath,testRuntimeClasspath
+com.zaxxer:HikariCP:3.4.5=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+commons-codec:commons-codec:1.14=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+io.github.classgraph:classgraph:4.8.69=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+io.grpc:grpc-alts:1.29.0=productionRuntimeClasspath,runtimeClasspath,testRuntimeClasspath
+io.grpc:grpc-alts:1.30.0=compileClasspath,testCompileClasspath
```
wchamber, line 70: what does it mean to have diffrent versions for `runtime` and `compile`? Should we be worried?

gradle.lockfile
zarsky-broad: After changing a dep, developers should run (for example) `./gradlew dependencies --write-locks` ([reference](https://docs.gradle.org/current/userguide/dependency_locking.html#lock_all_configurations_in_one_build_execution))

build.gradle
zarsky-broad: Yeah, it's https://docs.gradle.org/current/userguide/dependency_locking.html, I'll add a comment & put it in the readme

gradle.lockfile
wchamber: Can we enforce that somehow, similar to needing to run spotlessApply before committing?

gradle.lockfile
zarsky-broad: we shouldn't be worried, these are selected by the versions specified by the transitive deps that require them, which are different in the two configs.

the (simplified) inheritence:
```
$ gw dependencyInsight --dependency=grpc-alts

> Task :dependencyInsight

io.grpc:grpc-alts:{strictly 1.30.0} -> 1.30.0
\--- compileClasspath

io.grpc:grpc-alts:1.30.0
\--- com.google.api:gax-grpc:1.57.1
     +--- compileClasspath (requested com.google.api:gax-grpc:{strictly 1.57.1})
     \--- com.google.cloud:google-cloud-core-grpc:1.93.7
          +--- compileClasspath (requested com.google.cloud:google-cloud-core-grpc:{strictly 1.93.7})
          \--- org.springframework.cloud:spring-cloud-gcp-starter-trace:1.2.5.RELEASE
               \--- compileClasspath
```

vs

```
$ gw dependencyInsight --dependency=grpc-alts --configuration=runtimeClasspath

> Task :dependencyInsight

io.grpc:grpc-alts:{strictly 1.29.0} -> 1.29.0
\--- runtimeClasspath

io.grpc:grpc-alts:1.29.0
\--- com.google.cloud:libraries-bom:5.3.0
     +--- runtimeClasspath (requested com.google.cloud:libraries-bom:{strictly 5.3.0})
     \--- bio.terra:stairway:0.0.20
          \--- runtimeClasspath

io.grpc:grpc-alts:1.28.1 -> 1.29.0
\--- com.google.api:gax-grpc:1.56.0
     +--- runtimeClasspath (requested com.google.api:gax-grpc:{strictly 1.56.0})
     +--- com.google.cloud:libraries-bom:5.3.0
     |    +--- runtimeClasspath (requested com.google.cloud:libraries-bom:{strictly 5.3.0})
     |    \--- bio.terra:stairway:0.0.20
     |         \--- runtimeClasspath
     +--- com.google.cloud:google-cloud-pubsub:1.105.0
     |    +--- runtimeClasspath (requested com.google.cloud:google-cloud-pubsub:{strictly 1.105.0})
     |    +--- bio.terra:stairway:0.0.20 (requested com.google.cloud:google-cloud-pubsub:1.104.1) (*)
     |    \--- com.google.cloud:libraries-bom:5.3.0 (*)
     \--- com.google.cloud:google-cloud-core-grpc:1.93.4
          +--- runtimeClasspath (requested com.google.cloud:google-cloud-core-grpc:{strictly 1.93.4})
          +--- com.google.cloud:libraries-bom:5.3.0 (*)
          \--- org.springframework.cloud:spring-cloud-gcp-starter-trace:1.2.5.RELEASE (requested com.google.cloud:google-cloud-core-grpc:1.93.7)
               \--- runtimeClasspath
```

gradle.lockfile
zarsky-broad: See [here](https://docs.gradle.org/current/userguide/dependency_locking.html#running_a_build_with_lock_state_present), we could change the mode to strict but I don't _quite_ understand the difference

davidangb: I suggest we start with strict, and see if that causes us any problems. Though: I don't think strict mode is what @wchamber was talking about. I think we'd have to enforce that developers call `--write-locks` through PR review and such.

build.gradle
davidangb: +1 to adding info about dependency locking, and what to do if you need to change a dep version, to the README

gradle.lockfile
zarsky-broad: Working on making things clearer.
FWIW, here's the output when trying to run `unitTest` with some added unlocked dependencies:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileJava'.
> Could not resolve all files for configuration ':compileClasspath'.
   > Resolved 'ch.qos.logback.contrib:logback-json-classic:0.1.5' which is not part of the dependency lock state
   > Resolved 'com.google.cloud:google-cloud-logging:1.101.2' which is not part of the dependency lock state
   > Resolved 'com.google.api.grpc:proto-google-cloud-logging-v2:0.84.2' which is not part of the dependency lock state
   > Resolved 'ch.qos.logback.contrib:logback-json-core:0.1.5' which is not part of the dependency lock state
   > Resolved 'org.springframework.cloud:spring-cloud-gcp-logging:1.2.5.RELEASE' which is not part of the dependency lock state
   > Did not resolve 'com.google.auto.value:auto-value-annotations:1.7.2' which has been forced / substituted to a different version: '1.7.4'
   > Resolved 'org.springframework.cloud:spring-cloud-gcp-starter-logging:1.2.5.RELEASE' which is not part of the dependency lock state
   > Resolved 'org.checkerframework:checker-compat-qual:2.5.5' which is not part of the dependency lock state
   > Resolved 'com.google.cloud:google-cloud-logging-logback:0.118.0-alpha' which is not part of the dependency lock state
```

zarsky-broad: I'm leaning towards not enabling strict mode.
The main reason is that all the configurations we care about are already locked, which means that tools looking for synthetic configurations that don't actually exits get an error that there's no lockstate for them, rather than just seeing that the configurations don't exist. I discovered this with the [versions plugin](https://github.com/ben-manes/gradle-versions-plugin)

build.gradle
zarsky-broad: added to readme!

gradle.lockfile
```diff
@@ -0,0 +1,249 @@
+# This is a Gradle generated file for dependency locking.
+# Manual edits can break the build and are not advised.
```
wchamber, line 2: nit: maybe link to the readme, or copy the command here for the impatient?

wchamber (approved):

gradle.lockfile
zarsky-broad: When Gradle regenerates the file, anything we added to it would be lost, so that's not really possible.

davidangb (approved):

PR #113, opened 2020-10-16
Add support for running WM with a docker Postgres container in addition to local postgres

wchamber:
This allows developers to run a docker postgres container for WM to connect to when running in local development mode. This is helpful so that the developer's local postgres environment doesn't need to match WM's preferred postgres setup. This is also helpful when a developer is working on multiple services locally (but not simultaneously), preventing any database name conflicts. Copied from Janitor, which was copied from Sam.

```sh
./local_dev/run_postgres.sh start
```


davidangb (approved):
lgtm; did not test

local-dev/run_postgres.sh
```diff
@@ -0,0 +1,49 @@
...
+# Start up a postgres container with initial user/database setup.
+POSTGRES_VERSION=9.6
+start() {
+    echo "attempting to remove old $CONTAINER container..."
+    docker rm -f $CONTAINER || echo "docker rm failed. nothing to rm."
+
+    # start up postgres
+    echo "starting up postgres container..."
+    docker run --rm  --name $CONTAINER -e POSTGRES_PASSWORD=password -p "$POSTGRES_PORT:5432" \
+      -v $PWD/local-dev/local-postgres-init.sql:/docker-entrypoint-initdb.d/docker_postgres_init.sql \
```
zarsky-broad, line 11: Bind mounting is quite slow on mac and has a running cost, unfortunately; is there some other way we can attach files?

wchamber: Sorry, I'm not developing on mac, so I'm going to need help testing that any changes are good ones. I'd hope there are easy solutions here.
Is it the initial mounting that's slow (even for a single tiny file), or write propagation between host/container, or just the existence of the connection? I don't imagine that we mount many things, or that this file is something that gets modified during development. How slow is it to start up now? What's the constant running cost?

zloery (approved):

local-dev/run_postgres.sh
zarsky-broad: The issue is that, as long as the mount is active, everything chugs. It's an issue with macOS. As an alternative, consider:

```bash
docker create --name $CONTAINER -e POSTGRES_PASSWORD=password -p "$POSTGRES_PORT:5432" postgres:$POSTGRES_VERSION

docker cp $PWD/local-dev/local-postgres-init.sql $CONTAINER/docker-entrypoint-initdb.d/docker_postgres_init.sql

docker run --rm -d $CONTAINER
```

Also, is there a reason we have to spin up a second container to run the validation?

wchamber: Thanks, PTAL.
I think the original reasoning is that using a second container to run the validation allows the whole thing to start up without postgres being installed on the host. You could do all of your connections to the DB through docker, and never worry about having postgres installed on your machine. I don't have a strong opinion about this.

gjuggler: Side note on Docker performance with mounted volumes on OSX. This is something AoU struggled with quite a bit, since we used Docker to kick off our entire local API server. The performance degradation mostly seemed to scale with the # of files being mounted and how frequently they were being written to. (For example, we were blindly mounting the entire project directory, including the `build` subdirectory where intermediate outputs were constantly getting rewritten. This was bad.)

For a single-file mount like here, I wouldn't expect the performance hit to be too large. But just copying the script file makes sense as a best practice, where possible.

zarsky-broad: @gjuggler one solution we've used for that in the past is to use rsync with an intermediate container, since it doesn't have a performance penalty.

gjuggler: ack – the AoU team eventually adopted http://docker-sync.io/ which packages up a similar approach.

local-dev/run_postgres.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+# Start up a postgres container with initial user/database setup.
+POSTGRES_VERSION=9.6
+
+set -ex
+
+start() {
+    echo "attempting to remove old $CONTAINER container..."
+    docker rm -f $CONTAINER || echo "docker rm failed. nothing to rm."
```
zarsky-broad, line 9: Since this is expected to fail, I'm inclined not to print anything if it does.

```suggestion
    docker rm -f $CONTAINER
```

local-dev/run_postgres.sh
```diff
@@ -0,0 +1,51 @@
...
+        echo "postgres validation failed."
+        exit 1
+    fi
+
+}
+
+stop() {
+    echo "Stopping docker $CONTAINER container..."
+    docker stop $CONTAINER || echo "postgres stop failed. container already stopped."
+    docker rm -v $CONTAINER || echo "postgres rm -v failed.  container already destroyed."
```
zarsky-broad, line 33: Similar to above, with an explicit `exit 0` since we expect a failure.
```suggestion
    docker rm -v $CONTAINER
    exit 0
```

local-dev/run_postgres.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+# Start up a postgres container with initial user/database setup.
+POSTGRES_VERSION=9.6
+
+set -ex
```
zarsky-broad, line 5: I don't think we need this, really?
```suggestion
```

local-dev/run_postgres.sh
```diff
@@ -0,0 +1,51 @@
...
+    # start up postgres
+    echo "starting up postgres container..."
+    docker create --name $CONTAINER --rm -e POSTGRES_PASSWORD=password -p "$POSTGRES_PORT:5432" postgres:$POSTGRES_VERSION
+    docker cp $PWD/local-dev/local-postgres-init.sql $CONTAINER:/docker-entrypoint-initdb.d/docker_postgres_init.sql
+    docker start $CONTAINER
+
+    # validate postgres
+    echo "running postgres validation..."
+    docker run --rm --link $CONTAINER:postgres \
+      -v $PWD/local-dev/sql_validate.sh:/working/sql_validate.sh postgres:$POSTGRES_VERSION /working/sql_validate.sh
```
zarsky-broad, line 20: My thought with not using another container was this:
```suggestion
    docker exec $CONTAINER sh -c "$(cat $BASEDIR/local-dev/sql_validate.sh)"
```

We'd make a small tweak to the file, see below.

local-dev/sql_validate.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# validate postgres
+echo "sleeping for 5 seconds during postgres boot..."
+sleep 5
+export PGPASSWORD=wmpwd
+psql --username wmuser --host=postgres --port=5432 -d wm -c "SELECT VERSION();SELECT NOW()"
```
zarsky-broad, line 7: If we make the above tweak to reuse the container:
```suggestion
PGPASSWORD=wmpwd psql --username wmuser -d wm -c "SELECT VERSION();SELECT NOW()"
```

local-dev/run_postgres.sh
```diff
@@ -0,0 +1,51 @@
...
+set -ex
+
+start() {
+    echo "attempting to remove old $CONTAINER container..."
+    docker rm -f $CONTAINER || echo "docker rm failed. nothing to rm."
+
+    # start up postgres
+    echo "starting up postgres container..."
+    docker create --name $CONTAINER --rm -e POSTGRES_PASSWORD=password -p "$POSTGRES_PORT:5432" postgres:$POSTGRES_VERSION
+    docker cp $PWD/local-dev/local-postgres-init.sql $CONTAINER:/docker-entrypoint-initdb.d/docker_postgres_init.sql
```
zarsky-broad, line 14: To make this script not care where it's called from:

```suggestion
    docker cp $BASEDIR/local-dev/local-postgres-init.sql $CONTAINER:/docker-entrypoint-initdb.d/docker_postgres_init.sql
```

local-dev/run_postgres.sh
```diff
@@ -0,0 +1,51 @@
...
+POSTGRES_VERSION=9.6
+
+set -ex
+
+start() {
+    echo "attempting to remove old $CONTAINER container..."
+    docker rm -f $CONTAINER || echo "docker rm failed. nothing to rm."
+
+    # start up postgres
+    echo "starting up postgres container..."
```
zarsky-broad, line 12:
```suggestion
    echo "starting up postgres container..."
    BASEDIR=$(dirname "$0")
```

zarsky-broad (approved):
Working locally on macOS 10.15.7 (latest).
Some potential simplifications, otherwise good to go.

PR #116, opened 2020-10-19
PF-152 Use MapSqlParameterSource instead of a raw Map for sql parameters in DataReferenceDao.

wchamber:
Do the same on a couple of test cases in WorkspaceDaoTest.
MapSqlParameterSource is a more descriptive type for these variables. This standardizes our JdbcTemplate parameter types.


zloery (approved):
ty for this, LGTM assuming tests pass

PR #118, opened 2020-10-20
opt-in human readable logging

zarsky-broad:
Stackdriver formatted logging breaks things like clickable file names in stack traces in the IDE (in addition to being hard to parse).

I'm leaving stackdriver as the default here, but honestly I'd love to flip this, so deployments have to opt in to stackdriver logging.


davidangb:
Since production deployments require stackdriver/json logging, I think the default should be stackdriver logging. That way, we exercise those code paths as much as possible and shake out any potential issues with the logging stack such as https://logrhythm.com/blog/troubleshooting-delayed-syslog-messages/ that Leo just encountered.

I'm totally fine if we have a stanza in the config file for human-readable logging so that if a human wants to run a server, they can manually turn on human-readable logging.

zloery (approved):

zloery:
LGTM, and +1 to David's point about leaving the JSON format as the default

PR #122, opened 2020-10-23
Add test for getting a data reference

ungwudik:


zloery (approved):
LGTM assuming this passes, though I haven't run it

zloery:
Not sure if it's in scope for the ticket, but there's a closely related endpoint for getting a data reference by name instead of referenceId and that may need a test at some point as well

ungwudik:
> Not sure if it's in scope for the ticket, but there's a closely related endpoint for getting a data reference by name instead of referenceId and that may need a test at some point as well

Good call. This one https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/Workspace/getDataReferenceByName

It should be quick enough that I can add it here as well.

zloery (approved):

PR #123, opened 2020-10-23
PF-152 Add CRL dependency and configuration

wchamber:
Include CRL build dependencies for initial libraries needed for project creation within a workspace.
Create configuration to create CRL Cloud Object Wrapper (COW) beans.
Add janitor client credential grabbing for local test runs.


src/main/java/bio/terra/workspace/app/configuration/external/CrlConfiguration.java
```diff
@@ -0,0 +1,115 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+/** Configuration to use Terra Cloud Resource Library. */
+@Component
+@EnableConfigurationProperties
+@EnableTransactionManagement
+@ConfigurationProperties(prefix = "workspace.crl")
```
zloery, line 25: Are these settings supposed to be used outside of the test configuration, or is that a future PR?

zloery (approved):

src/main/java/bio/terra/workspace/app/configuration/external/CrlConfiguration.java
wchamber: The beans produced by the configuration are supposed to be used outside of tests in a future PR. So far, there's no configuration that needs to be set outside of testing.
But yes, this isn't used yet.

PR #124, opened 2020-10-23
add tracing to endpoints [AS-404]

zarsky-broad:
Added tracing for all authenticated endpoints, and the primary things that they call, i.e. sam auth and job things.

Questions:
- Do we want to trace unauthenticated endpoints?
- Is this too deep/shallow?

### DIFF
--- a/src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
+++ b/src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
@@ -17,6 +17,7 @@ import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.iam.SamService;
 import bio.terra.workspace.service.job.JobBuilder;
 import bio.terra.workspace.service.job.JobService;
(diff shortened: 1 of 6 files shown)


davidangb (approved):
what's in here looks good - assuming it all works! I haven't gone through the codebase to make sure this traces everything we need, but it looks like the right start at least.

I don't think we should trace unauthed endpoints. The /status endpoint would generate TONS of trace traffic that isn't worth it.

zarsky-broad:
Okay, so this doesn't work _quite_ how I'd thought.
The top level spans in the trace are auto-generated - this is because the opencensus spring jar [wires things up automatically](https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/spring/src/main/java/io/opencensus/contrib/spring/autoconfig/TraceWebAsyncClientAutoConfiguration.java) without actually telling anyone. This means it reports on all endpoints, including unauthed. However, there isn't a way to tell it to only autowire certain controllers. I'll keep working on this next week.

zarsky-broad:
Closing for now to think about this more.

PR #125, opened 2020-10-23
PF-152 Use the WM service account and user delegated SA for integration tests correctly.

wchamber:
Use the firecloud delegated service account to act as the test user in integration tests and use WM app service account to run WM application.
Previously, the common firecloud service account was used for both purposes.

Rename WorkspaceIntegrationTest config field names to make it clearer how they are used.

### DIFF
--- a/render_config.sh
+++ b/render_config.sh
@@ -1,25 +1,17 @@
 #!/bin/bash
 
 VAULT_TOKEN=${1:-$(cat "$HOME"/.vault-token)}
-TARGET_ENV=${2:-dev}
-
-FIRECLOUD_ACCOUNT_VAULT_PATH=secret/dsde/firecloud/$TARGET_ENV/common/firecloud-account.json
-SERVICE_ACCOUNT_OUTPUT_FILE_PATH=$(dirname "$0")/rendered/service-account.json
-
-if [[ ! "$TARGET_ENV" =~ ^(dev|alpha|perf|staging|prod)$ ]]; then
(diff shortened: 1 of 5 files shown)


render_config.sh
```diff
@@ -1,25 +1,17 @@
...
-TARGET_ENV=${2:-dev}
-
-FIRECLOUD_ACCOUNT_VAULT_PATH=secret/dsde/firecloud/$TARGET_ENV/common/firecloud-account.json
-SERVICE_ACCOUNT_OUTPUT_FILE_PATH=$(dirname "$0")/rendered/service-account.json
-
-if [[ ! "$TARGET_ENV" =~ ^(dev|alpha|perf|staging|prod)$ ]]; then
-    printf "\033[0;31m Unknown environment: $TARGET_ENV \n Must be one of [dev, alpha, perf, staging, prod] \n\033[0m"
-    exit 1
-fi
+WM_APP_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/terra/kernel/dev/dev/workspace/app-sa
```
zloery, line 4: Why hardcode the dev service account here?

zloery (approved):
LGTM with one question about render_config.sh

render_config.sh
wchamber: When would we want to vary it? This is the service account that WM runs as. I don't think we want to frequently be impersonating WM on other non-dev environments. I guess if we're trying to debug something that only happens in staging or perf? Should definitely never be prod. Seems overly flexible, so YAGNI.
Can change it back if you feel strongly.

PR #126, opened 2020-10-23
[AS-454] Improve error messaging when creating Data Repo snapshots in a workspace

Shakespeared:
Add additional error messages for various HTTP status errors when accessing a snapshot in a workspace.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
+++ b/src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
@@ -8,10 +8,13 @@ import bio.terra.workspace.app.configuration.external.DataRepoConfiguration;
 import bio.terra.workspace.app.configuration.spring.TraceInterceptorConfig;
 import bio.terra.workspace.common.exception.ValidationException;
 import bio.terra.workspace.generated.model.SystemStatusSystems;
+import bio.terra.workspace.service.datareference.exception.DataRepoAuthorizationException;
+import bio.terra.workspace.service.datareference.exception.DataRepoInternalServerErrorException;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import java.util.HashMap;
 import org.slf4j.MDC;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Component;
 
 @Component
@@ -60,7 +63,15 @@ public class DataRepoService {
       repositoryApi.retrieveSnapshot(snapshotId);
       return true;
     } catch (ApiException e) {
-      return false;
+      if (e.getCode() == HttpStatus.NOT_FOUND.value()) {
+        return false;
+      } else if (e.getCode() == HttpStatus.UNAUTHORIZED.value()) {
+        throw new DataRepoAuthorizationException(
+            "Not authorized to access Data Repo", e.getCause());
+      } else {
+        throw new DataRepoInternalServerErrorException(
+            "Data Repo returned the following error: " + e.getMessage(), e.getCause());
+      }
     }
   }
 
--- a/src/main/java/bio/terra/workspace/service/datareference/exception/DataRepoAuthorizationException.java
+++ b/src/main/java/bio/terra/workspace/service/datareference/exception/DataRepoAuthorizationException.java
@@ -1,0 +1,17 @@
+package bio.terra.workspace.service.datareference.exception;
+
+import bio.terra.workspace.common.exception.UnauthorizedException;
+
+public class DataRepoAuthorizationException extends UnauthorizedException {
+  public DataRepoAuthorizationException(String message) {
+    super(message);
+  }
+
+  public DataRepoAuthorizationException(String message, Throwable cause) {
+    super(message, cause);
+  }
+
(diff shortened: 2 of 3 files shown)


src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
```diff
@@ -60,7 +63,15 @@ public boolean snapshotExists(
...
-      return false;
+      if (e.getCode() == HttpStatus.NOT_FOUND.value()) {
+        return false;
+      } else if (e.getCode() == HttpStatus.UNAUTHORIZED.value()) {
+        throw new DataRepoAuthorizationException(
+            "Not authorized to access Data Repo", e.getCause());
+      } else {
+        throw new DataRepoInternalServerErrorException(
+            "Data Repo returned the following error: " + e.getMessage(), e.getCause());
+      }
```
davidangb, line 74: as a minor security issue for this PR - but still a standard best practice - we shouldn't distinguish between not-found and unauthorized. Doing so allows a malicious user to fish for existence/non-existence of a given resource. In this PR it's not that big a deal, but we should be in the habit of handling these the same. This is less helpful for developers but better for overall security.

I suggest this should return `false` for either `NOT_FOUND` or `UNAUTHORIZED` - and still bubble up any other exception - then the error message in `DataReferenceValidationUtils` should change to be something like "The given snapshot could not be found in the Data Repo instance provided or you do not have permission to it."

This means that `DataRepoAuthorizationException` may end up unused and you could delete it.

Final question: could the call to Data Repo return `HttpStatus.FORBIDDEN` too?

davidangb (commented):
yay for first workspace manager PR! See my comment inline.

src/main/java/bio/terra/workspace/service/datarepo/DataRepoService.java
gjuggler: Comment from the peanut gallery; this was an interesting detail that I hadn't considered before.

I totally agree that at an API level, returning 404 instead of 403 can contribute to a defense-in-depth security posture. It seems most useful for resources whose identifiers themselves might leak interesting information if their existence in the system can be inferred, e.g. email addresses or meaningful names.

A few points to add:

1) On a slightly rhetorical note: if Data Repo adopts the same philosophy in its own service, we should expect to always receive a 404 instead of a 403 here anyway. We may find it most useful to focus on ensuring WSM's other wholly-owned resources adopt this approach, and that this is socialized as a best practice across all services and/or encoded into a central library that converts exceptions into status codes.

2) This code is at the service level. It wouldn't be unreasonable to expect the response code obfuscation to happen at the API layer, while the service is fully transparent with its propagation of context. An argument could be made either way... which probably means this public method should be very explicit about whether it is doing any obfuscation, vs. relying on the controller layer for that. (Key point: it's a red flag to me that this public service method has no Javadoc — it should be simple to add some clarifying language.)

3) IMO, turning 401 --> 404 is a little less clear than 403 --> 404. It may be the case that Data Repo requires a non-empty access token in order to process any call, regardless of the requested resource. In that scenario, returning a 404 would be a real loss in terms of API usability, since the caller wouldn't get any hint that they forgot to include credentials. (This may well not be applicable to this call path – I just wanted to note.)

4) Regardless of where / how it's done, we should make sure to drop a server-side log whenever any status code transformation / obfuscation is taking place.

Shakespeared: @davidangb Thanks for the suggestion on how to handle the unauthorized vs not found cases, I had gone back and forth about how to treat those. I'm not sure if you can get `FORBIDDEN` from Data Repo… I will have to look into that! 

@gjuggler thanks for the detailed thoughts! For the purposes of this PR I think I lean towards doing the obfuscation here so we can get a quick fix, perhaps with a caveat that a better long term solution would be on the API side. However that's mostly due to my desire to resolve the PR so if someone has an opinion to the contrary I'm amenable to that.

davidangb: I agree that 401->404 is less clear and I don't think it's worth translating 401s to 404s.

I took a look through Data Repo and Rawls. What I found is that:
* Data Repo documentation says that they can return 401, 403, 404 from the get-snapshot endpoint (which is what this calls). I haven't spoken to a person to verify this.
* Rawls bubbles up any errors returned by Workspace Manager for the create-workspace endpoint (which is what this PR changes). Rawls is the public API for this call so it needs to do the right thing.

So I think we have two choices: 
1. Leave this PR as-is, and then create a second PR in Rawls to catch and wrap the errors so that both 403s and 404s get presented to the user as a 404.
2. Tweak this PR to change 403->404 and leave Rawls untouched.

From these two choices I think the first is the ideal path; as Greg said it puts the obfuscation in the right layer. @Shakespeared if you agree, let's figure out tactically what that means in terms of additional jira tickets, PRs, etc. etc.

Shakespeared: @davidangb yep, that makes sense to me, and sounds like the best overall solution. Let's chat about the follow-up to that some time this week.

davidangb (approved):
:+1:

PR #127, opened 2020-10-26
[PF-199] Run integration test against localhost instead of dev

wchamber:
Currently, the integration test runs against the dev environment after the PR is merged, but before the new image is pushed to dev. This PR makes it so that the integration test for a PR is run against a local instance of the service with the PR changes.
Also change the integration test workflow to run on PRs as well as after merges to dev.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -5,8 +5,6 @@ name: Run Integration Tests
 
 on:
   push:
-    branches:
-      - dev
     paths-ignore:
       - 'README.md'
       - '.github/**'
@@ -62,13 +60,17 @@ jobs:
       - name: Render configuration for tests
         run: ./render_config.sh ${{ steps.vault-token-step.outputs.vault-token }} dev
       - name: Ensure read permissions for rendered files
-        run: sudo chmod +r rendered/service-account.json
+        run: sudo chmod +r rendered
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
       - name: Initialize Postgres DB
         env:
           PGPASSWORD: postgres
         run: psql -h 127.0.0.1 -U postgres -f ./local-dev/local-postgres-init.sql
+      - name: Run local server (in background)
+        run: ./gradlew bootRun &
+      - name: Wait for local server to start
+        run: timeout 60 bash -c 'until echo > /dev/tcp/localhost/8080; do sleep 0.5; done'
       - name: Run integration tests
         env:
           TEST_ENV: dev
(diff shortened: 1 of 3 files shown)


.github/workflows/integration-test.yml
```diff
@@ -5,12 +5,9 @@ name: Run Integration Tests
 
 on:
   push:
-    branches:
-      - dev
-    paths-ignore:
-      - 'README.md'
-      - '.github/**'
-      - 'local-dev/**'
```
zloery, line 13: Why remove the paths-ignore section?

src/test/resources/application-integration-test.yml
```diff
@@ -8,6 +8,6 @@ workspace.integration-test:
   wsm-endpoints:
     workspaces: api/workspaces/v1
   wsm-urls:
-# swap the commented line to run integration tests against local
-    dev: https://workspace.dsde-dev.broadinstitute.org/
-#    dev: http://localhost:8080/
+# swap the commented line to run integration tests against dev
+#    dev: https://workspace.dsde-dev.broadinstitute.org/
+    dev: http://localhost:8080/
```
zloery, line 13: It feels kinda hacky to point at localhost but still call it dev. 

The slightly-less hacky solution would be changing the value of TEST_ENV in the github action, but that also requires copying values for `data-repo-instance-names` and `data-repo-snapshot-id`. I'm not sure that's worth it for a temporary solution like this

wchamber: I'm not that big a fan of TEST_ENV at all. I'd prefer just configuring the values instead of configuring a map of envs to values and then configuring then env. e.g. `data-repo-instance-name: terra`. We can have multiple application properties that we include on the fly if it's really interesting to run these tests against non dev environments.
As-is, what does it mean for there to be a local TEST_ENV? there's not a local DR to refer to, so we end up duplicating values. WDYT?

.github/workflows/integration-test.yml
wchamber: local-dev has the postgres init script. We actually care about if that changes, as that's actually used in this workflow for setting up the db.
The workflow itself is under .github. If we change the workflow, the test should run.
README.md doesn't matter, but neither does DEVELOPMENT.md. I'm not sure it's worth trying to keep an exhaustive list of not-actually-important files to save a few github action runs. So, meh.

src/test/resources/application-integration-test.yml
zloery: Yeah, that's the doubt I had. Given that it's ultimately only changing the wsm-url and this will be fairly short lived, I'm fine with the change as is.

zloery (approved):

PR #129, opened 2020-10-27
AS-525: auth token in header for create- and delete-workspace [risk: low]

davidangb:
This is a follow-on to broadinstitute/rawls#1298.

Highlights of this PR:
* for create- and delete-workspace, honor the auth token in the HTTP headers, and ignore any token sent in the request body
* in fact, remove the request-body token from the model definitions for the create- and delete-workspace API signatures
* since the token was the only member of the `DeleteWorkspaceRequestBody` class, remove that class entirely and define the delete API to not have a request body at all
* update all the runtime and test code to reflect these changes

Once these changes are in, we will publish a new workspace manager client lib, which we will adopt in Rawls in a future PR.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
+++ b/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
@@ -15,7 +15,6 @@ import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys;
 import io.opencensus.contrib.spring.aop.Traced;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.Tracer;
-import java.util.Optional;
 import java.util.UUID;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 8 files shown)


src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -70,9 +69,8 @@ public WorkspaceDescription getWorkspace(UUID id, AuthenticatedUserRequest userR
     return workspaceDao.getWorkspace(id);
   }
 
-  public void deleteWorkspace(UUID id, String userToken) {
+  public void deleteWorkspace(UUID id, AuthenticatedUserRequest userReq) {
```
davidangb, line 72: for consistency, I changed `deleteWorkspace` to accept a `AuthenticatedUserRequest`, since that's what the other two methods in this class accept.

src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java
```diff
@@ -134,26 +131,6 @@ public void deleteWorkspace(TestInfo testInfo) throws Exception {
...
-    DeleteWorkspaceRequestBody body = new DeleteWorkspaceRequestBody().authToken(token);
-    String jsonBody = testUtils.mapToJson(body);
-
-    WorkspaceResponse<?> deleteWorkspaceResponse =
-        workspaceManagerTestClient.delete(userEmail, path, jsonBody);
-
-    assertEquals(HttpStatus.UNAUTHORIZED, deleteWorkspaceResponse.getStatusCode());
-    assertTrue(deleteWorkspaceResponse.isErrorObject());
-  }
-
```
davidangb, line 156: this test is no longer necessary. This API now only looks at the token in the HTTP headers ... and any invalid token in the header will be rejected by the proxy and not even make it to the application.

zarsky-broad (approved):
Looks fabulous and passes the tests. As long as we're doing the Rawls PR dance right, I'm quite happy. 🎉

zloery (approved):
Nice, thanks for this!

ungwudik (approved):
Looks good!

PR #130, opened 2020-10-27
add publishing script and readme for client [no ticket; risk: low]

davidangb:
adds a script to ease publishing of the client lib to Artifactory. Plus a README for the client.

Script stolen from terra-rbs!

Tested by running locally and verifying in Artifactory.


zarsky-broad (approved):
TOL: I feel like there's got to be a way to do this with just one docker run, but I have no idea what it is, and it's mostly academic. ¯\_(ツ)_/¯

davidangb:
@zarsky-broad you _*could*_ read all of `ARTIFACTORY_ACCOUNT_PATH` as json in one docker execution, then extract the individual parts you need via `jq` or other parsing ... I'm with you on ¯\\_(ツ)_/¯ tho

PR #132, opened 2020-10-28
more detailed traces [AS-404]

zarsky-broad:
All endpoints are automatically traced by the opencensus spring package. This adds sub-spans for the major methods within them.

Since there's not a good way to only create traces for specific routes, this only adds request id's to spans for authenticated endpoints.


src/main/java/bio/terra/workspace/app/configuration/spring/TraceInterceptorConfig.java
```diff
@@ -51,8 +51,10 @@ public void addInterceptors(InterceptorRegistry registry) {
           public boolean preHandle(
               HttpServletRequest httpRequest, HttpServletResponse httpResponse, Object handler) {
 
-            // We don't need to do this for resources (swagger ui)
-            if (handler instanceof HandlerMethod) {
+            // We don't need to trace resources (swagger ui) or unauthenticated endpoints,
+            // but there isn't a good way with opencensus to filter what's captured, so we just
+            // don't bother to annotate them
+            if (handler instanceof HandlerMethod && httpRequest.getAuthType() != null) {
```
davidangb, line 57: Does this mean that if a user sent an auth token in headers to /status or /version, we'd still add the route, etc?

Also ... does the `getAuthType() != null` actually work, empirically? goes `getAuthType()` magically return non-null simply because a request has an Authorization header?

davidangb (approved):
`httpRequest.getRequestURI().startsWith("/api/")` looks better! :+1: assuming you have run this locally and see the traces doing the right thing.

PR #133, opened 2020-10-28
[PF-154] Add workspace stage feature toggle

zloery:
This change adds a column to the `workspace` table with an enum for tracking the stage of a workspace (e.g. either 'RAWLS_WORKSPACE' or 'MC_WORKSPACE') along with access methods to the workspaceDAO.

This also modifies the createWorkspace endpoint to accept an optional body parameter specifying the workspace stage (default is RAWLS_WORKSPACE) and the getWorkspace endpoint to include this information when querying about a workspace.


src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -3,6 +3,7 @@
 import bio.terra.workspace.common.exception.DuplicateWorkspaceException;
 import bio.terra.workspace.common.exception.WorkspaceNotFoundException;
 import bio.terra.workspace.generated.model.WorkspaceDescription;
+import bio.terra.workspace.generated.model.WorkspaceStageEnum;
```
wchamber, line 6: I'm still not wild about passing model classes down to the WorkspaceDao. This locks us into any changes to our representation of the workspaces internally being reflected in our public API, or vice versa. I think the cost of writing a conversion function to go `api model` <-> `internal enum representation` is tiny. I think the cost of this future proofing is low, and linking API & DB layers is a classic painful problem.
Examples:
- Something that might be a public enum might later have more state. 
- We might want to rename enums in the API. Renaming enums in the DB require DB migrations.

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -81,6 +84,8 @@ public WorkspaceDescription getWorkspace(UUID id) {
         desc.setSpendProfile(UUID.fromString(queryOutput.get("spend_profile").toString()));
       }
 
+      desc.setStage(WorkspaceStageEnum.fromValue(queryOutput.get("workspace_stage").toString()));
```
wchamber, line 87: similarly, not yet but I still strongly feel that we need to change this DAO to return something other than the `get workspace` request response. e.g., we're going to want to check the stage of a workspace to disable APIs. Why does that internal action generate a `get workspace` request response.
Doesn't have to be this PR, but it's fast approaching.

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -100,6 +101,52 @@ public void testGetExistingWorkspace() throws Exception {
     assertThat(desc.getId(), equalTo(workspaceId));
   }
 
+  @Test
+  public void testDefaultWorkspaceStage() throws Exception {
```
wchamber, line 105: nit: add the expected outcome as well as the scenario if the outcome isn't obvious, e.g. `testDefaultWorkspaceStageIsRawls`

src/main/resources/api/service_openapi.yaml
```diff
@@ -531,6 +535,11 @@ components:
           description: The ID of the Data Repo snapshot
           type: string
 
+    WorkspaceStageEnum:
```
wchamber, line 538: I'm not sold on `Stage`. I'm on the fence, could be convinced. WDYT about `Mode`?
Stage sounds like these workspaces are going to progress through these states. While that is the eventual migration plan, it's not going to be reflected in our code - migration probably happens outside of WM proper.

wchamber (changes requested):

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
zloery: Sounds good. It's a pretty minor difference to implement now, so if there's some expected benefit in the future that seems worthwhile. I haven't used the pattern of separate API model / internal representation before, so let me know if there are style improvements on where things should live, where the transform should happen, etc

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
zloery: I actually think this endpoint in particular is less of an issue, other than the continued thread of sharing API models with DB models. It's just pulling out every column of a workspace in the table, which we re-use directly as the response to a `getWorkspace` request.
It might make sense to create a specific method for just getting the enum if that's a concern, though

src/main/resources/api/service_openapi.yaml
zloery: My thinking with `Stage` (both here and in the doc) was that it makes it clear that there is a progression, even if moving through that really ends up happening outside WM. I didn't want to use something like `Mode` or `Type` because that feels like we're drawing a more permanent distinction between the workspaces, and I think that makes it more likely that it would creep into being more than a temporary feature toggle, which I'd like to avoid.

src/main/java/bio/terra/workspace/db/WorkspaceStageEnum.java
```diff
@@ -0,0 +1,58 @@
...
+import bio.terra.workspace.generated.model.WorkspaceStageEnumModel;
+
+/**
+ * Internal representation of the workspace stage used for gating MC Terra features from Rawls
+ * workspaces. This is the enum serialized into the Workspace db table, not the enum generated as
+ * part of the external API. Current values RAWLS_WORKSPACE or MC_WORKSPACE.
+ */
+public enum WorkspaceStageEnum {
+  RAWLS_WORKSPACE("RAWLS_WORKSPACE_INTERNAL"),
+  MC_WORKSPACE("MC_WORKSPACE_INTERNAL");
```
wchamber, line 13: I don't think the string's need to be different for the internal representation. Just having the option to fork them later is good. If we do change them, we can change the enum mapping functions to map multiple enum values to a single enum of a different type.

src/main/java/bio/terra/workspace/db/WorkspaceStageEnum.java
```diff
@@ -0,0 +1,58 @@
...
+    return String.valueOf(value);
+  }
+
+  public static WorkspaceStageEnum fromValue(String value) {
+    for (WorkspaceStageEnum b : WorkspaceStageEnum.values()) {
+      if (String.valueOf(b.value).equals(value)) {
+        return b;
+      }
+    }
+    return null;
```
wchamber, line 32: I think if we just have an enum without a private String that's fine and eliminates this code.

src/main/java/bio/terra/workspace/db/WorkspaceStageEnum.java
```diff
@@ -0,0 +1,58 @@
...
+  public static WorkspaceStageEnumModel toApiModel(WorkspaceStageEnum internalEnum) {
+    switch (internalEnum) {
+      case RAWLS_WORKSPACE:
+        return WorkspaceStageEnumModel.RAWLS_WORKSPACE;
+      case MC_WORKSPACE:
+        return WorkspaceStageEnumModel.MC_WORKSPACE;
+      default:
+        throw new InvalidWorkspaceStageException(
+            "Invalid workspace stage provided: " + internalEnum.toString());
+    }
```
wchamber, line 56: consider using a BiMap<WorkspaceStageEnumModel WorkspaceStage>. e.g. janitor model utils. (keeping conversion functions here are fine) https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/main/java/bio/terra/janitor/service/janitor/ModelUtils.java

src/main/java/bio/terra/workspace/db/WorkspaceStageEnum.java
```diff
@@ -0,0 +1,58 @@
...
+        return WorkspaceStageEnum.RAWLS_WORKSPACE;
+      case MC_WORKSPACE:
+        return WorkspaceStageEnum.MC_WORKSPACE;
+      default:
+        throw new InvalidWorkspaceStageException(
+            "Invalid workspace stage provided: " + modelEnum.toString());
+    }
+  }
+
+  public static WorkspaceStageEnumModel toApiModel(WorkspaceStageEnum internalEnum) {
```
wchamber, line 47: should this one be not-static?

src/main/java/bio/terra/workspace/db/exception/InvalidWorkspaceStageException.java
```diff
@@ -0,0 +1,26 @@
+package bio.terra.workspace.db.exception;
+
+import bio.terra.workspace.common.exception.BadRequestException;
+import java.util.List;
+
+public class InvalidWorkspaceStageException extends BadRequestException {
```
wchamber, line 6: do we need all the constructors, or even a separate exception class? I think the message would be enough to figure out what went wrong?

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -39,15 +40,18 @@ public WorkspaceDao(NamedParameterJdbcTemplate jdbcTemplate) {
   }
 
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
-  public String createWorkspace(UUID workspaceId, UUID spendProfile) {
+  public String createWorkspace(
+      UUID workspaceId, UUID spendProfile, WorkspaceStageEnumModel workspaceStage) {
```
wchamber, line 44: I think the workspaceDao should take the internal enum, not the model enum.

src/main/java/bio/terra/workspace/db/WorkspaceStageEnum.java
```diff
@@ -0,0 +1,58 @@
...
+
+import bio.terra.workspace.db.exception.InvalidWorkspaceStageException;
+import bio.terra.workspace.generated.model.WorkspaceStageEnumModel;
+
+/**
+ * Internal representation of the workspace stage used for gating MC Terra features from Rawls
+ * workspaces. This is the enum serialized into the Workspace db table, not the enum generated as
+ * part of the external API. Current values RAWLS_WORKSPACE or MC_WORKSPACE.
+ */
+public enum WorkspaceStageEnum {
```
wchamber, line 11: maybe `WorkspaceStage`? Don't think we need hungarian notation.

src/main/resources/api/service_openapi.yaml
```diff
@@ -531,6 +535,11 @@ components:
           description: The ID of the Data Repo snapshot
           type: string
 
+    WorkspaceStageEnumModel:
```
wchamber, line 538: `WorkspaceStage` is also fine, or `WorkspaceStageModel`. There shouldn't be too many classes that import both enums, but we at least don't need to name it Enum.

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
wchamber: All the conversion between the request/response models to the internal representation should happen at the request/response layer, i.e. the controllers. The serialization layers at the top/bottom of the service should be in charge of converting the internal representation to the serialized representation. The services/flights/DAOs should communicate with the internal representation. (The DAO handles an equivalent conversion to its own representation, the SQL queries/params/results.)

src/main/java/bio/terra/workspace/db/WorkspaceStageEnum.java
zloery: That's pretty useful, thanks!

src/main/java/bio/terra/workspace/db/exception/InvalidWorkspaceStageException.java
zloery: Where possible I prefer using more specific exception classes, I think it generally makes errors easier to trace and debug.

src/main/java/bio/terra/workspace/db/WorkspaceStageEnum.java
zloery: s.g. A few other enums from Swagger do the same, though there's no strong reason for it

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
zloery: That seems like a reasonable pattern, but at least at the moment a number of service methods directly take the input request. Refactoring those out seems like a larger effort

src/main/java/bio/terra/workspace/common/model/WorkspaceStage.java
```diff
@@ -0,0 +1,38 @@
...
+      return WorkspaceStage.RAWLS_WORKSPACE;
+    }
+    return stage;
+  }
+
+  public WorkspaceStageModel toApiModel() {
+    WorkspaceStageModel stage = stageMap.get(this);
+    if (stage == null) {
+      return WorkspaceStageModel.RAWLS_WORKSPACE;
+    }
```
wchamber, line 35: why do we need default values in the conversion?
If it's to support rows that already exist in the db, better to take the one time hit of doing a db migration, or handle the backwards compatibility null conversion in the DAO. If it's to support request serialization, I think we can specify open api default values. (Though, even then, we would only need the default from the model). If we never expect the arguments to be null, this ends up being confusing. If we add new values but forget to update this mapping, we'll silently use the defaults. Loudly failing early makes it more likely for us to find a bug before it's an issue.
It's much easier to reason about a non-null world, but it takes more effort up front to get there.

wchamber (changes requested):
last nit question

src/main/java/bio/terra/workspace/common/model/WorkspaceStage.java
zloery: I wanted requests with no value specified to default to RAWLS_WORKSPACE so that existing clients didn't need to know about the change. It's not a large change to update the client library Rawls is using, but it felt like Rawls shouldn't need to know about this toggle for features it isn't using. The default you specify in openAPI for request body schemas only affects the UI, an unspecified value still comes in as null (I think it's different for path/query params, but I'm not sure).

That said postgres supports default values for columns, so we could translate a null modelEnum to a null WorkspaceStageModel and when we write it to the DB we'd have RAWLS_WORKSPACE and wouldn't need the internal model -> api model default. I didn't like the idea of passing null around like that, but it would give a clearer warning in the case where we update the api enum without changing the model enum. I'm fine changing that if you think the warning is worth it

wchamber: Not needing to update the Rawls client sounds reasonable. Sorry, I was wrong about the defaults.
If the goal is to support backwards compatibility with existing client libraries, the right place to handle the unspecified null value is in the API layer where the request is received. We're trying to map existing create workspace requests -> RAWLS_WORKSPACE requests. That's a concern about the representation in the create workspace API and how we can change it over time. It's not a general conversion principle, or a stance that any default null value should be treated as RAWLS_WORKSPACE. If we added another endpoint that fetched all my workspaces in a separate stage, we wouldn't want 'null' to mean RAWLS_WORKSPACE. So this null/unspecified handling code should live in the controller that handles the API, and be used only in the endpoint we're concerned about.
This is a good, albeit minor, example of the benefits of having distinct layers. With a conversion in the API layer, none of the rest of the service needs to worry about the fact that the create workspace request has this support. When I'm reading the rest of the service, I don't need to worry about, oh, maybe this value was null and it got converted.

That sounds like we only need the default conversion for model -> internal.
I think the internal -> model default conversion is for the rows already in the DB. For that, I still think it's simpler to write liquibase to update all existing rows to RAWLS_WORKSPACE, as if they had all been created after this code change. Using postgres to interpret default null values does sound more confusing.

zloery: Sounds good, moved the default conversion into the controller endpoint and removed the internal->model conversion altogether. I also switched the liquibase to update all existing rows to RAWLS_WORKSPACE and not do anything special when workspace stage isn't specified.

wchamber (approved):
thanks!

davidangb (approved):
lgtm - with the change to the model in create- and get-workspace, please make sure to manually test the Rawls->WSM functionality and update Rawls with a new client as necessary!

zloery:
Manually testing the new client library with Rawls looks good! The Rawls update is in https://github.com/broadinstitute/rawls/pull/1305

PR #135, opened 2020-10-30
Fix data reference dao and add index on data reference table

zloery:
Currently, the `isControlled` and `deleteDataReference` operations in the data reference DAO only take reference IDs and not workspace IDs. While this is fine from a database perspective (as reference IDs are the primary key), it allows services (and by extension users) to call these operations to improperly access data references across workspace boundaries. 

Additionally, I added a joint index on the `reference_id` and `workspace_id` columns as we now have several operations that select by these two columns combined.

I originally considered defining `reference_id` and `workspace_id` as joint primary keys to avoid issues like this in the future, but after discussion we decided this wasn't needed and would incur a small performance hit in the future.


zloery:
+DA since it looks like Matt is out this week

davidangb:
Hm. I agree that conceptually, resources/references/applications are unique per-workspace. But I'm not sure it applies at the DB layer.

For resources and applications ... truth is we don't know exactly how we'll implement these, so any changes to the DB and its constraints/indexes are guesswork. I'm fine making the change in this PR but once we understand all the access patterns for real, we may have to roll back or make other changes.

For references ... we DO in fact generate a unique reference id for each row:
https://github.com/DataBiosphere/terra-workspace-manager/blob/5ff5aee37dfab934b2ab36ff7509f55b2d7dc72d/src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java#L89
So, the `reference_id` column is a fine primary key on its own. I do think adding an index on `workspace_id, reference_id` could be helpful, since we use both of them for select queries:
https://github.com/DataBiosphere/terra-workspace-manager/blob/5ff5aee37dfab934b2ab36ff7509f55b2d7dc72d/src/main/java/bio/terra/workspace/db/DataReferenceDao.java#L82

zloery:
Agreed that any of these could reasonably function as primary keys on their own as (I'm expecting) - my main concern is that treating them that way doesn't quite line up with how we think about them elsewhere, which can lead to bugs and inconsistencies. 
E.g. our current method for deleting a reference at the DAO level only takes the reference ID - you can currently delete references belonging to other workspaces (which should be fixed regardless of the change to DB schema here). I think that came out of misunderstanding whether reference IDs are unique globally or per-workspace

davidangb:
compound PKs/FKs are less performant than single-column PKs/FKs ... is it worth the performance hit to model the db in the way we think about them elsewhere, when just the single-column `reference_id` could suffice? The perf hit is not going to be big - perhaps imperceptible until we have much more data/requests; I'm just thinking ahead.

zloery:
Ah, I didn't know joint keys had particularly different performance than single-column ones. If the difference is enough that we would notice it at higher loads I'm inclined towards adding an index for that table, fixing deleteDataReference and isControlled, and leaving resources + applications untouched until we use them.

davidangb (approved):

PR #137, opened 2020-11-02
PF-155 Add service level methods for getting and creating google cloud context.

wchamber:
Add methods to WorkspaceService for creating and getting cloud context. Delete to follow. 

I'm ambivalent about adding these methods to WorkspaceService vs creating a CloudContextService. There's not much meat to them, but they cold be considered distinct. The WorkspaceService's workspace deletion flight will want to reference steps in the cloud context deletion service, so there will be cross package dependencies.

I would like to change the other tests in WorkspaceService to use its API instead of http endpoints in another PR.
Also removes a WorkspaceServiceTest `workspaceCreatedFromJobRequest` that duplicates `testGetExistingWorkspace` AFAICT.

### DIFF
--- a/src/test/java/bio/terra/workspace/common/BaseConnectedTest.java
+++ b/src/test/java/bio/terra/workspace/common/BaseConnectedTest.java
@@ -1,6 +1,7 @@
 package bio.terra.workspace.common;
 
 import org.junit.jupiter.api.Tag;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
 import org.springframework.test.context.ActiveProfiles;
 
 /**
@@ -11,4 +12,5 @@ import org.springframework.test.context.ActiveProfiles;
  */
 @Tag("connected")
 @ActiveProfiles("connected-test")
+@AutoConfigureMockMvc
 public class BaseConnectedTest extends BaseTest {}
--- a/src/test/java/bio/terra/workspace/common/BaseConnectedTest.java
+++ b/src/test/java/bio/terra/workspace/common/BaseConnectedTest.java
@@ -1,7 +1,6 @@
 package bio.terra.workspace.common;
 
 import org.junit.jupiter.api.Tag;
-import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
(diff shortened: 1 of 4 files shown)


src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -42,8 +45,9 @@
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.MvcResult;
 
-public class WorkspaceServiceTest extends BaseUnitTest {
-
+public class WorkspaceServiceTest extends BaseConnectedTest {
```
zloery, line 48: Do we want this to be a connected test when most dependencies are mocked out? Even if we move to calling services directly instead of via HTTP the existing tests here are really only testing WM and its database.

src/test/java/bio/terra/workspace/common/BaseConnectedTest.java
```diff
@@ -11,4 +12,5 @@
  */
 @Tag("connected")
 @ActiveProfiles("connected-test")
+@AutoConfigureMockMvc
```
zloery, line 15: Are we expecting to use MockMvc in connected other connected tests or is this just for WorkspaceServiceTest? If the latter, the annotation should be there (modulo keeping that as a connected test)

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
wchamber: The new tests are using HTTP, so they do need to be connected tests. I think it's fine for connected tests to have a mix of real and mocked/stubbed dependencies. I think it might be more confusing to have multiple `WorkspaceServiceTest.java` files each with partial method coverage, but that is an option. Soon, workspace deletion will want to delete projects as well, which is another connected use case.

src/test/java/bio/terra/workspace/common/BaseConnectedTest.java
wchamber: good point, done.

zloery (approved):

PR #141, opened 2020-11-06
PF-186 Add a spend service based on config with Sam authz check

wchamber:
Add a SpendProfileService. Base the "SpendProfiles" being handed out on configuration for spend profile ids & billing account id. Do a Sam `spend-profile` `link` authz check. This will serve as our short term Spend Profile Manager stub until the component is built and we integrate with it.
I will replace the existing UUID spendProfileId usage in the rest of the code/api with the String based strongly typed SpendProfileId in another PR.


src/main/java/bio/terra/workspace/service/spendprofile/SpendProfileService.java
```diff
@@ -0,0 +1,76 @@
...
+    this.samService = samService;
+    this.spendProfiles = Maps.uniqueIndex(spendProfiles, SpendProfile::id);
+  }
+
+  /**
+   * Authorize the user to link the Spend Profile. Returns the {@link SpendProfile} associated with
+   * the id if there is one and the user is authorized to link it. Otherwise, throws a {@link
+   * SpendUnauthorizedException}.
+   */
+  public SpendProfile authorizeLinking(SpendProfileId spendProfileId, String userAccessToken) {
```
zloery, line 44: Should this take an `AuthenticatedUserRequest` here instead of the token directly? That's how other services have encapsulated permissions objects

src/main/java/bio/terra/workspace/app/configuration/external/SpendProfileConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * <p>TODO: Replace the configuration based Spend Profile with integration with the to-be-built
+ * Spend Profile Component. For now, we manage spend profiles manually.
+ */
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "workspace.spend")
+public class SpendProfileConfiguration {
+  /** The Spend Profiles known to Workspace Manager. */
+  private List<SpendProfileModel> spendProfiles = new ArrayList<>();
```
zloery, line 21: I haven't created collections of non-primitive objects via config before - just to confirm, does this pattern work for creating objects?

src/test/java/bio/terra/workspace/service/spendprofile/SpendProfileServiceTest.java
```diff
@@ -0,0 +1,91 @@
...
+            mockSamService.isAuthorized(
+                Mockito.eq(USER_ACCESS_TOKEN),
+                Mockito.eq(SamUtils.SPEND_PROFILE_RESOURCE),
+                Mockito.any(),
+                Mockito.eq(SamUtils.SPEND_PROFILE_LINK_ACTION)))
+        .thenReturn(true);
+  }
+
+  @Test
+  public void authorizeLinking_Success() {
```
zloery, line 37: nit, afaik other tests in WM just use camelcase without underscores for names. I'd vote for consistency here, but it's not a strong opinion

zloery (approved):
LGTM w/ minor questions + nits

src/main/java/bio/terra/workspace/app/configuration/external/SpendProfileConfiguration.java
wchamber: Yep. Yaml for lists & maps works as you want it to. Nested pojo objects are also fine.

src/main/java/bio/terra/workspace/service/spendprofile/SpendProfileService.java
wchamber: I'll buy the consistency argument, but other than JobService, none of them need anything outside of the user token. (I also don't know what it means to have an AuthenticatedUserRequest with an Optional.empty token, which is the one field we need here and everywhere else.) Going ahead with AuthenticatedUserRequest for this PR.

src/test/java/bio/terra/workspace/service/spendprofile/SpendProfileServiceTest.java
wchamber: I've already added underscores in both WorkspaceDaoTest & MdcHookTest. But I also don't feel strongly.

PR #142, opened 2020-11-09
[PF-227] Fix stairway working map interactions in WorkspaceService

zloery:
This PR makes changes related to the way we're using Stairway's working map. Working map updates are persisted at step boundaries, so the previous pattern of "set flag to false, perform action, set flag to true" within a single step did not work.

This change adds an optional new `operationId` parameter to the create workspace request body. Requests with the same `operationId` are treated as logical duplicates. Clients should use this new field to ensure `createWorkspace` calls are idempotent. If an operation ID is not provided, the workspace Controller will generate a random one.

The reason this flight is treated differently from CreateDataReference is that workspace IDs are passed in externally, while data reference IDs are generated internally. For data references, this allows us to assume that the current flight is the only one working with a given ID, and we don't have to worry about deleting resources we didn't create. Because we can't make this assumption for workspace creation flights, we need another mechanism to support synchronization control.

Because this changes the generated classes, I'll update the client library after this is approved but before merging. I don't know the Rawls codebase well enough to say when operation IDs should be generated there.

(DataReferenceService changes moved to #150 as the WorkspaceService changes get more complex)


src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceAuthzStep.java
```diff
@@ -52,9 +55,14 @@ public StepResult undoStep(FlightContext flightContext) {
     FlightMap inputMap = flightContext.getInputParameters();
     // Only delete the Sam resource if we actually created it in the do step.
     FlightMap workingMap = flightContext.getWorkingMap();
-    if (workingMap.get(AUTHZ_COMPLETED_KEY, Boolean.class)) {
-      UUID workspaceID = inputMap.get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
+    UUID workspaceID = inputMap.get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
+    if (workingMap.get(AUTHZ_COMPLETED_KEY, Boolean.class) != null) {
```
yonghaoy, line 59: Where we set this value now? And do we still need this?

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
```diff
@@ -55,9 +58,14 @@ public StepResult doStep(FlightContext flightContext) throws RetryException {
   public StepResult undoStep(FlightContext flightContext) {
     FlightMap inputMap = flightContext.getInputParameters();
     FlightMap workingMap = flightContext.getWorkingMap();
-    if (workingMap.get(CREATE_WORKSPACE_COMPLETED_KEY, Boolean.class)) {
-      UUID workspaceId = inputMap.get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
+    UUID workspaceId = inputMap.get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
+    if (workingMap.get(CREATE_WORKSPACE_COMPLETED_KEY, Boolean.class) != null) {
```
yonghaoy, line 62: same quesiton for this.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
```diff
@@ -55,9 +58,14 @@ public StepResult doStep(FlightContext flightContext) throws RetryException {
...
       workspaceDao.deleteWorkspace(workspaceId);
+    } else {
+      logger.warn(
+          "Undoing a CreateWorkspaceStep that was not fully completed. This may have created an orphaned workspace "
+              + workspaceId.toString()
+              + " in WM's database.");
```
davidangb, line 68: I'd like to find a way to tie certain log messages into alerts - it would be nice to get an alert if this happens, so we can react. Maybe this is as simple as making this a `logger.error`? Or maybe we insert a magic string e.g. `[ALERT]` into the message? @ansingh7115 as our alert expert, do you have any ideas here?

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceAuthzStep.java
zloery: This value still gets set after creating a workspace resource in Sam, and we still need it to confirm that the doStep completed and created that resource.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
zloery: Same as the other step, that gets set in the `do` step as a way of confirming that a workspace was created in this flight.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
ansingh7115: We can create log-based metric and then set an alert on it
https://cloud.google.com/logging/docs/logs-based-metrics/

zloery: Sounds good - do we have an established pattern for what to put in the message contents for alerts (e.g. something like prefixing with `[ALERT]`?

davidangb: cool, so it sounds like no changes needed here!

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceAuthzStep.java
yonghaoy: ah, I see. Sorry, I just saw you remove `workingMap.put(AUTHZ_COMPLETED_KEY, false);` and thought this is not needed. Feels little bit wired that we are not actually using the `Boolean` type, this line works with any other types, all we need is to put something in working map. But I think it's fine.

Also, we still have the same problem as I discussed in my doc(working map is step boundary,)? 
Not famialr with the permormance requirement/use cases, but other options I can come up with if we want to get rid of working map here:
1. Check existances before calling delete
2. Call delete and catch 404 exception.

zloery: Yeah, it feels a little silly to only be using a "present/absent" flag, but I guess Boolean is the right type for that?

Agreed, this still has the issue of orphaning resources when there's a failure between the actual creation and marking the value in the working map - at least we have a warning for it now. I think both of your options run the risk of deleting resources that weren't created in this flight: since WorkspaceIDs are passed in externally via request, multiple flights might be running for the same ID. 

Without some kind of locking mechanism, I don't think we can guarantee that flights will both 1. undo everything they did and 2. won't undo work they didn't do. Since #2 is much worse than #1, that's the one this workingMap key prevents.

wchamber: drive-by locking mechanism comment: the workspaceId can be used as the jobId in job service for creating workspaces. Then, we can ensure that at most one create workspace request is running for a given id. In general, we want users to specify unique id for long running operations and de-dupe logical requests to the same operation. It's a little awkward atm to check the JobService for (is this jobId new, or does the user have access to it, or is the user not authorized.)

zloery: How does stairway handle multiple flights with the same ID - does it throw an exception for the second one, or just queue it and run it after the first one finishes? 

That does seem useful here but if we expect to use that to support concurrency elsewhere, it feels like what we really want is locking.

wchamber: Flight ids need to be unique until they're manually deleted from Stairway. atm, old flights accumulated in stairway db. It will throw an exception since it violates the unique id constraint of its db. We can write a unittest to determine exactly what it looks like atm.

src/main/java/bio/terra/workspace/service/datareference/flight/CreateDataReferenceStep.java
```diff
@@ -57,10 +53,15 @@ public StepResult undoStep(FlightContext flightContext) {
...
 
-    if (workingMap.get(CREATE_DATA_REFERENCE_COMPLETED_KEY, Boolean.class)) {
-      UUID referenceId = inputMap.get(DataReferenceFlightMapKeys.REFERENCE_ID, UUID.class);
-      UUID workspaceId = inputMap.get(DataReferenceFlightMapKeys.WORKSPACE_ID, UUID.class);
+    UUID referenceId = workingMap.get(DataReferenceFlightMapKeys.REFERENCE_ID, UUID.class);
+    UUID workspaceId = inputMap.get(DataReferenceFlightMapKeys.WORKSPACE_ID, UUID.class);
+
+    try {
       dataReferenceDao.deleteDataReference(workspaceId, referenceId);
+    } catch (DataAccessException notFoundEx) {
```
wchamber, line 61: DataAccessException could be things other than "not found." Anything other than "not found" is still an error. Use a narrower exception type, or consider changing the return type of deleteDataReference if we care about distinguishing between something vs nothing being deleted.

src/main/java/bio/terra/workspace/service/datareference/flight/GenerateReferenceIdStep.java
```diff
@@ -0,0 +1,27 @@
+package bio.terra.workspace.service.datareference.flight;
+
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.UUID;
+
+public class GenerateReferenceIdStep implements Step {
```
wchamber, line 10: add javadoc about the purpose of this step and why it's important to do as a separate step. Every public class should have some top level javadoc: https://google.github.io/styleguide/javaguide.html#s7.3-javadoc-where-required

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceAuthzStep.java
wchamber: I think Yonghao's right that we should either check existence before calling delete or handle 404 here. We should ensure that no more than one flight is trying to create a workspace with the same id simultaneously somewhere else, whether that's locking or using the workspaceId as the job/flight id.

I think we will want locking for workspaces, but I'm not sure how much that helps us here - we'd have to lock a workspace before it's created, or create the workspace in a-not-ready-to-get stage. Using JobService/Stairway unique guarantees seems easier, and we will also want that functionality for other operations, but open to other ideas.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
wchamber: ditto, I don't think this is sufficient. Let's do locking or unique operation ids.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
wchamber: For my 2c, I think we should only log.error when we want to be alerted that something is *wrong*. I think the logging level of error vs warning vs info already gets us there and we shouldn't have special alert prefixing to follow on top of that. We will need to be disciplined about the logging level and demote things that aren't actually errors.
So some alerting for any error message (with silencing on a per line level) would be great.

wchamber (changes requested):

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceAuthzStep.java
zloery: The old flights accumulating concerns me. I don't like the idea that WM won't ever be able to create a particular workspace because e.g. Sam was temporarily down when we tried to make the workspace. That's especially bad because we're taking WorkspaceIDs in from requests, since we can't just internally try again with a new ID and clients would now need to know that behavior.

wchamber: Good point. We can include a operation id in the create workspace request so that a client that's trying to do the same logical operation does the same thing, but at a certain point we do need to worry about different logical requests deciding on the same workspace id, or say that's random enough to not worry about.

zloery: I'm not really worried about different logical requests picking the same workspace ID - there aren't duplicate workspace IDs in existing Terra, and if someone tries to create one a 400 response makes sense. In general I expect WM createWorkspace calls to come from a single place (either Rawls or folder manager, depending) so I'm also less worried about systems that don't know they're making simultaneous calls.

Passing in an operation id and (I'm assuming) using that as the flight ID does solve this problem for WM, but it's really bumping it up a layer by expecting clients to manage their logical operations. That might be the right call, but it means a more complicated endpoint and changes to client behavior. I think locking is the right behavior here, even if it means locking on a not-totally-created workspace.

wchamber: While we'll also need locking, I will argue that operation ids are actually simpler for a client that's trying to handle failure. And we can always make the operation id optional and generate it if it's not specified.
If a client of WM doesn't generate it's only operation id, consider what happens when there's a network failure in the call to create the workspace. Now the client has to resend the request to create the workspace. What should they do when they receive a 400 for the second request? Do they try to track that they already send a request to create a workspace and treat 400 as ok in that special case? Do they error out and orphan the workspace? Do they infer that there operation has already begun and try to poll as if they received a 200?
If there is a generated operation id, WM can 200 duplicate requests, deduping the logical requests. The client behavior is always the same no matter which request succeeded in creating the operation in WM.
If the client is a stairway user, e.g. folder manager, generating and persisting the operation id before starting the operation is the familiar pattern, just as we're doing for data reference id generation. In this case, the client is pre-generating the workspace id. We could generate an additional operation id, but the workspace id fulfills the purpose of uniquely identifying the logical operation.
It's true that this chain of operation id generation has to stop at some client, e.g. the user's browser or the client not generating an id at all, but we should be careful when we cutoff the ability to specify an operation id.

zloery: That's a good point about idempotency, especially for clients using stairway. In that case, I'll update the whole request flow and probably bump the client version as well.

zloery:
After discussion on operation IDs for ensuring concurrency control on workspace creation, I'm moving this PR back to draft while I refactor. I'll reopen and add reviewers again when the changes are ready

wchamber (changes requested):
As this gets more extensive, consider breaking apart the data reference & workspace changes. Smaller, more granular PRs are easier to review and get merged in.

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -63,6 +64,11 @@ public void deleteWorkspace(String authToken, UUID id) {
     try {
       resourceApi.deleteResource(SamUtils.SAM_WORKSPACE_RESOURCE, id.toString());
     } catch (ApiException apiException) {
+      // Throw a different exception for 404 responses, as stairway steps will often ignore this but
+      // still care about other exceptions.
+      if (apiException.getCode() == 404) {
+        throw new SamResourceNotFoundException(apiException);
+      }
       throw new SamApiException(apiException);
```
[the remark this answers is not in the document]
zloery: I think the main advantage of wrapping this in our own exception type (extending ErrorReportException) is that we can provide cleaner exception handling in GlobalExceptionHandler. In general I think we should continue to treat SamApiExceptions as unchecked errors that bubble up to the global exception handler, which the other methods of SamService do.

The checked/unchecked distinction makes sense, and in that case I think it makes sense to make `SamResourceNotFoundException` checked. It seems reasonable to me to use SamApiException as an unchecked exception that we don't expect clients to handle, and to create new checked exceptions that we do expect them to handle.

src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -118,6 +120,16 @@ protected String submit(
...
       stairway.submit(jobId, flightClass, parameterMap);
+    } catch (DatabaseOperationException dbEx) {
+      String sqlState = (((SQLException) dbEx.getCause()).getSQLState());
+      // sqlState 23505 indicates a unique key violation, which in this case indicates a duplicate
+      // flightId. See https://www.postgresql.org/docs/10/errcodes-appendix.html for postgres
+      // error codes.
+      if (sqlState.equals("23505")) {
+        throw new DuplicateFlightIdException("Flight ID already exists in Stairway: " + jobId);
+      } else {
+        throw new InternalStairwayException(dbEx);
```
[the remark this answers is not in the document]
zloery: Agree that this is pretty hacky atm. I think the problem with checking for the ID before trying to submit is that this is inside our concurrency control logic - we can't be sure another request with the same ID isn't being processed at the same time. 
I think my ideal solution to this would be a checked exception or a distinct return value from Stairway on duplicate flightID. I'll check with Dan about those ideas later today.

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -47,9 +47,13 @@
           .email("fake@email.com")
           .subjectId("fakeID123");
 
+  /** A unique ID generated for each test to manage concurrent flights. */
```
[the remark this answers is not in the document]
zloery: Inlined things as part of the switch to `WorkspaceRequest`

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -84,21 +88,82 @@ public void testWorkspaceStagePersists() {
...
   public void testWithSpendProfile() {
     UUID workspaceId = UUID.randomUUID();
     Optional<SpendProfileId> spendProfileId = Optional.of(SpendProfileId.create("foo"));
     workspaceService.createWorkspace(
-        workspaceId, spendProfileId, WorkspaceStage.RAWLS_WORKSPACE, USER_REQUEST);
+        workspaceId,
+        spendProfileId,
+        WorkspaceStage.RAWLS_WORKSPACE,
+        UUID.randomUUID().toString(),
+        USER_REQUEST);
```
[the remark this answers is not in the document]
zloery: Agreed. I went ahead and made the `WorkspaceRequest` model we'd discussed before in the layering refactor and gave it a pre-filled builder method for convenience

src/main/java/bio/terra/workspace/service/job/JobService.java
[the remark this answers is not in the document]
wchamber: I don't know that we need to overly worry about the race between the id exists & resubmitting. If we do have concurrent requests that try to submit the same id, one of them will fail thanks to Stairway's guarantee at submission time, so no harm no foul. But any retry will work correctly. So it'd be a little nice for clients to never have the funny 503 error, we do get eventual correctness. I think the more common retry case is that the response to the first request gets lost. The second request won't be concurrent, so will work directly.
+1 checking with Dan.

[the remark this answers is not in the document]
zloery: Followup on this, I'm moving the duplicate flightID check to be a checked exception from Stairway in https://github.com/DataBiosphere/stairway/pull/42. I'll update WM here when that gets merged and published.

yonghaoy:
5 reviewers is too many for a PR, remove myself from reviewer list. Overrall the PR LGTM. Let me know if I can help on anything!

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -68,9 +69,19 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
     Optional<SpendProfileId> spendProfileId =
         Optional.ofNullable(body.getSpendProfile()).map(SpendProfileId::create);
+    // If clients do not provide an operation ID, we generate one instead.
+    String operationId =
+        body.getOperationId() != null ? body.getOperationId() : UUID.randomUUID().toString();
+
+    WorkspaceRequest internalRequest =
+        WorkspaceRequest.builder()
+            .workspaceId(body.getId())
+            .operationId(operationId)
```
wchamber, line 79: nit: We've been inconsistent so far (sorry), but we should probably pick a single term for 'job' vs 'operation'. No time like the present, so let's do 'jobId' here to correspond with the job service.

src/main/java/bio/terra/workspace/service/iam/SamService.java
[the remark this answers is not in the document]
wchamber: It's still a bit of a smell that this is the only method that throws the not found exception. If we want to handle this everywhere, the other methods should throw as well. If it is a special case exception, let's add the ApiException to SamApiException to expose the status code and check it only in the workspace delete.

src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -114,10 +115,13 @@ public JobBuilder newJob(
...
-  protected String submit(
-      Class<? extends Flight> flightClass, FlightMap parameterMap, String jobId) {
+  protected String submit(Class<? extends Flight> flightClass, FlightMap parameterMap, String jobId)
+      throws DuplicateFlightIdSubmittedException {
     try {
       stairway.submit(jobId, flightClass, parameterMap);
+    } catch (DuplicateFlightIdSubmittedException ex) {
+      // This needs to be thrown separately as it extends the more general StairwayException
+      throw ex;
     } catch (StairwayException | InterruptedException stairwayEx) {
```
wchamber, line 125: Probably not this PR, but it smells that JobService throws unchecked exceptions for some but not all of Stairway. It's leaking the Stairway API significantly, but still wrapping it. JobService should probably be pushed one way or another - more wrapping or less.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -48,29 +50,37 @@ public WorkspaceService(
 
   /** Create a workspace with the specified parameters. Returns workspaceID of the new workspace. */
   @Traced
-  public UUID createWorkspace(
-      UUID workspaceId,
-      Optional<SpendProfileId> spendProfileId,
-      WorkspaceStage workspaceStage,
-      AuthenticatedUserRequest userReq) {
+  public UUID createWorkspace(WorkspaceRequest internalRequest, AuthenticatedUserRequest userReq) {
```
wchamber, line 53: not wild about `internalRequest` variable name. What's wrong with `workspaceRequest` or `request`? At this layer, how is this more `internal` than anything else?

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -48,29 +50,37 @@ public WorkspaceService(
...
+    try {
+      return createJob.submitAndWait(UUID.class);
+    } catch (DuplicateFlightIdSubmittedException ex) {
+      // Indicates another request with the same operation ID already exists. Rather than running
+      // multiple concurrent create flights, all flights after the first will wait for and return
+      // the same result.
+      jobService.waitForJob(internalRequest.operationId());
+      return jobService
+          .retrieveJobResult(internalRequest.operationId(), UUID.class, userReq)
+          .getResult();
```
wchamber, line 82: should this be encapsulated by `JobService`? we could make it a part of submitAndWait, or another function that does this whole try catch. (Encapsulating would fix some of the leakage mentioned above.)

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -96,7 +106,12 @@ public void deleteWorkspace(UUID id, AuthenticatedUserRequest userReq) {
                 null, // Delete does not have a useful request body
                 userReq)
             .addParameter(WorkspaceFlightMapKeys.WORKSPACE_ID, id);
-    deleteJob.submitAndWait(null);
+    try {
+      deleteJob.submitAndWait(null);
+    } catch (DuplicateFlightIdSubmittedException ex) {
+      // FlightIDs used here are random UUIDs, so a collision is purely bad luck.
+      throw new InternalStairwayException(ex);
+    }
```
wchamber, line 114: if this is more common than not, it seems like our jobservice isn't pulling enough weight for us. Leaning more towards separate functions or parameter for allowing/failing duplicate job ids.

src/main/java/bio/terra/workspace/service/workspace/model/WorkspaceRequest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * Internal representation of a request to create a workspace.
+ *
+ * <p>While this object is nearly identical to the {@code Workspace} object, there is a semantic
+ * difference: a {@code Workspace} is a fully formed workspace, while a {@code WorkspaceRequest}
+ * only contains fields specified by clients. These objects may diverge in the future if properties
+ * of a Workspace are generated internally.
```
wchamber, line 14: the future is now with operation id. I think noting the semantic difference is enough. This is also a comment that's likely to not get updated to be correct when the future arrives.

src/main/java/bio/terra/workspace/service/workspace/model/WorkspaceRequest.java
```diff
@@ -0,0 +1,70 @@
...
+   *
+   * <p>This provides default values for operationID (random UUID), spend profile
+   * (Optional.empty()), and workspace stage (RAWLS_WORKSPACE).
+   */
+  public static WorkspaceRequest.Builder defaultRequestBuilder(UUID workspaceId) {
+    return builder()
+        .workspaceId(workspaceId)
+        .operationId(UUID.randomUUID().toString())
+        .spendProfileId(Optional.empty())
+        .workspaceStage(WorkspaceStage.RAWLS_WORKSPACE);
```
wchamber, line 68: I don't think this is a good idea to put in the `main` code. I think this is great for a test utility, but makes it more likely we don't fix places where this is created when we add more fields in the main code.
Tests don't care about details unrelated to what they're verifying. Main code should be explicit, unless there really are reasonable defaults. I don't think these are reasonable defaults in our main code.

src/main/resources/api/service_openapi.yaml
```diff
@@ -401,6 +401,14 @@ components:
           items:
             type: string
             format: uuid
+        operationId:
```
wchamber, line 404: jobId? and in comment

wchamber (changes requested):
thanks for the changes to Stairway.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
zloery: we're already using `jobId` in JobService, so I think that's the right standard

src/main/java/bio/terra/workspace/service/job/JobService.java
zloery: For the most part I think we should generally be wrapping Stairway (`InternalStairwayException` is a WM type) -  I'll change `DuplicateJobIdSubmittedException` to stay inside JobService.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
zloery: `internalRequest` makes sense to me at the controller level and I plumbed the same name through here. `workspaceRequest` sgtm

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
zloery: I'm adding a param here, I'd prefer not to break out another function when we already have both `submit` and `submitAndWait`. 

We could also consider a small refactor of JobBuilder to accept jobId as optional on new jobs and generate an ID if one isn't provided. That puts flightID generation all in one place instead of generating random UUIDs across WM, and also lets us make the distinction of "provided duplicate flightID -> idempotent behavior", "generated duplicate flightID -> (rare) error".

I think that's getting a little out of scope and this PR is already large though, so I'll leave that as a separate issue.

src/main/java/bio/terra/workspace/service/workspace/model/WorkspaceRequest.java
zloery: sg, moved this into WorkspaceServiceTest.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
zloery: moved into submit functions per below

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
wchamber: I like the idea of moving it into JobBuilder. Another PR sgtm.
(fwiw, another thing that's bothered me about the current JobBuilder API is that it wraps setting up the Flight input parameters so that the code that relies on the input parameters (the flight & its step) is in a totally separate place than the code that sets up the job and sets the parameters. Totally unrelated to what's here, but one of my pet peeves if/when you do take a look)

src/main/java/bio/terra/workspace/common/exception/SamApiException.java
```diff
@@ -35,4 +38,8 @@ public SamApiException(
       String message, Throwable cause, List<String> causes, HttpStatus statusCode) {
     super(message, cause, causes, statusCode);
   }
+
+  public int getApiExceptionStatus() {
```
wchamber, line 42: javadoc. int is not a descriptive return type, so worth a comment

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -306,11 +303,25 @@ public void createGoogleContextSpendWithoutBillingAccountThrows() {
...
+  /**
+   * Generate a fully built Workspace object with a random ID and default values in all fields.
    *
-   * <p>See {@code WorkspaceRequest.defaultRequestBuilder} for default values of each field.
+   * <p>See {@code defaultRequestBuilder} for default values of each field.
    */
   private WorkspaceRequest defaultWorkspaceRequest() {
-    return WorkspaceRequest.defaultRequestBuilder(UUID.randomUUID()).build();
+    return defaultRequestBuilder(UUID.randomUUID()).build();
   }
```
wchamber: multiple factory functions when there is a builder object is an anti-pattern that's caused me personal grief before. Let's pick one way to make a `default` `WorkspaceRequest.Builder` and customize that as appropriate. i.e. remove the (UUID workspaceId) argument method and call `defaultWorkspaceRequest().workspaceId(foo)` if needed.

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -58,7 +59,7 @@ public void createWorkspaceWithDefaults(String authToken, UUID id) {
     }
   }
 
-  public void deleteWorkspace(String authToken, UUID id) {
+  public void deleteWorkspace(String authToken, UUID id) throws SamResourceNotFoundException {
```
wchamber, line 62: I don't think this is throwing the exception anymore. let's remove the checked exception from the method, and I think the exception file itself.

src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -54,14 +54,15 @@ public JobBuilder addParameter(String keyName, Object val) {
   }
 
   // submits this job to stairway and returns the jobId immediately
-  public String submit() {
-    return jobServiceRef.submit(flightClass, jobParameterMap, jobId);
+  public String submit(boolean duplicateFlightOk) {
```
wchamber, line 57: Let's fix the comment to be javadoc on this method and explain what `boolean duplicateFlightOk` means. Ditto on submitAndWait.
If this is truly very short term until the next PR, we can leave this as a boolean. Otherwise, consider making an enum to make this readable at callsites and easier to document. Boolean parameters should be used exceedingly sparingly. go/java-practices/boolean-parameters
A TODO to move this into the JobBuilder would make me happy.

src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -115,9 +117,18 @@ public JobBuilder newJob(
...
-      Class<? extends Flight> flightClass, FlightMap parameterMap, String jobId) {
+      Class<? extends Flight> flightClass,
+      FlightMap parameterMap,
+      String jobId,
+      boolean duplicateFlightOk) {
     try {
       stairway.submit(jobId, flightClass, parameterMap);
+    } catch (DuplicateFlightIdSubmittedException ex) {
+      if (duplicateFlightOk) {
+        return jobId;
```
wchamber, line 128: probably worth some explanatory comment here or elsewhere on why this is nice behavior to have.
also, the catch ordering matters and that's not immediate obvious, so worth a warning comment.

wchamber (approved):
thanks! Last few nits

src/main/java/bio/terra/workspace/service/job/JobService.java
zloery: sg. I think it'll be worth adding a slightly longer explanation in the JobBuilder, but I'm less concerned about inside a protected method.

src/main/java/bio/terra/workspace/service/job/JobBuilder.java
zloery: Fixed javadoc and added TODO, I'd rather handle this more cleanly in JobBuilder (might still need an enum as part of that change, tbd).

src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: Yep, good catch.

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
zloery: I think that points towards leaving the builder and removing the `defaultWorkspaceRequest` since AutoValue objects are immutable, but sticking to a single factory function makes sense.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceAuthzStep.java
```diff
@@ -51,10 +47,15 @@ public StepResult doStep(FlightContext flightContext) throws RetryException {
...
     FlightMap inputMap = flightContext.getInputParameters();
     // Only delete the Sam resource if we actually created it in the do step.
-    FlightMap workingMap = flightContext.getWorkingMap();
-    if (workingMap.get(AUTHZ_COMPLETED_KEY, Boolean.class)) {
-      UUID workspaceID = inputMap.get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
+    UUID workspaceID = inputMap.get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
+    try {
       samService.deleteWorkspace(userReq.getRequiredToken(), workspaceID);
+    } catch (SamApiException ex) {
+      // Do nothing if the resource to delete is not found, this may not be the first time undo is
```
davidangb, line 54: same here - it could be worth logging the do-nothing case to see how often this happens.

davidangb (commented):
a few comments inline

build.gradle
```diff
@@ -67,7 +67,7 @@ repositories {
 dependencies {
 	// Terra deps
 	implementation group: "bio.terra", name: "datarepo-client", version: "1.0.1-SNAPSHOT"
-	implementation group: "bio.terra", name: "stairway", version: "0.0.20"
+	implementation group: "bio.terra", name: "stairway", version: "0.0.25-SNAPSHOT"
```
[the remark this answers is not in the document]
zloery: Not at the moment - I can talk to Dan about getting one published if needed though.
(side note, even SNAPSHOT versions for stairway get published by Github actions, so we don't have to worry as much about actual deps being overriden)

davidangb (approved):
lgtm, I did not test anything myself

zloery:
(Client library updated in #153, and Rawls dependency bumped in https://github.com/broadinstitute/rawls/pull/1317)

PR #143, opened 2020-11-09
Separate API objects from internal objects in WorkspaceService

zloery:
This PR creates a new `Workspace` object for internal representation of workspaces and separates that internal representation from API objects. All changes happen at the Controller level and below, so this does not require an update to the client library.

This is the first of a few changes to apply this layering pattern across all of WM.

### DIFF
--- a/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -66,14 +66,12 @@ public class WorkspaceApiController implements WorkspaceApi {
     WorkspaceStageModel requestStage = body.getStage();
     requestStage = (requestStage == null ? requestStage.RAWLS_WORKSPACE : requestStage);
     WorkspaceStage internalStage = WorkspaceStage.fromApiModel(requestStage);
-    Optional<SpendProfileId> spendProfile =
-        body.getSpendProfile() == null
-            ? Optional.empty()
-            : Optional.of(SpendProfileId.create(body.getSpendProfile()));
+    Optional<SpendProfileId> spendProfileId =
+        Optional.ofNullable(body.getSpendProfile()).map(SpendProfileId::create);
 
-    Workspace createdWorkspace =
-        workspaceService.createWorkspace(body.getId(), spendProfile, internalStage, userReq);
-    CreatedWorkspace responseWorkspace = new CreatedWorkspace().id(createdWorkspace.workspaceId());
+    UUID createdId =
+        workspaceService.createWorkspace(body.getId(), spendProfileId, internalStage, userReq);
+    CreatedWorkspace responseWorkspace = new CreatedWorkspace().id(createdId);
     return new ResponseEntity<>(responseWorkspace, HttpStatus.OK);
   }
 
--- a/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -1,5 +1,6 @@
 package bio.terra.workspace.app.controller;
 
+import bio.terra.workspace.common.model.Workspace;
 import bio.terra.workspace.common.model.WorkspaceStage;
 import bio.terra.workspace.common.utils.ControllerValidationUtils;
 import bio.terra.workspace.generated.controller.WorkspaceApi;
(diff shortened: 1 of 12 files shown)


src/main/java/bio/terra/workspace/common/model/Workspace.java
```diff
@@ -0,0 +1,47 @@
...
+
+  /** The globally unique identifier of this workspace */
+  public abstract UUID workspaceId();
+
+  /**
+   * The spend profile ID associated with this project, if one exists.
+   *
+   * In the future, this will correlate to a spend profile in the Spend Profile Manager. For now,
+   * it's just a unique identifier. Accounts with an associated GCP project must have a spend
+   * profile, while Rawls workspaces do not need one.
```
wchamber, line 26: nit: on the causal relation: `To associate a GCP project with a workspace, the workspace must have a spend profile. They are not needed otherwise.`

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -63,16 +66,27 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     WorkspaceStageModel requestStage = body.getStage();
     requestStage = (requestStage == null ? requestStage.RAWLS_WORKSPACE : requestStage);
     WorkspaceStage internalStage = WorkspaceStage.fromApiModel(requestStage);
-    return new ResponseEntity<>(
-        workspaceService.createWorkspace(body, internalStage, userReq), HttpStatus.OK);
+    Optional<SpendProfileId> spendProfileId =
+        Optional.ofNullable(body.getSpendProfile()).map(SpendProfileId::create);
+
+    UUID createdId =
+        workspaceService.createWorkspace(body.getId(), spendProfileId, internalStage, userReq);
```
wchamber, line 73: consider whether `createWorkspace` should take `(Workspace workspace, AuthenticatedUserRequest userReq)` as it's arguments. similarly for workspaceDao.createWorkspace. WorkspaceService API already includes the 'Workspace' type; right now there's nothing else needed or extraneous to creating a workspace. That could change later (more likely in WorkspaceService than in WorkspaceDao), but the parallel atm is correct.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -31,33 +31,33 @@ public WorkspaceService(JobService jobService, WorkspaceDao workspaceDao, SamSer
   }
 
   @Traced
-  public CreatedWorkspace createWorkspace(
-      CreateWorkspaceRequestBody body,
+  public UUID createWorkspace(
```
wchamber, line 34: add javadoc for this method. At least saying what is being returned since `UUID` isn't very informative.

wchamber (changes requested):
mostly lgtm, couple comments.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
zloery: I was mainly thinking about this from the perspective of "will the controller always have all information needed to make a workspace?" If not (e.g. if we generate something midway through a flight), I'm not thrilled with passing around a slightly-invalid workspace object. That doesn't really hold for workspaceDao though, I'll change it there.

wchamber: Totally agree on the slightly invalid workspace object being bad idea. It may be worth creating a WorkspaceRequest kind of internal rep that's distinct from the `Workspace` class when we get there, but I can't bring myself to have it yet when they don't differ. But yes, does need discipline later and agreed on workspaceDao.

src/test/java/bio/terra/workspace/db/WorkspaceDaoTest.java
```diff
@@ -137,18 +152,27 @@ public void deleteNonExistentWorkspaceFails() throws Exception {
 
   @Test
   public void duplicateWorkspaceFails() throws Exception {
-    workspaceDao.createWorkspace(workspaceId, Optional.empty(), WorkspaceStage.RAWLS_WORKSPACE);
+    Workspace workspace =
+        Workspace.builder()
+            .workspaceId(workspaceId)
+            .workspaceStage(WorkspaceStage.RAWLS_WORKSPACE)
+            .build();
```
wchamber, line 159: optional: For the tests where the exact shape of the `Workspace` isn't important (just that a `Workspace` exists), consider making a `defaultWorkspace` private variable or function to save some boilerplate and focus the lines in the test on what matters to them. As/if we add more fields to `Workspace` this becomes more helpful, whereas with 2 fields it's not as relevant. Ref: https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=java#constants

wchamber (approved):

PR #144, opened 2020-11-09
PF-186 Update the API for spend profile ids to be strings instead of UUIDs.

wchamber:
Change WM representation of a "spend profile id" from a UUID to a String. As-is, WM is dictating that spend profile ids have to look like UUIDs, but WM should be agnostic to SPM's id format.

This does change the API and client library, but nobody other than our own tests are using the spend profile id fields on create/get workspace. This will also invalidate any outstanding Stairway create workspace flights with the spend profile id set, but we really don't have any of those since the field is never used.
I will update Rawls after this is merged and a new client is built.
(couple misc spotlessApply changes in this PR. My fault from previous commit.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ sourceCompatibility = JavaVersion.VERSION_11
 
 allprojects {
 	group = "bio.terra"
-	version = "0.4.0-SNAPSHOT"
+	version = "0.5.0-SNAPSHOT"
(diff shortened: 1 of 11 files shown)


davidangb (approved):
:+1: changes look good, I didn't check if this is comprehensive

zloery (approved):
LGTM

PR #145, opened 2020-11-10
AS-369: Add more logging

ansingh7115:


src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -151,9 +181,14 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
       @Valid @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    logger.info(String.format("Getting data references in workspace %s for %s",
+            id.toString(), userReq.getEmail()));
     ControllerValidationUtils.validatePaginationParams(offset, limit);
     DataReferenceList enumerateResult =
-        dataReferenceService.enumerateDataReferences(id, offset, limit, getAuthenticatedInfo());
+        dataReferenceService.enumerateDataReferences(id, offset, limit, userReq);
+    logger.info(String.format("Getting data references in workspace %s for %s",
+            enumerateResult.toString(), userReq.getEmail()));
```
davidangb, line 191: nit: should be "Got data references ... "

davidangb (approved):
looks like this needs a rebase :(

one nit inline on a log message, otherwise looks good. We may need to tweak logging down the road but I think this is a good level to have for launch. I don't need to re-review if all you change is that nit.

Make sure platform foundation team sees this, so they know the level of logging we're trying to achieve?

for brainstorming/future thought:
* should we add a Marker (e.g. "AUDIT") to all of these log statements? https://stackoverflow.com/questions/16813032/what-are-markers-in-java-logging-frameworks-and-what-is-a-reason-to-use-them ... that might make it easier to filter down the road.
* if we ever decided that all of these log statements should be debug instead of info ... would we just go update the code? Or should we route all of them through some centralized log function so that we can update it in one place? Is this premature optimization?

PR #146, opened 2020-11-10
update deps

zarsky-broad:
No notable changes with any of the updates.

Change to openapi yaml file is due to change to swagger codegen for spring, which requires a min/max to be defined when providing a default value.

(all tests passed locally...?)


wchamber (approved):
Just curious, anything prompt this update?

zarsky-broad:
> Just curious, anything prompt this update?

Nope! I just like to stay on top of dep updates periodically. It's sort of a palate cleanser for me, so I do these from time to time.

zloery (approved):

PR #147, opened 2020-11-11
PF-186 Add spend authz check and billing account setup to create google context flight.

wchamber:
Check that a workspace has a valid spend profile id and that the user is authorized to link it before creating the google context. Set up that billing account on the project as a part of creating the google context. This allows the created project to spend money.
For now, only have tests with mocked sam spend link interaction.

### DIFF
--- a/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteGoogleContextFlightTest.java
+++ b/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteGoogleContextFlightTest.java
@@ -36,8 +36,6 @@ public class DeleteGoogleContextFlightTest extends BaseConnectedTest {
   @Test
   public void deleteContext() throws Exception {
     UUID workspaceId = createWorkspace();
-
-    // Both creating and deleting the google context happen to only need the workspace id as input.
     FlightMap createParameters = new FlightMap();
     createParameters.put(WorkspaceFlightMapKeys.WORKSPACE_ID, workspaceId);
     createParameters.put(
--- a/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteGoogleContextFlightTest.java
+++ b/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteGoogleContextFlightTest.java
@@ -9,6 +9,7 @@ import bio.terra.workspace.common.StairwayTestUtils;
 import bio.terra.workspace.common.model.WorkspaceStage;
 import bio.terra.workspace.db.WorkspaceDao;
 import bio.terra.workspace.service.job.JobService;
+import bio.terra.workspace.service.spendprofile.SpendConnectedTestUtils;
 import bio.terra.workspace.service.workspace.WorkspaceCloudContext;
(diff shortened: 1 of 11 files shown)


src/test/java/bio/terra/workspace/service/workspace/flight/DeleteGoogleContextFlightTest.java
```diff
@@ -30,21 +31,24 @@
   @Autowired private WorkspaceDao workspaceDao;
   @Autowired private CloudResourceManagerCow resourceManager;
   @Autowired private JobService jobService;
+  @Autowired private SpendConnectedTestUtils spendUtils;
 
   @Test
   public void deleteContext() throws Exception {
     UUID workspaceId = createWorkspace();
 
     // Both creating and deleting the google context happen to only need the workspace id as input.
```
zloery, line 40: nit: remove comment

src/test/resources/application-connected-test.yml
```diff
@@ -1,3 +1,13 @@
...
+  spend:
+    spend-profiles:
+      -
+        # TODO(PF-186): Create a meaningful spend profile id with a resource in Sam.
+        id: foo
+        # The billing account workspace-dev has access to.
+        billing-account-id: 01A82E-CA8A14-367457
+      -
+        # A special spend profile with no billing account associated with it.
+        id: no-billing-account
```
zloery, line 13: nothing is reading this account atm, do we need it?

zloery (approved):
LGTM, but as a heads up I just merged the WorkspaceService layering refactor in #143 so this may have conflicts now.

src/test/resources/application-connected-test.yml
wchamber: This is being used in the [WorkspaceServiceTest for handling a spend profile without a billing account](https://github.com/DataBiosphere/terra-workspace-manager/blob/8883fabc7226b8b43f59f8306b16bff967d2a59f/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java#L287). (The SpendProfileService is being created with this id so that it's available. If we didn't have this spend profile in the config, we would not be able to get a spend profile from the service without a billing account.) Currently, this special id is hardcoded in the spend connected test utils. I agree the connection isn't obvious. I'm open to other alternatives.

zloery: Ah good point - I saw it there but didn't follow the check all the way into the SpendProfileService.

wchamber:
@zloery So I think the Mockito flaky issue is due to the status service regularly (every 60s) calling the mocked Sam service. Mockito doesn't like concurrent modifications and calls to a mock (https://github.com/mockito/mockito/wiki/FAQ#is-mockito-thread-safe). How mockito decides to fail seems up in the air. The test that failed for this PR was the test that creates/deletes a project, taking minutes.
I'd like to disable the status service scheduled checks during tests in another PR to avoid concurrent modification of the mocked sam service. I don't think this PR is making anything worse.

zloery:
Sounds good. Disabling the status service for tests should be trivial, I'll make that change now

PR #149, opened 2020-11-12
Rework statusService to stop running scheduled calls in tests

zloery:
This change reworks how the StatusService works. Currently, the service uses Spring's `@Scheduled` annotation to check the status of dependencies every 60s. This causes problems in tests: dependencies are often mocked and Mockito (our mocking library) can't handle parallel calls to mocks. Long running or unlucky tests would sometimes hit the mocks at the same time as the Status service, causing errors.

The new approach creates a StatusService interface and uses our existing Spring profiles to determine whether we use a fake implementation (for tests) or a real scheduled one (for everything else). This also removes tests on Workspace Manager's implementation of BaseStatusService, as it was only testing status calls against mocks.


wchamber:
I'll counter with https://github.com/DataBiosphere/terra-workspace-manager/pull/148. I'll take a look at what you have & let's talk offline tomorrow.

src/main/java/bio/terra/workspace/service/status/BaseStatusService.java
```diff
@@ -1,23 +1,25 @@
...
+/**
+ * BaseStatusService is a Spring replacement for workbench-libs' HealthMonitor utilities.
+ *
+ * <p>It checks status information from subsystems asynchronously at regular intervals and provides
+ * a cached version of the latest statuses to support a high-traffic status endpoint. It also tracks
+ * time since the last update and returns an unhealthy status if subsystems are not checked after
+ * some amount of time, which indicates that something has gone wrong.
+ *
+ * <p>Specific services should extend this class with Component objects that register the
+ * appropriate subsystems.
```
wchamber, line 20: I don't really understand why specific services need to extend the class (is-a) instead of having a member variable BaseStatusService (has-a). But if we're saying this needs to be extended, should it be abstract?

src/main/java/bio/terra/workspace/service/status/StatusService.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * An interface for a service providing status information about this server.
+ *
+ * <p>This has two implementations in Workspace Manager. The first, ScheduledStatusService, is used
+ * for all deployments of the service. It checks all dependencies on a schedule and reports cached
+ * statuses via endpoints.
+ *
+ * <p>The second, FakeStatusService, is only used for tests and always returns an OK status with no
+ * subsystems. This is useful for preventing unexpected interactions between scheduled methods and
+ * tests.
```
wchamber, line 14: I don't think the interface is the right place for this comment. It probably won't get updated if this info changes, and the implementors can be found and have their own javadoc.
I think it's worth moving the comment about why we have a FakeStatusService to its class. I think BaseStatusService has most of what's needed in its javadoc already.

wchamber (commented):
I'm pretty ambivalent about the more spring-y profile enabled components vs manipulating the cron value. I do feel like we now have 4 classes instead of 1 just because we want to use `@Scheduled`, which is quite a bit of overhead.
Yesterday before doing the `cron =` I also experimented with adding a parameter to effectively disable the scheduling. It also feels like a lot. https://github.com/DataBiosphere/terra-workspace-manager/compare/wc-mockito-flake
WDYT? (talk offline for a few minutes?)

src/main/java/bio/terra/workspace/service/status/BaseStatusService.java
zloery: My thinking on this was that each service has a specific implementation of the BaseStatusService - each of those implementations is a BaseStatusService with a particular set of dependencies. A big driver for this pattern was separating the parts common to other services (which would live in a common library somewhere if we had one) from the actual WM dependencies. When this was written TDR was planning to use it as well, I don't know if that ever happened or got deprioritized.

Yeah, it should probably be abstract.

zloery:
Closing in favor of #148. The pattern of autowiring interfaces and deciding implementation with Spring profiles might be useful in the future, but we should avoid it where possible to limit complexity in the number of components we have.

PR #150, opened 2020-11-13
PF-227 Fix stairway working map interactions in DataReferenceService

zloery:
(These changes were originally in #142, I'm splitting them out here as the WorkspaceService-specific changes get more complex)

This PR makes changes related to the way we're using Stairway's working map. Working map updates are persisted at step boundaries, so the previous pattern of "set flag to false, perform action, set flag to true" within a single step did not work.

This breaks generating a data reference ID into a separate step, which allows us to handle undos more cleanly. Generating IDs as an early step in a flight is a best-practice when using Stairway.

This portion of the changes doesn't touch any objects in the interface layer, and so does not require an update to the client library.


zloery:
The connected test failures seem unrelated to these changes and are also happening for me locally on dev - investigating now.

wchamber (approved):

davidangb (approved):

PR #151, opened 2020-11-13
PF-153 Add endpoints create/delete google context.

wchamber:
Also add check for MC_WORKSPACE stage before creating the google context.

Things not in this PR.
- Jobs API. pending further discussion.
- Integration tests. Need Sam resources and means to use the client library for integration testing. I don't have strong feelings on the ordering of the tests vs this PR.
- Job id idempotency on create google context. Going to wait for Zach's work to add this for creating workspaces.

### DIFF
--- a/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -4,20 +4,15 @@ import bio.terra.workspace.common.model.Workspace;
 import bio.terra.workspace.common.model.WorkspaceStage;
 import bio.terra.workspace.common.utils.ControllerValidationUtils;
 import bio.terra.workspace.generated.controller.WorkspaceApi;
-import bio.terra.workspace.generated.model.CreateDataReferenceRequestBody;
-import bio.terra.workspace.generated.model.CreateWorkspaceRequestBody;
-import bio.terra.workspace.generated.model.CreatedWorkspace;
-import bio.terra.workspace.generated.model.DataReferenceDescription;
-import bio.terra.workspace.generated.model.DataReferenceList;
-import bio.terra.workspace.generated.model.ReferenceTypeEnum;
-import bio.terra.workspace.generated.model.WorkspaceDescription;
-import bio.terra.workspace.generated.model.WorkspaceStageModel;
+import bio.terra.workspace.generated.model.*;
 import bio.terra.workspace.service.datareference.DataReferenceService;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequestFactory;
 import bio.terra.workspace.service.job.JobService;
 import bio.terra.workspace.service.spendprofile.SpendProfileId;
+import bio.terra.workspace.service.workspace.WorkspaceCloudContext;
 import bio.terra.workspace.service.workspace.WorkspaceService;
+import java.net.URI;
 import java.util.Optional;
 import java.util.UUID;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -170,4 +169,25 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
         dataReferenceService.enumerateDataReferences(id, offset, limit, getAuthenticatedInfo());
     return ResponseEntity.ok(enumerateResult);
   }
+
+  @Override
+  public ResponseEntity<JobModel> createGoogleContext(
```
zloery, line 174: It's going to be weird to test this (for you and also for security team) prior to changing or at least re-enabling Jobs endpoints - the functionality is all there, but you have to poll `getWorkspace` for cloud context info instead. Doesn't need to be changed, but something to keep in mind

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -110,6 +111,10 @@ public WorkspaceCloudContext getCloudContext(UUID workspaceId, AuthenticatedUser
   @Traced
   public String createGoogleContext(UUID workspaceId, AuthenticatedUserRequest userReq) {
     samService.workspaceAuthz(userReq, workspaceId, SamUtils.SAM_WORKSPACE_WRITE_ACTION);
+    Workspace workspace = workspaceDao.getWorkspace(workspaceId);
```
zloery, line 114: There's also a `getWorkspaceStage` in the DAO which could be useful here

src/main/java/bio/terra/workspace/service/workspace/exceptions/StageDisabledException.java
```diff
@@ -0,0 +1,14 @@
+package bio.terra.workspace.service.workspace.exceptions;
+
+import bio.terra.workspace.common.exception.BadRequestException;
+import bio.terra.workspace.common.model.Workspace;
+
+/** Exception for when an operation on workspace is not allowed according to its stage. */
+public class StageDisabledException extends BadRequestException {
+  public StageDisabledException(Workspace workspace, String operationName) {
```
zloery, line 8: Maybe pass workspace ID and stage instead of a whole workspace? I could go either way on this one

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -110,6 +111,10 @@ public WorkspaceCloudContext getCloudContext(UUID workspaceId, AuthenticatedUser
...
+    if (!WorkspaceStage.MC_WORKSPACE.equals(workspace.workspaceStage())) {
```
zloery, line 115: Should this check be in `deleteGoogleContext` too? Right now it would just return success for a Rawls workspace which is fine, but slightly less informative than a `StageDisabledException`.

src/main/resources/api/service_openapi.yaml
```diff
@@ -175,6 +175,49 @@ paths:
...
+                type: string
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/JobModel'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '500':
```
zloery, line 207: Both operations can also throw a 404 if the workspace doesn't exist (or the user has no permission on it)

src/main/resources/api/service_openapi.yaml
```diff
@@ -540,6 +593,18 @@ components:
       type: string
       enum: ['RAWLS_WORKSPACE', 'MC_WORKSPACE']
 
+    CreateGoogleContextRequestBody:
+      description: Request body object for creating the Google context.
+      type: object
+      required: [name, cloningInstructions]
```
zloery, line 599: this doesn't have name or cloningInstructions fields

src/main/java/bio/terra/workspace/service/workspace/exceptions/StageDisabledException.java
wchamber: Yeah me too. Switching it.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
wchamber: Good point. done.

src/main/resources/api/service_openapi.yaml
wchamber: Done.

src/main/resources/api/service_openapi.yaml
wchamber: oops

zloery (approved):
LGTM

PR #154, opened 2020-11-19
[PF-226] Refactor DataReferenceService into layers

zloery:
This is the second and final PR separating Workspace Manager into logical "layers" of objects. This separates out the Swagger objects used in the external interface from internal objects used in services and DAOs, with all translation taking place in the Controller.

This PR makes the following notable changes:

1. Create new models for internal representation of data references, separate from the generated Swagger models. This also lets us model referenceObject subtypes more naturally.
2. Move layer translation logic and request validation from services to the Controller.
3. Move DataReferenceService to use our internal models instead of Swagger models.
4. Stop serializing a Swagger model to stairway as part of reference creation, and instead directly serialize all the fields we care about.
5. Remove DataReferenceDao logic for handling controlled resources. We already disallow this via endpoints, and I think it makes more sense to revisit this when we add controlled resources than to leave it half-implemented.
6. Add a serialization class for reference objects to DataReferenceDao. This will require recreating DBs in current deployments, but should prevent us from having to do so again in the future.
7. Update all DataReferenceDao tests to the new interface.
8. Rewrite DataReferenceServiceTest to use `DataReferenceService` directly instead of unit tests making mock HTTP requests.

While 5 and 8 could potentially be broken into a separate PR with some intermediate state, the rest are too closely tied together to happen piecemeal. I think that's more work than it's worth, but open to other opinions.

I'd also like to get thoughts on a few points:
1. Does the `ReferenceObject` and `SnapshotReference` model reflect what we want, and is this a reasonable way to structure objects?
2. In general, should we be serializing complex objects directly to stairway, or should we serialize each of their fields separately? Making autovalue classes serialize/deserialize neatly with Jackson is nontrivial, but the alternative is large blocks of individual fields being set and read.


src/main/java/bio/terra/workspace/service/datareference/model/SnapshotReference.java
```diff
@@ -0,0 +1,42 @@
...
+  public static SnapshotReference create(String instanceName, String snapshot) {
+    return new AutoValue_SnapshotReference(instanceName, snapshot);
+  }
+
+  public abstract String instanceName();
+
+  public abstract String snapshot();
+
+  @Override
+  public Map<String, String> getProperties() {
```
zloery, line 30: TODO: is it ok to return an ImmutableMap here? It seems like it leaves us open to accidentally trying to mutate this map in the future. OTOH Jackson doesn't know how to deserialize an ImmutableMap, so making this an `ImmutableMap` in `ReferenceObject` would require some working around.

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * An interface representing the subject of a data reference.
+ *
+ * <p>All referenced objects are read-only value objects described by a set of string key-value
+ * pairs. The set of keys available are specific to each referenced object type, as are validation
+ * rules. ReferenceObject implementations must be fully serializable by the output of getProperties.
+ */
+public interface ReferenceObject {
+  Map<String, String> getProperties();
```
wchamber, line 14: I don't really understand what getProperties is for. Is this for serialization only? Or is there something inherent about ReferenceObjects that they are maps of strings to strings?
If it's only for serialization in FlightMaps or Databases, we can accomplish that in other ways, e.g. toJson & fromJson abstract methods.

This is kind of a funny interface in that there's no behavior on it that's about what ReferenceObjects are for. Not necessarily wrong, I'm just not sure what WM actually does with ReferenceObjects given this interface. I think it's serialization at the request & db layers, but there aren't obvious methods for that.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -1,56 +1,67 @@
...
 
-  public DataReferenceValidationUtils(DataRepoService dataRepoService) {
-    this.dataRepoService = dataRepoService;
-  }
+  /**
+   * Names must be 1-63 characters long, and may consist of alphanumeric characters and underscores
+   * (but may not start with an underscore). These restrictions match TDR snapshot name
+   * restrictions.
+   */
+  public static final Pattern NAME_VALIDATION_PATTERN =
```
wchamber, line 24: Why do WM data reference names need to match snapshot name restrictions?

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -1,56 +1,67 @@
...
     }
   }
 
-  private void validateDataRepoReference(DataRepoSnapshot ref, AuthenticatedUserRequest userReq) {
-    if (StringUtils.isBlank(ref.getInstanceName()) || StringUtils.isBlank(ref.getSnapshot())) {
+  private void validateSnapshotReference(SnapshotReference ref, AuthenticatedUserRequest userReq) {
+    if (StringUtils.isBlank(ref.instanceName()) || StringUtils.isBlank(ref.snapshot())) {
       throw new InvalidDataReferenceException(
           "Invalid Data Repo Snapshot identifier: "
               + "instanceName and snapshot must both be provided.");
```
wchamber, line 62: should this be validation on SnapshotReference creation? What would it mean to have a SnapshotReference object that didn't pass this validation?

wchamber (changes requested):
Looks good, just some questions around serialization.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zloery: Yeah, this should probably be at object creation time.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zloery: Partially because we expect WM references to snapshots to frequently use the underlying snapshot's name (though this isn't required or enforced), partially because this seemed like a reasonable standard for human generated + readable names.

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
zloery: I think all ReferenceObjects really are maps of strings to strings, but in practice all we really care about here is serializing them. Agreed that it's a bit weird as is - I think having serialize/deserialize methods instead of directly exposing a map makes it clearer what this is for.

That said, I think using the same serialize/deserialize methods for database storage introduces the risk of bugs where changing fields of `SnapshotReference` breaks existing values. Using separate serialization classes does add more boilerplate though, so I'm not sure if this solution is overkill or not.

src/main/java/bio/terra/workspace/service/datareference/model/DataReference.java
```diff
@@ -0,0 +1,64 @@
...
+  public abstract DataReferenceType referenceType();
+
+  /** Instructions for how to clone this reference (if at all). */
+  public abstract CloningInstructions cloningInstructions();
+
+  /** The actual object being referenced. */
+  public abstract ReferenceObject referenceObject();
+
+  /** Convenience method for translating to an API model DataReferenceDescription object. */
+  public DataReferenceDescription toApiModel() {
```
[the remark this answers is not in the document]
zloery: From a modeling perspective that makes sense, but I think anything including the internal model could already separately include the api model, so the end result is "we need to be careful about layers while developing" which was always true. 
I don't think there's an easy way to put this inside the generated api models, and imo this is neater than a large util class for model translation.

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -63,56 +81,63 @@ public DataReferenceDescription getDataReferenceByName(
...
+    String description = "Create data reference in workspace " + referenceRequest.workspaceId();
 
     JobBuilder createJob =
         jobService
             .newJob(
                 description,
                 UUID.randomUUID().toString(),
                 CreateDataReferenceFlight.class,
-                body,
+                null,
```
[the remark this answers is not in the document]
zloery: Moving closer to builder-style construction sgtm. I think there's also an argument for getting rid of `request` as an argument altogether, I think at this point every call we have is passing `null`. Either way, out of scope for now

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -32,70 +31,61 @@
...
-      UUID resourceId,
-      String credentialId,
-      CloningInstructionsEnum cloningInstructions,
-      ReferenceTypeEnum referenceType,
-      DataRepoSnapshot reference) {
+  /** Create a data reference in a workspace and return the reference's ID. */
+  public String createDataReference(DataReferenceRequest request, UUID referenceId)
+      throws DuplicateDataReferenceException {
     String sql =
-        "INSERT INTO workspace_data_reference (workspace_id, reference_id, name, resource_id, credential_id, cloning_instructions, reference_type, reference) VALUES "
```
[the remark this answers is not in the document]
zloery: That's bullet 5 in the PR description, but tl;dr: they're fields for controlled resources + private resources respectively that aren't wired through to endpoints (and won't be in the near future), I think it's better to remove them now and add them back (possibly in a different shape) when they actually mean something.

src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -18,4 +29,28 @@ public static void validatePaginationParams(int offset, int limit) {
...
   }
+
+  /**
+   * Utility function for validating a CreateDataReferenceRequestBody.
+   *
+   * <p>CreateDataReferenceRequestBody is currently structured to allow several parameters for
+   * controlled and private resources that aren't supported in WM. This function throws exceptions
+   * if any of those fields are set, or if any required fields are missing.
+   */
+  public static void validateCreateDataReferenceRequestBody(CreateDataReferenceRequestBody body) {
```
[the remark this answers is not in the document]
zloery: SGTM, and that's a pretty handy way to think about naming in context. I'll leave `validatePaginationParams` as is along the same thinking.

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
```diff
@@ -11,19 +10,28 @@
...
+  /**
+   * Method for deserializing a ReferenceObject from a json string and its type.
+   *
+   * <p>This method is used for Stairway deserialization. Database deserialization should use the
+   * static {@Code DataReferenceDao.deserializeReferenceObject} method instead.
+   *
+   * <p>Implementations of this interface are expected to provide a static method for deserializing
+   * from json which should be referenced here. This should
+   */
+  static ReferenceObject fromJson(String jsonString, DataReferenceType type) {
```
wchamber, line 31: Can we remove the DataReferenceType argument and have symmetry with toJson? It'd be nice if we were able to tell the reference type from the json alone and did not need to store another field next to the json. As-is, it's funny that the serialization relies on some side piece of serialization to be deserialized later. This could take the form of a field in the json reserved for the DataReferenceType, or something else.

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
```diff
@@ -11,19 +10,28 @@
...
+  String toJson();
+
+  /**
+   * Method for deserializing a ReferenceObject from a json string and its type.
+   *
+   * <p>This method is used for Stairway deserialization. Database deserialization should use the
+   * static {@Code DataReferenceDao.deserializeReferenceObject} method instead.
+   *
+   * <p>Implementations of this interface are expected to provide a static method for deserializing
+   * from json which should be referenced here. This should
```
wchamber, line 29: nit: incomplete sentence

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
```diff
@@ -11,19 +10,28 @@
...
+   * {@Code DataReferenceDao.serializeReferenceObject} method instead.
    */
-  static ReferenceObject toReferenceObject(DataReferenceType type, Map<String, String> properties) {
+  String toJson();
+
+  /**
+   * Method for deserializing a ReferenceObject from a json string and its type.
+   *
+   * <p>This method is used for Stairway deserialization. Database deserialization should use the
+   * static {@Code DataReferenceDao.deserializeReferenceObject} method instead.
```
wchamber, line 26: I'm not totally convinced we need two serialization formats (dao & stairway) yet. One day, I could see us wanting to store different information in the DAO than in the stairway flight maps, but I'm not sure what feature would cause us to split. When we do split, we can split the code and do backwards compatibility support as needed, no different than today.

So far the only methods on this interface are that it supports json serialization that we need backwards compatibility for. It doesn't seem crazy to rely on that in the DAO as well as for Stairway.

At least, it's worth explaining why we need two json serialization formats for SnapshotReferences/ReferenceObjects.

src/main/resources/application.yml
```diff
@@ -72,3 +72,5 @@ workspace:
     upgrade-on-start: true
     uri: ${env.db.host}/${env.db.ws.name}
     username: ${env.db.ws.user}
+
+spring.profiles.include: human-readable-logging
```
wchamber, line 76: is this an intentional change?

src/test/java/bio/terra/workspace/service/datareference/DataReferenceServiceTest.java
```diff
@@ -63,25 +61,41 @@ public void testCreateDataReference() {
...
-  public void testCreateDataReferenceNameTooLongFails() {
-    UUID workspaceId = createDefaultWorkspace();
-    DataReferenceRequest request =
-        defaultReferenceRequest(workspaceId).name("1".repeat(100)).build();
-    assertThrows(
-        InvalidDataReferenceException.class,
-        () -> dataReferenceService.createDataReference(request, USER_REQUEST));
-  }
+  // TODO(zloery): reference validation happens at the controller level, so these need to be
+  //    rewritten as integration tests.
```
wchamber, line 65: I'm not sure we get a lot of value of keeping the commented out tests or the TODO here. When we fix the TODO, we're not going to change this file. Can we delete the code from this file and either a) file a ticket for adding DR snapshot validation integration tests or b) move this TODO to some integration test file? In the ticket, we can refer to git version that still has these tests.

src/main/java/bio/terra/workspace/service/datareference/model/SnapshotReference.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public static ReferenceObject fromJson(String jsonString) {
+    try {
+      // This must deserialize to a SnapshotReference and not an AutoValue_SnapshotReference, as
+      // the AutoValue class has no annotated constructors or factory methods.
+      return objectMapper.readValue(jsonString, SnapshotReference.class);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException("Error deserializing to SnapshotReference " + jsonString, e);
+    }
+  }
```
wchamber, line 56: Can we add unit tests for serialization/deserialization? (Sorry if I missed them)

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
wchamber: Can we add that info to the comment? As-is, it sounds like the restrictions must match the TDR snapshot name restrictions.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
wchamber: Sorry, I meant should this happen in the method SnapshotReference.create? Should we restrict bad SnapshotReference objects from even being created, or do we want to just verify at the controller that new incoming requests don't fail this validation. If we don't want them created, we have to think about what to do when we change validation in a non-backwards compatible way i.e. DAO reads will start failing. If we only want to check at the controller, we may not notice if we change validation in a non-backwards compatible way.

wchamber (changes requested):
Sorry for the slow review time.

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
wchamber: Missed this earlier, sorry. I think for both DAO and Stairway serialization, we need to support backwards compatibility and should have unit tests for the same - e.g. hard code the string value of all current/past versions and check they can be deserialized. So, as in other review comments, I'm not totally convinced we need 2 serialization formats for ReferenceObjects.

src/main/java/bio/terra/workspace/service/datareference/model/SnapshotReference.java
zloery: Ah, forgot to add those. fixed!

src/test/java/bio/terra/workspace/service/datareference/DataReferenceServiceTest.java
zloery: SGTM, created PF-245 to track this and pointed it at the existing code.

src/main/resources/application.yml
zloery: Whoops, should have removed that.

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
zloery: Yeah, this was something I was less sure about. I think there's a slightly higher risk of making backwards-incompatible changes if we use the same serialization in both places, but it's a pretty small risk and this way takes a lot of boilerplate.  If you're fine using the same serialization for both, I'll go with that.

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
zloery: It's a little weird this way, but my issue with storing the type in a json field is that every ReferenceObject implementation needs to serialize type to the same magic key inside `toJson`, and I don't know how to enforce that at the interface level. Reading that type and then appropriately deserializing in Jackson also seems really messy, and IMO that feels worse than explicitly requiring this extra information. Thoughts?

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zloery: Ah, gotcha. My concern with moving it to `create` was around deserialization, but agreed that introduces concerns around breaking validation. I think we can mitigate some amount of that with testing

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
wchamber: Yeah. I think we should mitigate the risk of backwards incompatible changes with unit tests commenting about what failures indicate. With that, for now it seems fine to me to write the serialization once.

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
wchamber: I'm not super concerned about enforcing it at the interface level. This interface and all its impls are tiedly coupled, so we're not really likely to change them independently. These are also pretty small classes. We can & should, however, write unit tests for each impl that `assertEquals(x, fromJson(x.toJson())` which will verify we're doing whatever magic key stuff correctly.

One option for enforcing this at the interface level is to make this an abstract class, have the `public final String toJson()` on it, and a `protected abstract JsonObjectType innerToJson` that sub classes override to fill out their fields. Also have a `protected abstract DataReferenceType dataReferenceType()`. Then the base class `toJson` calls both of those protected methods to construct the right json with the magic field key. We can runtime verify that the subclasses are setting the value of the magic field key in the value returned by innerToJson.
I think it's fine to do that with more protected methods, or to just comment about magic key values needing to be filled in. Both make a cleaner public interface, which is most important. Either way, we want tests.

src/main/resources/application.yml
```diff
@@ -71,6 +71,4 @@ workspace:
     password: ${env.db.ws.pass}
     upgrade-on-start: true
     uri: ${env.db.host}/${env.db.ws.name}
-    username: ${env.db.ws.user}
-
-spring.profiles.include: human-readable-logging
\ No newline at end of file
+    username: ${env.db.ws.user}
```
wchamber, line 74: nit: lost the trailing newline now.

src/test/java/bio/terra/workspace/service/datareference/DataReferenceServiceTest.java
```diff
@@ -193,6 +159,27 @@ public void testDeleteMissingDataReference() {
...
+                DataReferenceType.DATA_REPO_SNAPSHOT);
+    assertEquals("foo", snapshotReference.instanceName());
+    assertEquals("bar", snapshotReference.snapshot());
+  }
+
+  @Test
+  public void SnapshotReferenceDeserializationnBackwardsCompatible() throws Exception {
+    String serializedSnapshot = SnapshotReference.create("foo", "bar").toJson();
+    assertEquals("{\"instanceName\":\"foo\",\"snapshot\":\"bar\"}", serializedSnapshot);
+  }
```
wchamber: I think these should be on a ReferenceObjectTest or a SnapshotReferenceTest file. Unit tests generally go with a file dedicated to the class they're testing.

src/test/java/bio/terra/workspace/service/datareference/DataReferenceServiceTest.java
```diff
@@ -193,6 +159,27 @@ public void testDeleteMissingDataReference() {
...
+                "{\"instanceName\":\"foo\",\"snapshot\":\"bar\"}",
+                DataReferenceType.DATA_REPO_SNAPSHOT);
+    assertEquals("foo", snapshotReference.instanceName());
+    assertEquals("bar", snapshotReference.snapshot());
+  }
+
+  @Test
+  public void SnapshotReferenceDeserializationnBackwardsCompatible() throws Exception {
+    String serializedSnapshot = SnapshotReference.create("foo", "bar").toJson();
+    assertEquals("{\"instanceName\":\"foo\",\"snapshot\":\"bar\"}", serializedSnapshot);
```
wchamber, line 180: I don't think this helps with backwards compatibility. This just serves as change detection. The stored values are the json strings, not the objects themselves.

I would like to see `y = fromJson(x.toJson); assert x equals y`, which also isn't backwards compatibility, but checks that the current version can deserialize to what it started as. Good to have for anything writing its own serialize/deserialize.

wchamber (changes requested):
last few nits

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
zloery: Got it. For some reason I thought AutoValue interacted weirdly if you extended an abstract class, but maybe not. 

I also realized interfaces/implementations can be handled natively in Jackson though, so I think adding the field via annotation is probably the right fix here. Future implementation classes will need to update the `JsonSubTypes` annotation in `ReferenceObject`, but I think the comments on the class and `fromJson` should make that clear.

I'm leaning towards keeping the type enum as a column in the database so that we can query it in the future, but it also might be possible to do that entirely from the json object. I'm not familiar enough with json in Postgres to know what we can or can't do, though

zloery:
(You can ignore the failing checks for now, there's an ongoing issue with Github actions runners: https://www.githubstatus.com/)

src/test/java/bio/terra/workspace/service/datareference/model/ReferenceObjectTest.java
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Hard code serialized values to check that code changes do not break backwards compatibility of
+   * stored JSON values. If this test fails, your change may not work with existing databases.
+   */
+  @Test
+  public void SnapshotReferenceSerializationBackwardsCompatible() {
+    SnapshotReference snapshotReference =
+        (SnapshotReference)
+            ReferenceObject.fromJson(
+                "{\"@type\":\"AutoValue_SnapshotReference\",\"instanceName\":\"foo\",\"snapshot\":\"bar\"}");
```
wchamber, line 30: there is weirdness with auto value object mapper transformations. I think we would be better off doing the serialization manually. As-is, any class name change is going to necessitate backwards compatibility code. that's not a great place to be. That's some boilerplate, but it's fine to have 1 place of serialization boilerplate per class.

wchamber (approved):

src/test/java/bio/terra/workspace/service/datareference/model/ReferenceObjectTest.java
zloery: I think an easier way to avoid that issue is to explicitly set the type name we want to use instead of using the class name, which is straightforward with the annotation `JsonTypeName` annotation. I used `SnapshotReference` as the value here for clarity, but I also tested it with arbitrary strings, which work fine. I'd prefer to avoid manual serialization if possible, since it's a non-trivial amount of boilerplate to maintain

src/main/java/bio/terra/workspace/service/datareference/model/DataReference.java
```diff
@@ -0,0 +1,64 @@
...
+    return new DataReferenceDescription()
+        .referenceId(referenceId())
+        .name(name())
+        .workspaceId(workspaceId())
+        .referenceType(referenceType().toApiModel())
+        .reference(((SnapshotReference) referenceObject()).toApiModel())
+        .cloningInstructions(cloningInstructions().toApiModel());
+  }
```
davidangb, line 42: the `(SnapshotReference) referenceObject()` cast will have to change once we have more models. No change necessary in this PR, just thinking out loud.

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
```diff
@@ -0,0 +1,53 @@
...
+   *
+   * <p>This method is used for Stairway and database deserialization. This makes use of type
+   * information stored via the {@Code JsonTypeInfo} annotation, and parses that information using
+   * the {@Code JsonSubTypes} annotation.
+   */
+  static ReferenceObject fromJson(String jsonString) {
+    try {
+      return objectMapper.readValue(jsonString, ReferenceObject.class);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException("Unable to deserialize ReferenceObject from string " + jsonString);
```
davidangb, line 50: I'm moderately worried about including the raw `jsonString` in the exception here. It could include sensitive information, or it could be large and unwieldy. How about logging the error and the raw string,  but only including a generic message in the thrown exception?

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -1,56 +1,67 @@
...
-        return reference;
+        validateSnapshotReference((SnapshotReference) reference, userReq);
+        return;
       default:
         throw new InvalidDataReferenceException(
             "Invalid reference type specified. Valid types include: "
-                + DATA_REPO_SNAPSHOT.toString());
+                + DataReferenceType.DATA_REPO_SNAPSHOT.toString());
     }
   }
```
davidangb, line 56: once we have more reference types, this can throw a ClassCastException if the `referenceType` does not match the actual type of the `reference` - for instance, where we try to cast to a `SnapshotReference`. Maybe wrap this all in a try/catch and provide an explanatory error message for this case? (fine to defer this for a future PR, once we have more types)

davidangb (commented):
fairly minor comments inline. Nice work on this!

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zloery: Agreed, we should probably be checking if the provided type doesn't match the actual type. I'll defer until we have multiple types, but added a TODO here in the meantime.

src/main/java/bio/terra/workspace/service/datareference/model/ReferenceObject.java
zloery: Sounds good. I think this qualifies as an error, so I'm logging at that level

src/main/java/bio/terra/workspace/service/datareference/model/DataReference.java
zloery: Yep. I tried to stay away from making assumptions about how we'll handle multiple implementation types in the API model so I just left this as-is, but it'll definitely change in the future.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -171,109 +166,99 @@ public boolean deleteDataReference(UUID workspaceId, UUID referenceId) {
...
+            .build();
+      };
+
+  public static String serializeReferenceObject(ReferenceObject obj, DataReferenceType type) {
+    switch (type) {
+      case DATA_REPO_SNAPSHOT:
+        return SnapshotReferenceV1.from((SnapshotReference) obj).serialize();
+      default:
+        throw new InvalidDataReferenceException(
+            "Attempting to serialize invalid reference type to DB: " + type);
```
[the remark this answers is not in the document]
zloery: I ended up deleting this section since we don't need the DB specific serialization, but agreed that's an issue that also comes up elsewhere

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -32,70 +30,63 @@
...
-            .addValue("reference_type", referenceType == null ? null : referenceType.toString());
-    try {
-      params.addValue("reference", objectMapper.writeValueAsString(reference));
-    } catch (JsonProcessingException e) {
-      // TODO: add logger and print out the reference
-      throw new InvalidDataReferenceException(
-          "Couldn't convert reference to JSON. This... shouldn't happen.");
-    }
+            .addValue("name", request.name())
+            .addValue("cloning_instructions", request.cloningInstructions().toString())
```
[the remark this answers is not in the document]
zloery: Might as well, that's probably the best practice. I don't use that for Stairway serialization as that just relies on ObjectMapper, and because we don't persist those for very long.

src/main/java/bio/terra/workspace/service/datareference/model/CloningInstructions.java
```diff
@@ -33,4 +42,9 @@ public static CloningInstructions fromApiModel(CloningInstructionsEnum modelEnum
   public CloningInstructionsEnum toApiModel() {
     return instructionMap.get(this);
   }
+
+  /** Convert this to a String to be serialized to the DB. */
+  public String toSql() {
+    return sqlMap.get(this);
+  }
```
davidangb, line 49: kick me for scope creep ... now I'm wondering if we need a corresponding `fromSql` to use instead of relying on `valueOf` ... even if the initial implementation of `fromSql` delegates to `valueOf`.

davidangb (approved):
one comment about `fromSql` vs. `valueOf`, which I leave to you to decide if it's important enough ... :+1: otherwise!

src/main/java/bio/terra/workspace/service/datareference/model/CloningInstructions.java
zloery: Good point - for the sake of consistency that's the right thing to do. I think this is another good place for a `BiMap`

PR #156, opened 2020-12-03
[QA-1288] Use Test Runner library to write tests using client library

ichengchang:
This PR contains the first pass at using the Test Runner library to write client tests for Workspace Manager Service.

Added a workspace-manager-clienttests Gradle project.
Included a server specification for running tests in dev environment.
Wrote one test that hits the unauthenticated endpoint (ServiceStatus), and one that hits the authenticated endpoint (GetWorkspace).


workspace-manager-clienttests/src/main/resources/measurementlists/DemoList.json
```diff
@@ -0,0 +1,18 @@
+{
```
marikomedlock, line 1: Let's remove this file for now. This file is used for the collectMeasurement Gradle task, which is not tested outside Data Repo yet. I think this first PR should focus just on running tests (runTest Gradle task)

workspace-manager-clienttests/src/main/resources/servers/workspace-dev.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "workspace-dev",
+  "description": "Workspace Dev env.",
+
+  "samUri": "https://sam.dsde-dev.broadinstitute.org",
+  "workspaceManagerUri": "https://workspace.dsde-dev.broadinstitute.org",
```
marikomedlock, line 6: Can you add the appropriate Data Repo URL here also? Since WM will talk to Data Repo, I think it may be helpful for future tests. I think the URL you want is https://jade.datarepo-dev.broadinstitute.org

workspace-manager-clienttests/src/main/resources/serviceaccounts/delegate-user-sa.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "firecloud-dev@broad-dsde-dev.iam.gserviceaccount.com",
```
marikomedlock, line 2: I'm not familiar with this SA, could you give some background on where it comes from/what it's generally used for?

workspace-manager-clienttests/src/main/resources/serviceaccounts/dev-workspace-app-sa.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "workspace-dev@broad-dsde-dev.iam.gserviceaccount.com",
```
marikomedlock, line 2: Is this the SA that the WM app runs as? If so, I think we don't want to use that for Test Runner. Test Runner should have its own SA (we'll need to Terraform this later), but maybe for this first pass, just use the delegate-user-sa.json?

workspace-manager-clienttests/src/main/resources/serviceaccounts/dev-workspace-app-sa.json
```diff
@@ -0,0 +1,5 @@
...
+  "jsonKeyFilename": "service-account.json",
```
marikomedlock, line 3: Could you make this filename a little more descriptive?

workspace-manager-clienttests/src/main/resources/suites/BasicSmoke.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "BasicSmoke",
+  "description": "Smoke tests.",
+  "serverSpecificationFile": "workspace-dev.json",
+  "testConfigurationFiles": ["BasicUnAuthenticated.json", "BasicAuthenticated.json"]
```
marikomedlock, line 5: Typo in BasicUnAuthenticated file name.

workspace-manager-clienttests/tools/render-config.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+VAULT_TOKEN=${1:-$(cat "$HOME"/.vault-token)}
+WM_APP_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/terra/kernel/dev/dev/workspace/app-sa
+WM_APP_SERVICE_ACCOUNT_OUTPUT_PATH=$(dirname "$0")/../rendered/service-account.json
+USER_DELEGATED_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/firecloud/dev/common/firecloud-account.json
+USER_DELEGATED_SERVICE_ACCOUNT_OUTPUT_PATH=$(dirname "$0")/../rendered/user-delegated-service-account.json
+JANITOR_CLIENT_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/terra/kernel/integration/tools/crl_janitor/client-sa
+JANITOR_CLIENT_SERVICE_ACCOUNT_OUTPUT_PATH="$(dirname $0)"/../rendered/janitor-client-sa-account.json
```
marikomedlock, line 9: Do you need the Janitor client SA?

marikomedlock (changes requested):

workspace-manager-clienttests/settings.gradle
```diff
@@ -0,0 +1,18 @@
+pluginManagement {
+    repositories {
+        resolutionStrategy {
+            eachPlugin {
+                if (requested.id.namespace == 'bio.terra') {
+                    if (requested.id.id == 'bio.terra.test-runner-plugin' || requested.id.id == 'bio.terra.test-runner-echo-plugin') {
+                        useModule("bio.terra:terra-test-runner:0.0.1-SNAPSHOT")
```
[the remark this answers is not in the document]
ichengchang: Fixed.

workspace-manager-clienttests/settings.gradle
```diff
@@ -0,0 +1,18 @@
...
+                    }
+                }
+            }
+        }
+        maven {
+            url "https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot-local/"
+        }
```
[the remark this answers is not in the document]
ichengchang: Yes, we'll have to decide this later. Right now we aren't publishing to libs-release-local yet.

workspace-manager-clienttests/src/main/java/scripts/testscripts/ServiceStatus.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class ServiceStatus extends TestScript {
+    private static final Logger logger = LoggerFactory.getLogger(ServiceStatus.class);
+
+    /** Public constructor so that this class can be instantiated via reflection. */
+    public ServiceStatus() {
+        super();
+    }
+
+    public void userJourney(TestUserSpecification testUser) throws Exception {
```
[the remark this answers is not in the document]
ichengchang: Fixed

workspace-manager-clienttests/src/main/resources/measurementlists/DemoList.json
ichengchang: Removed DemoList.java

workspace-manager-clienttests/src/main/resources/suites/BasicSmoke.json
ichengchang: Fixed

workspace-manager-clienttests/tools/render-config.sh
ichengchang: This is part of the render-config.sh that was used by WSM. I just copied it and make it work inside the test subdirectory. I'd just leave it there.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,43 @@
...
+        ApiClient apiClient = WorkspaceManagerServiceUtils.getClientForTestUser(testUser, server);
+        WorkspaceApi workspaceApi = new WorkspaceApi(apiClient);
+
+        try {
+            WorkspaceDescription workspaceDescription = workspaceApi.getWorkspace(UUID.fromString("1a315d46-351c-11eb-adc1-0242ac120002"));
+            logger.debug("workspaceDescription: {}", workspaceDescription);
+        } catch (ApiException apiEx) {
+            logger.debug("Caught exception getting workspace ", apiEx);
+            assertThat("GET workspace did not throw not found exception", false);
+        }
```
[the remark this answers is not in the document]
ichengchang: I have updated the logic to expect an invalid workspace id as the true outcome.

workspace-manager-clienttests/src/main/resources/serviceaccounts/delegate-user-sa.json
ichengchang: Basically, this SA account is used as a user delegate service account for workspace manager service. See `tools.render-config.sh`

workspace-manager-clienttests/src/main/resources/serviceaccounts/dev-workspace-app-sa.json
ichengchang: Changed to `delegate-user-sa.json` for now until test runner has its own SA.

workspace-manager-clienttests/src/main/resources/serviceaccounts/dev-workspace-app-sa.json
ichengchang: Make the workspace manager SA names more descriptive in `render_config.sh` and corresponding json file in `serviceaccounts` directory.

workspace-manager-clienttests/src/main/resources/servers/workspace-dev.json
ichengchang: Added Data Repo URL to servers json and updated `build.gradle` to include Data Repo Client library as well.

gjuggler:
Ivan, one general comment on directory naming: would you consider renaming `workspace-manager-clienttests` to `integration-tests` or just `integration`?

I see that other examples in TDR and RBS use the "foo-bar-clienttests" pattern, but I think (1) the repetition of the service name is unnecessary for service-internal tests, and (2) terminology wise, we're aligning on `integration` to describe these type of tests.

I could imagine a counter-argument that `integration` is only about functional integration tests, while this directory may contain performance / upgrade / resilience tests as well. On the balance, I think anchoring with a well-understood word like `integration` is better than `clienttest`, but I'm curious to hear others' opinions. @marikomedlock WDYT?

marikomedlock:
Agreed about the service name being redundant. My only concern with using integration is that there are other integration tests (e.g. JUnit/Spring-based) in src/test of the main service Gradle project. Maybe testrunner-tests? No real preference from me.

gjuggler:
> Agreed about the service name being redundant. My only concern with using integration is that there are other integration tests (e.g. JUnit/Spring-based) in src/test of the main service Gradle project. Maybe testrunner-tests? No real preference from me.

Good point. I suspect our intent would be to migrate those over to use TestRunner eventually, so hopefully not a long-lasting source of confusion. In any case, this is just a folder name that should be easy to adjust over time. I have no issue with favoring consistency w/ other repos naming for now, and pulling the trigger on a rename at some point in the future.

workspace-manager-clienttests/README.md
```diff
@@ -0,0 +1,29 @@
+# workspace-manager-clienttests
+This Gradle project contains Test Runner tests written with the Workspace Manager Service client library.
+
+The Test Runner library [GitHub repository](https://github.com/DataBiosphere/terra-test-runner) has documentation for
+how to write and execute tests.
+
+#### Run a test
+From the workspace-manager-clienttests directory:
+```
+./gradlew  runTest --args="configs/BasicAuthenticated.json /tmp/TR"
```
gjuggler, line 10: nit: add the `./render-config.sh` call to this code example block. As a reader, I'd expect the first block under "Running a test" to show me the set of commands required to get started.

workspace-manager-clienttests/tools/render-config.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
gjuggler, line 1: I think this file should live in the clienttests/ directory rather than a tools/ subdirectory. I don't think we expect there to be so many supporting scripts / tools that they'll make the parent directory too full, and it's more confusing to have it separated from the build file.

workspace-manager-clienttests/tools/render-config.sh
```diff
@@ -0,0 +1,22 @@
...
+
+VAULT_TOKEN=${1:-$(cat "$HOME"/.vault-token)}
+WM_APP_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/terra/kernel/dev/dev/workspace/app-sa
+WM_APP_SERVICE_ACCOUNT_OUTPUT_PATH=$(dirname "$0")/../rendered/workspace-manager-app-service-account.json
```
gjuggler, line 5: This script failed for me when I first tried to run it, since there isn't a workspace-clienttests/rendered/ directory defined by the git repo. Maybe create one and add a .gitignore file?

gjuggler (commented):
Thanks Ivan for putting this together! I was able to set up and run the smoke tests on my machine (after one change -- see my comment about the `rendered` subdirectory).

Just a couple nits on documentation and the tools script.

workspace-manager-clienttests/tools/render-config.sh
marikomedlock: I'm not sure git will preserve an empty directory. The way I handled that in Buffer Service is to add a `mkdir -p ../rendered` to the script.

ichengchang: Added `.gitignore` to revered/ folder.

workspace-manager-clienttests/README.md
ichengchang: Updated `README.md` with instructions on `render-config.sh`.

workspace-manager-clienttests/tools/render-config.sh
ichengchang: Moved the `render-config.sh` file one level up.

.github/workflows/test-runner-perf.yml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          path: |
+            .gradle/caches
+            .gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Run perf tests using test runner library
+        run: |
+          cd ${GITHUB_WORKSPACE}/${workingDir}/workspace-manager-clienttests
+          ./gradlew runTest --args="configs/BasicAuthenticated.json /tmp/TestRunnerResults"
```
marikomedlock, line 47: Two comments here:
- I think we need to add a step to build the client library from the current branch's code. Otherwise, it's testing new code with an old library. That will cause confusion when changes affected the client library.
- I think this GH action should be in a separate PR. In particular, we should add this after or concurrently with the uploadResults Gradle task. Otherwise, the tests will run in this GH action runner and then we lose the results.

workspace-manager-clienttests/README.md
```diff
@@ -0,0 +1,35 @@
...
+Run the ./render-config.sh script (inside tools directory) before running tests. 
+Currently this script targets the dev environment for testing.
+
+From the workspace-manager-clienttests/tools directory:
+```
+./render-config.sh
+```
+
+Each service account JSON files in the resources/serviceaccounts directory of this project specifies a default file
+path for the client secret file. This default path should match where the render-config.sh script puts the secret.
```
marikomedlock, line 35: Nit: Move this information up into the block above. No need to repeat it.

workspace-manager-clienttests/build.gradle
```diff
@@ -0,0 +1,71 @@
...
+
+    // Google dependencies
+    compile "com.google.auth:google-auth-library-oauth2-http:${googleOauth2}"
+
+    // Data Repo client library
+    implementation "bio.terra:datarepo-client:${datarepoClient}"
+    compile "io.swagger.core.v3:swagger-annotations:${swaggerAnnotations}"
+
+    // Workspace Manager Service client library
+    implementation "bio.terra:workspace-manager-client:${workspaceManagerServiceClient}"
```
marikomedlock, line 61: I think we should have an environment variable override here to point to a local copy of the WM client library. See example in Data Repo.

workspace-manager-clienttests/render-config.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+VAULT_TOKEN=${1:-$(cat "$HOME"/.vault-token)}
+WM_APP_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/terra/kernel/dev/dev/workspace/app-sa
+WM_APP_SERVICE_ACCOUNT_OUTPUT_PATH=$(dirname "$0")/rendered/workspace-manager-app-service-account.json
+USER_DELEGATED_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/firecloud/dev/common/firecloud-account.json
+USER_DELEGATED_SERVICE_ACCOUNT_OUTPUT_PATH=$(dirname "$0")/rendered/user-delegated-service-account.json
+JANITOR_CLIENT_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/terra/kernel/integration/tools/crl_janitor/client-sa
+JANITOR_CLIENT_SERVICE_ACCOUNT_OUTPUT_PATH="$(dirname $0)"/rendered/janitor-client-sa-account.json
```
marikomedlock, line 9: I think we should remove this Janitor client SA unless the Test Runner needs it. This render-config script is specifically for the WM clienttests directory.

workspace-manager-clienttests/settings.gradle
```diff
@@ -0,0 +1,18 @@
...
+        gradlePluginPortal()
+    }
+}
+rootProject.name = 'workspace-manager-test-runner'
```
marikomedlock, line 18: Nit: Should this name match the directory name? That seems to be the convention in other repositories, though I've not done an exhaustive search.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,47 @@
...
+        ApiClient apiClient = WorkspaceManagerServiceUtils.getClientForTestUser(testUser, server);
+        WorkspaceApi workspaceApi = new WorkspaceApi(apiClient);
+
+        /**
+         * This GetWorkspace test expects a bogus workspace id resulting in
+         * workspace not found exception and http code 401
+         * **/
+        try {
+            WorkspaceDescription workspaceDescription = workspaceApi.getWorkspace(UUID.fromString("11111111-1111-1111-1111-111111111111"));
+            assertThat("GET workspace does not throw not found exception", true);
```
marikomedlock, line 37: Shouldn't this be assertThat(..., false)? I think we want this to fail if it doesn't throw an exception right?

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,47 @@
...
+    public void userJourney(TestUserSpecification testUser) throws Exception {
+        ApiClient apiClient = WorkspaceManagerServiceUtils.getClientForTestUser(testUser, server);
+        WorkspaceApi workspaceApi = new WorkspaceApi(apiClient);
+
+        /**
+         * This GetWorkspace test expects a bogus workspace id resulting in
+         * workspace not found exception and http code 401
+         * **/
+        try {
+            WorkspaceDescription workspaceDescription = workspaceApi.getWorkspace(UUID.fromString("11111111-1111-1111-1111-111111111111"));
```
marikomedlock, line 36: Nit: Could you specifically say "invalid-worspace-id" in the string or in a variable name?

marikomedlock (changes requested):

workspace-manager-clienttests/README.md
ichengchang: Fixed.

workspace-manager-clienttests/render-config.sh
ichengchang: Removed Janitor client SA.

workspace-manager-clienttests/settings.gradle
ichengchang: Fixed subproject name to match directory.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
ichengchang: Fixed.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
ichengchang: Added descriptive variable name.

workspace-manager-clienttests/build.gradle
ichengchang: Added env override to detect local WM client library.

.github/workflows/test-runner-perf.yml
ichengchang: I have created a separate PR for this: [](https://github.com/DataBiosphere/terra-workspace-manager/pull/163).

marikomedlock: It looks like it's still included in this one though?

marikomedlock (commented):
Thanks for all the changes.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    public void userJourney(TestUserSpecification testUser) throws Exception {
+        ApiClient apiClient = WorkspaceManagerServiceUtils.getClientForTestUser(testUser, server);
+        WorkspaceApi workspaceApi = new WorkspaceApi(apiClient);
+
+        /**
+         * This GetWorkspace test expects a bogus workspace id resulting in
+         * workspace not found exception and http code 401
+         * **/
```
davidangb, line 34: As discussed in person, this test really needs to create a workspace in some `BeforeAll` hook, then attempt to get that workspace.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,48 @@
...
+        try {
+            String invalidWorkspaceId = "11111111-1111-1111-1111-111111111111";
+            WorkspaceDescription workspaceDescription = workspaceApi.getWorkspace(UUID.fromString(invalidWorkspaceId));
+            assertThat("GET workspace does not throw not found exception", false);
+        } catch (ApiException apiEx) {
+            logger.debug("Caught exception getting workspace ", apiEx);
+            assertThat("GET workspace throws not found exception", true);
```
davidangb, line 41: at minimum, inspect the exception, a la
```
assertThat("GET workspace throws not found exception", apiEx.getCode() == 404);
```
ideally, use some pre-existing assertion/matcher/construct such as JUnit's `assertThrows` instead of manually writing try/catch with multiple assertions.

of course if this test changes to be looking for a 200, this is irrelevant for this test.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+public class GetWorkspace extends TestScript {
+    private static final Logger logger = LoggerFactory.getLogger(GetWorkspace.class);
+
+    /** Public constructor so that this class can be instantiated via reflection. */
+    public GetWorkspace() {
+        super();
+    }
```
davidangb, line 24: is this really necessary? Doesn't Java give you a no-arg public constructor by default?

workspace-manager-clienttests/src/main/java/scripts/testscripts/ServiceStatus.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    public void userJourney(TestUserSpecification testUser) throws Exception {
+        ApiClient apiClient = WorkspaceManagerServiceUtils.getClient(server);
+        UnauthenticatedApi unauthenticatedApi = new UnauthenticatedApi(apiClient);
+        SystemStatus systemStatus = unauthenticatedApi.serviceStatus();
+        logger.info("systemStatus: {}", systemStatus);
+
+        int httpCode = unauthenticatedApi.getApiClient().getStatusCode();
+        logger.info("Service status HTTP code: {}", httpCode);
```
davidangb, line 28: in a concurrent perf test, this is going to log a lot of data, and the two log lines can get interleaved with other parallel requests. Combine into one?
```
logger.info("Service status with code {}: {}", httpCode, systemStatus);
```

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * Build the Workspace Manager Service API client object for the given server specification.
+     *
+     * @param server the server we are testing against
+     * @return the API client object
+     */
+    public static ApiClient getClient(ServerSpecification server) throws IOException {
+        if (Strings.isNullOrEmpty(server.workspaceManagerUri)) {
+            throw new IllegalArgumentException("Buffer Service URI cannot be empty");
```
davidangb, line 32: workspace manager, not buffer service

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+        if (server.testRunnerServiceAccount == null) {
+            throw new IllegalArgumentException("Buffer Service client service account is required");
```
davidangb, line 35: workspace manager, not buffer service

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
```diff
@@ -0,0 +1,106 @@
...
+                "Fetching credentials and building Workspace Manager ApiClient object for test user: {}",
+                testUser.name);
+        ApiClient apiClient = new ApiClient();
+        apiClient.setBasePath(server.workspaceManagerUri);
+
+        apiClient.setAccessToken(userAccessToken.getTokenValue());
+
+        apiClientsForTestUsers.put(testUser, apiClient);
+        return apiClient;
+    }
```
davidangb, line 105: it feels dangerous to reuse ApiClients in a parallel/multithreaded test. Is there a good reason to cache clients and reuse them? I don't think ApiClients are threadsafe or reusable.

Consider `ServiceStatus` line 27:
```
int httpCode = unauthenticatedApi.getApiClient().getStatusCode();
```
when you invoke a request via the api client, it saves the http response code to a member variable of the ApiClient class. If you are reusing ApiClients across multiple requests, those requests can overwrite each other.

If I am wrong about this, please show me how!

workspace-manager-clienttests/src/main/resources/serviceaccounts/dev-workspace-app-sa.json
davidangb, line 2: please absolutely get dedicated SAs for these, instead of reusing existing ones and allowing their credentials to land on many people's laptops.  I'd love to see a dedicated SA for delegate-user-sa.json too

davidangb (changes requested):
a bunch of comments inline. I am critically concerned with the reuse/caching of ApiClient objects, but have other feedback too!

workspace-manager-clienttests/build.gradle
```diff
@@ -0,0 +1,84 @@
...
+        implementation files(project.findProperty("workspacemanagerclientjar"))
+        compile "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jackson}"
+        compile "org.glassfish.jersey.core:jersey-client:${jersey}"
+        compile "org.glassfish.jersey.media:jersey-media-json-jackson:${jersey}"
+        compile "org.glassfish.jersey.media:jersey-media-multipart:${jersey}"
+        compile "org.glassfish.jersey.inject:jersey-hk2:${jersey}"
+    } else {
+        implementation "bio.terra:workspace-manager-client:${workspacemanagerClient}"
+        compile "io.swagger.core.v3:swagger-annotations:${swaggerAnnotations}"
+    }
```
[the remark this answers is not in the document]
ichengchang: Refer to Mariko's comments above:

_I think we need to add a step to build the client library from the current branch's code. Otherwise, it's testing new code with an old library. That will cause confusion when changes affected the client library._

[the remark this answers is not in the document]
davidangb: apologies for not being clear ... what is the purpose of the `if` clause here? How/when does someone specify `workspacemanagerclientjar`? Should that be explained in the readme?

workspace-manager-clienttests/build.gradle
```diff
@@ -0,0 +1,84 @@
...
+
+    // Terra Test Runner Library
+    // implementation group: "bio.terra", name: "terra-test-runner", version: "0.0.1-SNAPSHOT"
+    implementation "bio.terra:terra-test-runner:${testRunner}"
+}
+
+group "bio.terra"
+version "1.0-SNAPSHOT"
+sourceCompatibility = JavaVersion.VERSION_1_8
```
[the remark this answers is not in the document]
ichengchang: Updated JDK version to 11.

workspace-manager-clienttests/README.md
```diff
@@ -0,0 +1,31 @@
...
+how to write and execute tests.
+
+#### Run a test
+The workspace manager tests require the appropriate service account keys to be available in the `rendered/` folder.
+
+Run the following command from the workspace-manager-clienttests directory to retrieve the required keys.
+
+```
+./render-config.sh
+```
```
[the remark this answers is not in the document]
ichengchang: Updated doc with `.vault-token `comment.

workspace-manager-clienttests/build.gradle
```diff
@@ -0,0 +1,84 @@
...
+        compile "org.glassfish.jersey.media:jersey-media-json-jackson:${jersey}"
+        compile "org.glassfish.jersey.media:jersey-media-multipart:${jersey}"
+        compile "org.glassfish.jersey.inject:jersey-hk2:${jersey}"
+    } else {
+        implementation "bio.terra:workspace-manager-client:${workspacemanagerClient}"
+        compile "io.swagger.core.v3:swagger-annotations:${swaggerAnnotations}"
+    }
+
+    // Terra Test Runner Library
+    // implementation group: "bio.terra", name: "terra-test-runner", version: "0.0.1-SNAPSHOT"
```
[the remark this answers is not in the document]
ichengchang: Deleted comment.

workspace-manager-clienttests/build.gradle
```diff
@@ -0,0 +1,84 @@
...
+
+    // Google dependencies
+    compile "com.google.auth:google-auth-library-oauth2-http:${googleOauth2}"
+
+    // Data Repo client library
+    implementation "bio.terra:datarepo-client:${datarepoClient}"
+    compile "io.swagger.core.v3:swagger-annotations:${swaggerAnnotations}"
+
+    // Workspace Manager Service client library
+    //implementation "bio.terra:workspace-manager-client:${workspacemanagerClient}"
```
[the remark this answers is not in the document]
ichengchang: Deleted comment.

workspace-manager-clienttests/src/main/java/scripts/testscripts/ServiceStatus.java
ichengchang: Merged multiple logs into one statement.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
ichengchang: Fixed typos.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
ichengchang: Fixed typos.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
ichengchang: Added setup and cleanup steps to GetWorkspace test.

workspace-manager-clienttests/README.md
```diff
@@ -0,0 +1,32 @@
...
+This Gradle project contains Test Runner tests written with the Workspace Manager Service client library.
+
+The Test Runner library [GitHub repository](https://github.com/DataBiosphere/terra-test-runner) has documentation for
+how to write and execute tests.
+
+#### Run a test
+The workspace manager tests require the appropriate service account keys to be available in the `rendered/` folder.
+
+Run the following command from the workspace-manager-clienttests directory to retrieve the required keys.
+Note: `render-config.sh` assumes a valid `.vault-token` in your `$HOME` directory.
```
ungwudik, line 11: Worth adding a reference for how to generate `.vault-token`? https://github.com/broadinstitute/dsde-toolbox#authenticating-to-vault

workspace-manager-clienttests/build.gradle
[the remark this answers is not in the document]
ungwudik: Un-resolved this comment...to make sure the open question is still visible.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,90 @@
...
+        WorkspaceApi workspaceApi = new WorkspaceApi(apiClient);
+
+        /**
+         * This GetWorkspace test expects a valid workspace id
+         * created from the setup step.
+         *
+         * Throw workspace not found exception if anything goes wrong
+         * **/
+        try {
+            // String invalidWorkspaceId = "11111111-1111-1111-1111-111111111111";
```
ungwudik, line 62: This comment no longer applies, correct? Since it looks like `workspaceApi.getWorkspace(workspace.getId())` now uses a randomly generated UUID.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,90 @@
...
+            WorkspaceDescription workspaceDescription = workspaceApi.getWorkspace(workspace.getId());
+            assertThat("GET workspace does not throw not found exception", true);
+        } catch (ApiException apiEx) {
```
ungwudik, line 65: Also, is it possible that other exceptions (apart from "not found") can be thrown here, in which case the assert message below may be incorrect?
`assertThat("GET workspace throws not found exception", true);`

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,90 @@
...
+        WorkspaceApi workspaceApi = new WorkspaceApi(apiClient);
+
+        try {
+            workspaceApi.deleteWorkspace(workspace.getId());
+        } catch (ApiException apiEx) {
+            logger.debug("Caught exception deleting workspace ", apiEx);
+        }
+
+        int httpCode = workspaceApi.getApiClient().getStatusCode();
+        logger.info("DELELE workspace HTTP code: {}", httpCode);
```
ungwudik, line 87: Nit: Misspelled "DELETE"

workspace-manager-clienttests/render-config.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+VAULT_TOKEN=${1:-$(cat "$HOME"/.vault-token)}
+WM_APP_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/terra/kernel/dev/dev/workspace/app-sa
+WM_APP_SERVICE_ACCOUNT_OUTPUT_PATH=$(dirname "$0")/rendered/workspace-manager-app-service-account.json
+USER_DELEGATED_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/firecloud/dev/common/firecloud-account.json
+USER_DELEGATED_SERVICE_ACCOUNT_OUTPUT_PATH=$(dirname "$0")/rendered/user-delegated-service-account.json
```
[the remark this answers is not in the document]
ichengchang: It is up to your team to define the appropriate environment for running test. I am using `dev` here as an example.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
ichengchang: For now, we need to do `try-catch` manually. We will look at what you have suggested (something similar to a JUnit's assertion framework) as a future enhancement.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
ichengchang: Removed the explicit no-arg constructor from test classes.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
ichengchang: After discussion with Mariko, we simply removed caching part from the code.

workspace-manager-clienttests/src/main/resources/serviceaccounts/dev-workspace-app-sa.json
ichengchang: This is something we will need to work in DevOps, still in progress. At some point we will create a new PR for this.

workspace-manager-clienttests/README.md
ichengchang: Added additional notes to PR on how to generate `vault-token`.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
ichengchang: Removed the comment.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
ichengchang: It is up to team to catch application level error as they see fit. The current code is just an example on how to use test runner library.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
ichengchang: Fixed typo.

workspace-manager-clienttests/README.md
```diff
@@ -0,0 +1,57 @@
...
+The Test Runner library [GitHub repository](https://github.com/DataBiosphere/terra-test-runner) has documentation for
+how to write and execute tests.
+
+#### Run a test
+The workspace manager tests require the appropriate service account keys to be available in the `rendered/` folder.
+
+Run the following command from the workspace-manager-clienttests directory to retrieve the required keys.
+Note: `render-config.sh` assumes a valid `.vault-token` in your `$HOME` directory. Please refer to the following document on how to generate a `vault-token`
+ 
+ https://github.com/broadinstitute/dsde-toolbox#authenticating-to-vault https://github.com/broadinstitute/dsde-toolbox#authenticating-to-vault
```
davidangb, line 13: nit: url is listed twice (paste error?)

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,87 @@
...
+    public void setup(List<TestUserSpecification> testUsers) throws Exception {
+        assertThat("There must be at least one test user in configs/testusers directory.", testUsers!=null && testUsers.size()>0);
+        id = UUID.randomUUID();
+        ApiClient apiClient = WorkspaceManagerServiceUtils.getClientForTestUser(testUsers.get(0), server);
+        WorkspaceApi workspaceApi = new WorkspaceApi(apiClient);
+        try {
+            CreateWorkspaceRequestBody requestBody = new CreateWorkspaceRequestBody();
+            requestBody.setId(id);
+            requestBody.setJobId("testrunner");
+            requestBody.setSpendProfile("testrunner");
```
davidangb, line 36: if you run in to problems using the string "testrunner" for a spend profile, just leave it out; it's not required

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
```diff
@@ -0,0 +1,87 @@
...
+         *
+         * Throw workspace not found exception if anything goes wrong
+         * **/
+        try {
+            WorkspaceDescription workspaceDescription = workspaceApi.getWorkspace(workspace.getId());
+            assertThat("GET workspace does not throw not found exception", true);
+        } catch (ApiException apiEx) {
+            logger.debug("Caught exception getting workspace ", apiEx);
+            assertThat("GET workspace throws not found exception", true);
+        }
```
davidangb, line 64: do you need the try/catch here at all any more? Can this just be:
```
  WorkspaceDescription workspaceDescription = workspaceApi.getWorkspace(workspace.getId());
  int httpCode = workspaceApi.getApiClient().getStatusCode();
  logger.info("GET workspace HTTP code: {}", httpCode);
  assertThat(httpCode, equalTo(200));
```
if the call to `getWorkspace` throws an exception, it'll fail the test, no try/catch needed, right?

If you opt to leave the try/catch in place, the `assertThat("GET workspace throws not found exception", true);` is wrong.

workspace-manager-clienttests/src/main/java/scripts/testscripts/ServiceStatus.java
```diff
@@ -0,0 +1,24 @@
...
+import scripts.utils.WorkspaceManagerServiceUtils;
+
+public class ServiceStatus extends TestScript {
+    private static final Logger logger = LoggerFactory.getLogger(ServiceStatus.class);
+
+    @Override
+    public void userJourney(TestUserSpecification testUser) throws Exception {
+        ApiClient apiClient = WorkspaceManagerServiceUtils.getClient(server);
+        UnauthenticatedApi unauthenticatedApi = new UnauthenticatedApi(apiClient);
+        SystemStatus systemStatus = unauthenticatedApi.serviceStatus();
```
davidangb, line 19: I think this block does not do what we want it to do.

`WorkspaceManagerServiceUtils.getClient(server)` returns an `ApiClient` **with an access token set**, where the access token belongs to the `testRunnerServiceAccount`. It then uses this `ApiClient` to make an unauthenticated request.

Instead, I think this block should be generating an `ApiClient` **without** an access token set, so it is truly making an authenticated request.

what I'm not sure about: does `UnauthenticatedApi` discard the access token somewhere within itself? Even if it does, it doesn't feel good relying on that, we shouldn't set an access token at all. Should this be using `WorkspaceManagerServiceUtils.getClientForTestUser(null, server)` instead? And if so, can `WorkspaceManagerServiceUtils.getClient` be deleted entirely since it isn't used?

workspace-manager-clienttests/rendered/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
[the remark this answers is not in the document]
ichengchang: Moved `.gitignore` from rendered folder one level up.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
```diff
@@ -0,0 +1,92 @@
...
+
+        // TODO: have ApiClients share an HTTP client, or one per each is ok?
+        // no cached ApiClient found, so build a new one here and add it to the cache before returning
+        logger.debug(
+                "Fetching credentials and building Workspace Manager ApiClient object for test user: {}",
+                testUser.name);
+        ApiClient apiClient = new ApiClient();
+        apiClient.setBasePath(server.workspaceManagerUri);
+
+        apiClient.setAccessToken(userAccessToken.getTokenValue());
```
davidangb, line 88: simplified implementation (untested):
```
        ApiClient apiClient = new ApiClient();
        apiClient.setBasePath(server.workspaceManagerUri);
        
        // if no test user is specified, then return a client object without an access token set
        // this is useful if the caller wants to make ONLY unauthenticated calls
        if (testUser != null) {
            logger.debug(
                    "Fetching credentials and building Workspace Manager ApiClient object for test user: {}",
                    testUser.name);

            // refresh the user token
            GoogleCredentials userCredential =
                    AuthenticationUtils.getDelegatedUserCredential(
                            testUser, AuthenticationUtils.userLoginScopes);
            AccessToken userAccessToken = AuthenticationUtils.getAccessToken(userCredential);
            
            apiClient.setAccessToken(userAccessToken.getTokenValue());
        }
```

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
```diff
@@ -0,0 +1,92 @@
...
+
+        return apiClient;
+    }
```
davidangb, line 91: or, refactoring both methods (including edits to javadoc/comments):
```
   private static ApiClient buildClient(AccessToken accessToken,  ServerSpecification server) throws IOException {
        if (Strings.isNullOrEmpty(server.workspaceManagerUri)) {
            throw new IllegalArgumentException("Workspace Manager Service URI cannot be empty");
        }

        // build the client object
        ApiClient apiClient = new ApiClient();
        apiClient.setBasePath(server.workspaceManagerUri);

        if (accessToken != null) {
            apiClient.setAccessToken(accessToken.getTokenValue());
        }

        return apiClient;
    }

    /**
     * Build the Workspace Manager Service API client object for the given server specification,
     * using an access token for the test runner service account.
     *
     * @param server the server we are testing against
     * @return the API client object
     */
    public static ApiClient getClient(ServerSpecification server) throws IOException {
        if (server.testRunnerServiceAccount == null) {
            throw new IllegalArgumentException("Workspace Manager Service client service account is required");
        }

        // refresh the client service account token
        GoogleCredentials serviceAccountCredential =
                AuthenticationUtils.getServiceAccountCredential(
                        server.testRunnerServiceAccount, AuthenticationUtils.userLoginScopes);
        AccessToken accessToken = AuthenticationUtils.getAccessToken(serviceAccountCredential);
        logger.debug(
                "Generated access token for workspace manager service client SA: {}",
                server.testRunnerServiceAccount.name);

        return buildClient(accessToken, server);

    }

    /**
     * Build the Workspace Manager API client object for the given test user and server specifications.
     * The test user's token is always refreshed
     *
     * @param testUser the test user whose credentials are supplied to the API client object
     * @param server the server we are testing against
     * @return the API client object for this user
     */
    public static ApiClient getClientForTestUser(
            TestUserSpecification testUser, ServerSpecification server) throws IOException {

        AccessToken accessToken = null;

        // if no test user is specified, then return a client object without an access token set
        // this is useful if the caller wants to make ONLY unauthenticated calls
        if (testUser != null) {
            logger.debug(
                    "Fetching credentials and building Workspace Manager ApiClient object for test user: {}",
                    testUser.name);

            // refresh the user token
            GoogleCredentials userCredential =
                    AuthenticationUtils.getDelegatedUserCredential(
                            testUser, AuthenticationUtils.userLoginScopes);
            accessToken = AuthenticationUtils.getAccessToken(userCredential);
        }

        return buildClient(accessToken, server);
    }
```

davidangb (changes requested):
sorry, one necessary change on `ServiceStatus` that I missed in the first review. I also have suggestions on how to refactor `WorkspaceManagerServiceUtils` that can go in this PR or a future PR (as long as we don't lose it), and a couple other small comments.

gjuggler (approved):
Approved with the understanding that we'll continue to follow-up on David's comments re: SA permissions. The priority here is to unblock other devs on the team to be able to kick the tires on TestRunner, and start building / executing their own integration & performance tests.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
marikomedlock: I don't think we'll need any change to TestRunner for that. You should just be able to import an assertion library, similar to how assertThat is imported in this file.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        // refresh the user token
+        GoogleCredentials userCredential =
+                AuthenticationUtils.getDelegatedUserCredential(
+                        testUser, AuthenticationUtils.userLoginScopes);
+        AccessToken userAccessToken = AuthenticationUtils.getAccessToken(userCredential);
+
+        // TODO: have ApiClients share an HTTP client, or one per each is ok?
+        // no cached ApiClient found, so build a new one here and add it to the cache before returning
```
marikomedlock, line 81: Nit: Remove this comment about caching

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
```diff
@@ -0,0 +1,92 @@
...
+        apiClient.setAccessToken(accessToken.getTokenValue());
+
+        return apiClient;
+    }
+
+    /**
+     * Build the Workspace Manager API client object for the given test user and server specifications. This
+     * class maintains a cache of API client objects, and will return the cached object if it already
+     * exists. The token is always refreshed, regardless of whether the API client object was found in
+     * the cache or not.
```
marikomedlock, line 55: Nit: Remove caching from this method comment.

workspace-manager-clienttests/src/main/java/scripts/testscripts/ServiceStatus.java
marikomedlock: Agreed that this should be calling WorkspaceManagerServiceUtils.getClient(null, server) instead, or some convenience wrapper (e.g. WorkspaceManagerServiceUtils.getClientWithoutAccessToken) for that.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private WorkspaceManagerServiceUtils() {}
+
+    /**
+     * Build the Workspace Manager Service API client object for the given server specification.
+     *
+     * @param server the server we are testing against
+     * @return the API client object
+     */
+    public static ApiClient getClient(ServerSpecification server) throws IOException {
```
marikomedlock, line 26: Perhaps we can rename this to getClientForTestRunnerSA instead and update the comment to include that it's setting the access token for the Test Runner SA specified for the given server?

marikomedlock (commented):
A couple suggestions to make the getClient* methods easier to understand, and a few nits.

Also, I agree we should punt getting new SAs to a follow-on PR.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
ichengchang: Set spend profile to null.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetWorkspace.java
ichengchang: Removed try-catch, all exceptions will be thrown.

workspace-manager-clienttests/src/main/java/scripts/testscripts/ServiceStatus.java
ichengchang: Refactored build client code.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
ichengchang: Refactored build client code.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
ichengchang: Refactored build client code.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
ichengchang: Refactored build client code.

workspace-manager-clienttests/README.md
ichengchang: Removed duplicate links.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
ichengchang: Caching comment removed.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceManagerServiceUtils.java
ichengchang: Caching comment removed.

gjuggler:
[nit] I just updated the PR formatting to align w/ conventions

davidangb (approved):
:+1: approving, on the assumption that any outstanding comments are tracked elsewhere for followup

.github/workflows/test-runner-perf.yml
ichengchang: I think this is all set. Nothing more needs to be done here.

workspace-manager-clienttests/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Each service account JSON files in the resources/serviceaccounts directory of this project specifies a default file
+path for the client secret file. This default path should match where the render-config.sh script puts the secret.
+
+#### Use a local Workspace Manager client JAR file
+The version of the Workspace Manager client JAR file is specified in the build.gradle file in this sub-project. This JAR file is
+fetched from the Broad Institute Maven repository. You can override this to use a local version of the Workspace Manager client
+JAR file by specifying a Gradle project property, either with a command line argument
+
+`./gradlew -Pworkspacemanagerclientjar=~/terra-workspace-manager/workspace-manager-client/build/libs/workspace-manager-client-0.5.0-SNAPSHOT.jar run --args="configs/BasicUnauthenticated.json`
```
marikomedlock, line 41: Nit: Should be "runTest" not "run" and the args part is missing the output directory and ending double quote.

marikomedlock (approved):
LGTM. One nit about what looks like a bad copy/paste in the README.

Thanks for putting this all together!

workspace-manager-clienttests/README.md
ichengchang: All set. Fixed minor typos.

araina2018:
@davidangb -David, noted your comment on tracking outstanding comments. Will follow up.

PR #158, opened 2020-12-04
migrate reference column to jsonb [no ticket; risk: low]

davidangb:
tested locally, and this properly migrates pre-existing data in the table (in my tests)

jsonb generally supports more functionality and could be useful for us in the future, allowing us to query against individual parts of the embedded json.


src/main/resources/db/changesets/20201204_reference_jsonb.yaml
```diff
@@ -0,0 +1,10 @@
+# Add an index on workspace_id and reference_id as they're frequently looked up together
```
zloery, line 1: nit: update comment

zloery (approved):
LGTM, one comment to change

wchamber (commented):
There's a few other `json` columns in the initial schema. Can we change all of them at once? I don't think there's much downside.

src/main/resources/db/changesets/20201204_reference_jsonb.yaml
davidangb: lol thank you, good catch 🤷

davidangb:
@wchamber I think I got them all, how's this?

wchamber (approved):

zloery (approved):

PR #160, opened 2020-12-04
scan image before push [AS-548]

zarsky-broad:


davidangb:
@zarsky-broad is there a way to see the Trivy results, even if no vulnerabilities were found? It would be good to see the "success" result.

.github/workflows/build.yml
```diff
@@ -78,16 +78,38 @@ jobs:
...
+        format: 'template'
+        template: '@/contrib/sarif.tpl'
+        output: 'trivy-results.sarif'
+        severity: 'CRITICAL'
+        exit-code: '1'
+    - name: Upload Trivy scan results to GitHub Security tab
+      uses: github/codeql-action/upload-sarif@v1
+      with:
+        sarif_file: 'trivy-results.sarif'
+    - name: Abort if vulnerabilities found
```
gmalkov, line 108: IIRC when a step fails, the default behavior is to not run any further steps unless they have `if: failure()` or `if: always()`, so this step should not be needed? Does the `scan-image` step always succeed without this step?

gmalkov (approved):
LGTM

.github/workflows/build.yml
gmalkov: Ah I missed the `continue-on-error` and I see now how that is used to always upload results. NVM.

zarsky-broad:
> @zarsky-broad is there a way to see the Trivy results, even if no vulnerabilities were found? It would be good to see the "success" result.

I don't think so - as far as I'm able to tell, the option to fail on vulnerabilities found is tied to the level of vulnerabilities to scan for; in other words, we can't show vulnerabilities without failing on them.

If we wanted, we could add another step to perform the scan again with a text output to the end, after the docker image is pushed, but that's all I can think of.

.github/workflows/build.yml
```diff
@@ -78,16 +78,38 @@ jobs:
...
+    - name: Run Trivy vulnerability scanner
+      id: scan-image
+      uses: aquasecurity/trivy-action@master
+      continue-on-error: true
+      with:
+        image-ref: ${{ steps.image-name.outputs.name }}
+        format: 'template'
+        template: '@/contrib/sarif.tpl'
+        output: 'trivy-results.sarif'
+        severity: 'CRITICAL'
```
zbedo, line 102: Can you add  `ignore-unfixed: true` so it will not report Criticals that are not fixed yet, which means even if you update all packages, you can't fix these vulnerabilities.

zarsky-broad: Done, look good?

davidangb:
@zarsky-broad I guess my question is really: how can we be sure that the trivy scan is executing correctly, if we can't see any results/it's not reporting any vulnerabilities? Brainstorming: could we intentionally introduce a problem (say, by extending a known-bad base image) so we can see it complain?

zarsky-broad:
> @zarsky-broad I guess my question is really: how can we be sure that the trivy scan is executing correctly, if we can't see any results/it's not reporting any vulnerabilities? Brainstorming: could we intentionally introduce a problem (say, by extending a known-bad base image) so we can see it complain?

@zbedo Do you have a known bad java 11 base image we could try?

zbedo:
@zarsky-broad  Can you try this `adoptopenjdk/openjdk11:jdk-11.28-alpine` ?

zarsky-broad:
@davidangb @zbedo well, it didn't fail, but we got these two lines which may explain why:
```
 2020-12-08T00:18:28.894Z	WARN	This OS version is no longer supported by the distribution: alpine 3.8.1
2020-12-08T00:18:28.894Z	WARN	The vulnerability detection may be insufficient because security updates are not provided
```

davidangb (approved):
approving - though I'm still a bit uncomfortable that we'll properly get success/failure in absence/presence of vulnerabilities. I probably won't be comfortable until this has run for a while and found at least one problem.

PR #161, opened 2020-12-04
simplify test workflow

zarsky-broad:
Combining the test github workflows into one matrixed workflow, so we don't have so much repeated code


gjuggler:
This looks like a clever approach, but I'm wondering whether it squares with the intended use of matrix. The [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) talk mostly about matrix-ing over OSes, languages, and tool versions. But this doesn't involve any of those use cases; it's about general code reuse where you want to share some setup behavior across various distinct workflow steps.

Over in AoU land (which uses CircleCI), we define reusable "commands" which are then called (sometimes with parameters) by concrete "jobs" ([code ref](https://github.com/all-of-us/workbench/blob/master/.circleci/config.yml#L102)). Does something similar not exist in GHA?

If possible, it would be great to add a comment pointing to some reference (maybe a stackdriver, github issue, or blog post?) demonstrating this pattern in use.

gjuggler:
> This looks like a clever approach, but I'm wondering whether it squares with the intended use of matrix. The [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) talk mostly about matrix-ing over OSes, languages, and tool versions. But this doesn't involve any of those use cases; it's about general code reuse where you want to share some setup behavior across various distinct workflow steps.
> 
> Over in AoU land (which uses CircleCI), we define reusable "commands" which are then called (sometimes with parameters) by concrete "jobs" ([code ref](https://github.com/all-of-us/workbench/blob/master/.circleci/config.yml#L102)). Does something similar not exist in GHA?
> 
> If possible, it would be great to add a comment pointing to some reference (maybe a stackdriver, github issue, or blog post?) demonstrating this pattern in use.

Related link drop: this discussion https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/14 seems to be very targeted at this use case. Composite actions was what GitHub built to address that scenario, but I feel like that may be too complex for our needs (e.g. we're talking about reuse within a single workflow, not across actions).

Have you explored whether it would be possible to follow this example, splitting logical steps into different "jobs" and using `needs` to indicate dependencies between jobs?

davidangb (approved):
:+1:, modulo anything that comes from the discussion Greg started

zarsky-broad:
@gjuggler To the best of my understanding, `needs` can be used to specify an order, but each job still runs on its own fresh instance, so they don't share any setup.

My only other idea is to write an action to automate all of this, but it looks like I can't do that locally, that would have to be in its own repo. We could write just a bash script to do the shared parts of the setup, but that seems to defeat the point.

gjuggler:
> @gjuggler To the best of my understanding, `needs` can be used to specify an order, but each job still runs on its own fresh instance, so they don't share any setup.
> 
> My only other idea is to write an action to automate all of this, but it looks like I can't do that locally, that would have to be in its own repo. We could write just a bash script to do the shared parts of the setup, but that seems to defeat the point.

Bah, that's too bad. One more pointer in the GitHub support space: https://github.community/t/support-for-yaml-anchors/16128/46 (if they were to support YAML anchors, that would probably solve ~90% of our needs here).

@ddietterich has had some poor experiences with actions defined in a separate repo; I suspect we'll want to avoid making that jump unless our needs become notably more complex.

I didn't research it extensively, but it [looks possible](https://github.community/t/path-to-action-in-the-same-repository-as-workflow/16952/2) to reference an action from within the same repo? If that's the case case, you could frame the reusable logic as a [composite action](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-composite-run-steps-action) which (1) seems more in line with how GitHub wants the framework to be used, and (2) avoids pitfalls of a fully-separate actions repo.

zarsky-broad:
@gjuggler I've just spent some time experimenting with creating a local action. It's definitely doable.
The problem is that actions [currently] can't just call other actions, which means that any code calling actions would have to be duplicated across the 3 jobs. The major one for this is the cache action to save deps, which is a lot of code to have to keep in sync.
The other issue, which is more minor, is that I can't create a service within an action, so the postgres container would have to be manually `docker run`, which is irksome.

zarsky-broad:
I did refactor the config rendering step into it's own little action, since it's a lot of steps and each one needed a conditional.

zarsky-broad:
Update: actions don't have access to secrets, I'd have to pass those through manually and then I'd be in charge of masking them, which I don't want.

gjuggler:
Thanks Isaac for digging in here. Sad stories, and reasonable workarounds, all around.

I'd love if you could comment above the strategy.matrix section with some notes summarizing the situation here, ideally with a breadcrumb or two for someone who is wondering what's going on. Maybe a link to https://github.community/t/support-for-yaml-anchors/16128/46 as the best "this would work if it existed" feature request.

.github/workflows/test.yml
```diff
@@ -0,0 +1,98 @@
...
+    branches: [ '**' ]
+
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
```
gjuggler, line 19: Pulling my earlier comment into the appropriate code location:

I'd love if you could comment above the strategy.matrix section with some notes summarizing the situation here, ideally with a breadcrumb or two for someone who is wondering what's going on. Maybe a link to https://github.community/t/support-for-yaml-anchors/16128/46 as the best "this would work if it existed" feature request.

gjuggler (approved):
Thanks Isaac. LGTM modulo the request for an add'l comment qualifying the use of matrix strategy.

Also – is there a ticket to associate this PR with? If not, I'd tend to actively acknowledge this with a [no ticket] in the PR description so it's obvious to folks that this is unaccounted cleanup / fix-up work.

PR #162, opened 2020-12-07
explicitly set db isolation levels to read committed [AS-457]

Shakespeared:
I downgraded the existing isolation levels in the workspace DAO, and added annotations explicitly setting the isolation level in data reference DAO. This way it is easy to see or change the isolation level of any given transaction. 

Let me know if any of the annotations are redundant or you disagree with setting them explicitly to read committed rather than the built-in default (the db default for postgres is already read committed, but it's more opaque which is why I've set it individually).

### DIFF
--- a/src/main/java/bio/terra/workspace/db/DataReferenceDao.java
+++ b/src/main/java/bio/terra/workspace/db/DataReferenceDao.java
@@ -19,6 +19,9 @@ import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Isolation;
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
 
 @Component
 public class DataReferenceDao {
@@ -39,6 +42,7 @@ public class DataReferenceDao {
   private Logger logger = LoggerFactory.getLogger(DataReferenceDao.class);
 
   /** Create a data reference in a workspace and return the reference's ID. */
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.READ_COMMITTED)
   public String createDataReference(DataReferenceRequest request, UUID referenceId)
       throws DuplicateDataReferenceException {
     String sql =
@@ -68,6 +72,7 @@ public class DataReferenceDao {
   }
 
   /** Retrieve a data reference by ID from the DB. */
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.READ_COMMITTED)
   public DataReference getDataReference(UUID workspaceId, UUID referenceId) {
     String sql =
         "SELECT workspace_id, reference_id, name, cloning_instructions, reference_type, reference from workspace_data_reference where workspace_id = :workspace_id AND reference_id = :reference_id";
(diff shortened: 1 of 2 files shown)


davidangb (approved):
:+1:!

Some optional thoughts, none is a blocker for this PR:
* would it be worth setting a variable somewhere - even read from config - that centrally defines our default transaction level as `Isolation.READ_COMMITTED`, and then using that variable in all the `@Transactional`  annotations? If we ever wanted to change our default in the future, that would make it a one-line change.
* I reviewed all the methods marked as `@Transactional` and they are all single-statement operations. Nowhere do we issue multiple statements in a transaction - if we did, they would be candidates for a stricter isolation level.
* I also didn't see any database operations we missed - everything now has `@Transactional`.
  * well, except for a few tests. Are those worth annotating too?
  * and, except for the `connection.isValid` call inside `WorkspaceManagerStatusService`. I don't think that needs an annotation, though.

Shakespeared:
@davidangb Anu and I did look into creating a config variable for the default, but ran into some issues with getting it to read correctly in the annotation. I'm not sure exactly why it wasn't working (Anu might remember in more detail) but we poked at it for a while and then decided it seemed like it would require adding more complexity than it would remove. However I'd be totally open to looking into that again if you feel strongly or have a way in mind to set the config up so it can be accessed in both DAOs.

Yeah I did not annotate any of the tests - happy to add that if you think that's worthwhile.

davidangb:
@Shakespeared nope, I don't feel strongly about centralizing a variable, nor about adding annotations to tests. All good!

Shakespeared:
@davidangb Great! To me it seems more likely that we'd want to elevate the isolation level of a particular transaction (if it ever involved multiple statements for instance) than wanting to change the default very often, so it makes sense to me to just set it explicitly in each transaction and not worry about a centralized default. 

I'll wait and see if anyone from the channel has comments, then merge. Thanks for the review!

ddietterich:
I still think this change is at best useless and at worst harmful. I have exactly the opposite position of @davidangb. I would rather be serializable by default and spot fix any performance problems. This introduces the potential of hard-to-debug consistency problems - even in single statements - with no measurable benefit.

Shakespeared:
@ddietterich It sounds like the basic options here are prioritizing baseline consistency and spot fixing performance, or prioritizing baseline performance and spot fixing consistency. Without more data on the db performance it's hard to know if this will make a measurable improvement or not. Personally, I'm inclined to try it and see. but I don't have a strongly held opinion either way.

ddietterich:
I would characterize this as introducing potential unreliability in favor of an unmeasured, unproven performance benefit. I do not want that in the code base.

Database performance is dependent on the workload. As is nearly always the case in software engineering, it is bad practice to make premature optimization for performance without empirical data to back it up.

gjuggler:
@Shakespeared it sounds like we may need to discuss this topic a bit more within the architecture group, with an aim to come back with more concrete guidance intended to apply across all (or most?) MC Terra services.

Normally I wouldn't want to block forward progress, but I think this PR isn't critical for related / ongoing work. If it sounds reasonable, would you mind holding this on ice until we loop back with some agreed-upon guidance?

Shakespeared:
@gjuggler Sure thing, that makes sense. I'm happy to hang onto this, it isn't an urgent issue on our end so I can drop it out of our current sprint if necessary. Should I close this PR for now, change it to draft, or leave it as is for now?

gjuggler:
> @gjuggler Sure thing, that makes sense. I'm happy to hang onto this, it isn't an urgent issue on our end so I can drop it out of our current sprint if necessary. Should I close this PR for now, change it to draft, or leave it as is for now?

Thanks! Switching it to draft sounds best.

Shakespeared:
> Thanks! Switching it to draft sounds best.

@gjuggler Sure thing! Changed it to a draft PR, keep me posted so I can move this forward once there's agreement on how to do so

Shakespeared:
Closing this PR because there's no consensus yet, and the PR is now out of date.

PR #164, opened 2020-12-10
[PF-224] Add single-user permissions API

zloery:
This change adds a basic permissions API for adding or removing roles to single users, and for retrieving the current state of permissions on a workspace. This makes use of the workspace stage feature toggle, and all three new endpoints are explicitly disallowed for Rawls workspaces (Rawls handles permissions on those workspaces). 

We should consider separating gated endpoints into a separate controller (which groups them separately in Swagger UI), but I'd prefer to do that in a separate PR for clarity.


src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -129,6 +136,95 @@ public void workspaceAuthz(AuthenticatedUserRequest userReq, UUID workspaceId, S
               userReq.getEmail(), action, workspaceId.toString()));
   }
 
+  /**
+   * Wrapper around Sam client to grant a role to the provided user.
+   *
+   * <p>This operation is only available to MC_TERRA stage workspaces, as Rawls manages permissions
```
ddietterich, line 142: What is an MC_TERRA stage workspace?
I think you mean an MC_WORKSPACE?

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -129,6 +136,95 @@ public void workspaceAuthz(AuthenticatedUserRequest userReq, UUID workspaceId, S
...
+   * directly on other workspaces.
+   * Trying to remove a role that a user does not have will succeed, though Sam will error if the
+   * email is not a registered user.
+   */
+  public void removeWorkspaceRole(
+      UUID workspaceId, AuthenticatedUserRequest userReq, IamRole role, String email) {
+    WorkspaceStage stage = workspaceDao.getWorkspaceStage(workspaceId);
+    if (!WorkspaceStage.MC_WORKSPACE.equals(stage)) {
```
ddietterich, line 182: Just so I understand... since it is an enum, can't you use `==` ?

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -129,6 +136,95 @@ public void workspaceAuthz(AuthenticatedUserRequest userReq, UUID workspaceId, S
...
+   * <p>This operation is only available to MC_TERRA stage workspaces, as Rawls manages permissions
+   * directly on other workspaces.
+   *
+   * @param workspaceId The workspace this operation takes place in
+   * @param userReq Credentials of the user requesting this operation. Only owners have permission
+   *     to modify roles in a workspace.
+   * @param role The role being granted.
+   * @param email The user being granted a role.
+   */
+  public void grantWorkspaceRole(
```
ddietterich, line 151: The grant and the revoke methods are nearly identical, is there a way to DRY it up?
One idea: factoring out stage checking and throwing into a method.

src/main/java/bio/terra/workspace/service/iam/model/RoleBinding.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.google.auto.value.AutoValue;
+import java.util.List;
+
+/** A binding between an IAM role and users/groups. */
+@AutoValue
+public abstract class RoleBinding {
+
+  /** The role granted to each of the bound users. */
+  public abstract IamRole role();
```
ddietterich, line 11: Again, my wanting to understand. Does this mean that the getter is just `role()`?  Could this be renamed to `getRole()` or does it have to be the eventual name of the class member?

src/main/resources/api/service_openapi.yaml
```diff
@@ -336,6 +382,22 @@ components:
...
+      in: path
+      description: An IAM role
+      required: true
+      schema:
+        $ref: '#/components/schemas/IamRole'
+
+    MemberEmail:
+      name: memberEmail
+      in: path
+      description: A user or group's email. Used for adding or removing IAM permissions
```
ddietterich, line 396: Validation? Or you count on Sam for that?

ddietterich (commented):
Many comments and questions. Mostly minor.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -268,6 +272,46 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
 
+  @Override
+  public ResponseEntity<Void> grantRole(
+      @PathVariable("id") UUID id,
+      @PathVariable("role") IamRole role,
+      @PathVariable("memberEmail") String memberEmail) {
+    samService.grantWorkspaceRole(
+        id,
+        getAuthenticatedInfo(),
+        bio.terra.workspace.service.iam.model.IamRole.fromApiModel(role),
```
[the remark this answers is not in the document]
zloery: We can't import both the generated `IamRole` and the internal `IamRole`, so one ends up fully specified. That felt fine to me since the Controller is explicitly the only place that will see both, but I could also change one or both names. Thoguhts?

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -268,6 +272,46 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+        memberEmail);
+    return new ResponseEntity<>(HttpStatus.valueOf(204));
```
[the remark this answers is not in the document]
zloery: Done, and changed the other controlled methods that do the same.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -268,6 +272,46 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
   }
   */
 
+  @Override
+  public ResponseEntity<Void> grantRole(
+      @PathVariable("id") UUID id,
```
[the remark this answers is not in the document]
zloery: Yep, that formatting comes from OpenAPI, and it throws an exception if the input isn't shaped properly.

src/main/java/bio/terra/workspace/service/iam/SamUtils.java
```diff
@@ -1,4 +1,4 @@
-package bio.terra.workspace.common.utils;
+package bio.terra.workspace.service.iam;
```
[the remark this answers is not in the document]
zloery: Agreed that utils is a weird name for a bunch of constants. I think this makes sense as a model since other parts of the code call into SamService with those values (e.g. calling `samService.isAuthorized` with a resource type), though I'm not 100% on the name.

src/main/java/bio/terra/workspace/service/iam/model/IamRole.java
```diff
@@ -11,17 +11,31 @@
...
 
   private static final BiMap<IamRole, String> samMap = EnumHashBiMap.create(IamRole.class);
+  private static final BiMap<IamRole, bio.terra.workspace.generated.model.IamRole> apiMap =
+      EnumBiMap.create(IamRole.class, bio.terra.workspace.generated.model.IamRole.class);
 
   static {
     samMap.put(IamRole.READER, "reader");
     samMap.put(IamRole.WRITER, "writer");
     samMap.put(IamRole.OWNER, "owner");
+
```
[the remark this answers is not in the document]
zloery: Currently we don't use generated classes anywhere inside services/DAOs, which we did intentionally to decouple WM's external interface from our internal representation. That approach makes a lot of sense to me for models or things that change, but you end up with more boilerplate and it's overkill for enums like this, though I still prefer this to bringing in generated classes just for here.

Another option is condensing the Sam names into the enum values themselves (e.g. what we do in `JobMapKeys`), but it felt a little off hardcoding values here to match Sam.

src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: Nice catch, thanks!

src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: yep, either works. I usually default to `equals`, is there a reason to use one over the other?

src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: That gets used a few other places as well, moved it to a common method

src/main/java/bio/terra/workspace/service/iam/model/RoleBinding.java
zloery: Since this is an AutoValue class, the getter is just `role()` and can't be changed. 
That said AutoValue has a bit of magic where if every field is prefixed with `get_` or `is_` it knows to strip them. So you can't change them arbitrarily, but you can pick which style you want.

src/main/resources/api/service_openapi.yaml
zloery: I was a little split on this one, though right now I just rely on Sam for that. Passing around a user-provided and unvalidated string doesn't feel great, but validating in two places introduces the risk of disagreeing on what's considered a valid email address (e.g. Sam has some list of blocked domains).

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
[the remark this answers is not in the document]
ddietterich: Got it. I didn't notice that.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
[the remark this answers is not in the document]
ddietterich: Cool! Good to know.

src/main/java/bio/terra/workspace/service/iam/SamService.java
ddietterich: Nice - I like the name

src/main/java/bio/terra/workspace/service/iam/model/IamRole.java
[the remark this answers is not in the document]
ddietterich: What about giving the enums member values like:
```
public enum IamRole m{
    READER("reader", bio.terra.workspace.generated.model.IamRole.READER), 
    ...

private final String samRole;
private final bio.terra.workspace.generated.model.IamRole apiRole;

IamRole(String samRole, bio.terra.workspace.generated.model.IamRole apiRole) {
    this.samRole = samRole;
    this.apiRole = apiRole;
}

public static IamRole fromApiModel(bio.terra.workspace.generated.model.IamRole apiModel) {
    Optional<IamRole> result =
        IamRole.values().stream().filter(x -> x.apiRole.equals(apiModel)).findFirst();
   // throw if null? return null? not sure what you want the semantics to be.
}
```

src/main/java/bio/terra/workspace/service/iam/model/RoleBinding.java
ddietterich: Thanks for explaining.

src/main/resources/api/service_openapi.yaml
ddietterich: My concern would not be about the email itself. It would be someone injecting additional URL elements in their address that would change the Sam semantics. So making sure the string does contain a `@` and doesn't contain `?` or `/` or other special URL characters. I'm happy to have Sam deal with whether the email is actually reasonable.

ddietterich (commented):
One more thought on the enums and one thought on the email validation.

src/main/resources/api/service_openapi.yaml
zloery: Makes sense. Once we start checking for `@` and excluding some special characters (but e.g. not `+`) though I think it makes sense to just use the same regex as Sam, with a comment indicating that's where it came from.

src/main/java/bio/terra/workspace/service/iam/model/IamRole.java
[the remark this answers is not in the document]
zloery: Done. I went with an exception when enums don't match, since that means someone changed a definition and something is broken.

src/main/resources/api/service_openapi.yaml
zloery: Also worth noting that unit tests and connected tests don't exercise the controller, so this sort of validation would be tested in a client test. I've added that to my backlog of tests to write

ddietterich (approved):
LGTM - thank you!

MatthewBemis (approved):
👍

PR #165, opened 2020-12-10
[PF-224] Create Sam workspaces with empty reader/writer policies

zloery:
This change modifies the call `SamService` makes when creating a workspace resource in Sam to include empty `reader` and `writer` policies in addition to the existing `owner` policy. Sam doesn't allow creating new policies on existing workspaces, so this is the only way to get modifiable `reader` and `writer` policies. 

This was previously broken in the Sam client library, but is fixed as of https://github.com/broadinstitute/sam/pull/493. This change bumps the Sam client library version to include the fix.

`createWorkspaceWithDefaults` is only called during workspace creation if a Sam resource doesn't already exist. Rawls already creates Sam resources, so this is only used on workspaces not managed by Rawls.

For tests, this also adds methods to `UserAccessUtils` for getting auth objects and modifies connected tests to use them. Previously, these tests only used an access token with no email.


src/main/java/bio/terra/workspace/service/iam/model/IamRole.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.common.collect.EnumHashBiMap;
+
+/** Internal representation of IAM roles. */
+public enum IamRole {
+  READER,
+  WRITER,
+  OWNER;
+
+  private static final BiMap<IamRole, String> samMap = EnumHashBiMap.create(IamRole.class);
+  private static final BiMap<IamRole, bio.terra.workspace.generated.model.IamRole> apiMap =
```
zloery, line 14: Technically this isn't used in this change and could go in #164 instead - I just left it here since I'm expected both PRs to go in at similar times.

davidangb (approved):
lgtm!

MatthewBemis (approved):

PR #167, opened 2020-12-18
[no ticket] Add config resolution priority link to readme

zloery:
No ticket, this is a documentation change only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,12 @@ In general, we use type-safe configuration parameters as shown here:
 That allows proper typing of parameters read from property files or environment variables. Parameters are
 then accessed with normal accessor methods. You should never need to use an `@Value` annotation.
 
+Be aware that environment variables will override values in our YAML configuration.
+This should not be used for configuration as it makes the source of values harder to track,
+but it may be useful for debugging unexpected configurations. See Spring Boot's 
+[Externalized Configuration documentation](https://docs.spring.io/spring-boot/docs/2.4.0/reference/html/spring-boot-features.html#boot-features-external-config)
+for the exact priority order of configurations.
+
 ### Initialization
 When the applications starts, Spring wires up the components based on the profiles in place.
 Setting different profiles allows different components to be included. This technique is used


yonghaoy (approved):
Not related with the PR, just two small findings:
I saw WSM use env variables in application.yaml and helm already: https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/resources/application.yml#L8
And helm file:
https://github.com/broadinstitute/terra-helm/blob/master/charts/workspacemanager/templates/deployment.yaml#L115

I feel it is possible to confuse Spring sometimes and may let Spring by accidentally override some values. For example  STAIRWAY_DATABASE_NAME, it can also let Spring to override some config like: stairway.database-name


What we did in Buffer Service and Janitor is to let env variable match Spring config name. For example, we have
buffer.stairway.db.password in spring config, 
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/application.yml#L19
In helm we just use BUFFER_STAIRWAY_DB_PASSWORD
https://github.com/broadinstitute/terra-helm/blob/master/charts/buffer/templates/deployment.yaml#L49

So I think it might be easier to just to use the spring's path as env name. 

Tip 2 not sure if you know, the spring's magic doesn't need even require the value presented in application.yml. It can by pass that file and associate with your java code directly. 
For example: we have default value as false for cleanupAfterHandout and a setter in java code. 
 https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/app/configuration/CrlConfiguration.java#L104
Our application.yml doesn't have that at all, but instead we set in helm directly:
https://github.com/broadinstitute/terra-helm/blob/master/charts/buffer/templates/deployment.yaml#L69

zloery:
For 1, yeah that can happen. I think we explicitly want to avoid that, which is why we split all the configs we expect to read env vars apart in application.yml. So far we've avoided that (especially because most vars are nested under `workspace:` - the risk of it happening accidentally was why I wanted to add it to the README.

For 2 yeah, I think we do that already with some spend profile stuff. I'm not thrilled with doing that since it makes it a lot harder to track where config values are coming from. I'd definitely rather have a few values in config that don't really change much if they can ever be modified

gjuggler:
> For 1, yeah that can happen. I think we explicitly want to avoid that, which is why we split all the configs we expect to read env vars apart in application.yml. So far we've avoided that (especially because most vars are nested under `workspace:` - the risk of it happening accidentally was why I wanted to add it to the README.
> 
> For 2 yeah, I think we do that already with some spend profile stuff. I'm not thrilled with doing that since it makes it a lot harder to track where config values are coming from. I'd definitely rather have a few values in config that don't really change much if they can ever be modified

Re: point 1, injecting my opinion / long term goal: I'd like to eventually move away from using individual env vars for setting application-level configs, in favor of Spring's support for directly reading app config from k8s ConfigMap values ([docs](https://cloud.spring.io/spring-cloud-static/spring-cloud-kubernetes/1.0.0.M2/multi/multi__configmap_propertysource.html)). This will give us a path towards being able to modify certain app config values without re-deploying at the k8s level, which will be a useful operational hook.

If we make that shift, the env-mapping boilerplate in WSM would be unnecessary.

zloery:
SGTM - the main benefit of explicitly listing the variables that get overridden is having them all in one place, so if that place is a ConfigMap instead that's fine. I'll still merge this README change for now since it's useful information

PR #168, opened 2020-12-18
QA-1327: Testrunner test steps are launched successfully through GitHub actions

ichengchang:
The `test-runner-integration.yml` demonstrates automation of test runner scripts through GitHub Actions.


.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,56 @@
...
+            .gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Run GetWorkspace tests using test runner library
+        run: |
+          cd ${GITHUB_WORKSPACE}/${workingDir}/workspace-manager-client
+          ../gradlew clean
+          ../gradlew build
+          cd ${GITHUB_WORKSPACE}/${workingDir}/workspace-manager-clienttests
+          ./gradlew runTest --args="configs/BasicAuthenticated.json /tmp/TestRunnerResults"
```
davidangb, line 56: this will run an arbitrary number of concurrent perf tests against the dev workspace manager (and dev sam, etc). Since this action is triggered on every `push`, we can have many tests running at once - as many as we have commits to open PRs plus merges to dev, etc. Is that what we really want to do? It feels like that could cause lots of variability in test results.

gjuggler: I don't think there's any intention to be running perf tests (yet) on every PR push. I'd expect this execution mode to be scoped only to integration-style tests.

There's a separate discussion, picking up some steam, around how exactly we handle issues like this (contention for integration environments from test runners) while we're still waiting on preview environments. See [Service Test Topologies](https://docs.google.com/document/d/13jwti5yTZAKpV2OIM6Rr3e79CdiFQF--8VnuFvEFSXM/edit?resourcekey=0-ByHOWp_X5-bqKSNlqL4CPQ#heading=h.htiffe9pigic) which I'm currently touching up in prep to share more widely and start moving forward on. Might be a good topic for the WSM sync tomorrow.

davidangb: As is, this PR has
```
on:
  push:
    ...
```
so it *will* run perf tests on every push. If that's not the intention, this PR will have to change its trigger. I am considering the `configs/BasicAuthenticated.json` as a perf test; are you suggesting it's not?

gjuggler: Right – good point. My working assumption here is that the suite of tests to be run during PR review won't include performance tests, so if that's not reflected here, we should fix it. One thing @ichengchang that might help would be to be very explicit in the config names, whether a given suite of tests is Integration vs. Performance. Perhaps a common prefix, or folder, e.g.

configs/IntegrationBasic.json, or
configs/integration/BasicAuthenticated.json

(and eventually...)

configs/PerfSmokeTests.json
configs/PerfFullSuite.json, or
configs/perf/SmokeTests.json

ddietterich: I would prefer the folder structure.

.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Set up AdoptOpenJDK 11
+        uses: joschi/setup-jdk@v2
+        with:
+          java-version: 11
+      - name: Get Vault token
+        id: vault-token-step
+        env:
+          VAULT_ADDR: https://clotho.broadinstitute.org:8200
+        run: |
+          VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
```
jaycarlton, line 27: Could this chunk be its own script for readability?

.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,55 @@
...
+            .gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Run GetWorkspace tests using test runner library
+        run: |
+          cd ${GITHUB_WORKSPACE}/${workingDir}/workspace-manager-client
+          ../gradlew clean
+          ../gradlew build
+          cd ${GITHUB_WORKSPACE}/${workingDir}/workspace-manager-clienttests
+          ./gradlew runTest --args="configs/integration/BasicAuthenticated.json /tmp/TestRunnerResults"
```
jaycarlton, line 55: nit: newline at EOF here and below

jaycarlton: It seems odd to have  to call gradlew three times in succession. Could we not have one job that cleans, builds, and runs the tests? Or is this because we're in a different gradle project for the tests?

workspace-manager-clienttests/src/main/resources/configs/integration/BasicAuthenticated.json
```diff
@@ -0,0 +1,17 @@
+{
```
jaycarlton, line 1: why is integration under client tests? I thought those were an independent category. Now `GetWorkspace` is in two config files. Is that intentional?

jaycarlton (approved):

.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up AdoptOpenJDK 11
+        uses: joschi/setup-jdk@v2
+        with:
+          java-version: 11
+      - name: Get Vault token
+        id: vault-token-step
+        env:
+          VAULT_ADDR: https://clotho.broadinstitute.org:8200
```
zarsky-broad, line 25: Why is this provided via env if it's just substituted in once after?

.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,55 @@
...
+            .gradle/caches
+            .gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Run GetWorkspace tests using test runner library
+        run: |
+          cd ${GITHUB_WORKSPACE}/${workingDir}/workspace-manager-client
+          ../gradlew clean
+          ../gradlew build
+          cd ${GITHUB_WORKSPACE}/${workingDir}/workspace-manager-clienttests
```
zarsky-broad, line 54: Is this supposed to just build the client? This will build everything. We also don't need to run clean before. 

```suggestion
          ./gradlew :workspace-manager-client:build
          cd workspace-manager-clienttests
```

.github/workflows/test-runner-integration.yml
zarsky-broad: Should the client tests be their own independent project? Or are they a subproject? If the latter, they should not have their own gradle invocation. If the former, why?

zarsky-broad (changes requested):

jaycarlton:
https://github.com/DataBiosphere/terra-workspace-manager/pull/168#discussion_r555314217

We have a ticket to make it a sub-project. A standalone project makes sense to me too.

.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,56 @@
...
+              secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+          echo ::set-output name=vault-token::$VAULT_TOKEN
+          echo ::add-mask::$VAULT_TOKEN
+      - name: "Configure Test Runner"
+        run: |
+          cd ${GITHUB_WORKSPACE}/${workingDir}/workspace-manager-clienttests
+          chmod +x render-config.sh
+          ./render-config.sh ${{ steps.vault-token-step.outputs.vault-token }}
+          chmod +r rendered
+          ls -l rendered
```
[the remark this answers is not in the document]
ichengchang: Removed the `ls -l ` but kept `chmod +r` to be consistent with the `test.yml` workflow.

.github/workflows/test-runner-integration.yml
ichengchang: Created separate subfolders (`integration`, `perf`) as suggested.

.github/workflows/test-runner-integration.yml
ichengchang: This was done the same way as workflow `test.yml` out of consistency.

.github/workflows/test-runner-integration.yml
ichengchang: The first two `gradlew` operations have been merged into a single `gradlew` as suggested by @zarsky-broad. Right now it's just `./gradlew :workspace-manager-client:build`, or we can add back the `clean` task to the build process `./gradlew :workspace-manager-client:clean :workspace-manager-client:build`. This step builds a local client jar. 

Developers can specify the location of the local client jar through either a `gradle` property `workspacemanagerclientjar` or `env` `ORG_GRADLE_PROJECT_workspacemanagerclientjar` as described in the `README`.

The client tests were set up as independent project with their own `gradle`, avoiding having to make the decision to touch the main project `build.gradle` and `settings.gradle`. So far, this set up is sufficient. Team should decide if they want it to be.a subproject managed by the main `build.gradle`.

.github/workflows/test-runner-integration.yml
ichengchang: We could move this code into its own reusable script and make the yml file more readable. Right now there are not that many workflows so I am not too worrying about it.

workspace-manager-clienttests/src/main/resources/configs/integration/BasicAuthenticated.json
ichengchang: The following updates were made after following up with @marikomedlock.

We will keep to the convention of having separate subfolders under configs/ to group tests under different test categories. 

For example, under the `configs/` folder, there are two subfolders: `integration` and `perf`.
Inside each subfolder (`integration` or `perf`), we have the test configuration files. 
The test configuration files determine the target scenario, user load, which server config to use etc.
So even the same scenario `GetWorkspace` was used, the target load and server config can be different for `integration` vs `perf`. 

Also the `suites/` folder is important, the `json` files inside this folder defines how the tests are grouped together. The sample GitHub action workflows `test-runner-nightly-perf.yml`, `test-runner-integration.yml` use the `json` definitions in `suites/` folder to launch the appropriate tests.

Hope this makes it all clear.

jaycarlton: Is there ever a reason to have more than one test script in a test config file (as opposed to putting them in suites)?

.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,55 @@
...
+          path: |
+            .gradle/caches
+            .gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Run Basic Intergation tests using test runner library
+        run: |
+          ./gradlew :workspace-manager-client:build
+          cd workspace-manager-clienttests
+          ls src/main/resources/configs/integration
```
jaycarlton, line 53: Are the `ls` lines just for debugging? Maybe a separate script with dedicated logging would be clearer if you want to keep this kind of info.

.github/workflows/test-runner-nightly-perf.yml
```diff
@@ -0,0 +1,54 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Sample Nightly Perf Tests
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: '0 14 * * *' # run at 2 PM UTC, 9 AM EST.
```
jaycarlton, line 9: Why 9 AM EST? That's going to overlap with the start of the workday for Cambridge.

.github/workflows/test-runner-nightly-perf.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Set up AdoptOpenJDK 11
+        uses: joschi/setup-jdk@v2
+        with:
+          java-version: 11
+      - name: Get Vault token
+        id: vault-token-step
+        env:
+          VAULT_ADDR: https://clotho.broadinstitute.org:8200
+        run: |
+          VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
```
jaycarlton, line 26: Can you pull this docker business into its own script or action? It looks like we're dropping two abstraction levels from the surrounding config. I understand if GH actions doesn't have great support here.

.github/workflows/test-runner-nightly-perf.yml
```diff
@@ -0,0 +1,54 @@
...
+            -e "VAULT_ADDR=${VAULT_ADDR}" \
+            vault:1.1.0 \
+            vault write -field token \
+              auth/approle/login role_id=${{ secrets.VAULT_APPROLE_ROLE_ID }} \
+              secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+          echo ::set-output name=vault-token::$VAULT_TOKEN
```
jaycarlton, line 32: Why `echo` without having something receiving the output?

.github/workflows/test-runner-nightly-perf.yml
```diff
@@ -0,0 +1,54 @@
...
+            .gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Run Basic Intergation tests using test runner library
+        run: |
+          ./gradlew :workspace-manager-client:build
+          cd workspace-manager-clienttests
+          ls src/main/resources/configs/integration
+          ls src/main/resources/suites
+          ./gradlew runTest --args="suites/BasicNightlyPerf.json /tmp/TestRunnerResults"
```
jaycarlton, line 54: Maybe parameterize the test output directory so we can find it easily for uploading results or debugging.

Likewise, can the list of suites be taken as an action-level input?

workspace-manager-clienttests/README.md
```diff
@@ -19,13 +19,13 @@ Note: `render-config.sh` assumes a valid `.vault-token` in your `$HOME` director
 To run the test, use the following command from the workspace-manager-clienttests directory.
 
 ```
-./gradlew  runTest --args="configs/BasicAuthenticated.json /tmp/TR"
+./gradlew  runTest --args="integration/BasicAuthenticated.json /tmp/TR"
```
jaycarlton, line 22: Aside: right now all `BasicAuthenticated` has is `GetWorkspace`, so I'm wondering if it needs a new name. It's a config, and not a suite, so maybe that's the confusion. Maybe change `Basic` to `Smoke`, since that's kinda what this test does, and it's referred to as a smoke test somewhere else.

workspace-manager-clienttests/README.md
```diff
@@ -38,13 +38,13 @@ The version of the Workspace Manager client JAR file is specified in the build.g
...
 
-`./gradlew -Pworkspacemanagerclientjar=~/terra-workspace-manager/workspace-manager-client/build/libs/workspace-manager-client-0.5.0-SNAPSHOT.jar runTest --args="configs/BasicUnauthenticated.json"
+`./gradlew -Pworkspacemanagerclientjar=~/terra-workspace-manager/workspace-manager-client/build/libs/workspace-manager-client-0.5.0-SNAPSHOT.jar runTest --args="integration/BasicAuthenticated.json"
 
 or an environment variable.
 
 ```
 export ORG_GRADLE_PROJECT_workspacemanagerclientjar=~/terra-workspace-manager/workspace-manager-client/build/libs/workspace-manager-client-0.5.0-SNAPSHOT.jar
-./gradlew runTest --args="configs/BasicUnauthenticated.json /tmp/TestRunnerResults"
+./gradlew runTest --args="integration/BasicAuthenticated.json /tmp/TestRunnerResults"
```
jaycarlton, line 47: So should we rename the directory from `-clienttests` to `-integrationtests`? I'm confused as to the degree these terms are interchangeable.

workspace-manager-clienttests/src/main/resources/configs/integration/BasicAuthenticated.json
```diff
@@ -0,0 +1,17 @@
...
+  "name": "BasicAuthenticated",
+  "description": "Get workspace once. Authentication required.",
+  "serverSpecificationFile": "workspace-dev.json",
+  "kubernetes": {},
+  "application": {},
+  "testScripts": [
+    {
+      "name": "GetWorkspace",
+      "numberOfUserJourneyThreadsToRun": 1,
```
jaycarlton, line 10: I thought this was 10 before.

workspace-manager-clienttests/src/main/resources/suites/BasicNightlyPerf.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "BasicNightlyPerf",
+  "description": "Nightly perf tests.",
```
jaycarlton, line 3: If the schedule is independent of this config, then I'd simply name it `Perf Tests`.

jaycarlton (approved):

.github/workflows/test-runner-nightly-perf.yml
zloery: `echo ::set-output...` and `echo ::add-mask...` are commands that get read by the Github runner itself, [see docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions). `set-output` makes output available for other steps, and `add-mask` prevents the specified value from being logged.

.github/workflows/test-runner-integration.yml
ichengchang: Removed debugging lines.

.github/workflows/test-runner-nightly-perf.yml
ichengchang: Team can decide when's an appropriate time to trigger the test. I updated it to 1AM ET.

.github/workflows/test-runner-nightly-perf.yml
ichengchang: According to documentation [Workflow commands for GutHub Actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#using-workflow-commands-to-access-toolkit-functions).

The `echo` command is used to send the workflow command (in this case `set-output`) to the runner over `stdout`, which is equivalent to executing an API `core.setOutput('vault-token', $VAULT_TOKEN)`. So it's not using `echo` is the usual sense.

.github/workflows/test-runner-nightly-perf.yml
ichengchang: Right now, this is what everybody's else is doing. I'll create a separate ticket for future discussions across teams.

.github/workflows/test-runner-nightly-perf.yml
ichengchang: We are working towards a solution to upload test results to BigQuery tables, so parameterization of the output location is under consideration.

workspace-manager-clienttests/src/main/resources/suites/BasicNightlyPerf.json
ichengchang: Renamed the config and updated README to remove the suggestion of schedule dependency.

workspace-manager-clienttests/src/main/resources/configs/integration/BasicAuthenticated.json
ichengchang: Yes, this was changed from 10 to 1 when I separate out the configs into `integration` and `perf`.

.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,53 @@
...
+          chmod +r rendered
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            .gradle/caches
+            .gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Run Basic Intergation tests using test runner library
```
marikomedlock, line 49: Nit: Build the WSM client and run the Basic Integration test suite.

.github/workflows/test-runner-nightly-perf.yml
```diff
@@ -0,0 +1,52 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Sample Nightly Perf Tests
```
marikomedlock, line 4: Nit: Not sure if the Sample was a leftover or intended?

.github/workflows/test-runner-nightly-perf.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: |
+          VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
+            -e "VAULT_ADDR=${VAULT_ADDR}" \
+            vault:1.1.0 \
+            vault write -field token \
+              auth/approle/login role_id=${{ secrets.VAULT_APPROLE_ROLE_ID }} \
+              secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+          echo ::set-output name=vault-token::$VAULT_TOKEN
+          echo ::add-mask::$VAULT_TOKEN
+      - name: "Configure Test Runner"
```
marikomedlock, line 34: Same comment as above. This section renders the configs for Test Runner to use.

.github/workflows/test-runner-nightly-perf.yml
```diff
@@ -0,0 +1,52 @@
...
+          chmod +r rendered
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            .gradle/caches
+            .gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Run Basic Intergation tests using test runner library
```
marikomedlock, line 48: Same comment as above. I just think it'd be good to note that we're building the client library here. Also, this says Integration, I think it should be Run Basic Perf test suite.

.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: |
+          ./gradlew :workspace-manager-client:build
```
marikomedlock, line 51: I see that we're building a local version of the WSM client library here. Is it actually getting used by Test Runner though? I think you need to set an env var for it to override the default client library specified in the build.gradle file.

.github/workflows/test-runner-nightly-perf.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: |
+          ./gradlew :workspace-manager-client:build
```
marikomedlock, line 50: Same comment as above. Is this local version of the client library actually being used?

workspace-manager-clienttests/README.md
```diff
@@ -16,16 +16,25 @@ Note: `render-config.sh` assumes a valid `.vault-token` in your `$HOME` director
...
 
-To run the test, use the following command from the workspace-manager-clienttests directory.
+To run the tests, use one of the following commands from the workspace-manager-clienttests directory.
+NB: The `BasicIntegration.json` and `BasicNightlyPerf.json` are sample test suite configurations for Integration and Perf tests.
+See `test-runner-integration.yml` and `test-runner-nightly-perf.yml` in `.github/workflows` for use cases.
 
+Integration Test
 ```
-./gradlew  runTest --args="configs/BasicAuthenticated.json /tmp/TR"
+./gradlew  runTest --args="suites/BasicIntegration.json /tmp/TR"
```
marikomedlock, line 25: I think the README should include both instructions: how to run an individual test and how to run a suite. For developers writing tests, it will often be helpful to run a single test.

workspace-manager-clienttests/README.md
```diff
@@ -16,16 +16,25 @@ Note: `render-config.sh` assumes a valid `.vault-token` in your `$HOME` director
...
 ```
 
-The default server that this test will run against is specified in the resources/configs/BasicAuthenticated.json file.
+Perf Test
+```
+./gradlew  runTest --args="suites/BasicPerf.json /tmp/TR"
+```
+
+The default server that the test will run against is specified in the resources/configs/integration/BasicAuthenticated.json file, using integration test as example.
```
marikomedlock, line 33: Nit: the default server is specified in the test config file.

workspace-manager-clienttests/README.md
```diff
@@ -38,13 +47,13 @@ The version of the Workspace Manager client JAR file is specified in the build.g
 fetched from the Broad Institute Maven repository. You can override this to use a local version of the Workspace Manager client
 JAR file by specifying a Gradle project property, either with a command line argument
 
-`./gradlew -Pworkspacemanagerclientjar=~/terra-workspace-manager/workspace-manager-client/build/libs/workspace-manager-client-0.5.0-SNAPSHOT.jar runTest --args="configs/BasicUnauthenticated.json"
+`./gradlew -Pworkspacemanagerclientjar=~/terra-workspace-manager/workspace-manager-client/build/libs/workspace-manager-client-0.5.0-SNAPSHOT.jar runTest --args="suites/BasicIntegration.json"
```
marikomedlock, line 50: I think the single test config run command is more helpful. Or include both. But just the suite run command is probably not sufficient for all use cases.

workspace-manager-clienttests/README.md
```diff
@@ -38,13 +47,13 @@ The version of the Workspace Manager client JAR file is specified in the build.g
...
 
 or an environment variable.
 
 ```
 export ORG_GRADLE_PROJECT_workspacemanagerclientjar=~/terra-workspace-manager/workspace-manager-client/build/libs/workspace-manager-client-0.5.0-SNAPSHOT.jar
-./gradlew runTest --args="configs/BasicUnauthenticated.json /tmp/TestRunnerResults"
+./gradlew runTest --args="suites/BasicIntegration.json /tmp/TestRunnerResults"
```
marikomedlock, line 56: Nit: Same as above. I'd just keep the single test config run command.

workspace-manager-clienttests/src/main/resources/configs/integration/BasicAuthenticated.json
marikomedlock: Yes, there is. A test script implements a particular user journey. A test config specifies which user journeys to run and how to scale them. A test config can run several different user journeys concurrently to test a varying load (e.g. we want to test that the server survives 100 workspace create requests and 200 workspace get requests at the same time). Let me know if it's not clear -- happy to talk on a call also.

workspace-manager-clienttests/README.md
ichengchang: The reference to `client` presents the ability of reusing the published or local client jar for writing test cases. Whether it is `smoke`, `integration`, or `perf` test, developers should be able to write the test using the same test runner framework. The client jar we are leveraging here is defined in the corresponding `build.gradle`. 

```
    // Workspace Manager client library
    // Gradle project property "workspacemanagerclientjar" overrides the fetch from Maven
    if (project.hasProperty("workspacemanagerclientjar")) {
        implementation files(project.findProperty("workspacemanagerclientjar"))
        compile "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jackson}"
        compile "org.glassfish.jersey.core:jersey-client:${jersey}"
        compile "org.glassfish.jersey.media:jersey-media-json-jackson:${jersey}"
        compile "org.glassfish.jersey.media:jersey-media-multipart:${jersey}"
        compile "org.glassfish.jersey.inject:jersey-hk2:${jersey}"
    } else {
        implementation "bio.terra:workspace-manager-client:${workspacemanagerClient}"
        compile "io.swagger.core.v3:swagger-annotations:${swaggerAnnotations}"
    }
```

.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,53 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Sample Integration Tests
```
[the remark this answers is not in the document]
ichengchang: Updated workflow comments as suggested.

.github/workflows/test-runner-integration.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: |
+          VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
+            -e "VAULT_ADDR=${VAULT_ADDR}" \
+            vault:1.1.0 \
+            vault write -field token \
+              auth/approle/login role_id=${{ secrets.VAULT_APPROLE_ROLE_ID }} \
+              secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+          echo ::set-output name=vault-token::$VAULT_TOKEN
+          echo ::add-mask::$VAULT_TOKEN
+      - name: "Configure Test Runner"
```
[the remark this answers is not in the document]
ichengchang: Updated workflow comments as suggested.

.github/workflows/test-runner-integration.yml
ichengchang: Updated comments as suggested.

.github/workflows/test-runner-nightly-perf.yml
ichengchang: Replaced this with comments suggested.

.github/workflows/test-runner-nightly-perf.yml
ichengchang: Updated the comments as suggested earlier.

workspace-manager-clienttests/README.md
ichengchang: Updated comments.

workspace-manager-clienttests/README.md
ichengchang: I have added both test modes: single test and test suite in separate sections.
Developers should be able to follow the guidelines to create their own use cases, whether for individual tests or grouping multiple tests into a test suite.

workspace-manager-clienttests/README.md
ichengchang: Updated. With the added sections on individual versus test suite modes. It should be clear which mode we are using as an example here,

.github/workflows/test-runner-nightly-perf.yml
ichengchang: Updated comments in workflow yaml.

workspace-manager-clienttests/README.md
ichengchang: Both individual test and test suite modes are included in README now.

.github/workflows/test-runner-nightly-perf.yml
ichengchang: Updated comments.

.github/workflows/test-runner-integration.yml
ichengchang: I have expanded this workflow to include steps on using local client jar. Updated README which references this workflow.

workspace-manager-clienttests/README.md
ichengchang: We can look at it this way. The `GetWorkspace` example has a simple user journey of retrieving a valid workspace. The user journey can be expanded to involve other APIs (e.g. Data Repo) but simply not in scope for this ticket.

For example, [WorkspaceIntegrationTest](https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/test/java/bio/terra/workspace/integration/WorkspaceIntegrationTest.java) can be refactored using the test runner framework. The PF team might already been looking at it.

.github/workflows/test-runner-integration.yml
ichengchang: Closing this thread as resolved.

zarsky-broad (approved):

PR #169, opened 2021-01-04
Tn pf 301 fix flight structure

tnadler:
Fixing flight structure, per https://docs.google.com/document/d/1qJz0e1zzp28jV_3bVN9zwbu582sfUXDMd4Mw19v66Io/edit?ts=5fdcd66e&resourcekey=0-3gLFZURTsw6zPjLiEs2agA#

### DIFF
--- a/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextRBSFlight.java
+++ b/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextRBSFlight.java
@@ -1,5 +1,6 @@
 package bio.terra.workspace.service.workspace.flight;
 
+import bio.terra.cloudres.google.cloudresourcemanager.CloudResourceManagerCow;
 import bio.terra.stairway.Flight;
 import bio.terra.stairway.FlightMap;
 import bio.terra.stairway.RetryRule;
@@ -13,12 +14,14 @@ public class CreateGoogleContextRBSFlight extends Flight {
     super(inputParameters, applicationContext);
     ApplicationContext appContext = (ApplicationContext) applicationContext;
     BufferService bufferService = appContext.getBean(BufferService.class);
-    // TODO(tovanadler): Update the retry rules once we properly design the flight.
+    CloudResourceManagerCow resourceManager = appContext.getBean(CloudResourceManagerCow.class);
+
     RetryRule retryRule =
         new RetryRuleExponentialBackoff(
             /* initialIntervalSeconds= */ 1,
-            /* maxIntervalSeconds= */ 5*60,
+            /* maxIntervalSeconds= */ 5 * 60,
             /* maxOperationTimeSeconds= */ 16);
-    addStep(new PullProjectFromPoolStep(bufferService), retryRule);
+    addStep(new GenerateResourceIdStep());
+    addStep(new PullProjectFromPoolStep(bufferService, resourceManager), retryRule);
   }
 }
--- a/src/main/java/bio/terra/workspace/service/workspace/flight/GenerateResourceIdStep.java
+++ b/src/main/java/bio/terra/workspace/service/workspace/flight/GenerateResourceIdStep.java
@@ -1,0 +1,34 @@
+package bio.terra.workspace.service.workspace.flight;
+
+import static bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.RBS_RESOURCE_ID;
+
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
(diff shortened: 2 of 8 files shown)


src/main/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextRBSFlight.java
```diff
@@ -1,24 +1,44 @@
...
+            /* maxIntervalSeconds= */ 8,
             /* maxOperationTimeSeconds= */ 16);
-    addStep(new PullProjectFromPoolStep(bufferService), retryRule);
+
+    addStep(new GenerateResourceIdStep());
+    addStep(new PullProjectFromPoolStep(bufferService, resourceManager), rbsRetryRule);
+
+    // TODO(tovanadler): Enable this once we fix configurations.
+    // addStep(new SetProjectBillingStep(billingClient));
+    // addStep(new StoreGoogleContextStep(workspaceDao, transactionTemplate), storeRetryRule);
```
zloery, line 42: It looks like StoreGoogleContextStep needs a workspace ID in the flight input, this will cause your test to fail for now

src/main/java/bio/terra/workspace/service/workspace/flight/GenerateResourceIdStep.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    return StepResult.getStepResultSuccess();
+  }
+
+  /** Generate a pseudo-random resource id */
+  public static String randomResourceId() {
+    return "wm-" + Long.valueOf(UUID.randomUUID().getMostSignificantBits()).toString();
```
zloery, line 32: I'm not familiar with RBS so this is mostly for my knowledge - is this an established format for resource ID requests?

zloery (approved):
LGTM with one issue that will likely break tests

src/main/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextRBSFlight.java
tnadler: Yup, just commented it out. Meant to before.

zloery: jk, looks like you beat me to it

src/main/java/bio/terra/workspace/service/workspace/flight/GenerateResourceIdStep.java
tnadler: @yonghaoy I am actually not sure, this just seemed reasonable. Thoughts?

yonghaoy: any unique random string works! I am not famillar with `getMostSignificantBits()` method. Is that large enough to avoid collision? If yes, this method LG!

src/main/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextRBSFlight.java
```diff
@@ -1,24 +1,44 @@
...
 import bio.terra.workspace.service.buffer.BufferService;
 import org.springframework.context.ApplicationContext;
+import org.springframework.transaction.support.TransactionTemplate;
 
 // NOTE: DO NOT USE. Currently just a shell class to exercise connection to Buffer Service.
 public class CreateGoogleContextRBSFlight extends Flight {
   public CreateGoogleContextRBSFlight(FlightMap inputParameters, Object applicationContext) {
     super(inputParameters, applicationContext);
     ApplicationContext appContext = (ApplicationContext) applicationContext;
     BufferService bufferService = appContext.getBean(BufferService.class);
```
yonghaoy, line 19: How the bean get Buffer Service? When we create the BufferService, where is the code to set access token in request header?

yonghaoy: Example code without dependency injection: https://github.com/DataBiosphere/terra-resource-buffer/blob/mm-PF-237-use-test-runner-library/buffer-clienttests/src/main/java/scripts/utils/BufferServiceUtils.java#L55-L67

With Spring DI, that should be a bean like: https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java#L33

yonghaoy (approved):

PR #171, opened 2021-01-05
Increate default timeouts

tnadler:


yonghaoy (approved):

ddietterich (changes requested):
I'd like to understand the rationale here better. I think we need to write this down and get some other eyes on it.

If an http request is taking longer to respond than 30-60 seconds, then there is a protocol problem. Waiting 5 minutes delays failure detection of services that should be retried. To elaborate: suppose we are talking to TDR. We will get a network timeout if a TDR pod goes down. We need to retry in order to failover to a new pod. It should not take us 5 minutes to fail over. It should be more like 30 seconds.

tnadler:
@yonghaoy I would love your thoughts here. I had initially just wanted to increase the retry timeout on the flight itself but yonghao was concerned some operations were quite long lived themselves.

tnadler:
I have lost the thread on this. @yonghaoy @dd - shall I revert this?

ddietterich:
I suggest closing this. It isn't the approach we want to take.

PR #172, opened 2021-01-05
[PF-159] Sync Sam groups to GCP project on context creation

zloery:
This change adds extra steps to the `CreateGoogleContextFlight` to sync Sam policies (now reader, writer, and owner as of #165) to google groups, then grants those google groups GCP roles on the new project. This does not affect existing GCP roles on the project, and so is safe to use with projects from RBS which may have pre-existing roles.

This is only triggered by the `createGoogleContext` endpoint (which uses workspace stage feature gating), so this feature is inaccessible to Rawls-managed workspaces.

Currently, the IAM role -> GCP role mapping is hardcoded in `CloudSyncRoleMapping`. This is a medium term solution and liable to change until a more permanent design for cloud sync functionality is finished.


src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -0,0 +1,24 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.util.List;
+
+public class CloudSyncRoleMapping {
+  public static final ImmutableMap<IamRole, List<String>> cloudSyncRoleMap =
+      ImmutableMap.of(
+          IamRole.OWNER,
+              ImmutableList.of(
+                  "roles/viewer",
```
ddietterich, line 13: I thought in your spec you had identified only roles/viewer on the project as the permission for all workspace roles. I didn't think we wanted uncontrolled objects created in the workspace project. For example, giving dataEditor on the project lets the workspace owner create BigQuery datasets in the project. Similarly, storage.objectAdmin.

src/main/java/bio/terra/workspace/service/workspace/flight/GoogleCloudSyncStep.java
```diff
@@ -0,0 +1,103 @@
...
+      Map<IamRole, String> samEmailMap =
+          Map.of(
+              IamRole.READER,
+                  workingMap.get(WorkspaceFlightMapKeys.IAM_READER_GROUP_EMAIL, String.class),
+              IamRole.WRITER,
+                  workingMap.get(WorkspaceFlightMapKeys.IAM_WRITER_GROUP_EMAIL, String.class),
+              IamRole.OWNER,
+                  workingMap.get(WorkspaceFlightMapKeys.IAM_OWNER_GROUP_EMAIL, String.class));
+      List<Binding> newBindings = new ArrayList<>();
+      for (IamRole iamRole : IamRole.values()) {
```
ddietterich, line 71: This code assumes that the list of IamRole.values() is identical to the roles in the map. Might there be more roles than there are sync'd roles?

Also, it doesn't make sense to me to build the map and then take it apart to build the bindings.Seems like it'd be easier to directly build the bindings.  Maybe a method to build a binding for a role. What am I missing?

src/main/java/bio/terra/workspace/service/workspace/flight/GoogleCloudSyncStep.java
```diff
@@ -0,0 +1,103 @@
...
+      Policy newPolicy =
+          new Policy()
+              .setVersion(currentPolicy.getVersion())
+              .setBindings(newBindings)
+              .setEtag(currentPolicy.getEtag());
+      SetIamPolicyRequest iamPolicyRequest = new SetIamPolicyRequest().setPolicy(newPolicy);
+      logger.info("Setting new Cloud Context IAM policy: " + iamPolicyRequest.toPrettyString());
+      resourceManagerCow.projects().setIamPolicy(gcpProjectName, iamPolicyRequest).execute();
+    } catch (IOException e) {
+      throw new RetryException("Error setting IAM permissions", e);
```
ddietterich, line 90: Should this be a more specific exception? I don't understand the pattern for making new exceptions in WM.

src/main/java/bio/terra/workspace/service/workspace/flight/SyncSamGroupsStep.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * A {@Code Step} which synchronizes Sam policies with google groups and stores the group names in
+ * the Stairway working map.
+ */
+public class SyncSamGroupsStep implements Step {
+
+  private final SamService samService;
+
+  @Autowired
```
ddietterich, line 23: I don't think you can use @Autowired in steps. They are dynamically created by Stairway and won't get injection. I noticed it here, but you use it in other steps as well.

src/main/java/bio/terra/workspace/service/workspace/flight/GoogleCloudSyncStep.java
```diff
@@ -0,0 +1,103 @@
...
+ * CloudSyncRoleMapping}. Note that the bindings list sent to GCP may contain multiple entries with
+ * the same role. This is valid, though GCP will condense them into one binding per role internally.
+ */
+public class GoogleCloudSyncStep implements Step {
+
+  CloudResourceManagerCow resourceManagerCow;
+
+  private Logger logger = LoggerFactory.getLogger(GoogleCloudSyncStep.class);
+
+  @Autowired
```
ddietterich, line 44: Another case of Step using `@Autowired`. See comment below

ddietterich (changes requested):
I'm concerned about the assigned roles.
Also, I don't think `@Autowired` works in steps so am wondering how this works in test.

src/main/java/bio/terra/workspace/service/workspace/flight/SyncSamGroupsStep.java
zloery: Whoops, you're right you can't use `@Autowired` in steps. The step instances are really being created in the Flight, so those annotations aren't actually doing anything.

src/main/java/bio/terra/workspace/service/workspace/flight/GoogleCloudSyncStep.java
zloery: In general I try to create a specific exception when it would provide some useful context for debugging or logs. This could reasonably be something like `RetryableCrlException`, so I added that for this + other steps.

src/main/java/bio/terra/workspace/service/workspace/flight/GoogleCloudSyncStep.java
zloery: Looking at this again, that was mostly trying to be clever and avoid copying code. I think just building the bindings ends up being clearer and avoiding the potential mismatch from reading `IamRole.values()`.

src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
zloery: After more discussion with Greg we decided that we'll likely be letting users create their own resources in the workspace project for the initial pilot/CLI usage, especially before controlled resources are figured out. Agreed that it's incompatible with enforcing policies, and going to change in the future.

MatthewBemis (approved):
LGTM

ddietterich (approved):
LGTM

PR #174, opened 2021-01-06
Use devsec trivy action & blessed base

zarsky-broad:


dinvlad (approved):

dinvlad:
Might want to rename the workflow to `Tag, Build, Scan and Push Image`

dinvlad:
Also, this makes it such that the workflow will fail if the scan fails - if you think that's gonna be disruptive for your deployments, feel free to move it into a separate trivy.yml workflow

zarsky-broad:
@dinvlad we were failing the run that way previously as well. I'm not really concerned about the naming change, we were scanning previously and had the name the same, I think of scanning as part of the build process.

build.gradle
```diff
@@ -220,7 +220,7 @@ task setGitVersion(type: WriteProperties) {
 
 jib {
 	from {
-		image = "adoptopenjdk:11-jre-hotspot"
+		image = "us.gcr.io/broad-dsp-gcr-public/base/jre:11-alpine"
```
davidangb, line 223: I believe that both of these are hotspot/jre 11, correct? Is there any change in OS, from ??? to alpine?

I could go reverse-engineer all of this, but I believe you have already looked into it and have the info ...

gjuggler: I'd also love to see some comments / add'l context on this change, possibly with comments left for posterity.

Thinking more broadly – do we expect all Terra repos to adopt this as a base image? What is the source repo for that image in case someone wants to dig further into details? (A comment with a github link would be great.)

dinvlad: They're here, for your reference!
https://github.com/broadinstitute/dsp-appsec-blessed-images/tree/main/jre

However, our intention for these images was not to have teams depend on the *very particular* versions of all software, but rather just to provide a few general "flavors" (e.g. Debian, Alpine, Distroless) of base images that can be relied upon as a "turn-key" solution. Otherwise, it would be very hard to satisfy everyone's requirements simultaneously :-)

Some more thoughts on that reasoning here: https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/README.md

If your team requires some very particular versions, I'd say it might be best to use your own base, instead of the "blessed" ones ;) The whole purpose of "blessed" images is to make it general and easy enough to set up Dockerfiles with a "secure" base. But you're always welcome to use another one, if you need!

That being said, I think we can commit to providing AdoptOpenJDK Hotspot flavors for the moment (for Debian and Alpine), seeing that those upstream images are both best-maintained and most widely used across teams (GraalVM is probably the only exception). I can document that in the README for JRE.

zarsky-broad: This is ubuntu -> alpine, but we switched from distroless -> ubuntu not too long ago without issue, so I don't forsee anything weird here. They're both hotspot/jre, yes.

dinvlad: Yep, the only difference *might* be in performance, as Alpine relies on an alternative to Glibc. Might be worth benchmarking them, although may not matter here.

gjuggler (approved):

davidangb (approved):

PR #182, opened 2021-01-11
[PF-328] data reference enumeration test

jaycarlton:
Create, read (two ways), and delete a data reference. Uses the new test-fixture version of the base class, which needs to be merged first. It's at #181.

PR #183, opened 2021-01-11
[PF-329] Add TestRunner client tests covering existing integration tests

zloery:
This change adds new TestRunner client tests covering workspace CRUD endpoints. Combined with the tests in #182, this covers the existing integration test cases.

This also removes calls to `assertHttpOk` in `WorkspaceTestScriptBase`, as these will fail on tests with empty doSetup/doCleanup methods.

Note this does not disable/remove the existing integration tests, as we don't have the TestRunner tests running automatically yet. Once we have these tests running properly, we should remove the integration tests.


workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteDataReference.java
```diff
@@ -0,0 +1,60 @@
...
+public class CreateGetDeleteDataReference extends WorkspaceTestScriptBase {
+  private static final Logger logger = LoggerFactory.getLogger(CreateGetDeleteDataReference.class);
+  private UUID workspaceId;
+
+  @Override
+  public void doSetup(List<TestUserSpecification> testUsers, WorkspaceApi workspaceApi) throws ApiException {
+    workspaceId = UUID.randomUUID();
+    CreateWorkspaceRequestBody requestBody = new CreateWorkspaceRequestBody().id(workspaceId);
+    workspaceApi.createWorkspace(requestBody);
+    WorkspaceManagerServiceUtils.assertHttpOk(workspaceApi, "CREATE workspace");
```
jaycarlton, line 37: I think right now the base class does this for you. It's a bit magical though.

workspace-manager-clienttests/src/main/resources/configs/BasicIntegrationTest.json
```diff
@@ -0,0 +1,31 @@
...
+    {
+      "name": "GetDataReferenceByName",
+      "numberOfUserJourneyThreadsToRun": 1,
+      "userJourneyThreadPoolSize": 1,
+      "expectedTimeForEach": 5,
+      "expectedTimeForEachUnit": "SECONDS"
+    }
+  ],
+  "testUserFiles": ["william.json"]
+}
```
zarsky-broad, line 31:
```suggestion
}

```

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetDataReferenceByName.java
```diff
@@ -0,0 +1,56 @@
...
+import bio.terra.workspace.model.WorkspaceDescription;
+import java.util.List;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scripts.utils.DataReferenceUtils;
+import scripts.utils.WorkspaceManagerServiceUtils;
+import scripts.utils.WorkspaceTestScriptBase;
+
+public class GetDataReferenceByName extends WorkspaceTestScriptBase {
```
jaycarlton, line 25: Hmm. I have something very similar at https://github.com/DataBiosphere/terra-workspace-manager/pull/182/files for PF-328

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetDataReferenceByName.java
```diff
@@ -0,0 +1,56 @@
...
+  private UUID workspaceId;
+  private DataReferenceDescription dataReference;
+
+  @Override
+  public void doSetup(List<TestUserSpecification> testUsers, WorkspaceApi workspaceApi) throws ApiException {
+    workspaceId = UUID.randomUUID();
+    // First, create a workspace.
+    CreateWorkspaceRequestBody requestBody = new CreateWorkspaceRequestBody().id(workspaceId);
+    workspaceApi.createWorkspace(requestBody);
+    WorkspaceManagerServiceUtils.assertHttpOk(workspaceApi, "CREATE workspace");
```
jaycarlton, line 36: We should probably rename this `assertHttpSuccess()`, since OK is only for 200.

davidangb:
fyi @ansingh7115 since you are currently looking at integration tests

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteDataReference.java
zloery: Which part? if it's calling `assertHttpSuccess`, I removed that from the base class as it caused empty setups/cleanups to fail.

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetDataReferenceByName.java
zloery: Agreed. I think we only need one as long as it's in the integration test suite

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteDataReference.java
jaycarlton: Yes, killing that in the base class is the way to go. It also prevented calling get after delete to verify a 404.

workspace-manager-clienttests/src/main/java/scripts/utils/DataReferenceUtils.java
```diff
@@ -0,0 +1,37 @@
...
+    // support this test in other DataRepo environments.
+    // Data Repo makes a reasonable effort to maintain their dev environment, so this should be a
+    // very rare occurrence.
+    DataRepoSnapshot snapshotReference =
+        new DataRepoSnapshot()
+            .snapshot("97b5559a-2f8f-4df3-89ae-5a249173ee0c")
+            .instanceName("terra");
+    // Names need to be unique per reference type within a workspace, use a random name here
+    // to support multiple test threads running at once.
+    String dataReferenceName = "snapshot_" + UUID.randomUUID().toString().replace("-", "_");
```
jaycarlton, line 28: FYI I have a util method for this on my branch.

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceTestScriptBase.java
```diff
@@ -32,7 +32,6 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
       logger.debug("Caught exception creating setup ", apiEx);
       throw(apiEx);
     }
-    WorkspaceManagerServiceUtils.assertHttpOk(workspaceApi, "setup");
```
jaycarlton, line 35: 👏🏻

jaycarlton (approved):

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetDataReferenceByName.java
```diff
@@ -0,0 +1,34 @@
...
+import bio.terra.testrunner.runner.config.TestUserSpecification;
+import bio.terra.workspace.api.WorkspaceApi;
+import bio.terra.workspace.client.ApiException;
+import bio.terra.workspace.model.CreateDataReferenceRequestBody;
+import bio.terra.workspace.model.DataReferenceDescription;
+import java.util.List;
+import scripts.utils.ClientTestUtils;
+import scripts.utils.WorkspaceFixtureTestScriptBase;
+
+public class GetDataReferenceByName extends WorkspaceFixtureTestScriptBase {
```
ungwudik, line 15: Should we remove this class from this PR, given the comment here? https://github.com/DataBiosphere/terra-workspace-manager/pull/183#discussion_r555317348

Separate but related - I'm not quite clear on the intended structure of these test classes. On one hand, it looks like we want to write CRUD-like tests for a resource in one test class (example `CreateGetDeleteWorkspace.java` for Workspaces). On the other hand, `GetDataReferenceByName.java` seems to live by itself here. It looks like that was addressed in [#182](https://github.com/DataBiosphere/terra-workspace-manager/pull/182) with [DataReferenceLifecycle.java](https://github.com/DataBiosphere/terra-workspace-manager/pull/182/files#diff-65909047383aed3785c17cb9d2864086cda6b7ccdd58473140e5dff5497e0490). However, the same [#182](https://github.com/DataBiosphere/terra-workspace-manager/pull/182) also had [EnumerateDataReferences.java](https://github.com/DataBiosphere/terra-workspace-manager/pull/182/files#diff-03c853b3e5fec3083a8555b059e2ceec5c9cf80f0d4bdb399b92a84128e86a19) test class which seems to break out on its own, unlike the other DataReference-related tests.

zloery: Yep, I need to remove that one.

I think we want:
1. tests that cover each of the basic CRUD endpoints we have (e.g. `CreateGetDeleteWorkspace`, `DataReferenceLifecycle`. These tend to get lumped into a single test since they all exercise the same "create -> read -> delete" combination of endpoints.
2. tests that cover individual endpoints, either more exhaustively or because they aren't included in #1. 

In general I prefer the smaller and more focused tests, but I think individually testing each part of a CRUD api ends up with a lot of duplicate tests.

ungwudik: That's fair, IMO.

We should watch out for how this approach plays into other things, though...such as reporting test results. In a scenario where `GET workspace` assertion fails, for example, I wonder how that will propagate into test reports/dashboards for stakeholders to consume. Are we going to see that `doUserJourney` (i.e. the test) failed and then go to the logs to see what assertion is the culprit? This is in contrast to when the CRUD actions are tested separately, thereby making it easy to quickly identify what action (i.e. `GET workspace` in this case) failed.

Hopefully, I'm not overthinking this. Let me know what your thoughts are, either here or Slack/Meet/etc.

ungwudik (approved):

ddietterich (approved):
lgtm

workspace-manager-clienttests/src/main/java/scripts/testscripts/GetDataReferenceByName.java
zloery: Agreed, you'll have to dig into the test results a bit to see which action actually failed. At the same time, given that the tests for the actions are so similar (e.g. testing GET and DELETE require a call to create in setup) the tests will likely either pass or fail as a group - even if we separate them into separate tests, if they regularly fail together we'd still need to dig deeper into results to see causes.

PR #184, opened 2021-01-12
[PF-356] Move memberEmail to request body for grantRole

zloery:
Currently, the `grantRole` endpoint is at `PUT /api/workspaces/v1/{id}/roles/{role}/members/{memberEmail}` and does not take a body. The swagger generated client (using Jersey2) enforces its own rules about HTTP requests and disallows PUT requests with an empty body, which means we can't call this endpoint from the client.

This changes PUT to POST, which may have empty request bodies.

As this is a change to the API, I've bumped the project version to 0.8.0 and will publish a new client library when this change is merged.

This also includes unrelated lint changes to `RetryableCrlException` and `GoogleCloudSyncStep` because I pushed without linting in #172. Sorry!

### DIFF
--- a/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -276,7 +276,7 @@ public class WorkspaceApiController implements WorkspaceApi {
   public ResponseEntity<Void> grantRole(
       @PathVariable("id") UUID id,
       @PathVariable("role") IamRole role,
-      @PathVariable("memberEmail") String memberEmail) {
+      @RequestBody String memberEmail) {
     ControllerValidationUtils.validateEmail(memberEmail);
     samService.grantWorkspaceRole(
         id,
--- a/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -276,7 +276,7 @@ public class WorkspaceApiController implements WorkspaceApi {
   public ResponseEntity<Void> grantRole(
       @PathVariable("id") UUID id,
       @PathVariable("role") IamRole role,
-      @RequestBody String memberEmail) {
+      @PathVariable("memberEmail") String memberEmail) {
     ControllerValidationUtils.validateEmail(memberEmail);
     samService.grantWorkspaceRole(
         id,
--- a/src/main/resources/api/service_openapi.yaml
+++ b/src/main/resources/api/service_openapi.yaml
@@ -714,7 +714,7 @@ components:
         $ref: '#/components/schemas/RoleBinding'
 
     MemberEmail:
-      description: A user or group's email. Used for adding or removing IAM permissions
+      description: A user or group's email.
       type: string
 
   responses:
--- a/src/main/resources/api/service_openapi.yaml
+++ b/src/main/resources/api/service_openapi.yaml
@@ -235,15 +235,20 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
-  /api/workspaces/v1/{id}/roles/{role}/members/{memberEmail}:
+  /api/workspaces/v1/{id}/roles/{role}/members:
     parameters:
     - $ref: '#/components/parameters/Id'
     - $ref: '#/components/parameters/Role'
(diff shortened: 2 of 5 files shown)


zarsky-broad:
I'm hesitant to do this, since it doesn't match the pattern in [sam](https://sam.dsde-dev.broadinstitute.org/#/Group/AddEmailToGroup), let me look at the jersey config...

ddietterich:
My understanding is that the Jersey rules are enforcing the http standard's rules. I'm not keen on building non-standard APIs.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -276,7 +276,7 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
   public ResponseEntity<Void> grantRole(
       @PathVariable("id") UUID id,
       @PathVariable("role") IamRole role,
-      @PathVariable("memberEmail") String memberEmail) {
+      @RequestBody String memberEmail) {
```
ddietterich, line 279: I wonder if the body should be JSON. That would allow upward change, if needed.
I wonder if the role should be in the path or also moved to the JSON.

The asymmetry with using DELETE is annoying. Maybe we make a PUT that does add or delete so we can have a single expression of a change rather than two different API syntaxes.

ddietterich (commented):
We probably need a more general conversation about this one and get some consistency across the components about how we make these endpoints.

Is this blocking anything in the short term?

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
jaycarlton: Why not POST & DELETE?

I agree re: putting role into the body and making it a JSON object.

src/main/resources/api/service_openapi.yaml
```diff
@@ -393,7 +404,7 @@ components:
     MemberEmail:
```
jaycarlton, line 404: Maybe  change to `GrantRoleRequest` or similar, as more fields may be added in future.

src/main/resources/api/service_openapi.yaml
```diff
@@ -702,6 +713,10 @@ components:
       items:
         $ref: '#/components/schemas/RoleBinding'
 
+    MemberEmail:
```
jaycarlton, line 716: It might make sense (later) to have a `ValidEmailAddress` type instead of string, to indicate that you've got a real email address (for tighter typing). Might be on the cute side though.

zloery:
> My understanding is that the Jersey rules are enforcing the http standard's rules. I'm not keen on building non-standard APIs.

At least from cursory googling I can't tell if this is a rule or a grey area that Jersey takes a stance on. The Sam client (non-Jersey) is happy to [send empty PUT requests](https://sam.dsde-dev.broadinstitute.org/#/Resources/addUserToPolicy), and that's the endpoint WM eventually calls. Having to disable a Jersey check might be a strong enough signal that we don't want to do this, but that would mean being inconsistent with the shape from Sam.

> Is this blocking anything in the short term?

It blocks us from writing integration tests on these endpoints (PF-331), but not any functionality until something like the CLI needs to add users to a workspace.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
jaycarlton: There's a good discussion [here](https://stackoverflow.com/a/1233569/12345554). PATCH is another option they mention.

ddietterich:
I think the best way to resolve this is to use POST and the original all-in-the-URL formulation.
In REST, when you create a new thing in a collection you use POST, when you update a thing you use PUT, and when you remove the thing you use DELETE. We are adding a new member to the collection here, so I think POST is correct.

I think it is worth writing this down so we don't revisit in the future.

zloery:
I was under the impression that POST had the same restriction on body as PUT - turns out I was totally wrong, empty POST bodies are allowed by both Swagger UI and the client library (verified by building a local client and running client tests).

ddietterich (approved):
lgtm

jaycarlton (approved):
lgtm

zloery:
Client version `0.8.0-SNAPSHOT` is now published and available

PR #185, opened 2021-01-14
[PF-356] Move memberEmail from url to request body on grantRole endpoint

zloery:
Previously in #184, I changed the `grantRole` endpoint from PUT to POST, leaving all parameters in the URL. This is allowed by both the server and client, but GCP automatically rejects POST requests without a `Content-length` header (which Jersey does not include for empty bodies).

This change moves the `memberEmail` parameter from the URL to the request body for `grantRole`. The `removeRole` endpoint is unchanged, as DELETE requests should not have bodies.

From an API design standpoint this makes the most sense to me: we are adding one member to a role's collection of members. This endpoint is not creating a new role, so I did not want to move role to the request body.

While this does mean our POST and DELETE endpoints are no longer symmetrical, this is the same approach we've taken for creating/deleting workspaces and data references.

In addition to testing this with a locally built client and server, I verified this against my preview environment to ensure there were no complaints from the proxy or GCP ingress.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ sourceCompatibility = JavaVersion.VERSION_11
 
 allprojects {
 	group = "bio.terra"
-	version = "0.8.0-SNAPSHOT"
+	version = "0.9.0-SNAPSHOT"
 	ext {
 		artifactGroup = "${group}.workspace"
 		swaggerOutputDir = "${buildDir}/swagger-code"
--- a/src/main/resources/api/service_openapi.yaml
+++ b/src/main/resources/api/service_openapi.yaml
@@ -235,15 +235,20 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
-  /api/workspaces/v1/{id}/roles/{role}/members/{memberEmail}:
+  /api/workspaces/v1/{id}/roles/{role}/members:
     parameters:
     - $ref: '#/components/parameters/Id'
     - $ref: '#/components/parameters/Role'
-    - $ref: '#/components/parameters/MemberEmail'
     post:
       summary: Grant an IAM role to a user or group.
       operationId: grantRole
       tags: [Workspace]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/GrantRoleRequestBody'
       responses:
         '204':
           description: Role granted successfully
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -15,7 +15,7 @@ sourceCompatibility = JavaVersion.VERSION_11
 
 allprojects {
 	group = "bio.terra"
-	version = "0.8.0-SNAPSHOT"
+	version = "0.9.0-SNAPSHOT"
```
jaycarlton, line 18: Why not 0.8.1 if it's a bugfix?

src/main/resources/api/service_openapi.yaml
```diff
@@ -702,6 +712,14 @@ components:
       items:
         $ref: '#/components/schemas/RoleBinding'
 
+    GrantRoleRequestBody:
+      description: The body for a request to grant a role to a single user.
+      type: object
+      required: [memberEmail]
```
jaycarlton, line 718: thinking out loud: you could make a batch endpoint that takes a list of emails. It would look very similar to this. That can come later.

src/main/resources/api/service_openapi.yaml
```diff
@@ -702,6 +712,14 @@ components:
...
+      properties:
+        memberEmail:
+          type: string
```
jaycarlton, line 721: A side benefit of using the body is that you don't have to escape the email address in the URL.

jaycarlton (approved):
LGTM

build.gradle
zloery: Because we're changing an API in a way that's incompatible between versions. I don't really like putting a change like that in a bugfix version (though iiuc, versioning rules are generally much looser for major version 0)

jaycarlton: Sure, but it's not a breaking change unless it worked before. 😉

src/main/resources/api/service_openapi.yaml
zloery: +1, though I believe the jersey client and swagger-ui both handle that for us. still nice though

src/main/resources/api/service_openapi.yaml
zloery: we actually [talked about writing a batch api](https://docs.google.com/document/d/1mC2AhU3Rw8zvOTYtxdLyfQtcrIHjQjXBgvmMBD-I7As/edit?usp=sharing) that looks pretty similar to this one, though it raised some interesting questions around transactions and how they're supported in Sam

ddietterich (approved):
lgtm

zloery:
Client library version `0.9.0-SNAPSHOT` is published from this and now available.

PR #186, opened 2021-01-15
[PF-356] Modify Jersey2 ApiClient to force content-length on empty POST body

gjuggler:
We're encountering some issues with the Swagger-generated Jersey2 client, where a POST method without any body parameters will result in HTTP requests with a missing Content-Length header. This works fine locally, but for deployed servers, some proxy along the way rejects these requests with a message "POST requests must have a Content-Length header".

Most suggestions on the internet suggest putting an empty string to force Jersey to calculate a content-length ([ref](https://stackoverflow.com/a/39520816)). We can do this by modifying the Swagger codegen template.

Normally, I probably wouldn't go to such lengths just to fix an awkward bug like this. However, we can eventually migrate these overridden templates to `terra-common-lib` and share them across all Terra services, which compounds the value of the fix. Plus, we have other good reasons to want to have more control over out auto-generated API client code (e.g. we may eventually want to create custom client code for easier consumption of async APIs).

Note: the only Terra-specific change in the Swagger template code is within ApiClient.mustace; look for the block marked "Begin Terra change". I've included the api.mustache template as well for completeness.


workspace-manager-client/src/main/resources/swaggercodegen/libraries/jersey2/api.mustache
```diff
@@ -0,0 +1,133 @@
...
+  Copied from the https://github.com/swagger-api/swagger-codegen-generators tagged release 1.0.24:
+  https://github.com/swagger-api/swagger-codegen-generators/blob/v1.0.24/src/main/resources/handlebars/Java/libraries/jersey2/api.mustache
+
+  This file may include Terra-specific modifications to the Swagger-codegen generated API client
+  code. This file (api.mustache) is responsible for generating each typed API class, e.g.
+  bio.terra.*.api.FooApi.
+
+  All Terra changes or additions should be surrounded by comment blocks of the form:
+
+  {{! Begin Terra changes }}
```
jaycarlton, line 11: is the trick to use a partial for the terra snippets applicable here?

ddietterich:
It makes sense to me.

workspace-manager-client/src/main/resources/swaggercodegen/libraries/jersey2/api.mustache
gjuggler: I could go either way, but I'm inclined to leave as-is for simplicity's sake (fewer files and less indirection). It's likely that we'll ultimately want to manage & publish these template overrides from terra-common-lib, so that will provide another chance to assess the ideal structure.

jaycarlton (commented):
@gjuggler I'm having trouble finding the place where you actually made the change. Maybe if you could point that out or (maybe overkill) split into one PR to establish these files here and another to fix the bug.

workspace-manager-client/src/main/resources/swaggercodegen/libraries/jersey2/api.mustache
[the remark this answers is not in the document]
gjuggler: Sorry – the only change is actually in ApiClient.mustache (as noted in the PR description). I updated the comment on this file, api.mustache, to clarify that no changes have been made (yet).

ddietterich (approved):
lgtm
Is this complete?

gjuggler:
> lgtm
> Is this complete?

Yep, I think so. Proof from my machine that this gets included into the swagger-built API client sources:

Screen Shot 2021-01-26 at 6 07 11 PM

I'll file a follow-up ticket tracking the effort to wrap common Swagger templates into terra-common-lib, which I think would be the next logical step. But this at least captures "here's what it takes to override Swagger templates for a single repo" which is a helpful pattern to have an example of.

zloery (approved):

gjuggler:
> I'll file a follow-up ticket tracking the effort to wrap common Swagger templates into terra-common-lib, which I think would be the next logical step. But this at least captures "here's what it takes to override Swagger templates for a single repo" which is a helpful pattern to have an example of.

Filed https://broadworkbench.atlassian.net/browse/PF-396 to track pulling this into terra-common-lib.

PR #187, opened 2021-01-15
[PF-364] Fix error codes from permissions API

zloery:
Currently, calling any permissions API endpoint on a workspace that doesn't exist will return a 500. To be consistent with other endpoints, this should be a 403 or 404, depending on whether the workspace exists. This does allow users to find out whether workspaces exist or not, but WM is not expected to obscure this.

This change introduces an additional call to Sam for each permissions API endpoint. While this is extra overhead, it keeps responses consistent and provides useful information instead of an incorrect 500 error.


src/main/resources/api/service_openapi.yaml
```diff
@@ -247,9 +247,7 @@ paths:
       responses:
         '204':
           description: Role granted successfully
-        '403':
-          $ref: '#/components/responses/PermissionDenied'
-        '404':
+        '401':
           $ref: '#/components/responses/NotFound'
```
jaycarlton, line 251: It seems like 401 should give a different response type than 404. I get why you avoid doing 404 though.

If not found is the predominant situation, then why not return 404 in all cases? If that's more likely to be clear.

jaycarlton (approved):
Thanks. One less 500 to worry about.

src/main/resources/api/service_openapi.yaml
```diff
@@ -247,9 +247,7 @@ paths:
       responses:
         '204':
           description: Role granted successfully
-        '403':
-          $ref: '#/components/responses/PermissionDenied'
-        '404':
+        '401':
```
ddietterich, line 250: 401 is not the right response. UNAUTHORIZED means that your credentials were invalid - not that you do not have permission. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401. It would be misleading to return.

We can confirm with others, but my understanding is that for MC Terra APIs we do not need to hide the difference between NOT FOUND and FORBIDDEN. I know that in secure APIs, the best practice is to say NOT FOUND uniformly to not leak existence info.

src/main/resources/api/service_openapi.yaml
zloery: Yeah, this is one I'm not sure about. We intentionally give the same response to stuff that should be a 401 vs. 404, and we expect to see both cases. It's hard to say one is predominant here.

I picked the `NotFound` object here because its description explicitly calls out "Not found (or unauthorized)". The only difference between the response types here is the description string that gets used in swagger-ui - otherwise, they're both just ErrorReport objects.

ddietterich (changes requested):
We should not use 401 and I'm not sure we need to hide FORBIDDEN

src/main/resources/api/service_openapi.yaml
ddietterich: Sam does not hide the difference, FWIW

zloery: Ah, I was under the impression that we were trying to avoid leaking existence info. Which parts of Sam don't hide the difference? querying "can I do action X on resource Y" returns `false` for things that don't exist, or things where it's actually false

ddietterich: I was looking at:
    GET /api/resources/v1/{resourceTypeName}/{resourceId}/policies
but there are a fair number that return both 403 and 404.

davidangb: I think for CWB, our approach has been that WSM doesn't need to hide any error codes, and we should handle leak-avoidance at a higher level. I'm not sure if that approach holds true for your non-CWB uses ... is WSM still a low-level service in that architecture?

ddietterich: I checked with Sarah and Denis. There is no compliance reason for us to be hiding the state.

I think hiding it is a negative, since it makes it harder for the caller to debug what has gone wrong.

zloery:
After #193, failed calls to `workspaceAuthz` in this change will now throw 403 response errors instead of 401, making this behavior consistent with the rest of WM.

zloery:
Updated to reflect changes from #198. Permissions API endpoints now use either 403 or 404 as appropriate.

davidangb (approved):
looks good, thanks!

ddietterich (approved):
lgtm

PR #188, opened 2021-01-19
[PF-363] Handle calls to create duplicate context better

zloery:
Currently, making a call to create a google context on a workspace which already has one will result in partially executing the stairway flight (including allocating a new GCP project) before failing (and deleting the new GCP project). This is a waste of time and projects, and we can avoid it and provide a more useful error code with a simple check.


ddietterich (approved):
lgtm

davidangb (approved):

PR #191, opened 2021-01-20
[PF-326] Update Workspace Manager environment variable names to work with Helm

tnadler:
Also add some logging.

### DIFF
--- a/src/test/resources/application-test.yml
+++ b/src/test/resources/application-test.yml
@@ -2,7 +2,7 @@ workspace:
   buffer:
     pool-id: workspace_manager_v1
     instance-url: https://buffer.tools.integ.envs.broadinstitute.org
-    buffer-client-credential-file-path: rendered/buffer-client-sa-account.json
+    client-credential-file-path: rendered/buffer-client-sa-account.json
 
   crl:
     testing-mode: true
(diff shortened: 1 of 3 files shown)


src/test/resources/application-test.yml
```diff
@@ -2,7 +2,7 @@ workspace:
   buffer:
     pool-id: workspace_manager_v1
     instance-url: https://buffer.tools.integ.envs.broadinstitute.org
-    buffer-client-credential-file-path: rendered/buffer-client-sa-account.json
+    client-credential-file-path: rendered/buffer-client-sa-account.json
```
jaycarlton, line 5: Is there reflection somewhere that maps this name into a java object?

jaycarlton (approved):

zloery (approved):

src/test/resources/application-test.yml
zloery: This will be read using spring's `ConfigurationProperties` annotation (https://github.com/DataBiosphere/terra-workspace-manager/blob/a9cd0addddadce40fd75b0311dfe93eaecb8bd57/src/main/java/bio/terra/workspace/app/configuration/external/BufferServiceConfiguration.java#L15) which turns the path into a String

PR #193, opened 2021-01-20
[PF-372] Change 401 responses to 403

zloery:
Currently, lots of things in WM return an HTTP 401 response to indicate a user does not have permission for an action. Technically [401 means authentication info is missing or unreadable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) while [403 means permission is denied](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403). 

Note that Sam does not distinguish between "resource not found" and "permission denied" in calls to its `resourcePermission` endpoint. While WM does not explicitly need to hide this difference, it uses that endpoint to check permissions. Because WM cannot distinguish between 403 and 404 responses from Sam, WM will return 403 in this case, with an error message indicating either case is possible.

### DIFF
--- a/src/main/resources/api/service_openapi.yaml
+++ b/src/main/resources/api/service_openapi.yaml
@@ -56,7 +56,7 @@ paths:
           $ref: '#/components/responses/CreatedWorkspaceResponse'
         '400':
           $ref: '#/components/responses/BadRequest'
-        '401':
+        '403':
           $ref: '#/components/responses/PermissionDenied'
         '500':
           $ref: '#/components/responses/NotFound'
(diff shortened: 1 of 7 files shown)


ddietterich (approved):
lgtm

src/main/resources/api/service_openapi.yaml
```diff
@@ -90,8 +88,8 @@ paths:
           description: Success
         '400':
           $ref: '#/components/responses/BadRequest'
-        '401':
-          $ref: '#/components/responses/NotFound'
+        '403':
```
jaycarlton, line 91: FWIW If not found is the more common case, I'd probably prefer 404 to 403 where there's ambiguity.

jaycarlton (approved):

PR #194, opened 2021-01-20
no ticket - put new tests in perf and integration suites

jaycarlton:
Put all the new tests into both suites for perf and integration. Perf gets 10 threads; integration gets 1 to start.


ddietterich (approved):
lgtm

zloery (approved):
LGTM, thanks for organizing these!

PR #195, opened 2021-01-20
PF-390 - Replace bean lookups with autowired class and accessors

ddietterich:
Small change that cleans up flights.

Currently, flights do dynamic bean lookups and then cast into the right class.

This change creates a class that autowires all of the beans needed in flights. The class instance is used as the application context in Stairway. Flights use accessors of the class to get the beans they need. So no more casting and no more dynamic lookups.


src/main/java/bio/terra/workspace/common/utils/FlightApplicationContext.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * The purpose of FlightApplicationContext is to provide a clean interface for flights to get access
+ * to singleton Spring components. This avoids the use of dynamic bean lookups in flights and
+ * casting the lookup result. Instead, flights make calls to accessors in this class. Spring will
+ * wire up the underlying methods once at startup avoiding the bean lookup. The objects will be
+ * properly types without casting.
+ */
+@Component
+public class FlightApplicationContext {
```
jaycarlton, line 24: I don't know if this is a real `ApplicationContext` since it's not implementing that [interface](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/ApplicationContext.html). This is more of a bean bag.

src/main/java/bio/terra/workspace/service/datareference/flight/CreateDataReferenceFlight.java
```diff
@@ -2,20 +2,19 @@
...
 
 public class CreateDataReferenceFlight extends Flight {
 
   public CreateDataReferenceFlight(FlightMap inputParameters, Object applicationContext) {
     super(inputParameters, applicationContext);
 
-    ApplicationContext appContext = (ApplicationContext) applicationContext;
-    DataReferenceDao dataReferenceDao = (DataReferenceDao) appContext.getBean("dataReferenceDao");
-    ObjectMapper objectMapper = (ObjectMapper) appContext.getBean("objectMapper");
+    FlightApplicationContext appContext =
```
jaycarlton, line 12: So is this constructor constrained somehow such that you can't simply wire in `dataReferenceDao` and `objectMapper`? There's also field autowiring, but I've read that's frowned upon.

src/main/java/bio/terra/workspace/common/utils/FlightApplicationContext.java
```diff
@@ -0,0 +1,103 @@
...
+  private final DataReferenceDao dataReferenceDao;
+  private final GoogleWorkspaceConfiguration googleWorkspaceConfiguration;
+  private final ObjectMapper objectMapper;
+  private final SamService samService;
+  private final ServiceUsageCow serviceUsage;
+  private final TransactionTemplate transactionTemplate;
+  private final WorkspaceDao workspaceDao;
+
+  @Autowired
+  public FlightApplicationContext(
```
jaycarlton, line 37: I'm concerned this class could get arbitrarily large, and it's building beans that may never be needed at runtime depending on the actual dependency graph.

src/main/java/bio/terra/workspace/common/utils/FlightApplicationContext.java
```diff
@@ -0,0 +1,103 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.support.TransactionTemplate;
+
+/**
+ * The purpose of FlightApplicationContext is to provide a clean interface for flights to get access
+ * to singleton Spring components. This avoids the use of dynamic bean lookups in flights and
+ * casting the lookup result. Instead, flights make calls to accessors in this class. Spring will
+ * wire up the underlying methods once at startup avoiding the bean lookup. The objects will be
+ * properly types without casting.
```
jaycarlton, line 21: *proper

jaycarlton (commented):
I have a funny feeling that you're reimplementing part of Spring, but I don't know which bit. It's still an improvement, though.

src/main/java/bio/terra/workspace/common/utils/FlightApplicationContext.java
ddietterich: True. Is `FlightBeanBag` too silly?

src/main/java/bio/terra/workspace/service/datareference/flight/CreateDataReferenceFlight.java
ddietterich: Spring only autowires classes that are @Component or similar - they are singletons.
Things that  you `new` cannot use autowire.

src/main/java/bio/terra/workspace/common/utils/FlightApplicationContext.java
ddietterich: The only beans I put in there are ones referenced in flights, so they are needed at runtime.

src/main/java/bio/terra/workspace/service/datareference/flight/CreateDataReferenceFlight.java
jaycarlton: oh right, flights aren't singletons.

jaycarlton (approved):

src/main/java/bio/terra/workspace/common/utils/FlightApplicationContext.java
jaycarlton: As of right now  yes, but you'll have to remember to remove the ones that aren't needed anymore.

zloery (commented):
Generally LGTM, though it looks like this breaks unit tests: they now need to load the `ClientConfig` for using CRL with Janitor, which tries to read the rendered secret janitor-client-sa-account.json on creation, but we don't render secrets for unit tests.

One way to fix this is turning off Janitor for unit tests, which is controlled by the `workspace.crl.testing-mode` config. Given that unit tests don't create real cloud resources, this wouldn't be a change to behavior

src/main/java/bio/terra/workspace/common/utils/FlightBeanBag.java
```diff
@@ -0,0 +1,87 @@
...
+import bio.terra.workspace.service.buffer.BufferService;
+import bio.terra.workspace.service.crl.CrlService;
+import bio.terra.workspace.service.iam.SamService;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.support.TransactionTemplate;
+
+/**
+ * The purpose of FlightApplicationContext is to provide a clean interface for flights to get access
```
zloery, line 15: nit: renamed to FlightBeanBag

zloery (approved):
LGTM

PR #198, opened 2021-01-22
AS-592: return 404 from workspace-specific APIs if the workspace doesn't exist [risk: low]

davidangb:
This would eliminate the need for broadinstitute/rawls#1347.

In this PR, we change workspace- and data reference-related APIs, which all operate in the context of a single workspace, to always return a 404 if the workspace doesn't exist. It still returns a 403 if the workspace exists but you don't have permission; and returns an appropriate code (generally 200/204) if you do have permission.

In the case of 403,  it changes the error message from "User %s is not authorized to %s workspace %s or it does not exist" to "User %s is not authorized to %s workspace %s". 

For those who haven't followed all the conversation, we realize the 403/404 differentiation leaks existence of a workspace to those who don't have permission to see it. Because Workspace Manager is a low-level service, this is ok; we obscure  error messages at higher layers of the stack.

Do I need to duplicate the new tests using testrunner syntaxes?

### DIFF
--- a/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
+++ b/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
@@ -124,6 +124,17 @@ public class WorkspaceDao {
     return WorkspaceStage.valueOf(jdbcTemplate.queryForObject(sql, params, String.class));
   }
 
+  // TODO: do all three variants of assertMcWorkspace belong in WorkspaceDAO?
+  private void assertMcWorkspace(WorkspaceStage stage, UUID workspaceId, String action) {
+    if (!WorkspaceStage.MC_WORKSPACE.equals(stage)) {
+      throw new StageDisabledException(workspaceId, stage, action);
+    }
+  }
+
+  public void assertMcWorkspace(Workspace workspace, String action) {
+    assertMcWorkspace(workspace.workspaceStage(), workspace.workspaceId(), action);
+  }
+
   /**
    * Check that the given workspace has stage MC_WORKSPACE, and throw an exception otherwise.
    *
@@ -132,9 +143,7 @@ public class WorkspaceDao {
    */
   public void assertMcWorkspace(UUID workspaceId, String action) {
     WorkspaceStage stage = getWorkspaceStage(workspaceId);
-    if (!WorkspaceStage.MC_WORKSPACE.equals(stage)) {
-      throw new StageDisabledException(workspaceId, stage, action);
-    }
+    assertMcWorkspace(stage, workspaceId, action);
   }
 
   /** Retrieves the cloud context of the workspace. */
--- a/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
+++ b/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
@@ -124,26 +124,32 @@ public class WorkspaceDao {
     return WorkspaceStage.valueOf(jdbcTemplate.queryForObject(sql, params, String.class));
   }
 
-  // TODO: do all three variants of assertMcWorkspace belong in WorkspaceDAO?
-  private void assertMcWorkspace(WorkspaceStage stage, UUID workspaceId, String action) {
+  // underlying implementation for the two public assertMcWorkspace() methods
+  private void assertMcWorkspace(WorkspaceStage stage, UUID workspaceId, String actionMessage) {
     if (!WorkspaceStage.MC_WORKSPACE.equals(stage)) {
-      throw new StageDisabledException(workspaceId, stage, action);
+      throw new StageDisabledException(workspaceId, stage, actionMessage);
     }
-  }
-
-  public void assertMcWorkspace(Workspace workspace, String action) {
-    assertMcWorkspace(workspace.workspaceStage(), workspace.workspaceId(), action);
   }
 
   /**
    * Check that the given workspace has stage MC_WORKSPACE, and throw an exception otherwise.
    *
-   * @param workspaceId ID of the workspace to check
-   * @param action The action being performed, used in the exception message if needed
+   * @param workspace the workspace to check
+   * @param actionMessage The action being performed, used only in the exception message if needed
    */
-  public void assertMcWorkspace(UUID workspaceId, String action) {
+  public void assertMcWorkspace(Workspace workspace, String actionMessage) {
+    assertMcWorkspace(workspace.workspaceStage(), workspace.workspaceId(), actionMessage);
+  }
+
+  /**
+   * Check that the given workspace id has stage MC_WORKSPACE, and throw an exception otherwise.
+   *
+   * @param workspaceId ID of the workspace to check
+   * @param actionMessage The action being performed, used only in the exception message if needed
+   */
+  public void assertMcWorkspace(UUID workspaceId, String actionMessage) {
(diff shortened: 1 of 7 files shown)


jaycarlton:
Setting up your tests in TestRunner would be appreciated. It could be a separate PR though.

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -124,6 +124,17 @@ public WorkspaceStage getWorkspaceStage(UUID workspaceId) {
     return WorkspaceStage.valueOf(jdbcTemplate.queryForObject(sql, params, String.class));
   }
 
+  // TODO: do all three variants of assertMcWorkspace belong in WorkspaceDAO?
+  private void assertMcWorkspace(WorkspaceStage stage, UUID workspaceId, String action) {
```
jaycarlton, line 128: nit: It would be nice to make the actions an enum with a string description constant, but that can be another PR.

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -53,7 +53,8 @@ public DataReferenceService(
   public DataReference getDataReference(
       UUID workspaceId, UUID referenceId, AuthenticatedUserRequest userReq) {
 
-    samService.workspaceAuthz(userReq, workspaceId, SamConstants.SAM_WORKSPACE_READ_ACTION);
+    workspaceService.authorizedGetWorkspace(
```
jaycarlton, line 56: For assertions that throw, I like to prefix the function with `validate`, though I don't know if there's a prevailing convention here.

More importantly, this method checks more than workspace read, so maybe `validateWorkspaceOperation()` would be better.

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -78,7 +80,7 @@ public DataReference getDataReferenceByName(
   public DataReference createDataReference(
```
jaycarlton, line 80: For public methods, maybe add to the javadoc that we validate workspace visibility and modifiability first, then validate the data reference operation, etc.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -68,18 +68,42 @@ public UUID createWorkspace(WorkspaceRequest workspaceRequest, AuthenticatedUser
...
+   * @param userReq the user's authenticated request
+   * @param workspaceId id of the workspace in question
+   * @param action the action to authorize against the workspace
+   * @return the workspace, if it exists and the user is permitted to perform the specified action.
+   */
+  @Traced
+  public Workspace authorizedGetWorkspace(
+      AuthenticatedUserRequest userReq, UUID workspaceId, String action) {
+    Workspace workspace = workspaceDao.getWorkspace(workspaceId);
+    samService.workspaceAuthzOnly(userReq, workspaceId, action);
```
jaycarlton, line 93: why is the auth check after the `getWorkspace()` call here?

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -97,20 +121,20 @@ public void deleteWorkspace(UUID id, AuthenticatedUserRequest userReq) {
...
-    workspaceDao.assertMcWorkspace(workspaceId, "createGoogleContext");
+    Workspace workspace =
+        authorizedGetWorkspace(userReq, workspaceId, SamConstants.SAM_WORKSPACE_WRITE_ACTION);
+    workspaceDao.assertMcWorkspace(workspace, "createGoogleContext");
     if (!workspaceDao.getCloudContext(workspaceId).googleProjectId().isEmpty()) {
       throw new DuplicateGoogleContextException(workspaceId);
     }
-    Optional<SpendProfileId> spendProfileId =
-        workspaceDao.getWorkspace(workspaceId).spendProfileId();
+    Optional<SpendProfileId> spendProfileId = workspace.spendProfileId();
```
jaycarlton, line 137: nit: `SpendProfileId spendProfileId = workspace.spendProfileId().orElseThrow(() -> new MissingSpend...())`

src/main/resources/api/service_openapi.yaml
```diff
@@ -114,6 +118,8 @@ paths:
           $ref: '#/components/responses/BadRequest'
         '403':
           $ref: '#/components/responses/PermissionDenied'
+        '404':
```
jaycarlton, line 121: aside: I wish OpenAPI gave us a way to avoid re-declaring the map between error and type everywhere.

jaycarlton (approved):

zloery (approved):
LGTM
You shouldn't need to do anything for TestRunner - all the tests you touch here are unit and connected tests, and TestRunner is only for client library tests.

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
davidangb: I hate naming! I want to also indicate in the name that this method retrieves and returns a `Workspace` object, after validating permissions. I'll try to come up with something better.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
davidangb: this is the core of the entire PR. As discussed in Slack, consensus is that we always want to return 404 if the workspace doesn't exist, regardless of what Sam returns from the permission check. Consider the use case that triggered this PR: deleting a non-existent workspace. In that use case, Sam returns a 403 since it can find no permission for the user. However, we want to return 404. I implemented this logic simply by  putting `getWorkspace()` first, letting it throw 404 if the workspace doesn't exist,  then performing the authz check which  will throw 403  or pass.

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
davidangb: instead of making this an enum, I have clarified its usage in the javadoc, and renamed the argument to reflect its true usage. Does that work?

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
davidangb: renamed to `validateWorkspaceAndAction`, how's that? The codebase uses "action" consistently instead of "operation" for the thing in question here.

ungwudik (approved):
Looks good to me.

zloery (approved):
new changes lgtm

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
jaycarlton: sounds good

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
jaycarlton: Please add a comment to that effect.

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
jaycarlton: sure. we could change later if we find we want a pie chart of actions or something.

PR #199, opened 2021-01-25
PF-394 Extract CRL beans into a service

ddietterich:
Split testmode into use-crl and use-janitor
Move those settings into connected and integration resource files
Add a unit-test tag and resource file. It sets use-crl to false.
Bump the Stairway version to pick up the new logging stuff


src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -0,0 +1,116 @@
...
+import java.security.GeneralSecurityException;
+import java.time.Duration;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CrlService {
+  /** The client name required by CRL. */
+  public static final String CLIENT_NAME = "workspace";
+  /** How long to keep the resource before Janitor do the cleanup. */
```
zloery, line 25: nit: do -> does

src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -74,6 +74,7 @@ public JobService(
         Stairway.newBuilder()
             .applicationContext(applicationContext)
             .exceptionSerializer(serializer)
+            .keepFlightLog(true)
```
zloery, line 77: to confirm: this means stairway will no longer delete flights from it's DB once they're finished?

src/main/java/bio/terra/workspace/service/job/StairwayExceptionSerializer.java
```diff
@@ -11,10 +11,13 @@
 import java.lang.reflect.InvocationTargetException;
 import java.util.List;
 import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
 
 public class StairwayExceptionSerializer implements ExceptionSerializer {
-  private ObjectMapper objectMapper;
+  private static final Logger logger = LoggerFactory.getLogger(StairwayExceptionSerializer.class);
```
zloery, line 19: Do we need this logger if we aren't logging anything yet?

build.gradle
```diff
@@ -67,7 +67,7 @@ repositories {
 dependencies {
 	// Terra deps
 	implementation group: "bio.terra", name: "datarepo-client", version: "1.0.1-SNAPSHOT"
-	implementation group: "bio.terra", name: "stairway", version: "0.0.25-SNAPSHOT"
+	implementation group: "bio.terra", name: "stairway", version: "0.0.35-SNAPSHOT"
```
zloery, line 70: worth double-checking since this is a 10 release jump: are there any major/breaking changes we should know about, or has stairway been pretty stable?

zloery (approved):
LGTM

src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -0,0 +1,116 @@
...
+  private final CloudBillingClientCow crlBillingClientCow;
+  private final ServiceUsageCow crlServiceUsageCow;
+
+  @Autowired
+  public CrlService(CrlConfiguration crlConfig) {
+    this.crlConfig = crlConfig;
+
+    ClientConfig clientConfig;
+    if (crlConfig.useCrl()) {
+      var creds = buildCredentials();
```
wchamber, line 40: Do we have guidelines on when to use `var`? My personal preference, fwiw, would be to use it when the type is obvious entirely e.g. `var x = new Foo();` or long, uninteresting, and somewhat obvious. I don't think the return value of a member function is either, and there are many possible "credentials" types in general.

src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * @return CRL {@link CloudResourceManagerCow} which wrappers Google Cloud Resource Manager API
+   */
+  public CloudResourceManagerCow cloudResourceManagerCow() {
+    assertCrlInUse();
+    return crlResourceManagerCow;
+  }
+
+  /** The CRL {@link CloudBillingClientCow} which wrappers Google Billing API. */
```
wchamber, line 66: nit: use the same comment style for all of these Cows?

src/main/java/bio/terra/workspace/service/crl/exception/CrlInternalException.java
```diff
@@ -0,0 +1,18 @@
+package bio.terra.workspace.service.crl.exception;
+
+import bio.terra.workspace.common.exception.InternalServerErrorException;
+
+public class CrlInternalException extends InternalServerErrorException {
```
wchamber, line 5: nit: javadoc comment on what this exception is for

src/main/java/bio/terra/workspace/service/crl/exception/CrlSecurityException.java
```diff
@@ -0,0 +1,18 @@
+package bio.terra.workspace.service.crl.exception;
+
+import bio.terra.workspace.common.exception.ForbiddenException;
+
+public class CrlSecurityException extends ForbiddenException {
```
wchamber, line 5: nit: ditto comment

src/test/resources/application-unit-test.yml
```diff
@@ -0,0 +1,4 @@
+workspace:
+  crl:
+    use-crl: false
+    use-janitor: false
```
wchamber, line 4: worth a quick comment? `// Unit tests should not be communicating with clouds, so they should not use CRL or Janitor.`

wchamber (approved):

src/test/resources/application-unit-test.yml
```diff
@@ -0,0 +1,4 @@
+workspace:
```
jaycarlton, line 1: can you just leave this file out altogether if you want all defaults?

jaycarlton (approved):

src/main/java/bio/terra/workspace/service/job/JobService.java
ddietterich: Stairway does not delete flights at all right now, but it does delete the flight logs. This stops it from deleting the flight logs. The plan is to have a retention policy set here (like 90d) and delete after that.

src/main/java/bio/terra/workspace/service/job/StairwayExceptionSerializer.java
ddietterich: No, I did some logging for debugging. Removed the logging, but not the logger. I'll remove it.

build.gradle
ddietterich: The numbers bump pretty fast, since every merge causes a version bump. This is all upward compatible: addition of debuginfo, retry logging, and the hook enhancements.

src/main/java/bio/terra/workspace/service/crl/CrlService.java
ddietterich: We don't, but point taken. I agree if it is not obvious what the type is without looking elsewhere, then you shouldn't use var. Will fix.

src/main/java/bio/terra/workspace/service/crl/CrlService.java
ddietterich: I inherited those comments from the existing implementation. Happy to clean them up!

src/main/java/bio/terra/workspace/app/configuration/external/CrlConfiguration.java
```diff
@@ -44,72 +33,43 @@
   /** pubsub topic id to publish track resource to Janitor */
   private String janitorTrackResourceTopicId;
 
-  public void setTestingMode(boolean testingMode) {
-    this.testingMode = testingMode;
+  public boolean useCrl() {
```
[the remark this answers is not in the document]
ddietterich: Funny - I removed the get because I thought it read oddly. I put it back.

src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -0,0 +1,116 @@
...
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.time.Duration;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CrlService {
+  /** The client name required by CRL. */
+  public static final String CLIENT_NAME = "workspace";
```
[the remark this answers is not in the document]
ddietterich: This constant is also used as the name of a resource in Sam. I don't think we should share the value for two different purposes. This used to be in the CrlConfiguration.java, but I moved it here, since it was the only place it was used.

In fact, neither of the statics are used outside of this module. I will make them private.

src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -0,0 +1,116 @@
...
+      crlResourceManagerCow = null;
+      crlBillingClientCow = null;
+      crlServiceUsageCow = null;
+    }
+  }
+
+  /**
+   * @return CRL {@link CloudResourceManagerCow} which wrappers Google Cloud Resource Manager API
+   */
+  public CloudResourceManagerCow cloudResourceManagerCow() {
```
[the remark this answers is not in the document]
ddietterich: I debated that. These were the names of the beans before I moved them.
I agree with you. I'll prefix them properly.

src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -0,0 +1,116 @@
...
+  private ClientConfig buildClientConfig() {
+    var builder = ClientConfig.Builder.newBuilder().setClient(CLIENT_NAME);
+    if (crlConfig.useJanitor()) {
+      builder.setCleanupConfig(
+          CleanupConfig.builder()
+              .setCleanupId(CLIENT_NAME + "-test")
+              .setTimeToLive(TEST_RESOURCE_TIME_TO_LIVE)
+              .setJanitorProjectId(crlConfig.getJanitorTrackResourceProjectId())
+              .setJanitorTopicName(crlConfig.getJanitorTrackResourceTopicId())
+              .setCredentials(
```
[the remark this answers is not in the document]
ddietterich: We are working with two sets of credentials: the ones that the WM is running as (getApplicationDefault()) and the special credentials we use with janitor to tell it we are allowed to use it. In both cases, failure to make the credentials throws.

I think "OrDie" is a bit dramatic. We are not exiting - just throwing an exception.

And as I look at it, I will remove the naked RuntimeException throw and replace it with the new security exception I added.

src/test/resources/application-unit-test.yml
ddietterich: Not sure what you mean. The default configuration - which is the production config - is in CrlConfiguration.java and says use CRL and do not use Janitor.

This file is included by the Spring framework when the active profile is "unit-test" and it sets the config properly for unit tests. The BaseUnitTest.java file sets the active profile.

jaycarlton: sounds good

PR #200, opened 2021-01-25
PF-341 Temporarily add 'roles/notebooks.admin' for workspace owners and writers

wchamber:
Add a notebooks admin role for workspace owners and writers for the Enterprise alpha. When we introduce controlled resources for notebooks, we will revise the permissions. This initial role allows users to demo notebooks.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
+++ b/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
@@ -13,12 +13,16 @@ public class CloudSyncRoleMapping {
                   "roles/viewer",
                   "roles/bigquery.dataEditor",
                   "roles/lifesciences.editor",
+                  // TODO(wchambers): Revise notebooks permissions when there are controlled resources for notebooks.
+                  "roles/notebooks.admin",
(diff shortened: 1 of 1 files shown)


zloery (approved):
LGTM as a pre-controlled-resources solution. That role grants a pretty wide variety of `compute.x` permissions that we probably won't want in the future.

Do readers need any permission, or are we only concerned with writers/owners for now?

wchamber:
> LGTM as a pre-controlled-resources solution. That role grants a pretty wide variety of `compute.x` permissions that we probably won't want in the future.
> 
> Do readers need any permission, or are we only concerned with writers/owners for now?

All of the `compute.x` permissions are read permissions. the only writes are `notebooks.x`. This is still too broad, but yes, we won't want it in the future.

It looks to me that `roles/viewers` is sufficient for what we'd want viewers to have, i.e. in includes things like `notebooks.instances.get`. Obviously, if we remove that, we'd have to add more granular permissions.

wchamber:
(I'm assuming that the test runner test failures are expected and I'm not going to examine them before merging this PR)

ddietterich (approved):
lgtm

PR #201, opened 2021-01-26
PF-359 - new path to service account

jaycarlton:
Use the new `wsmtest` service account when rendering the config.


ddietterich (approved):
No idea if that is the right path, but it looks like a path all right!

render_config.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 VAULT_TOKEN=${1:-$(cat "$HOME"/.vault-token)}
-WM_APP_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/terra/kernel/dev/dev/workspace/app-sa
```
zloery, line 4: This is the service account that gets used for running connected tests. I think you'll also need to update the `google.folder-id` property in the connected test configuration to point to the wsmtest projects folder.

workspace-manager-clienttests/render-config.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 VAULT_TOKEN=${1:-$(cat "$HOME"/.vault-token)}
-WM_APP_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/terra/kernel/dev/dev/workspace/app-sa
+WM_APP_SERVICE_ACCOUNT_VAULT_PATH=secret/dsde/terra/kernel/integration/wsmtest/workspace/app-sa
```
zloery, line 4: Do the client tests work with this change? (The GHA is broken, but you should still be able to run them manually).
The SA rendered from this path gets read in `/workspace-manager-clienttests/src/main/resources/serviceaccounts/dev-workspace-app-sa.json`, though I don't know if anything actually uses that credential

render_config.sh
jaycarlton: Was I supposed to use a new folder ID for that? I copied it from somewhere I think.

workspace-manager-clienttests/render-config.sh
jaycarlton: They were broken earlier today. Still not working for me.

render_config.sh
zloery: Terraform generates one, and it should be visible in the atlantis output or from the gcp console. I just double checked, it looks like the wrong one is in helmfile (sorry about that, I completely missed it in review).

Connected tests have a [hardcoded value](https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/test/resources/application-connected-test.yml#L3) instead of reading from helm that you'll need to update.

jaycarlton: Where do I need to update that?

workspace-manager-clienttests/render-config.sh
jaycarlton: (I wasn't on the VPN and saw the discussion of another issue giving 403s, so thought it was the same issue).

render_config.sh
zloery: https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/test/resources/application-connected-test.yml#L3, you should point it to the wsmtest folder ID instead (if that's where we want connected tests to create projects)

src/test/resources/application-connected-test.yml
```diff
@@ -1,6 +1,6 @@
 workspace:
-  # workspace-dev folder id to match dev service account used in testing.
-  google.folder-id: 7275817190
+  # workspace-wsmtest folder id to match dev service account used in testing.
```
zloery, line 2: nit: wsmtest service account

zloery (approved):

PR #208, opened 2021-02-01
PF-385: remove credentialId from CreateDataReferenceRequestBody

jaycarlton:
removing un-designed credential argument and code to ensure it's not  present.


ddietterich (approved):
lgtm

zloery (approved):
This change should technically also involve publishing a new client library version, though no clients use this field so it's not particularly important. It's also present in #212, though I'm fine breaking it out here.

PR #212, opened 2021-02-01
[PF-308] Add GCS bucket and bigquery data references

zloery:
pair programmed with @jaycarlton 

This change adds support for GCS bucket and BigQuery dataset data references in addition to the existing Data Repo snapshot references. It also introduces a new pattern for reference endpoints broken out by reference type. Cleanup of the old endpoints after migration from the old API to the new API is tracked in PF-404.

To preserve backwards compatibility, the existing endpoints and interface objects are left unchanged. This means the existing GET and enumerate endpoints will not work with bucket and dataset references, as that would require changing the response object.

This includes basic reference validation for the new types, similar to what we do for TDR snapshots. In this case we make a GET call for the specified bucket or dataset on behalf of the user, which requires that the user have appropriate permission on the object. This does not do any further validation, such as examining bucket contents or BQ tables.

This change is **not** feature gated behind the MC_WORKSPACE. Rawls will have access to these new reference types.

This requires publishing a new client library version, which I'll do after approvals but before merging.


build.gradle
```diff
@@ -124,6 +127,20 @@ if (hasProperty("buildScan")) {
...
+	cloudResourceSchema { transitive = false }
+}
+dependencies {
+	cloudResourceSchema group: 'bio.terra.cloud-resource-lib', name: 'cloud-resource-schema'
+}
+task unzipCloudResourceSchema(type: Copy) {
+	from zipTree(configurations.cloudResourceSchema.singleFile).matching {
+		include 'cloud_resources_uid.yaml'
+	}
+	into "${buildDir}/crlSchema"
```
jaycarlton, line 141: The thing with this is that now the yaml file depends on a build step and a new file in the local filesystem several directories up. It's not clear  how all our openAPI tools treat this, and it appears Swagger UI doesn't handle it well.

src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -55,26 +61,41 @@ public CrlService(CrlConfiguration crlConfig) {
...
+    return new BigQueryCow(
+        clientConfig, BigQueryOptions.newBuilder().setCredentials(creds).build());
+  }
+
+  /** @return CRL {@link StorageCow} which wraps Google Cloud Storage API */
+  public StorageCow createStorageCow(AuthenticatedUserRequest userReq) {
+    assertCrlInUse();
+    AccessToken accessToken = new AccessToken(userReq.getRequiredToken(), null);
+    GoogleCredentials creds = GoogleCredentials.create(accessToken);
+    return new StorageCow(clientConfig, StorageOptions.newBuilder().setCredentials(creds).build());
```
jaycarlton, line 96: we could de-dupe the three lines above with a helper method

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -87,6 +87,16 @@ public DataReference createDataReference(
...
     workspaceService.validateWorkspaceAndAction(
         userReq, referenceRequest.workspaceId(), SamConstants.SAM_WORKSPACE_WRITE_ACTION);
+    try {
+      dataReferenceDao.getDataReferenceByName(
+          referenceRequest.workspaceId(),
+          referenceRequest.referenceType(),
+          referenceRequest.name());
+      throw new DuplicateDataReferenceException(
+          "A reference with the specified name and type already exists in this workspace.");
+    } catch (DataReferenceNotFoundException e) {
```
jaycarlton, line 97: rename `e` to `expected` to help document this.

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -87,6 +87,16 @@ public DataReference createDataReference(
 
     workspaceService.validateWorkspaceAndAction(
         userReq, referenceRequest.workspaceId(), SamConstants.SAM_WORKSPACE_WRITE_ACTION);
+    try {
+      dataReferenceDao.getDataReferenceByName(
+          referenceRequest.workspaceId(),
+          referenceRequest.referenceType(),
+          referenceRequest.name());
+      throw new DuplicateDataReferenceException(
```
jaycarlton, line 95: it would be good to include the name and type and workspace iD in the error message.

src/main/java/bio/terra/workspace/service/datareference/model/BigQueryDatasetReference.java
```diff
@@ -0,0 +1,49 @@
...
+ * using a constant string instead of the actual class name, changing the name of this class will
+ * not break backwards compatibility with existing serialized objects. This string does not need to
+ * match the class name - it only matches for clarity.
+ */
+@AutoValue
+@JsonTypeName("BigQueryDatasetReference")
+public abstract class BigQueryDatasetReference implements ReferenceObject {
+
+  @JsonCreator
+  public static BigQueryDatasetReference create(
```
jaycarlton, line 24: Can we make builders for these (in a separate PR)? I think you can still call the autovalue constructor in `build()`. Ideally they'd give  you that for free like Immutables does.

src/main/java/bio/terra/workspace/service/datareference/model/BigQueryDatasetReference.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @JsonProperty("datasetName")
+  public abstract String datasetName();
+
+  @Override
+  public String toJson() {
+    try {
+      return objectMapper.writeValueAsString(this);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException("Error serializing BigQueryReference", e);
```
jaycarlton, line 41: It would be good to have a more specific exception class

src/main/java/bio/terra/workspace/service/datareference/model/BigQueryDatasetReference.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @JsonCreator
+  public static BigQueryDatasetReference create(
+      @JsonProperty("projectId") String projectId,
+      @JsonProperty("datasetName") String datasetName) {
+    return new AutoValue_BigQueryDatasetReference(projectId, datasetName);
+  }
+
+  @JsonProperty("projectId")
+  public abstract String projectId();
```
jaycarlton, line 31: Some of our getters start with `get` and others don't it seems.

src/main/java/bio/terra/workspace/service/datareference/model/DataReference.java
```diff
@@ -32,13 +33,35 @@
...
+            .workspaceId(workspaceId())
+            .referenceType(referenceType().toApiModel())
+            .referenceInfo(new DataReferenceInfo())
+            .cloningInstructions(cloningInstructions().toApiModel());
+    switch (referenceType()) {
+      case DATA_REPO_SNAPSHOT:
+        // TODO: we populate both the deprecated reference field and the new referenceInfo fields
+        // until all clients have migrated to use the new field.
+        reference.reference(((SnapshotReference) referenceObject()).toApiModel());
+        reference
```
jaycarlton, line 49: this can all be one chain I think.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -68,4 +84,42 @@ private void validateSnapshotReference(SnapshotReference ref, AuthenticatedUserR
...
+   */
+  private void validateGoogleBucket(GoogleBucketReference ref, AuthenticatedUserRequest userReq) {
+    try {
+      // StorageCow.get() returns null if the bucket does not exist or a user does not have access,
+      // which fails validation.
+      Optional.ofNullable(crlService.createStorageCow(userReq).get(ref.bucketName()))
+          .orElseThrow(
+              () ->
+                  new InvalidDataReferenceException(
+                      "Could not access specified GCS bucket. Ensure the name is correct and that you have access."));
```
jaycarlton, line 100: include the bucket name in these messages.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -68,4 +84,42 @@ private void validateSnapshotReference(SnapshotReference ref, AuthenticatedUserR
...
+      BigQueryDatasetReference ref, AuthenticatedUserRequest userReq) {
+    try {
+      DatasetId datasetId = DatasetId.of(ref.projectId(), ref.datasetName());
+      // BigQueryCow.get() returns null if the bucket does not exist or a user does not have access,
+      // which fails validation.
+      Optional.ofNullable(crlService.createBigQueryCow(userReq).getDataset(datasetId))
+          .orElseThrow(
+              () ->
+                  new InvalidDataReferenceException(
+                      "Could not access specified BigQuery dataset. Ensure the name and GCP project are correct and that you have access."));
```
jaycarlton, line 120: include dataset name and project ID

src/test/java/bio/terra/workspace/common/utils/ControllerValidationUtilsTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+  // TODO(PF-404): we're supporting the existing CreateDataReferenceRequest.reference field to
+  // migrate clients, but this test should be cleaned up with that field.
+  @Test
+  public void testCreateDataReferenceDeprecatedReference() {
+    DataRepoSnapshot snapshotReference =
+        new DataRepoSnapshot().snapshot("snapshot-name").instanceName("instance-name");
+    CreateDataReferenceRequestBody deprecatedRequest =
+        new CreateDataReferenceRequestBody()
+            .referenceType(ReferenceTypeEnum.DATA_REPO_SNAPSHOT)
```
jaycarlton, line 78: I think it's not too late to rename `ReferenceTypeEnum` to `ReferenceType`. Clients have access to that name but since the ordinals are the same it should just work, right?

src/test/java/bio/terra/workspace/service/datareference/DataReferenceServiceTest.java
```diff
@@ -89,6 +131,23 @@ public void testGetDataReferenceByName() {
     assertThat(ref.name(), equalTo(request.name()));
   }
 
+  @Test
+  public void testGetDataReferenceByWrongTypeMissing() {
+    UUID workspaceId = createDefaultWorkspace();
+    DataReferenceRequest request =
+        defaultReferenceRequest(workspaceId, DataReferenceType.DATA_REPO_SNAPSHOT).build();
+    String snapshotName = request.name();
+    dataReferenceService.createDataReference(request, USER_REQUEST).referenceId();
```
jaycarlton, line 140: why aren't we doing anything with the reference id?

src/test/java/bio/terra/workspace/service/datareference/model/ReferenceObjectTest.java
```diff
@@ -31,4 +31,52 @@ public void SnapshotReferenceSerializationBackwardsCompatible() {
...
+  /**
+   * Hard code serialized values to check that code changes do not break backwards compatibility of
+   * stored JSON values. If this test fails, your change may not work with existing databases.
+   */
+  @Test
+  public void BigQueryDatasetReferenceSerializationBackwardsCompatible() {
+    BigQueryDatasetReference datasetReference =
+        (BigQueryDatasetReference)
+            ReferenceObject.fromJson(
+                "{\"@type\":\"BigQueryDatasetReference\",\"projectId\":\"foo\",\"datasetName\":\"bar\"}");
```
jaycarlton, line 78: any more complicated than this and we should just have JSON files in the test references

jaycarlton (commented):
some notes

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
zloery: My concern with that is that `name` is a user provided string, and the input from security has been to avoid directly returning user-provided strings where possible. A common pattern we use instead is returning an error message without the values and logging the actual values, which I've added here.

src/main/java/bio/terra/workspace/service/datareference/model/BigQueryDatasetReference.java
zloery: I thin all of our autovalue classes don't start with `get`, while generated swagger classes do. I'm not thrilled with it and we can change the autovalue classes if we want for consistency, but that might be more hassle than it's worth.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zloery: same logging/response issue as above

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zloery: same logging/response issue as above

src/test/java/bio/terra/workspace/service/datareference/DataReferenceServiceTest.java
zloery: whoops, it was previously an unused variable and I removed the var but not `.referenceId()`. We should be calling `createDataReference()` and ignoring the return value.

src/test/java/bio/terra/workspace/service/datareference/model/ReferenceObjectTest.java
zloery: I think these are simple enough JSON structures to leave as-is, and I think they provide a quick smoke test level of "do my changes break serialization"

build.gradle
zloery: I think this is a larger discussion at an MC terra level, which was where we decided to share yaml via artifactory. it's worth surfacing those issues, but I think it's out of scope for this PR

src/main/java/bio/terra/workspace/service/datareference/model/BigQueryDatasetReference.java
zloery: Do we need builders, or is that for consistency with other autovalue classes? I'm not sure what benefit it brings otherwise

build.gradle
jaycarlton: sounds good. I want to try the other way to see if that's even viable. Don't know the URL I'd use for it though.

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
jaycarlton: Sounds good.

src/main/java/bio/terra/workspace/service/datareference/model/BigQueryDatasetReference.java
jaycarlton: It's for style, safety, & immutability. Looks like you do get it for free:
```
 static final class Builder extends DataReference.Builder { // ...
 ```
 💯

File: src/test/java/bio/terra/workspace/service/datareference/DataReferenceServiceTest.java (hunk 12 above)
Comment (jaycarlton) [reply src/test/java/bio/terra/workspace/service/datareference/DataReferenceServiceTest.java]: glad we're ignoring the right value 😥

File: src/test/java/bio/terra/workspace/service/datareference/model/ReferenceObjectTest.java (hunk 13 above)
Comment (jaycarlton) [reply src/test/java/bio/terra/workspace/service/datareference/model/ReferenceObjectTest.java:78]: You must really like backslashes.

**Review by jaycarlton — APPROVED** [review]
Thanks for the detailed responses!

**Comment by davidangb** [comment]
fyi @MatthewBemis for BigQuery references and @zarsky-broad for APIs related to managing references

File: src/main/java/bio/terra/workspace/service/datareference/model/DataReference.java [hunk 14]
```diff
@@ -32,13 +33,35 @@
...
+            .referenceId(referenceId())
+            .name(name())
+            .workspaceId(workspaceId())
+            .referenceType(referenceType().toApiModel())
+            .referenceInfo(new DataReferenceInfo())
+            .cloningInstructions(cloningInstructions().toApiModel());
+    switch (referenceType()) {
+      case DATA_REPO_SNAPSHOT:
+        // TODO: we populate both the deprecated reference field and the new referenceInfo fields
+        // until all clients have migrated to use the new field.
```
davidangb, line 47: this will require more than just adopting the new client library - for instance, the Terra UI is looking inside the `reference` key to find snapshot ids. We can migrate but it'll be more work.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -63,9 +84,55 @@ private void validateSnapshotReference(SnapshotReference ref, AuthenticatedUserR
...
+        logger.warn(
+            String.format(
+                "Dataset %s not found in project %s", ref.datasetName(), ref.projectId()));
+        throw new InvalidDataReferenceException(
+            "Could not access specified BigQuery dataset. Ensure the name and GCP project are correct and that you have access.");
+      }
+    } catch (BigQueryException e) {
+      throw new InvalidDataReferenceException("Error while trying to access BigQuery dataset", e);
+    }
+  }
```
davidangb, line 137: I wonder if each `ReferenceObject`  subclass should define their own `validate()` method inside their own class  - similar to how `toJson()` is defined. Then we'd just delegate to their own implementations. This might require passing a `CrlService` or `DataRepoService` to it.

src/main/resources/api/service_openapi.yaml
```diff
@@ -582,11 +582,7 @@ components:
     ReferenceTypeEnum:
       type: string
       description: The type of a reference, if an uncontrolled resource
-      # TODO: swagger-codegen always looks for and removes a common prefix for
-      # enum values, which breaks enums that only have a single value.
-      # PLACEHOLDER_VALUE should be removed when we have an actual second value
-      # to use.
-      enum: ['DATA_REPO_SNAPSHOT', 'PLACEHOLDER_VALUE'] #eventually include GCS bucket, etc.
+      enum: ['DATA_REPO_SNAPSHOT', 'GOOGLE_BUCKET', 'BIG_QUERY_DATASET']
```
davidangb, line 585: glad to have `PLACEHOLDER_VALUE` out of here!

src/main/resources/api/service_openapi.yaml
```diff
@@ -690,6 +669,27 @@ components:
...
+#      description: A reference to a BigQuery dataset.
+#      type: object
+#      required: [projectId, datasetName]
+#      properties:
+#        projectId:
+#          description: ID of the Google project holding this dataset
+#          type: string
+#        datasetName:
+#          description: The name of this dataset
+#          type: string
```
davidangb, line 691: delete this block?

davidangb (approved):
:+1:  overall, some non-blocking comments inline.

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
ddietterich: And there is nothing dangerous about returning the reference type or workspaceId is there?
So if we fixed all of that, you could skip the logging and just throw and let the global exception handler do the logging.

src/main/java/bio/terra/workspace/service/datareference/model/BigQueryDatasetReference.java
ddietterich: I want to re-open the question of whether we should use AutoValue classes at all.
I was neutral on them at first, but they don't work well with Jackson - which means they don't work well as Stairway parameters. That means that we not only do translation from the API classes to internal ones, but we have to do a second explosion/translation to have them work in Stairway. Instead of being simple, they become littered with Json annotations.

Given that IntelliJ can auto-magically generate and regenerate the internals of a value class, I am not seeing a big win.

src/main/java/bio/terra/workspace/service/datareference/model/DataReferenceType.java
```diff
@@ -18,8 +22,12 @@
 
   static {
     typeMap.put(DataReferenceType.DATA_REPO_SNAPSHOT, ReferenceTypeEnum.DATA_REPO_SNAPSHOT);
+    typeMap.put(DataReferenceType.GOOGLE_BUCKET, ReferenceTypeEnum.GOOGLE_BUCKET);
+    typeMap.put(DataReferenceType.BIG_QUERY_DATASET, ReferenceTypeEnum.BIG_QUERY_DATASET);
 
     sqlMap.put(DATA_REPO_SNAPSHOT, "DATA_REPO_SNAPSHOT");
```
ddietterich, line 28: This map is unnecessary. The Enum's built-in `name()` and `validOf` methods will do the same thing. And you can convert to them without changing your interface.

src/main/java/bio/terra/workspace/service/datareference/model/DataReferenceType.java
```diff
@@ -18,8 +22,12 @@
 
   static {
     typeMap.put(DataReferenceType.DATA_REPO_SNAPSHOT, ReferenceTypeEnum.DATA_REPO_SNAPSHOT);
+    typeMap.put(DataReferenceType.GOOGLE_BUCKET, ReferenceTypeEnum.GOOGLE_BUCKET);
```
ddietterich, line 25: And this map is overkill. Maybe I will write you an "enum" class that will use a list search - which is generally faster than a hash for lists smaller than 10 items...

I know, it probably isn't important and it is all static. If we were running on a PDP-11/70 you would thank me :)

src/main/java/bio/terra/workspace/service/datareference/model/DataReference.java
zloery: Hm, good to know. I think we would still want to migrate, but it sounds like it's going to be worth coordinating more on this. I've currently been tracking this as PF-404, but we may end up with AS tickets as well.

src/main/resources/api/service_openapi.yaml
zloery: Same, that truncation is still one of the strangest opinions I've seen swagger take.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zloery: Having to pass the services to the objects was the reason I didn't want to take that approach - it feels like we're drifting further away from these being value classes. (They already kind of weren't with things like `toApiModel()`, but adding more members makes serialization/deserialization trickier).

ddietterich: What is the benefit of them being value classes if we have to write switch statements to dispatch methods on them?

I think it would be a good separation of concerns to have the reference objects encapsulate the knowledge of the cloud resource they reference.

ddietterich (changes requested):
I think there is an upward compatibility issue and maybe a validation issue.

src/main/java/bio/terra/workspace/service/datareference/model/DataReferenceType.java
jaycarlton: The idea of the sql map is to have invariant names, so that even if they refactor the enum names in Java, we can serialize correctly. `.name()` is too smart for this purpose, and would return the new name.

src/main/java/bio/terra/workspace/service/datareference/model/DataReferenceType.java
jaycarlton: https://skn.noip.me/pdp11/pdp11.html

jaycarlton:
https://github.com/DataBiosphere/terra-workspace-manager/pull/212#discussion_r568979736

Yeah, and we could have a parent class with just the type as a public member in case we need it.

src/main/resources/api/service_openapi.yaml
```diff
@@ -598,23 +594,20 @@ components:
 
     CreateDataReferenceRequestBody:
       type: object
-      required: [name, cloningInstructions]
+      required: [name, referenceType, cloningInstructions]
```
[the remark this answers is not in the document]
jaycarlton: I think that's a good point. default it to data repo snapshot and leave it optional.

ddietterich:
As posted in slack:
> Small brainstorm especially of interest to @zloery and @jaycarlton. What about making the new references endpoints use the path `/api/workspaces/v1/resources/references` and keep all of the WM resources within one collection. We would leave the current datareferences as is for backward compatibility.

It might make the new version cleaner because we could require referenceType and remove the reference field. And we can make a general GET for /resources that returns everything we’ve got. I know this is a late-breaking change, but I just thought of it after Jay’s suggestion on the controlled resources spec.

src/main/resources/api/service_openapi.yaml
```diff
@@ -690,6 +669,27 @@ components:
           description: The ID of the Data Repo snapshot
           type: string
 
+#    GoogleBucket:
```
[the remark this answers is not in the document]
zloery: Nope, we replaced those definitions with shared ones from the common library and I forgot to remove them

src/main/resources/api/service_openapi.yaml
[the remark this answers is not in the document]
zloery: Rawls is already specifying type, and this was previously enforced by rejecting requests with no `referenceType` field in `ControllerValidationUtils`. 
We did that validation as part of the app logic instead of putting it in Swagger because the request shape was (kind of) also used for controlled resource requests, so only common fields between the two were made required. Now that we're removing all the unused controlled resource stuff here, we can move `referenceType` to required. I'd also like to move `referenceInfo` there once we stop using the `reference` field.

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -146,20 +148,29 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
-    SnapshotReference snapshot =
-        SnapshotReference.create(
-            body.getReference().getInstanceName(), body.getReference().getSnapshot());
-    dataReferenceValidation.validateReferenceObject(snapshot, referenceType, userReq);
+    final DataReferenceType referenceType = DataReferenceType.fromApiModel(body.getReferenceType());
+
+    final DataReferenceInfo referenceInfo = body.getReferenceInfo();
+    final ReferenceObject referenceObject;
+    if (body.getReference() != null) {
+      referenceObject =
```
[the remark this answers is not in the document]
zloery: We also previously errored in ControllerValidationUtils if referenceType was not specified, so that is actually compatible.

src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -42,24 +43,48 @@ public static void validatePaginationParams(int offset, int limit) {
...
+    if (referenceType == null) {
+      throw new ValidationException("Invalid reference type specified");
     }
-    if (body.getReferenceType() == null || body.getReference() == null) {
-      throw new InvalidDataReferenceException(
-          "Data reference must contain a reference type and a reference description");
+    // TODO(PF-404): requests using the deprecated reference field will not set the referenceInfo
+    // field. referenceInfo can be made required and this check can be removed once clients migrate.
+    DataReferenceInfo info =
+        body.getReferenceInfo() == null ? new DataReferenceInfo() : body.getReferenceInfo();
```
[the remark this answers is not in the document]
zloery: This handles the existing case where a client specifies the old `reference` field but does not specify the new `referenceInfo` field. You can't have bucket or dataset references this way (as `getGoogleBucket()` and `getBigQueryDataset()` on this reference will be null), but you can specify a snapshot in the old `reference` field this way, and we want to support that until we move clients. That's why we check either the `referenceInfo` or the `reference` field when validating snapshots here.
(As above, existing clients do already use the `DATA_REPO_SNAPSHOT` referenceType, so that hasn't changed)

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
zloery: Makes sense, I'll add bucket and dataset name validation and then return the names as part of the error response.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zloery: The value is that they can be serialized into the database and back out, which is how we store the actual references. If we require the reference objects to have a `DataRepoService` field, that gets more complicated - do we serialize the `DataRepoService` field with the reference? and if not, how do you deserialize into a valid reference object?

src/main/java/bio/terra/workspace/service/datareference/model/DataReferenceType.java
zloery: Cutting out the `typeMap` seems fine. I'm a little less sure about  `sqlMap` for the reason Jay stated, since this value does get serialized directly.

zloery:
re: moving to `/api/workspaces/v1/resources/references`, I'm definitely thinking about it.

I'm also thinking about our model of resources - the DB structure suggests that we have data references to all resources (both controlled and uncontrolled), so should the common GET be on `.../datareferences`? It's also not clear to me what the referenceId of a reference to a controlled resource really though. probably worth more discussion

src/main/java/bio/terra/workspace/service/datareference/model/BigQueryDatasetReference.java
zloery: ack. We use AutoValue throughout WM so I think changing it is out of scope here, but I opened PF-433 to track that effort and can pick that up separately.

zloery:
Per earlier discussion, we'll break out each type into a separate endpoint, while still supporting the existing data reference APIs for compatibility in the short-term. Moving this to a draft while I work on that

zarsky-broad:
FWIW: https://github.com/OpenAPITools/openapi-generator/pull/5661#issuecomment-772873049

zloery:
...huh. I'm wary of blocking on something that's been in progress for almost a year, but I want to keep an eye on that

gjuggler:
> ...huh. I'm wary of blocking on something that's been in progress for almost a year, but I want to keep an eye on that. thanks for the pointer!

Also, don't we use swagger-codegen and not openapi-generator? ::shakes fist at whoever forked the projects, however good their intentions may have been::

zloery:
Update: I've broken out the create, get, get-by-name, and delete endpoints for each resource type separately, and the new interface no longer uses any polymorphic types. I'm not totally sure this structure is right (delete and get-by-id don't really depend on type and could be methods on `/api/workspaces/v1/{id}/resources/references`, but I think this is ready for more eyes.
I haven't replaced the enumerate endpoint yet, which we will need to do before fully migrating.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
ddietterich: I think there is a middle ground. I wrote a doc for further discussion: [Workspace Manager Object Hierarchy](https://docs.google.com/document/d/1yOW5oYeNctijQghEA4PWFMHxHc4p5I5EpUMojDZq14Q/edit#heading=h.x6w7yp6besie)

src/main/resources/api/service_openapi.yaml
```diff
@@ -191,6 +194,228 @@ paths:
...
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '500':
+          $ref: '#/components/responses/ServerError'
+
+  /api/workspaces/v1/{id}/resources/references/datarepo/snapshot/{referenceId}:
```
ddietterich, line 222: I think we should discuss what level we should provide delete, get and enumerate support.
I am sure we need a top-level GET by name and id and enumerate across all resource types.
I am not sure we need per-type GET. The pro is, it is a simpler result. The con is, it is lots of endpoints that might never be used.
I do not think we need per-type DELETE.
I think we need per-type PATCH for controlled resources. Maybe for references, it is a wholesale replacement? A PUT?

src/main/resources/api/service_openapi.yaml
```diff
@@ -191,6 +194,228 @@ paths:
...
+        '204':
+          description: OK
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '500':
+          $ref: '#/components/responses/ServerError'
+
+  /api/workspaces/v1/{id}/resources/references/datarepo/snapshot/name/{name}:
```
ddietterich, line 253: Is name URL safe? Or you just assume it is encoded?

ddietterich (commented):
lotta good stuff in here! I want to discuss object model before we merge on it and maybe remove some endpoints.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -64,8 +111,54 @@ private void validateSnapshotReference(SnapshotReference ref, AuthenticatedUserR
...
+        throw new InvalidDataReferenceException(
+            String.format(
+                "Could not access GCS bucket %s. Ensure the name is correct and that you have access.",
+                ref.bucketName()));
+      }
+    } catch (StorageException e) {
+      throw new InvalidDataReferenceException(
+          String.format("Error while trying to access GCS bucket %s", ref.bucketName()), e);
+    }
+  }
```
davidangb, line 140: I haven't looked at the innards of CRL - do you know what happens here if the bucket is requester-pays?

src/main/java/bio/terra/workspace/app/controller/GoogleReferenceController.java
```diff
@@ -0,0 +1,245 @@
...
+    this.dataReferenceService = dataReferenceService;
+    this.authenticatedUserRequestFactory = authenticatedUserRequestFactory;
+    this.request = request;
+  }
+
+  private final Logger logger = LoggerFactory.getLogger(GoogleReferenceController.class);
+
+  private AuthenticatedUserRequest getAuthenticatedInfo() {
+    return authenticatedUserRequestFactory.from(request);
+  }
```
davidangb, line 53: `DataRepoReferenceController` and `GoogleReferenceController` share boilerplate - is there opportunity for a `ReferenceController` base class?

src/main/resources/api/service_openapi.yaml
```diff
@@ -191,6 +194,228 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/workspaces/v1/{id}/resources/references/datarepo/snapshots:
+    parameters:
+    - $ref: '#/components/parameters/Id'
```
davidangb, line 199: not for this PR: I really wish this was named `workspaceId` instead of the totally-ambiguous `id`

src/main/resources/api/service_openapi.yaml
```diff
@@ -754,6 +1071,27 @@ components:
...
+          schema:
+            $ref: '#/components/schemas/GoogleBucketReference'
+
+    BigQueryDatasetReferenceResponse:
+      description: Response containing a reference to a BiqQuery dataset.
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/BigQueryDatasetReference'
+
```
davidangb, line 1094: this will also need updates based on #221  - adding description to `DataReferenceMetadata` ( I assume that's where it would go ), plus patch APIs for each reference type.

davidangb (commented):
see comments inline!

src/main/java/bio/terra/workspace/app/controller/DataRepoReferenceController.java
```diff
@@ -0,0 +1,146 @@
...
+  private AuthenticatedUserRequest getAuthenticatedInfo() {
+    return authenticatedUserRequestFactory.from(request);
+  }
+
+  @Override
+  public ResponseEntity<DataRepoSnapshotReference> createDataRepoSnapshotReference(
+      UUID id, @Valid CreateDataRepoSnapshotReferenceRequestBody body) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    logger.info(
+        String.format(
```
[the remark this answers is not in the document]
zloery: Nice, thanks!

src/main/resources/api/service_openapi.yaml
zloery: Names must be URL safe and they're checked for validity in Controller methods

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -234,16 +246,21 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
-            "Got data references in workspace %s for %s", id.toString(), userReq.getEmail()));
+            "Got snapshot data references in workspace %s for %s",
+            id.toString(), userReq.getEmail()));
     DataReferenceList responseList = new DataReferenceList();
     for (DataReference ref : enumerateResult) {
-      responseList.addResourcesItem(ref.toApiModel());
+      // TODO(PF-404): this is a workaround until clients migrate off this endpoint.
+      if (ref.referenceType() == DataReferenceType.DATA_REPO_SNAPSHOT) {
+        responseList.addResourcesItem(ref.toApiModel());
+      }
```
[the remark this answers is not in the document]
zloery: Agreed. I think the main way to mitigate this is not adopting the new endpoints (including creating non-snapshot references) until the enumerate is ready, but that's a risk.

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
```diff
@@ -24,15 +34,47 @@
...
+  public static final Pattern BUCKET_NAME_VALIDATION_PATTERN =
+      Pattern.compile("^[a-z0-9][-_.a-z0-9]{1,220}[a-z0-9]$");
+
+  /**
+   * BigQuery datasets must be 1-1024 characters, using letters (upper or lowercase), numbers, and
+   * underscores.
+   */
+  public static final Pattern BQ_DATASET_NAME_VALIDATION_PATTERN =
+      Pattern.compile("^[_a-zA-Z0-9]{1,1024}$");
+
```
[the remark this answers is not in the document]
zloery: Generally the client libs don't validate the names, you'll just get a 40X response from GCP. Agreed that rolling our own regexes for a "necessary but not sufficient" check is a little messy, but I think it's better than passing around totally unverified strings

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
zloery: I haven't seen that case specifically, but CRL is a really thin wrapper around the Google client library. If we're calling GET on a bucket (which I think is a class B operation), it looks like this will fail with a `400 UserProjectMissing` and we'd end up throwing. Is that a usecase we need to support?

src/main/java/bio/terra/workspace/app/controller/GoogleReferenceController.java
zloery: Done. I'm not sure I took the cleanest approach given the way I'm passing beans around, but it does cut down on boilerplate. I might try to revisit this a bit to shrink the argument list

src/main/resources/api/service_openapi.yaml
zloery: it really should be, that's tripped people up in client tests before when `id` gets mixed up between workspaces and references.

zloery:
Update: the current approach is per-type create, patch, and get (and get-by-name) endpoints. Delete and enumerate will be shared at a `.../resources` level, out of scope for this PR. Reference descriptions are also merged in, PTAL.

src/main/java/bio/terra/workspace/app/controller/common/ReferenceController.java
```diff
@@ -0,0 +1,131 @@
...
+      throw new InvalidDataReferenceException(
+          "Must specify name or referenceDescription to update.");
+    }
+
+    if (body.getName() != null) {
+      DataReferenceValidationUtils.validateReferenceName(body.getName());
+    }
+
+    logger.info(
+        String.format(
```
ddietterich, line 121: You don't need the almost all the time. You can use placeholders {} in the log message.

build.gradle
```diff
@@ -125,6 +128,20 @@ if (hasProperty("buildScan")) {
...
+	cloudResourceSchema { transitive = false }
+}
+dependencies {
+	cloudResourceSchema group: 'bio.terra.cloud-resource-lib', name: 'cloud-resource-schema'
+}
+task unzipCloudResourceSchema(type: Copy) {
+	from zipTree(configurations.cloudResourceSchema.singleFile).matching {
+		include 'cloud_resources_uid.yaml'
+	}
+	into "${resourceDir}/api/common"
```
ddietterich, line 142: Why do we need to unpack this into the source tree?
Is an alternative to copy our resources into the build tree?

ddietterich (commented):
Looking good. A scattering of comments.

I want to understand moving the untar of the common swagger into the source tree. I'd like to avoid that if possible

src/main/java/bio/terra/workspace/service/datareference/utils/DataReferenceValidationUtils.java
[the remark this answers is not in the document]
jaycarlton: Devil's advocate: what are we ever going to use the string for except to pass to Google to make a resource? What's the harm in late validation?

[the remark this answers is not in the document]
ddietterich: We want to send it back to the user in an error message. Security doesn't want us putting unvalidated user input into an output string.

src/test/java/bio/terra/workspace/service/datareference/DataReferenceValidationUtilsTest.java
```diff
@@ -0,0 +1,54 @@
...
+  @Test
+  public void validCharInReferenceNameOk() {
+    DataReferenceValidationUtils.validateReferenceName("valid_name");
+  }
+
+  @Test
+  public void testInvalidCharInBucketName() {
+    assertThrows(
+        InvalidDataReferenceException.class,
+        () -> DataReferenceValidationUtils.validateBucketName("INVALIDBUCKETNAME"));
```
jaycarlton, line 28: It would be nice to have this validation at a higher level so controlled resources can use it too.

jaycarlton (approved):

src/main/java/bio/terra/workspace/app/controller/GoogleReferenceController.java
```diff
@@ -0,0 +1,184 @@
...
+    return new ResponseEntity<>(response, HttpStatus.OK);
+  }
+
+  @Override
+  public ResponseEntity<GoogleBucketReference> getGoogleBucketReference(UUID id, UUID referenceId) {
+    DataReference ref = getReference(id, referenceId, getAuthenticatedInfo(), dataReferenceService);
+    GoogleBucketReference response =
+        new GoogleBucketReference()
+            .bucket(
+                ((bio.terra.workspace.service.datareference.model.GoogleBucketReference)
```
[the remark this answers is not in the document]
zloery: Agreed, but I think making that change (especially if we add a global prefix) end up out of scope for this PR.

src/main/java/bio/terra/workspace/app/controller/common/ReferenceController.java
zloery: Ah, I missed this one. thanks!

build.gradle
zloery: I unpacked it into the source tree so that our swagger-ui (also in the source dir) can parse references to it. otherwise, those schemas won't show up in swagger-ui, which breaks example values.

build.gradle
```diff
@@ -230,9 +247,12 @@ jib {
 	}
 }
 
-// Internal dependencies
-
 sourceSets.main.java.srcDir "${swaggerOutputDir}/src/main/java"
+
+// Internal dependencies
+// Note The Open API schema depends on an external library - cloud-resource-schema, so need to unzip it first.
+swaggerSources.server.code.dependsOn tasks.unzipCloudResourceSchema
```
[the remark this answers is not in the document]
zloery: This took longer to investigate and fix than I want to admit, but you're right that swagger client tasks didn't depend on this. Should be fixed now

src/main/java/bio/terra/workspace/app/controller/GoogleReferenceController.java
```diff
@@ -0,0 +1,184 @@
...
+                ((bio.terra.workspace.service.datareference.model.BigQueryDatasetReference)
+                        ref.referenceObject())
+                    .toApiModel())
+            .metadata(ControllerTranslationUtils.metadataFromDataReference(ref));
+    return new ResponseEntity<>(response, HttpStatus.OK);
+  }
+
+  @Override
+  public ResponseEntity<Void> updateGoogleBucketReference(
+      UUID id, UUID referenceId, UpdateDataReferenceRequestBody body) {
```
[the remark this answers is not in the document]
zloery: Done

src/test/java/bio/terra/workspace/service/datareference/DataReferenceValidationUtilsTest.java
zloery: This is in a validation util class called by the controller, I think it's actually reasonable for the various controlled resource controllers to use this bean as well

build.gradle
ddietterich: We can leave it for now. I'll make a ticket.
Another approach might be to move all of the parts into a build subdirectory and do the codegen from there.
I'll make a ticket.

ddietterich (approved):

zloery:
Published version `0.12.0-SNAPSHOT` from this.

PR #220, opened 2021-02-05
PF-408 Controlled Google Bucket Framing

jaycarlton:
This includes:
* API
* controller
* flight
* steps
* dao
* service
* dao unit test

ControlledGcsBucketResource


src/main/resources/api/service_openapi.yaml
```diff
@@ -739,6 +756,162 @@ components:
...
+          description: Asyncronous job ID for this create operation.
+          type: string
+        googleBucket:
+          $ref: '#/components/schemas/GoogleBucketCreationParameters'
+
+    CreatedControlledGoogleBucket:
+      description: Response Payload for requesting a new controlled resource.
+      type: object
+      properties:
+        resourceId:
```
ddietterich, line 773: This will need a JobReport. See the response form in this "template" for an idea of what this might look like.
https://docs.google.com/document/d/1PTd4xvmV9xnEkWaIgFc6d3VUyJwwymQTFdhpKrdsUKw/edit?ts=60142fef#heading=h.cabe38m95n2e

src/main/resources/api/service_openapi.yaml
```diff
@@ -739,6 +756,162 @@ components:
...
+        Type and availability of objects in a bucket, described at https://cloud.google.com/storage/docs/storage-classes.
+        This list does not include legacy types that apply to some existing buckets. For those, see
+        GoogleBucketLegacyStorageClass.
+      type: string
+      enum:
+        - STANDARD
+        - NEARLINE
+        - COLDLINE
+        - ARCHIVE
+    GoogleBucketLegacyStorageClass:
```
ddietterich, line 811: We won't need to support this, since we only make new bucket.

src/main/resources/api/service_openapi.yaml
```diff
@@ -768,6 +941,13 @@ components:
           schema:
             $ref: '#/components/schemas/RoleBindingList'
 
+    CreatedControlledGoogleBucketResponse:
```
ddietterich, line 944: I am thinking we want a single structure that we return on resource retrieve and enumeration. Should we use that same structure for these responses?
The benefit is that the client has one object to navigate. The bad part is having to navigate when there is no reason.

ddietterich (commented):
Comments

src/main/java/bio/terra/workspace/app/controller/ControlledGoogleResourceApiController.java
```diff
@@ -0,0 +1,70 @@
...
+  private final JobService jobService;
+
+  @Autowired
+  public ControlledGoogleResourceApiController(
+      AuthenticatedUserRequestFactory authenticatedUserRequestFactory,
+      ControlledGoogleResourceService controlledResourceService,
+      JobService jobService) {
+    this.authenticatedUserRequestFactory = authenticatedUserRequestFactory;
+    this.controlledResourceService = controlledResourceService;
+    this.request = getRequest().orElse(null);
```
jaycarlton, line 39: I'm not sure about this. I noticed IJ complaining something wasn't Kosher about injecting the request and saw this method in the parent class.

src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -79,4 +80,10 @@ public static void validateCloudContext(CloudContext context) {
           "Invalid cloud context. Currently, only Google contexts are supported.");
     }
   }
+
+  public static void validateGoogleBucket(GoogleBucketCreationParameters params) {
+    if (params.getName() == null) {
```
jaycarlton, line 85: I'm not sure how much validation I get for free from the generated code's required attributes.

src/main/java/bio/terra/workspace/db/ControlledResourceDao.java
```diff
@@ -0,0 +1,83 @@
...
+            + "(workspace_id, resource_id, associated_app, is_visible, owner, attributes) values "
+            + "(:workspaceId, :resourceId, :associatedApp, :isVisible, :owner, :attributes)";
+    final var params =
+        new MapSqlParameterSource()
+            .addValue("workspaceId", controlledResource.workspaceId())
+            .addValue("resourceId", controlledResource.resourceId())
+            .addValue("isVisible", controlledResource.isVisible())
+            .addValue("associatedApp", controlledResource.associatedApp().orElse(null))
+            .addValue("owner", controlledResource.owner().orElse(null))
+            .addValue("attributes", controlledResource.attributes().orElse(null));
```
jaycarlton, line 39: I suppose I need to build a map into JSON for this field. Not sure I need it yet though.

src/main/java/bio/terra/workspace/db/ControlledResourceDao.java
```diff
@@ -0,0 +1,83 @@
...
+        "SELECT workspace_id, resource_id, associated_app, is_visible, owner, attributes "
+            + "FROM workspace_resource "
+            + "WHERE workspace_id = :workspaceId AND resource_id = :resourceId";
+    final MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue("workspaceId", workspaceId.toString())
+            .addValue("resourceId", resourceId.toString());
+    final Map<String, Object> columnToValue;
+    try {
+      columnToValue = jdbcTemplate.queryForMap(sql, params);
```
jaycarlton, line 68: `queryForMap` will throw if there aren't exactly 1 rows in the result set. We know there won't be more than one because the resource id is unique.

src/main/java/bio/terra/workspace/db/ControlledResourceDao.java
```diff
@@ -0,0 +1,83 @@
...
+      throw new DuplicateControlledResourceException(
+          String.format(
+              "Resource %s already exists in workspace %s",
+              controlledResource.resourceId(), controlledResource.workspaceId()),
+          e);
+    }
+  }
+
+  public Optional<ControlledResourceMetadata> getControlledResource(
+      UUID workspaceId, UUID resourceId) {
```
jaycarlton, line 57: Should I even take in the workspace id here if the resource id is sufficient?

src/main/java/bio/terra/workspace/service/controlledresource/ControlledGoogleResourceService.java
```diff
@@ -0,0 +1,55 @@
...
+        jobService.newJob(
+            description,
+            requestBody.getCommon().getJobControl().getId(),
+            CreateControlledGoogleBucketFlight.class,
+            requestBody,
+            userRequest);
+    final GoogleBucketCreationParameters params = requestBody.getGoogleBucket();
+    jobBuilder.addParameter(WorkspaceFlightMapKeys.WORKSPACE_ID, workspaceId);
+    jobBuilder.addParameter(
+        WorkspaceFlightMapKeys.IAM_OWNER_GROUP_EMAIL, userRequest.getEmail()); // IS THIS RIGHT?
```
jaycarlton, line 42: I think we want a new user email flight key

src/main/java/bio/terra/workspace/service/controlledresource/model/ControlledResourceMetadata.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public abstract Optional<String> owner();
+
+  public abstract Optional<Object> attributes();
+
+  public static ControlledResourceMetadata.Builder builder() {
+    return new AutoValue_ControlledResourceMetadata.Builder();
+  }
+
+  @AutoValue.Builder
```
jaycarlton, line 25: This is way more work than Immutables.

src/main/java/bio/terra/workspace/service/workspace/flight/DeleteWorkspaceStateStep.java
```diff
@@ -9,14 +9,12 @@
...
 import bio.terra.workspace.db.WorkspaceDao;
 import java.util.UUID;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 
 public class DeleteWorkspaceStateStep implements Step {
 
   private final WorkspaceDao workspaceDao;
 
-  @Autowired
```
jaycarlton, line 19: Can't use autowired here b/c it's not a bean.

src/main/java/bio/terra/workspace/service/workspace/flight/GenerateControlledResourceIdStep.java
```diff
@@ -0,0 +1,24 @@
...
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    flightContext
+        .getWorkingMap()
+        .put(WorkspaceFlightMapKeys.CONTROLLED_RESOURCE_ID, UUID.randomUUID());
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
```
jaycarlton, line 21: should this `undoStep` technically clear the entry from the map?

src/main/resources/api/service_openapi.yaml
jaycarlton: I think it's straightforward for our union type to include by reference each of the individual response types.

src/main/resources/api/service_openapi.yaml
jaycarlton: Oh.

src/main/java/bio/terra/workspace/app/controller/ControlledGoogleResourceApiController.java
```diff
@@ -0,0 +1,70 @@
...
+    final CreatedControlledGoogleBucket response = fetchGoogleBucketResult(jobId);
+    return new ResponseEntity<>(
+        response, HttpStatus.valueOf(response.getJobReport().getStatusCode()));
+  }
+
+  private CreatedControlledGoogleBucket fetchGoogleBucketResult(String jobId) {
+    final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    final AsyncJobResult<CreatedControlledGoogleBucket> jobResult =
+        jobService.retrieveAsyncJobResult(jobId, CreatedControlledGoogleBucket.class, userRequest);
+    return jobResult.getResult();
```
ddietterich, line 64: Won't that be null if the job is not complete?

src/main/java/bio/terra/workspace/db/ControlledResourceDao.java
ddietterich: Might be redundant. Depends on what indexes are defined in the database.

src/main/java/bio/terra/workspace/db/ControlledResourceDao.java
ddietterich: Yes, but it leaves you casting later instead of being able to use the result methods like getString("workspace_id").

src/main/java/bio/terra/workspace/service/controlledresource/ControlledGoogleResourceService.java
ddietterich: You are passing the request body right into the flight, so can't you pull out everything you need from that? I'm confused.

src/main/java/bio/terra/workspace/service/controlledresource/model/ControlledResourceMetadata.java
ddietterich: But they both have the same downside: they are not serializable with Jackson without extra Json annotations.

src/main/java/bio/terra/workspace/service/workspace/flight/GenerateControlledResourceIdStep.java
ddietterich: That isn't the common practice.

src/main/java/bio/terra/workspace/service/workspace/flight/GoogleBucketFlightMapKeys.java
```diff
@@ -0,0 +1,33 @@
...
+    this.klass = klass;
+  }
+
+  public String getKey() {
+    return key;
+  }
+
+  public Class<? extends Object> getKlass() {
+    return klass;
+  }
```
ddietterich, line 32: Could you avoid exposing this by making a method like:
```
<T> T getFromMap(FlightMap map, GoogleBucketFlightMapKeys key) {
  return map.get(key.getName(), key.getKlass());
}
```
??

src/main/java/bio/terra/workspace/service/workspace/flight/GoogleBucketFlightMapKeys.java
```diff
@@ -0,0 +1,33 @@
...
+import bio.terra.workspace.generated.model.GoogleBucketCreationParameters;
+import bio.terra.workspace.generated.model.GoogleBucketDefaultStorageClass;
+import bio.terra.workspace.generated.model.GoogleBucketLifecycle;
+
+/**
+ * Keys into the flight parameter map along with their types. Corresponds to
+ * GoogleBucketCreationParameters
+ */
+public enum GoogleBucketFlightMapKeys {
+  NAME("name", String.class),
```
ddietterich, line 12: This is an interesting approach. Should we put this into the flight class itself?

src/main/java/bio/terra/workspace/service/workspace/flight/StoreGoogleBucketMetadataStep.java
```diff
@@ -0,0 +1,45 @@
...
+  public StoreGoogleBucketMetadataStep(ControlledResourceDao controlledResourceDao) {
+    this.controlledResourceDao = controlledResourceDao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    final FlightMap workingMap = flightContext.getWorkingMap();
+    final ControlledResourceMetadata controlledResourceMetadata =
+        ControlledResourceMetadata.builder()
```
ddietterich, line 25: I think we should build the intermediate structure in the controller - not passing the request body through. That is the general practice.

If you did that AND you didn't use @AutoValue to make it, you could pass that structure in on the JobBuilder as the request parameter. That parameter becomes the REQUEST entry in the inputParameters FlightMap. That would save all of this to-ing and fro-ing with individual parameters.

I think some of these parameters are from the inputParameter map and not from the working map.

ddietterich (commented):
My understanding is that this is a work-in-progress.
I've reviewed and commented.

src/main/resources/api/service_openapi.yaml
```diff
@@ -783,6 +824,191 @@ components:
...
+      type: object
+      properties:
+        resourceId:
+          description: UUID of a newly-created resource. Null if not created yet.
+          type: string
+          format: uuid
+        jobReport:
+          $ref: '#/components/schemas/JobReport'
+        googleBucket:
+          $ref: '#/components/schemas/GoogleBucketCreationParameters' # TODO: should this include all parameters, not just the settable ones we expose?
```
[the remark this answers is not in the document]
jaycarlton: Nope, just verifying we should do it the easy way.

jaycarlton:
https://github.com/DataBiosphere/terra-workspace-manager/pull/220#discussion_r574051699

For some reason, static analysis complains about this injection as being not from a bean.
Screen Shot 2021-02-11 at 9 38 50 AM

src/main/java/bio/terra/workspace/app/controller/ControlledGoogleResourceApiController.java
jaycarlton: Screen Shot 2021-02-11 at 9 38 50 AM

jaycarlton: switched back to injecting.

src/main/java/bio/terra/workspace/db/ControlledResourceDao.java
jaycarlton: True. So I'd be better off with a result mapper version like `public <T> T queryForObject(String sql, Map<String, ?> paramMap, RowMapper<T>rowMapper)`?

src/main/java/bio/terra/workspace/db/ControlledResourceDao.java
jaycarlton: resource ID is the 
Screen Shot 2021-02-11 at 9 49 05 AM
PK.

src/main/java/bio/terra/workspace/service/controlledresource/ControlledGoogleResourceService.java
jaycarlton: I had it the other way first. Is it OK to just give it the whole thing, or is that too much coupling?

src/main/java/bio/terra/workspace/service/controlledresource/model/ControlledResourceMetadata.java
jaycarlton: They've spent a lot of time getting it tight though. https://immutables.github.io/json.html#jackson. Do POJOs not also need annotations to work with Jackson?

src/main/java/bio/terra/workspace/service/workspace/flight/GenerateControlledResourceIdStep.java
jaycarlton: So only effects on the external environment then.

src/main/java/bio/terra/workspace/app/controller/ControlledGoogleResourceApiController.java
```diff
@@ -0,0 +1,70 @@
...
+    final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    final String jobId = controlledResourceService.createBucket(id, body, userRequest);
+    return getCreateBucketResult(id, jobId);
+  }
+
+  @Override
+  public ResponseEntity<CreatedControlledGoogleBucket> getCreateBucketResult(
+      UUID id, String jobId) {
+    final CreatedControlledGoogleBucket response = fetchGoogleBucketResult(jobId);
+    return new ResponseEntity<>(
```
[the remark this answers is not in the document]
jaycarlton: Yeah, I was doing the whole object in all four cases (200 vs 202 x 2 endpoints). But the objects contains a JobControl.

src/main/java/bio/terra/workspace/service/workspace/flight/GoogleBucketFlightMapKeys.java
jaycarlton: What I"d like to do is extend an interface like `FlightMapKey<Class<?> T>` and have several enums that all support lookup into the map. The map class would then support `<T> T get(FlightMapKey<T> key)`. Assuming I can get the enum inheritance to work right. I think I've done it before.

Your flight map  has code I've not seen before but that I think I've been missing around generic casting.

src/main/java/bio/terra/workspace/service/workspace/flight/GoogleBucketFlightMapKeys.java
jaycarlton: Potentially.

src/main/java/bio/terra/workspace/service/workspace/flight/GoogleBucketFlightMapKeys.java
jaycarlton: I don't think it would work. The type inference gets lost if you're not pointing to a specific member of the enum, and even then, it seems weak.

src/main/java/bio/terra/workspace/service/resource/controlled/WsmResourceType.java
```diff
@@ -0,0 +1,20 @@
+package bio.terra.workspace.service.resource.controlled;
+
+import bio.terra.workspace.service.datareference.model.DataReferenceType;
+
+// TODO: should this merge with DataReferenceType? References and controlled resources support
+// different types.
+public enum WsmResourceType {
```
ddietterich, line 7: I think we should have one enumeration for resource type and it should be cloud-specific. Buckets are the tiny exception where an object is fairly common. I think it will actually make code more complex if we have to check two enumerations to figure out what thing we are dealing with.

src/main/java/bio/terra/workspace/service/workspace/flight/StoreControlledResourceMetadataStep.java
```diff
@@ -0,0 +1,54 @@
...
+    dataReferenceDao.createDataReference(resource.toDataReferenceRequest(resourceId), referenceId);
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    final FlightMap workingMap = flightContext.getWorkingMap();
+    final UUID resourceId = workingMap.get(ControlledResourceKeys.RESOURCE_ID, UUID.class);
+    final boolean deleted = controlledResourceDao.deleteControlledResource(resourceId);
+    return deleted
```
ddietterich, line 50: It should not be an error if the object does not exist. There are two cases:
1. You get an error in the do() method before the object is created; the undo would incorrectly fail
2. undo() could have deleted it and, before Stairway managed to write its log record, the pod could fail. This undo method would be rerun and error.

src/test/java/bio/terra/workspace/common/fixtures/ControlledResourceFixtures.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.UUID;
+
+/** A series of static objects useful for testing controlled resources. */
+public class ControlledResourceFixtures {
+
+  public static final UUID WORKSPACE_ID = UUID.fromString("00000000-fcf0-4981-bb96-6b8dd634e7c0");
+  public static final UUID RESOURCE_ID = UUID.fromString("11111111-fcf0-4981-bb96-6b8dd634e7c0");
+  public static final UUID DATA_REFERENCE_ID =
+      UUID.fromString("33333333-fcf0-4981-bb96-6b8dd634e7c0");
+  public static final String OWNER_EMAIL = "jay@all-the-bits-thats-fit-to-blit.dev";
```
ddietterich, line 24: 😄

ddietterich (approved):
This looks really good.
I have a couple of comments.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateControlledResourceFlight.java
```diff
@@ -0,0 +1,57 @@
...
+  public CreateControlledResourceFlight(FlightMap inputParameters, Object applicationContext) {
+    super(inputParameters, applicationContext);
+    final FlightBeanBag flightBeanBag = FlightBeanBag.getFromObject(applicationContext);
+
+    // Generate a new resource UUID
+    addStep(new GenerateControlledResourceIdStep());
+    // Genderate reference ID for the workspace_data_reference table
+    addStep(new GenerateReferenceIdStep());
+
+    // store the resource metadata in CloudSQL
```
zloery, line 26: nit: this stores the resource metadata in WSM's underlying database, which happens to be CloudSQL in most environments. But that isn't necessarily be true, e.g. locally it's just a postgres DB.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateControlledResourceFlight.java
```diff
@@ -0,0 +1,57 @@
...
+            throw new IllegalStateException(
+                String.format("Unrecognized resource type %s", resource.getResourceType()));
+        }
+        break;
+      case AZURE:
+      default:
+        throw new IllegalStateException(
+            String.format("Unrecognized cloud platform %s", resource.getCloudPlatform()));
+    }
+    // Step 2: create the cloud resource via CRL
```
zloery, line 51: nit: step 2 is above this comment

src/main/java/bio/terra/workspace/service/workspace/flight/CreateControlledResourceFlight.java
```diff
@@ -0,0 +1,57 @@
+package bio.terra.workspace.service.workspace.flight;
```
zloery, line 1: I think this flight and its steps should be in `bio.terra.workspace.service.resource...`, not the workspace package

src/main/java/bio/terra/workspace/service/workspace/flight/CreateSamResourceStep.java
```diff
@@ -0,0 +1,27 @@
+package bio.terra.workspace.service.workspace.flight;
+
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import bio.terra.workspace.service.iam.SamService;
+
+public class CreateSamResourceStep implements Step {
```
zloery, line 9: nit: add a todo with the appropriate ticket to fill this out

src/main/java/bio/terra/workspace/service/workspace/flight/FlightMapKey.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.workspace.service.workspace.flight;
+
+public interface FlightMapKey {
+  String getKey();
+
+  Class<?> getKlass();
+
+  //  default <T> T getFromMap(FlightMap flightMap) {
```
zloery, line 8: remove comment?

src/main/java/bio/terra/workspace/service/workspace/flight/FlightMapKey.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.workspace.service.workspace.flight;
+
+public interface FlightMapKey {
```
zloery, line 3: Where is this interface used/going to be used?

src/main/resources/api/service_openapi.yaml
```diff
@@ -1153,6 +1194,192 @@ components:
...
+      type: object
+      properties:
+        common:
+          $ref: '#/components/schemas/ControlledResourceCommonFields'
+        googleBucket:
+          $ref: '#/components/schemas/GoogleBucketCreationParameters'
+
+    ControlledResourceCommonFields:
+      type: object
+      required: [ name, cloningInstructions, jobControl, accessScope ]
```
zloery, line 1208: At least currently, the async spec expects `jobControl` to be a top-level field in the request. I don't like pushing it down another layer, even if that means a duplicate field on more async request bodies.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1153,6 +1194,192 @@ components:
...
+          description: A valid bucket name per https://cloud.google.com/storage/docs/naming-buckets.
+          type: string
+        location:
+          description: A valid bucket location per https://cloud.google.com/storage/docs/locations.
+          type: string
+        defaultStorageClass:
+          $ref: '#/components/schemas/GoogleBucketDefaultStorageClass'
+        lifecycle:
+          $ref: '#/components/schemas/GoogleBucketLifecycle'
+    GoogleBucketStoredAttributes: # todo: use bucket UID type here
```
zloery, line 1272: is this todo tracked in a ticket?

zloery (commented):
Overall this looks good, I have a bunch of nits but one larger question: do we actually need `ControlledResourceDbModel`, or can we just pass a `ControlledResource` to the ControlledResourceDAO along with a resource ID? 
Likewise, we could pass a `WsmResource` and ID to the ReferenceDao, though that's out of scope for this PR and would require a bit of jiggling for resource types.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateControlledResourceFlight.java
jaycarlton: good point

src/main/java/bio/terra/workspace/service/workspace/flight/FlightMapKey.java
jaycarlton: It's not anymore. Deleted. It was almost a good idea, but generics aren't quite smart enough for it to work.

src/main/java/bio/terra/workspace/service/workspace/flight/GenerateControlledResourceIdStep.java
jaycarlton: Discussed offline, and this isn't necessary, as the sequence isn't restarted in the middle.

src/main/java/bio/terra/workspace/service/workspace/flight/StoreControlledResourceMetadataStep.java
jaycarlton: Good catch!

src/main/resources/api/service_openapi.yaml
jaycarlton: @ddietterich do you have a preference here? I know we wanted to bundle all the common properties we could.

src/main/resources/api/service_openapi.yaml
jaycarlton: I don't think I want to do this anymore. It locks us into only exposing the fields on that (unrelated) type, currently just name, or breaking the API to add another.

src/test/java/bio/terra/workspace/common/fixtures/ControlledResourceFixtures.java
jaycarlton: It was going to be a .it domain, but for Italy you need paperwork saying you're an Italian business.

jaycarlton (commented):
> Overall this looks good, I have a bunch of nits but one larger question: do we actually need `ControlledResourceDbModel`, or can we just pass a `ControlledResource` to the ControlledResourceDAO along with a resource ID?
I like having a tight type to bind to each table in the DB. We can't, for example, create a whole WsmResource just from one query, so the intermediate type is at least needed there I think.

> Likewise, we could pass a `WsmResource` and ID to the ReferenceDao, though that's out of scope for this PR and would require a bit of jiggling for resource types.
From a Law of Demeter standpoint, I don't want to pass a huge class into a tiny method; I like the extra immutable-ish POJO type for each table.

src/main/java/bio/terra/workspace/service/resource/WsmResource.java
```diff
@@ -0,0 +1,100 @@
...
+   * Validate the state of to this object. Subclasses should override this method, calling super()
+   * first to validate parent class properties (even if those are abstract). This will prevent
+   * different resource type concrete classes from repeating the same validation code.
+   */
+  public void validate() {
+    if (Strings.isNullOrEmpty(getName())
+        || getWorkspaceId() == null
+        || getCloningInstructions() == null
+        || getStewardshipType() == null
+        || Strings.isNullOrEmpty(getOwner())
```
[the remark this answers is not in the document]
zloery: Alternatively you could push this down to ControlledResource, since it only goes in the controlled resource table. then again, I'm not exactly sure where this ends up after schema churn

src/main/resources/api/service_openapi.yaml
zloery: Agreed that replacing `GoogleBucketStoredAttributes` with the UID makes for a fragile interface, but you could make the `bucketName` property a `GoogleBucketUid` object instead of a string

src/main/java/bio/terra/workspace/service/resource/controlled/WsmResourceType.java
ddietterich: We should maintain the cloud type enum. That is useful on its own and the revised schema needs that. I suggest we put the cloud type as a field on the resourceType enum so we can carry it as a single thing.

src/main/java/bio/terra/workspace/db/ControlledResourceDao.java
```diff
@@ -0,0 +1,88 @@
...
+public class ControlledResourceDao {
+  private final Logger logger = LoggerFactory.getLogger(ControlledResourceDao.class);
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+
+  @Autowired
+  public ControlledResourceDao(NamedParameterJdbcTemplate jdbcTemplate) {
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  public void createControlledResource(ControlledResourceDbModel controlledResource) {
```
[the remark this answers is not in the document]
ddietterich: We should only have @Transactional on the top-most public method that runs the transaction. We should avoid coding cases where we share a method both as a top-level transaction method and a subroutine of another top-level transaction method.

Update transactions should be serializable, unless we find a reason not to do that. Read transactions should be marked read-only.

I know @gjuggler has a doc coming out on this.

The write transactions need to be retried. I'll make a ticket to add retrying. In fact, the retrying code from TDR would be a good candidate for common library.

src/main/resources/api/service_openapi.yaml
ddietterich: I think it is OK to have the JobControl one level down. My intent in the async spec was to provide an example, but not dictate the request structure.

@zloery Can you say more about why you would prefer it at the top level?

src/main/java/bio/terra/workspace/app/controller/ControlledGoogleResourceApiController.java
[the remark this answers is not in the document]
ddietterich: I think we discussed this off-thread and agreed to return the full object in these cases. I'm going to update the Async spec.

ddietterich:
I may have lost the thread, but I wanted to understand the answer to @zloery question about the need for ControlledResourceDbModel. Looking again at the code, it seems the DAO could call the `setJsonAttributes()` method itself and get the information it needs without that class.

src/main/java/bio/terra/workspace/service/resource/controlled/WsmResourceType.java
jaycarlton: I like that solution. I didn't make an internal cloud platform enum yet; it's just using the API one. Is that something we need?

src/main/java/bio/terra/workspace/db/ControlledResourceDao.java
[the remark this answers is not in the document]
jaycarlton: Maybe having two daos here is a bad abstraction? If we want the tables to work as one, then maybe we need to merge the create methods into one dao, with one transaction.

src/main/resources/api/service_openapi.yaml
jaycarlton: I suspect we'll revisit this structure a bit after the schema migration.

src/main/resources/api/service_openapi.yaml
jaycarlton: That doesn't buy me any type safety or useful abstraction, and just puts the bucket name one level down under a nonstandard "uid" property. I don't feel like it's enough to warrant reuse on its own.

jaycarlton:
> I may have lost the thread, but I wanted to understand the answer to @zloery question about the need for ControlledResourceDbModel. Looking again at the code, it seems the DAO could call the `setJsonAttributes()` method itself and get the information it needs without that class.

Maybe, but on retrieve, it's nice to have a DTO class that's 1:1 with the table. That said, I'm going to need to extend it to include fields from the other table.

How about we revisit this after the schema change?

ddietterich (approved):
I am at that point where I think we need to get this to pass tests and merge it. We can make incremental fixes from there.

zloery (approved):
LGTM. I'd like to revisit the need for `ControlledResourceDbModel`, but agreed that it's probably worth waiting until after the schema changes settle to do so

PR #221, opened 2021-02-08
Add reference descriptions and make them and names updateable [AS-512]

zarsky-broad:
Once this goes through, I'll publish an updated client and update Rawls.

One note: `description` is a protected key for flights, so I've standardized on `referenceDescription` so the name can be the same everywhere.


src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -135,6 +137,45 @@ public boolean isControlled(UUID workspaceId, UUID referenceId) {
...
+            ", ",
+            "UPDATE workspace_data_reference SET ",
+            " WHERE reference_id = :id AND workspace_id = :workspace_id");
+
+    if (name != null) {
+      updateStatement.add("name = :name");
+    }
+    if (referenceDescription != null) {
+      updateStatement.add("reference_description = :reference_description");
+    }
```
davidangb, line 160: if name and referenceDescription are both null, this should not go to the database at all ... especially since it will generate invalid SQL `UPDATE workspace_data_reference SET  WHERE reference_id = :id AND workspace_id = :workspace_id`

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -135,6 +137,45 @@ public boolean isControlled(UUID workspaceId, UUID referenceId) {
...
+            .addValue("id", referenceId.toString())
+            .addValue("workspace_id", workspaceId.toString())
+            .addValue("name", name)
+            .addValue("reference_description", referenceDescription);
+
+    StringJoiner updateStatement =
+        new StringJoiner(
+            ", ",
+            "UPDATE workspace_data_reference SET ",
+            " WHERE reference_id = :id AND workspace_id = :workspace_id");
```
davidangb, line 153: using a `StringJoiner` here gets the job done and is a clever choice ... but it feels less readable imho than a `StringBuilder`. I would prefer readability ... but I also won't die on this hill.

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
```diff
@@ -128,6 +132,24 @@ public DataReference createDataReference(
...
+      AuthenticatedUserRequest userReq) {
+    workspaceService.validateWorkspaceAndAction(
+        userReq, workspaceId, SamConstants.SAM_WORKSPACE_WRITE_ACTION);
+
+    if (!dataReferenceDao.updateDataReference(
+        workspaceId,
+        referenceId,
+        updateRequest.getName(),
+        updateRequest.getReferenceDescription())) {
+      throw new DataReferenceNotFoundException("Data Reference not found.");
```
davidangb, line 149: I think you can also reach here if both name and reference are null; `DataReferenceNotFoundException` is misleading in that case.

src/test/java/bio/terra/workspace/db/DataReferenceDaoTest.java
```diff
@@ -113,6 +115,50 @@ public void verifyGetDataReferenceNotInWorkspaceNotFound() {
...
+    DataReference updatedDescriptionReference = currentReference.get();
+    assertThat(updatedDescriptionReference.name(), equalTo(updatedNameReference1.name()));
+    assertThat(updatedDescriptionReference.referenceDescription(), equalTo(updatedDescription));
+
+    // Update both
+    updateReference.apply(updatedName2, updatedDescription2);
+    DataReference updatedBothReference = currentReference.get();
+    assertThat(updatedBothReference.name(), equalTo(updatedName2));
+    assertThat(updatedBothReference.referenceDescription(), equalTo(updatedDescription2));
+  }
```
davidangb, line 160: add a test case for updating when both name and description are null.

Ideally, all of the cases in this method would be distinct tests - `verifyUpdateDataReferenceName`, `verifyUpdateDataReferenceDescription`, `verifyUpdateDataReferenceNameAndDescription`, etc - so we can see individually which cases pass and which fail, if any fail.

src/test/java/bio/terra/workspace/service/datareference/DataReferenceServiceTest.java
```diff
@@ -130,6 +133,52 @@ public void enumerateFailsUnauthorized() {
...
+
+    // Update both
+    updateReference.accept(
+        new UpdateDataReferenceRequestBody()
+            .name(updatedName2)
+            .referenceDescription(updatedDescription2));
+    DataReference updatedBothReference = currentReference.get();
+    assertThat(updatedBothReference.name(), equalTo(updatedName2));
+    assertThat(updatedBothReference.referenceDescription(), equalTo(updatedDescription2));
+  }
```
davidangb, line 180: see previous comment - add case for both null, consider breaking out individual tests

src/main/resources/api/service_openapi.yaml
```diff
@@ -158,6 +158,25 @@ paths:
...
+              $ref: '#/components/schemas/UpdateDataReferenceRequestBody'
+      responses:
+        '204':
+          description: OK
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '500':
+          $ref: '#/components/responses/ServerError'
```
davidangb, line 179: This satisfies the current Jira ticket, but from an API design perspective it's awkward that we can create and get entire data references, but we can only edit the name/description of those references - we can't edit the innards, such as pointing an existing reference at a different snapshot id. I'm not sure the best way to handle this - possibly just explaining it in the API description.

davidangb (commented):
first-pass comments inline!

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
```diff
@@ -135,6 +137,45 @@ public boolean isControlled(UUID workspaceId, UUID referenceId) {
...
+      updateStatement.add("name = :name");
+    }
+    if (referenceDescription != null) {
+      updateStatement.add("reference_description = :reference_description");
+    }
+
+    int rowsAffected = jdbcTemplate.update(updateStatement.toString(), params);
+    boolean updated = rowsAffected > 0;
+
+    if (updated)
```
ddietterich, line 165: nit: I think this can be debug logging and can just report the updated or not flag rather than all of this.

ddietterich (commented):
Comment, suggestions, and nits

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -209,6 +210,29 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+      @PathVariable("referenceId") UUID referenceId,
+      @RequestBody UpdateDataReferenceRequestBody body) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+
+    if (body.getName() != null) DataReferenceValidationUtils.validateReferenceName(body.getName());
+
+    logger.info(
+        String.format(
+            "Updating data reference by id %s in workspace %s for %s with body %s",
+            referenceId.toString(), id.toString(), userReq.getEmail(), body.toString()));
```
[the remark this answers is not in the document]
zarsky-broad: This is copied from the other methods here, so I'd want to update them all for consistency, I'm inclined to leave this one as-is and then update all of them in a follow-on.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
zarsky-broad: I liked the `StringJoiner` because it's extensible for the future, if we ever have other updatable properties.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
zarsky-broad: I'm going to add a guard for that at the controller level.

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
zarsky-broad: This is copied from the delete endpoint below, so I'm inclined to leave it as-is for now and change both in a follow-on.

src/main/java/bio/terra/workspace/service/datareference/DataReferenceService.java
zarsky-broad: Adding a guard at the controller level.

build.gradle
```diff
@@ -111,6 +111,8 @@ dependencies {
 
 	annotationProcessor group: "com.google.auto.value", name: "auto-value", version: "1.7.4"
 	annotationProcessor group: "org.springframework.boot", name: "spring-boot-configuration-processor"
+
+	developmentOnly group: "org.springframework.boot", name: "spring-boot-devtools"
```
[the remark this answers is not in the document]
zarsky-broad: Ah, I was experimenting, I didn't, I'll pull it back out

src/main/java/bio/terra/workspace/db/DataReferenceDao.java
zarsky-broad: (plus, using `StringBuilder` would mean doing a whole nested case thing to add the comma...

ddietterich (approved):
LGTM

davidangb (approved):
thanks for the changes!

PR #222, opened 2021-02-10
Tidy up [post- AS-512]

zarsky-broad:
- removed `String.format` nested in a logger
- fixed a javadoc tag (`@Code` -> `@code`)
- normalized tests
  - removed `public`, matching JUnit docs
  - removed some unthrown exceptions
  - also added a bit of nesting, so **better reviewed with whitespace hidden**


davidangb:
code changes look ok; I see that some tests are failing, don't know if those are due to PR changes or not!

davidangb (approved):

zloery (approved):

PR #224, opened 2021-02-12
[PF-450] Switch WSM to use bio.terra.common.logging

gjuggler:
This PR switches WSM to use the logging setup that's been extracted into terra-common-lib.


src/main/java/bio/terra/workspace/app/configuration/external/TracingConfiguration.java
```diff
@@ -21,11 +23,14 @@ public void setProbability(Double probability) {
     this.probability = probability;
   }
 
-  public String getProjectId() {
```
gjuggler, line 24: This is a small drive-by change: I don't think we need to be manually specifying a tracing project ID here. We should use the default project ID (from app default credentials) for tracing, and Google provides reasonable helper methods to extract this.

ddietterich: Is this always supposed to be the project where the service is running?

ddietterich: That is, TDR and WSM work across many projects. We don't expect this ID to change for that. If we want to keep track of the project we are working on, we log that separately. Yes?

gjuggler: Yes – this line was only configuring the target Google Project that OpenCensus would export service-level trace information to, which is (as far as I can tell) 100% of the time the Google Project that's hosting the k8s cluster.

src/main/java/bio/terra/workspace/service/buffer/BufferService.java
```diff
@@ -40,9 +38,6 @@ public BufferService(BufferServiceConfiguration bufferServiceConfiguration) {
 
   private ApiClient getApiClient(String accessToken) {
     ApiClient client = new ApiClient();
-    client.addDefaultHeader(
```
gjuggler, line 43: I'm proposing we remove these one-off header additions for now. This should come back as part of a more comprehensive tracing & MDC propagation story, which we are poised to tackle with the common-lib tracing / monitoring module.

ddietterich (approved):
lgtm

zloery (approved):

PR #225, opened 2021-02-16
[PF-387] Explicitly track Sam workspace resource ownership

zloery:
This change adds a new field to our workspace representation indicating whether WSM owns the corresponding Sam resource or not. This is specified at creation time (defaulting to "not owned", Rawls' behavior for compatibility) and is used during workspace deletion to know whether we need to delete the Sam resource or not. As this is a temporary field used for Rawls integration, it is not returned on calls to retrieve a workspace.

I'm also looking for more opinions on whether this change is beneficial or not. It adds an additional integration wart that will eventually need to be removed, but it also helps prevent corrupt resource layouts. Specifically, we currently assume Rawls will always delete its workspace Sam resource before calling delete workspace on WSM. If this assumption is violated, WSM will delete a Rawls-managed Sam resource, which could lead to unexpected behavior. This assumption holds for the current Rawls implementation, but we may not want to rely on this assumption.

### DIFF
--- a/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -90,6 +90,8 @@ public class WorkspaceApiController implements WorkspaceApi {
     // RAWLS_WORKSPACE.
     WorkspaceStageModel requestStage = body.getStage();
     requestStage = (requestStage == null ? WorkspaceStageModel.RAWLS_WORKSPACE : requestStage);
+    boolean isSamResourceOwner =
+        (body.isSamResourceOwner() == null ? false : body.isSamResourceOwner());
     WorkspaceStage internalStage = WorkspaceStage.fromApiModel(requestStage);
     Optional<SpendProfileId> spendProfileId =
         Optional.ofNullable(body.getSpendProfile()).map(SpendProfileId::create);
@@ -102,6 +104,7 @@ public class WorkspaceApiController implements WorkspaceApi {
             .jobId(jobId)
             .spendProfileId(spendProfileId)
             .workspaceStage(internalStage)
+            .isSamResourceOwner(isSamResourceOwner)
             .build();
     UUID createdId = workspaceService.createWorkspace(internalRequest, userReq);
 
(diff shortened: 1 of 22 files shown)


zloery:
Client tests pass with a locally-published client jar, but won't pass on GHA without a published version. I'll publish the library and update the test dependency if we decide to go forward with this.

davidangb:
how is this new column different from the Rawls vs. MC Terra workspace stage? Could you just inspect the stage instead of using this new column?

You did say "This assumption holds for the current Rawls implementation, but we may not want to rely on this assumption" - which is valid, but I think until/unless we find a case in which the assumption is false, this new column could be a premature optimization.

zloery:
My biggest concern was situations like tests or calls from swagger-ui which default to the RAWLS_WORKSPACE stage, but aren't actually owning a Sam resource and would stop working. I think there are also a number of existing RAWLS_WORKSPACEs from tests/swagger-ui that would end up orphaned.

That said... having written this out, I've already changed our test calls and existing ones would still hit the same issue. Littering in dev Sam might be acceptable here.

I'm also thinking of this as less of an optimization and more of a risk mitigation - the case I'm concerned about is someone changing Rawls' behavior without knowing the (complicated) connection it has with WSM here. But it's a fair point that we don't expect to make changes like that at the moment

ddietterich:
The reason I raised this in the first place is because when I reviewed the code it looked fragile. If there were a software bug, we would not detect it during this delete process. So I think of this as a software quality issue.

Essentially, we need the feature flag (or its surrogate) as a parameter to create workspace so it is reliably set. Alternately, we can make a separate endpoint for the "stage" workspace creation.

zloery:
Thanks for the comments - after more discussion and trying out using workspace stage instead of this new field, I think that's the better approach. Given how different it the approach is, I've opened #233 instead.

Poking around in this space more though, I also realized the current approach leaves behind zombie rows every time Rawls deletes a workspace that's backed by WSM. This shouldn't cause problems beyond littering (as it doesn't affect MC_WORKSPACE workspaces, and Rawls always generates random workspace IDs) but it's worth cleaning up while we can. I added more detail to the PR description and we can discuss more as needed.

PR #226, opened 2021-02-16
rename referenceDescription to description [AS-512]

zarsky-broad:
Also ensures that an omitted description is a blank string instead of null.

Once this is good to go I'll publish the new client.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/datareference/flight/DataReferenceFlightMapKeys.java
+++ b/src/main/java/bio/terra/workspace/service/datareference/flight/DataReferenceFlightMapKeys.java
@@ -4,7 +4,7 @@ public final class DataReferenceFlightMapKeys {
   public static final String REFERENCE_ID = "referenceId";
   public static final String WORKSPACE_ID = "workspaceId";
   public static final String NAME = "name";
-  public static final String REFERENCE_DESCRIPTION = "referenceDescription";
+  public static final String DESCRIPTION = "WSM:description";
   public static final String REFERENCE_TYPE = "referenceType";
   public static final String CLONING_INSTRUCTIONS = "cloningInstructions";
   public static final String REFERENCE_OBJECT = "referenceObject";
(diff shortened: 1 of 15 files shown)


src/main/java/bio/terra/workspace/service/datareference/flight/DataReferenceFlightMapKeys.java
```diff
@@ -4,7 +4,7 @@
   public static final String REFERENCE_ID = "referenceId";
   public static final String WORKSPACE_ID = "workspaceId";
   public static final String NAME = "name";
-  public static final String REFERENCE_DESCRIPTION = "referenceDescription";
+  public static final String DESCRIPTION = "WSM:description";
```
davidangb, line 7: nit: I'd leave this as "referenceDescription"; clarifying that this is "WSM" doesn't describe much. I don't feel very strongly about this.

src/main/resources/api/service_openapi.yaml
```diff
@@ -924,7 +924,7 @@ components:
       properties:
         name:
           $ref: "#/components/schemas/Name"
-        referenceDescription:
+        description:
           type: string
```
davidangb, line 928: can you document here that the default value is the empty string? (does openapi allow that?)

src/main/java/bio/terra/workspace/service/datareference/flight/DataReferenceFlightMapKeys.java
zarsky-broad: Yeah, this is all because we can't use `description` as a key, I'll add a clarifying comment.

zarsky-broad: (I think the prefix with the colon makes it clearer that it's just namespace-ish)

ansingh7115 (approved):
LGTM

davidangb (approved):

src/main/java/bio/terra/workspace/service/datareference/flight/DataReferenceFlightMapKeys.java
ddietterich: The root problem here is that we do not pass the DataReferenceRequest into the flight as the request, but (pretty needlessly) break it into little pieces.

As I understand it, we got here by:
* AutoValue is good - use that
* AutoValue class is a pain to use as a Stairway input parameter
* Dismantle class into its atoms

Anyway, not for this PR. I expect to change the usage of AutoValue classes one way or the other so we do not have to dismantle them to launch a flight.

PR #228, opened 2021-02-19
[PF-276] Change to use common lib liquibase migrate code

ddietterich:
- Remove the WSM migrate service
- Added in the common migrate service; requires adding to the component scan
- Add documentation to describe running the connected tests; please review
- Made unitTest the default test, so that the build target succeeds consistently, instead of failing consistently


yonghaoy (approved):

DEVELOPMENT.md
```diff
@@ -78,6 +78,12 @@ To run unit tests:
 ./gradlew unitTest
 ```
   
+To run connected tests:
+**DO YOU HAVE TO RUN render_config.sh for connected tests??**
```
zloery, line 82: yes, you need to run render_config to get the credentials needed to run these tests

zloery (approved):

DEVELOPMENT.md
jaycarlton: Can we wrap that step into a dependent gradle task so we don't have to remember?

jaycarlton (approved):

PR #231, opened 2021-02-23
[No ticket] Always upload test results on failure

zloery:
No ticket because this only affects test infrastructure.

Currently, when a test run fails the GHA workflow will skip steps after "run tests", and so the test report is not uploaded. We should always make the test report available to make debugging failed tests easier.


ansingh7115 (approved):

zarsky-broad (approved):
This might be better than the pattern I've used in other services, where I ignore the failure, upload, and then manually set a failed status...

PR #232, opened 2021-02-24
[PF-415] Step to create GCS Bucket

jaycarlton:
Next step of the flight is to create a GCS bucket.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateControlledResourceFlight.java
+++ b/src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateControlledResourceFlight.java
@@ -33,14 +33,16 @@ public class CreateControlledResourceFlight extends Flight {
     // create the cloud resource via CRL
     final ControlledResource resource =
         inputParameters.get(JobMapKeys.REQUEST.getKeyName(), ControlledResource.class);
-    final AuthenticatedUserRequest userRequest = inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    final AuthenticatedUserRequest userRequest =
+        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
 
     switch (resource.getResourceType()) {
       case GCS_BUCKET:
-        addStep(new CreateGcsBucketStep(
-            flightBeanBag.getCrlService(),
-            (ControlledGcsBucketResource) resource,
-            userRequest));
+        addStep(
+            new CreateGcsBucketStep(
+                flightBeanBag.getCrlService(),
+                (ControlledGcsBucketResource) resource,
+                userRequest));
         break;
       case BIGQUERY_DATASET:
       default:
--- a/src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateControlledResourceFlight.java
+++ b/src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateControlledResourceFlight.java
@@ -4,8 +4,10 @@ import bio.terra.stairway.Flight;
 import bio.terra.stairway.FlightMap;
 import bio.terra.workspace.common.utils.FlightBeanBag;
 import bio.terra.workspace.service.datareference.flight.GenerateReferenceIdStep;
+import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.job.JobMapKeys;
 import bio.terra.workspace.service.resource.controlled.ControlledResource;
+import bio.terra.workspace.service.resource.controlled.gcp.ControlledGcsBucketResource;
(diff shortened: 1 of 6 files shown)


src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateGcsBucketStep.java
```diff
@@ -1,28 +1,136 @@
...
+        default:
+          throw new IllegalStateException(
+              "Unrecognized lifecycle action type " + lifecycleRuleAction.getType());
+      }
+    }
+
+    private static LifecycleCondition toGcsApi(GoogleBucketLifecycleRuleCondition condition) {
+      final LifecycleCondition.Builder resultBuilder = LifecycleCondition.newBuilder();
+
+      // TODO: some conditions aren't on the Google api object
```
ddietterich, line 115: Not sure what that means. Is there more work to do in this PR? If you are suggesting to do it later (which is fine), then maybe make a Jira ticket and put the ticket number on the TODO. I want to keep track of the loose ends.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateGcsBucketStep.java
```diff
@@ -1,28 +1,136 @@
...
+      Optional.ofNullable(condition.getAge()).ifPresent(resultBuilder::setAge);
```
ddietterich, line 116: Reverse polish assignment! OMG!!

I don't think you have to do so much work.
```
resultBuilder.setAge(condition.getAge())
```
works just fine.

ddietterich (approved):
LGTM

Please check my comment on Optional use for the Lifecycle stuff. It seems like overkill to me.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateControlledResourceFlight.java
```diff
@@ -31,10 +33,16 @@ public CreateControlledResourceFlight(FlightMap inputParameters, Object applicat
...
+    final AuthenticatedUserRequest userRequest =
+        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
 
     switch (resource.getResourceType()) {
       case GCS_BUCKET:
-        addStep(new CreateGcsBucketStep(flightBeanBag.getCrlService()));
+        addStep(
+            new CreateGcsBucketStep(
+                flightBeanBag.getCrlService(),
+                (ControlledGcsBucketResource) resource,
```
tnadler, line 44: (nit) any reason not to do this cast above where the variable is defined? And how are you confident it will succeed?

tnadler (approved):

src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateControlledResourceFlight.java
jaycarlton: Above we don't know which resource type we have. We check that in the switch, so I know this will succeed on this branch.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateGcsBucketStep.java
jaycarlton: It's natural once you get used to it. The alternative won't work if `setAge()` assumes a non-null argument. Probably not an issue in this case, assuming no side effects happen when setting this value.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateGcsBucketStep.java
jaycarlton: The list documented contained presumably obsolete conditions that don't seem to be available in the Google API for new buckets.

I'll check into this this afternoon.

jaycarlton: Turns out we need a new version of the storage library, meaning a new CRL release. It turns out there may be breakage in the new version, if the tests are to be believed, so I'm pulling this work into its own ticket so it doesn't block this thread.

jaycarlton: Work to update CRL and fill in the rest of these is [PF-506](https://broadworkbench.atlassian.net/browse/PF-506).

src/main/java/bio/terra/workspace/service/resource/controlled/flight/CreateGcsBucketStep.java
jaycarlton: I just remembered the situation where this was necessary. For a BigQuery insert object, I had to make a map of all the non-null columns, as the map didn't support null. So this pattern was a good fit for that. I checked and this builder doesn't care if it gets nulls.

PR #234, opened 2021-02-25
AS-523: Ignore buffer tests on non-dev environments 

ansingh7115:


src/test/java/bio/terra/workspace/common/BaseConnectedTest.java
```diff
@@ -11,4 +11,7 @@
  */
 @Tag("connected")
 @ActiveProfiles("connected-test")
-public class BaseConnectedTest extends BaseTest {}
+public class BaseConnectedTest extends BaseTest {
+
+  public static final String bufferServiceDisabledEnvsRegEx = "alpha|staging|prod";
```
zloery, line 16: Do we run tests in prod?

zloery, line 16: iiuc we'll remove this env-based disabling as we deploy buffer service to other environments. 
I realize it's a lot of overhead, but for the sake of this not getting lost could you create a ticket for removing this and add a todo here pointing to that ticket?

zloery (approved):
LGTM with nits

build.gradle
```diff
@@ -203,6 +203,10 @@ task unitTest(type: Test) {
 }
 
 task connectedTest(type: Test) {
+	environment = [
+			'TEST_ENV': System.getenv("TEST_ENV"),
+			'GOOGLE_APPLICATION_CREDENTIALS': "${googleCredentialsFile}"
```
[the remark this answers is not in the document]
ansingh7115: When I remove that it gives the following error:

```
Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
```

src/test/java/bio/terra/workspace/common/BaseConnectedTest.java
ansingh7115: No - I decided to put that there just in case

ansingh7115: Is it alright to put that in the PF board? Since you folks will be the ones working on buffer service?

zloery: sgtm, just made PF-507 for this

build.gradle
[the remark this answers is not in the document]
gjuggler: Ah, what's probably happening is that L182 is reassigning the entire "environment" variable, which is a Map<String, Object> ([docs](https://docs.gradle.org/3.3/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:environment)).

Could you try changing that reassignment into a map addition? Maybe then the TEST_ENV will propagate without any explicit effort on our part...

L182:
```
environment.put('GOOGLE_APPLICATION_CREDENTIALS', "${googleCredentialsFile}");
```

gjuggler (approved):
If you're willing to give it a try, I proposed a possible simplification to the env variable propagation. If that doesn't work, going ahead w/ the current solution seems fine (we can pursue a cleanup in another ticket).

src/test/java/bio/terra/workspace/common/BaseConnectedTest.java
ansingh7115: Thank you!

build.gradle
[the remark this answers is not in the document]
ansingh7115: Adding it there didn't work - it was still being overwritten, but adding `environment.put('TEST_ENV'...)` to the connected and integration tasks worked!

gjuggler (approved):

PR #235, opened 2021-02-25
[PF-410] first cut at get bucket

jaycarlton:
This is just a database call, so we don't need a flight.

### DIFF
--- a/src/main/java/bio/terra/workspace/common/exception/ControlledResourceNotFoundException.java
+++ b/src/main/java/bio/terra/workspace/common/exception/ControlledResourceNotFoundException.java
@@ -1,0 +1,26 @@
+package bio.terra.workspace.common.exception;
+
+import java.util.List;
+
+public class ControlledResourceNotFoundException extends NotFoundException {
+
+  public ControlledResourceNotFoundException(String message) {
+    super(message);
+  }
+
+  public ControlledResourceNotFoundException(String message, Throwable cause) {
+    super(message, cause);
+  }
+
+  public ControlledResourceNotFoundException(Throwable cause) {
+    super(cause);
+  }
+
+  public ControlledResourceNotFoundException(String message, List<String> causes) {
+    super(message, causes);
+  }
+
+  public ControlledResourceNotFoundException(String message, Throwable cause, List<String> causes) {
+    super(message, cause, causes);
+  }
+}
--- a/src/main/java/bio/terra/workspace/service/resource/controlled/gcp/ControlledGcpResourceService.java
+++ b/src/main/java/bio/terra/workspace/service/resource/controlled/gcp/ControlledGcpResourceService.java
@@ -1,12 +1,18 @@
 package bio.terra.workspace.service.resource.controlled.gcp;
 
+import bio.terra.workspace.common.exception.ControlledResourceNotFoundException;
+import bio.terra.workspace.db.ControlledResourceDao;
+import bio.terra.workspace.generated.model.GoogleBucketStoredAttributes;
 import bio.terra.workspace.generated.model.JobControl;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.job.JobBuilder;
 import bio.terra.workspace.service.job.JobService;
+import bio.terra.workspace.service.resource.controlled.ControlledResourceDbModel;
 import bio.terra.workspace.service.resource.controlled.flight.CreateControlledResourceFlight;
 import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys;
 import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.ControlledResourceKeys;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import java.util.UUID;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
@@ -15,10 +21,13 @@ import org.springframework.stereotype.Component;
 public class ControlledGcpResourceService {
 
   private final JobService jobService;
+  private final ControlledResourceDao controlledResourceDao;
 
   @Autowired
-  public ControlledGcpResourceService(JobService jobService) {
+  public ControlledGcpResourceService(JobService jobService,
+      ControlledResourceDao controlledResourceDao) {
     this.jobService = jobService;
+    this.controlledResourceDao = controlledResourceDao;
   }
 
   public String createGcsBucket(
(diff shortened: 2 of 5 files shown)


src/main/java/bio/terra/workspace/common/exception/ControlledResourceNotFoundException.java
```diff
@@ -0,0 +1,26 @@
+package bio.terra.workspace.common.exception;
```
zloery, line 1: I'm not sure we need to separate `ControlledResourceNotFoundException` from `DataReferenceNotFoundException`. Thoughts on just renaming the existing exception to `ResourceNotFoundException` and using that?

src/main/java/bio/terra/workspace/service/resource/controlled/gcp/ControlledGcpResourceService.java
```diff
@@ -37,4 +46,16 @@ public String createGcsBucket(
...
     return jobBuilder.submit();
   }
+
+  /**
+   * Retrieve the bucket metadata from the DAO(s)
+   * @param workspaceId - workspace containing this bucket
+   * @param resourceId - resource ID for the bucket
+   * @return attributes we store about this bucket (i.e. its name)
+   */
+  public GoogleBucketStoredAttributes getBucket(UUID workspaceId, UUID resourceId) {
```
zloery, line 56: This needs an authz check like `validateWorkspaceAndAction` from workspace service, otherwise it's open to everyone instead of just workspace readers.

zloery (commented):
Can you add tests for the new endpoint?

src/main/java/bio/terra/workspace/common/exception/ControlledResourceNotFoundException.java
jaycarlton: That works, if we're settled on resource as the top-level name.

src/main/java/bio/terra/workspace/service/resource/controlled/gcp/ControlledGcpResourceService.java
jaycarlton: Yeah, spoke with @ddietterich a bit on that. I got the impression we were waiting on something to do this.

```
We do need to check permissions.
10:01
But, we still need the new workspace model to do it right. Sigh…
10:02
That is, the new Sam model for workspace permissions that Zach is doing
```

zloery: Ah good point - we should really check with Sam whether the user has read permission on the specified resource, not general read access to the workspace. Can you add a todo pointing to PF-413?

jaycarlton:
> Can you add tests for the new endpoint?

For a real test, I'd like the ability to create and delete the bucket, and we don't have that capability. A unit test seemed to be more of an exercise in mocking without much to assert but I'm open to suggestions.

src/main/resources/api/service_openapi.yaml
```diff
@@ -702,7 +716,16 @@ components:
...
       in: path
-      description: A UUID used to identify a data reference in a workspace
+      description: A UUID used to identify a data reference in a workspace.
+      required: true
+      schema:
+        type: string
+        format: uuid
+
+    ResourceId:
+      name: referenceId
```
marikomedlock, line 726: resourceId?

src/main/resources/api/service_openapi.yaml
```diff
@@ -643,7 +643,21 @@ paths:
...
-
+  /api/workspaces/v1/{workspaceId}/resources/controlled/google/buckets/{resourceId}:
+    parameters:
+      - $ref: '#/components/parameters/WorkspaceId'
+      - $ref: '#/components/parameters/ResourceId'
+    get:
+      summary: get a controlled GCS bucket resource
+      operationId: getBucket
+      tags: [ControlledGoogleResource]
+      responses:
```
marikomedlock, line 654: Should there also be a 404 here?

src/main/java/bio/terra/workspace/service/resource/controlled/gcp/ControlledGcsBucketResource.java
```diff
@@ -104,4 +104,13 @@ public boolean equals(Object o) {
   public int hashCode() {
     return Objects.hash(super.hashCode());
   }
+
+  public static GoogleBucketStoredAttributes attributesToOutputApiModel(String jsonAttributes) {
+    try {
+      return OBJECT_MAPPER.readValue(jsonAttributes, GoogleBucketStoredAttributes.class);
```
marikomedlock, line 110: If I understand correctly, this method is for converting a json-formatted string field in the dao DB to a Java ApiModel object.
- Please add a method comment.
- It feels like this conversion should be in the dao class. Since the conversion from DB  row(s) to Java class (ControlledResourceDbModel) happens there -- is there a reason to pull out the conversion of this particular field?

src/main/resources/api/service_openapi.yaml
jaycarlton: Good catch. These are going to be merged in the new schema.

src/main/resources/api/service_openapi.yaml
jaycarlton: yep

src/main/java/bio/terra/workspace/service/resource/controlled/gcp/ControlledGcsBucketResource.java
jaycarlton: I put it in this class to be a bookend to the `getJsonAttributes()` instance method it has. The knowledge of this part of the conversion in that direction lives in this class already. 

There will be an opportunity to revisit this with the new schema changes.

src/main/resources/api/service_openapi.yaml
jaycarlton: @zloery I haven't been using the `@PathVariable` annotation in my controller. Should I? It's there for me in the generated parent class.

zloery: Given that the annotation is in the generated api class, I don't think we need to explicitly add it to the controller.
I'm pretty sure we manually copied it in as we were implementing controllers for the first generated api class and it's been copied + pasted since.

marikomedlock (approved):

zloery:
> > Can you add tests for the new endpoint?
> 
> For a real test, I'd like the ability to create and delete the bucket, and we don't have that capability. A unit test seemed to be more of an exercise in mocking without much to assert but I'm open to suggestions.

ah, good point. in that case, can you add an issue to the PF board so this doesn't get lost?

zloery (approved):

jaycarlton:
> > > Can you add tests for the new endpoint?
> > 
> > 
> > For a real test, I'd like the ability to create and delete the bucket, and we don't have that capability. A unit test seemed to be more of an exercise in mocking without much to assert but I'm open to suggestions.
> 
> ah, good point. in that case, can you add an issue to the PF board so this doesn't get lost?

Sure. Added https://broadworkbench.atlassian.net/browse/PF-512

PR #236, opened 2021-02-25
[PF-505] workspaceId instead of Id

jaycarlton:


davidangb (approved):
lgtm! I see that at least one test failed, I haven't tried to investigate those, just make sure those aren't "real" failures due to this PR.

jaycarlton:
The test failure is repeatable, and only on this branch apparently. I just did a fake PR using dev and it passed everything.

zarsky-broad (approved):
Fine with this, but only with a version bump, since this will require a change in the use of the swagger client lib.

zloery (commented):
I don't think this actually needs a version bump - the names of path variables aren't visible to clients, so nothing needs to change on their end.
The test failures are probably the controllers specifying `PathVariable("id")`, which also need to be changed. But that should also fail the client tests and those passed, so I'm gonna do a bit of digging

zarsky-broad:
> I don't think this actually needs a version bump - the names of path variables aren't visible to clients, so nothing needs to change on their end.
Ah, I was getting confused with `CreateWorkspaceRequestBody`, nvm

jaycarlton:
publiished new version just now

PR #237, opened 2021-02-25
Revert "[PF-450] Switch WSM to use bio.terra.common.logging"

gjuggler:
Reverts DataBiosphere/terra-workspace-manager#224

This is causing servlet exceptions for non-routed requests, like Swagger UI.


zloery (approved):
LGTM - I didn't review in depth as this is a clean rollback.

wchamber (approved):

PR #240, opened 2021-02-25
[PF-450] Switch WSM to use bio.terra.common.logging (roll-forward)

gjuggler:
This PR duplicates https://github.com/DataBiosphere/terra-workspace-manager/pull/224, but with a fix to the tracing interceptor to avoid 500 errors on static resource requests.

*The only change from the previously-merged PR is https://github.com/DataBiosphere/terra-workspace-manager/commit/18392c755c2854c25fc74db934349b39bafc5b4e.* This review should only cover this commit, since all other code was previously reviewed as part of #224.

The error that I'd introduced in #224 was that I'd removed the line that exited early for non-API requests ([code ref](https://github.com/DataBiosphere/terra-workspace-manager/pull/224/files#diff-a503b7bd09604bc3ab9fe0a5064ea1b6b09e468a34af0349f434b7a2b66d0ca0L57)).

I believe it was still correct to remove that early-exit — since OpenCensus is already adding spans for all inbound requests, we may as well annotate all spans in a consistent manner.

However, I'd left in the unconditional class cast, which only worked for methods handled by our Spring controllers (e.g. things under the /api/ prefix).

This PR adds a small fix to only add the "route" annotation when we have a HandlerMethod instance available. (Heads up @wchamber, I expect this, or its replacement, will eventually land in the common tracing module.)

*Side note*
One interesting wrinkle in identifying this bug was that our logs in dev WSM were quite vague. This is what was caught from stdout by the Cloud Logging agent:

Screen Shot 2021-02-25 at 4 58 14 PM

I was able to get a much more meaningful error message when looking at my personal env:

Screen Shot 2021-02-25 at 4 59 40 PM

I'm not quite sure why the dev logs were so heavily truncated.


zloery (approved):
LGTM. I'm surprised by the difference between dev and personal logs - I was under the impression logging should be the same between the two

ansingh7115 (approved):
Assuming that the changes in 18392c7 were manually tested to make sure the MDC Id is being set properly!

PR #241, opened 2021-02-25
[PF-306] Add nightly integration tests, make client tests run locally

zloery:
The current setup for TestRunner integration tests runs on commits to a PR and builds a local client, but runs against the `wsmtest` environment server, meaning server-side changes are not captured in these PR checks. They also don't run nightly like the perf tests, which we want.

This change switches the existing `test-runner-nightly-integration` action to run the nightly integration tests in the wsmtest environment. It also switches to using the currently specified version of the client library, rather than building a client with potentially unpublished changes from head. The tests will run at 7 AM UTC since the perf tests also use the `wsmtest` environment and run at 6 AM UTC.

Additionally, I added a new `test-runner-pr-integration` action. On push to a PR, this action will run client tests against a local server (with local changes) using the currently specified client version. It will **not** build a local client with unpublished changes. Publishing is currently manual, and letting the test pass with unpublished changes could lead to failing releases.
(Side note: we might want to consider automatically publishing new versions of the client library on merge to dev. I believe TDR does this already, so there may be shared actions we can use).

### DIFF
--- a/.github/workflows/test-runner-nightly-integration.yml
+++ b/.github/workflows/test-runner-nightly-integration.yml
@@ -6,25 +6,12 @@ name: Test Runner Nightly Integration Tests
 on:
   workflow_dispatch: {}
   schedule:
-  - cron: '0 6 * * *' # run at 6 AM UTC, 1 AM ET.
+  - cron: '0 7 * * *' # run at 7 AM UTC, 2 AM ET.
 
 jobs:
   test-runner-integration:
 
     runs-on: ubuntu-latest
-
-    services:
-      postgres:
-        image: postgres:12.3
-        env:
-          POSTGRES_PASSWORD: postgres
-        options: >-
-          --health-cmd pg_isready
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 5
-        ports:
-          - 5432:5432
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/test-runner-nightly-integration.yml
+++ b/.github/workflows/test-runner-nightly-integration.yml
@@ -4,10 +4,9 @@
 name: Test Runner Nightly Integration Tests
 
 on:
-  push:
-    branches: [ dev ]
-  pull_request:
-      branches: [ '**' ]
+  workflow_dispatch: {}
+  schedule:
+  - cron: '0 6 * * *' # run at 6 AM UTC, 1 AM ET.
 
 jobs:
   test-runner-integration:
@@ -29,11 +28,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      # This step is needed when running test using a local server.
-      - name: Initialize Postgres DB
-        env:
-          PGPASSWORD: postgres
-        run: psql -h 127.0.0.1 -U postgres -f ./local-dev/local-postgres-init.sql
       - name: Set up AdoptOpenJDK 11
         uses: joschi/setup-jdk@v2
         with:
@@ -49,8 +43,6 @@ jobs:
           restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
-
-      # These steps aren't needed for unit tests
       - name: Get Vault token
         id: vault-token-step
         env:
@@ -69,22 +61,6 @@ jobs:
           chmod +x render_config.sh
           ./render_config.sh dev ${{ steps.vault-token-step.outputs.vault-token }}
           chmod +r rendered
-
-      # Spin up a local server.
-      - name: Run local server (in background)
-        run: ./gradlew bootRun &
-      - name: Wait for local server to start
-        # With a timeout of 60 seconds, try to connect to localhost:8080 every 1 second.
-        run: timeout 60 bash -c 'until echo > /dev/tcp/localhost/8080; do sleep 1; done'
-
-      - name: Build the WSM client
-        run: |
-          echo "Building WSM client library"
-          ./gradlew :workspace-manager-client:clean :workspace-manager-client:assemble
-          cd workspace-manager-clienttests
-          export ORG_GRADLE_PROJECT_workspacemanagerclientjar=$(find .. -type f -name "workspace-manager-client*.jar")
-          echo "ORG_GRADLE_PROJECT_workspacemanagerclientjar = ${ORG_GRADLE_PROJECT_workspacemanagerclientjar}"
-
       # We can also set the env var in the GH action script to specify which server we are testing against.
       # This will be particularly helpful once we switch over to testing against any test env -- then it's just a change to the env var.
       # Please refer to https://github.com/DataBiosphere/terra-test-runner#override-the-server-from-the-command-line for more information.
--- a/.github/workflows/test-runner-pr-integration.yml
+++ b/.github/workflows/test-runner-pr-integration.yml
@@ -1,0 +1,103 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Test Runner Nightly Integration Tests
+
+on:
+  push:
+    branches: [ dev ]
+  pull_request:
+    branches: [ '**' ]
+
+jobs:
+  test-runner-integration:
+
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:12.3
+        env:
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+        - 5432:5432
+
+    steps:
+    - uses: actions/checkout@v2
+    # This step is needed when running test using a local server.
+    - name: Initialize Postgres DB
+      env:
+        PGPASSWORD: postgres
+      run: psql -h 127.0.0.1 -U postgres -f ./local-dev/local-postgres-init.sql
+    - name: Set up AdoptOpenJDK 11
+      uses: joschi/setup-jdk@v2
+      with:
+        java-version: 11
+    # This step caches Gradle packages for reuse in the same VM instance.
+    - name: Cache Gradle packages
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+        key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+        restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+
+    - name: Get Vault token
+      id: vault-token-step
+      env:
+        VAULT_ADDR: https://clotho.broadinstitute.org:8200
+      run: |
+        VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
+          -e "VAULT_ADDR=${VAULT_ADDR}" \
+          vault:1.1.0 \
+          vault write -field token \
+            auth/approle/login role_id=${{ secrets.VAULT_APPROLE_ROLE_ID }} \
+            secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+        echo ::set-output name=vault-token::$VAULT_TOKEN
+        echo ::add-mask::$VAULT_TOKEN
+    - name: Set permissions and render test configurations
+      run: |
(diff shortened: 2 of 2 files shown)


davidangb:
@ansingh7115 @ungwudik take a look?

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
...
+        image: postgres:12.3
+        env:
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+        - 5432:5432
```
jaycarlton, line 28: nit: consider making variables for the ports for readability.

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
...
+    # This step is needed when running test using a local server.
+    - name: Initialize Postgres DB
+      env:
+        PGPASSWORD: postgres
+      run: psql -h 127.0.0.1 -U postgres -f ./local-dev/local-postgres-init.sql
+    - name: Set up AdoptOpenJDK 11
+      uses: joschi/setup-jdk@v2
+      with:
+        java-version: 11
+    # This step caches Gradle packages for reuse in the same VM instance.
```
jaycarlton, line 41: Can you confirm this won't use stale artifacts? I.e. it knows how to update them between runs?

Is it an issue if a manual run caches newer artifacts and an hourly run picks those up instead of the right ones for the current build?

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
...
+    - name: Cache Gradle packages
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.gradle/caches
```
jaycarlton, line 46: nit: This seems odd: it's not an array and it's on two lines. Other plalces arguments used `>-`.

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
...
+      run: |
+        VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
+          -e "VAULT_ADDR=${VAULT_ADDR}" \
+          vault:1.1.0 \
+          vault write -field token \
+            auth/approle/login role_id=${{ secrets.VAULT_APPROLE_ROLE_ID }} \
+            secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+        echo ::set-output name=vault-token::$VAULT_TOKEN
+        echo ::add-mask::$VAULT_TOKEN
+    - name: Set permissions and render test configurations
```
jaycarlton, line 66: nit: I wish we had a better name for this than "render configurations".  This could be its own little action so you can share it with the other workflows.

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
...
+        chmod +x render_config.sh
+        ./render_config.sh dev ${{ steps.vault-token-step.outputs.vault-token }}
+        chmod +r rendered
+
+    # Spin up a local server.
+    - name: Run local server (in background)
+      run: ./gradlew bootRun &
+    - name: Wait for local server to start
+      # With a timeout of 60 seconds, try to connect to localhost:8080 every 1 second.
+      run: timeout 60 bash -c 'until echo > /dev/tcp/localhost/8080; do sleep 1; done'
```
jaycarlton, line 77: I wonder if there's a community action for this.

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
...
+        echo "Test Runner config"
+        ./render-config.sh ${{ steps.vault-token-step.outputs.vault-token }}
+        echo "Running test suite"
+        export TEST_RUNNER_SERVER_SPECIFICATION_FILE="workspace-local.json"
+        ./gradlew runTest --args="suites/BasicIntegration.json /tmp/TestRunnerResults"
+
+    - name: Upload test results to Google Bucket
+      if: ${{ always() }}
+      run: |
+        cd workspace-manager-clienttests
```
jaycarlton, line 94: if you're already in this directory, does this fail? I'd be more comfortable using the full path to the JSON filie, since we can run this step in situations where previous steps (and `cd`s) failed.

It may be better to use full paths throughout, so you don't rely on local context if you don't have to.

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
...
+        echo "Upload test results to Google Bucket"
+        ./gradlew uploadResults --args="CompressDirectoryToBucket.json /tmp/TestRunnerResults"
+
+    - name: Upload test results to BigQuery Test Runner Dataset
+      if: ${{ always() }}
+      run: |
+        cd workspace-manager-clienttests
+        echo "Upload test results to BigQuery Test Runner Dataset"
+        ./gradlew uploadResults --args="SummariesToBigQuery.json /tmp/TestRunnerResults"
```
jaycarlton, line 103: I'd kinda prefer we make a variable for a named directory for the results folder for every run so we don't accidentally upload stale results from a previous one. (Granted the files inside that directory are named with timestamps).

Should this step or a cleanup step delete the output dir after successful upload?

jaycarlton (commented):
This is looking good. Good catches about the servers. Lots of questions, a few opinions, and nits on shareability.

.github/workflows/test-runner-nightly-integration.yml
```diff
@@ -4,36 +4,17 @@
 name: Test Runner Nightly Integration Tests
 
 on:
-  push:
-    branches: [ dev ]
-  pull_request:
-      branches: [ '**' ]
+  workflow_dispatch: {}
```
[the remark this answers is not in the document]
zloery: TIL you can send manual dispatches from the actions page. We don't have great concurrency control so I'm not thrilled with recommending people do manual tests - this is mostly to support existing systems (jenkins) kicking off runs.

.github/workflows/test-runner-nightly-integration.yml
```diff
@@ -49,8 +30,6 @@ jobs:
           restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
-
-      # These steps aren't needed for unit tests
       - name: Get Vault token
```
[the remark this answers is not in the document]
zloery: I think single lines of bash can be easier to read than remote actions, but this step is common enough we should probably refactor. It's used enough places that the refactor is out of scope here, though

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
```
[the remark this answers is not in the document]
zloery: Done, and removed parts that aren't really relevant here

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
...
+
+name: Test Runner PR Integration Tests
+
+on:
+  push:
+    branches: [ dev ]
```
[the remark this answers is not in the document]
zloery: We don't allow direct pushes to dev, but that flow also runs when you merge a PR into dev to test the new head (also, you'll have a ✅  or ❌  icon next to the latests commit box at the top of the repo). Results show up on the Actions tab of the repo.
It seems like it could catch issues with unpublished changes but might be pretty low signal. I'm inclined to leave it in for consistency, but could take it out

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Test Runner PR Integration Tests
```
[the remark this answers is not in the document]
zloery: Renamed to `PR Client Library Integration Tests`

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -0,0 +1,103 @@
...
+
+on:
+  push:
+    branches: [ dev ]
+  pull_request:
+    branches: [ '**' ]
+
+jobs:
+  test-runner-integration:
```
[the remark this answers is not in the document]
zloery: done

.github/workflows/test-runner-nightly-integration.yml
```diff
@@ -4,36 +4,17 @@
 name: Test Runner Nightly Integration Tests
 
 on:
-  push:
-    branches: [ dev ]
-  pull_request:
-      branches: [ '**' ]
+  workflow_dispatch: {}
+  schedule:
+  - cron: '0 7 * * *' # run at 7 AM UTC, 2 AM ET.
```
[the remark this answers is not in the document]
zloery: Right now, results show up in the repo's Actions page. I think we want to surface nightly failures somehow, but I'm not sure what we currently do beyond uploading results - @ungwudik or @ansingh7115, do you have any more context on surfacing nightly test failures?

.github/workflows/test-runner-pr-integration.yml
zloery: I'm inclined to leave this since it's the default postgres port, but I could go either way

.github/workflows/test-runner-pr-integration.yml
zloery: this is only caching dependencies and gradle packages, and it uses the hash of our build file as (part of) the cache key, so any changes in dependencies will result in ignoring the cache.
(Branch name is also part of the cache key, so the nightly run on dev won't pick up a PR run from a separate branch)

.github/workflows/test-runner-pr-integration.yml
zloery: iiuc `|` preserves linebreaks while `>` (or `>-` to trim tailing linebreak) doesn't. in this case, this action expects a multi-line string for multiple path values.

.github/workflows/test-runner-pr-integration.yml
zloery: I'm actually not against the one-line script here. I think even with an action, we'll need basically the same line

.github/workflows/test-runner-pr-integration.yml
zloery: I think gradle does depend on local context, though I'm not sure on that one

.github/workflows/test-runner-pr-integration.yml
zloery: Each runner starts in a fresh virtual environment, so this is handled automatically without us cleaning up.
(Side note, the github-provided cache step we use above is explicitly an exception to this)

.github/workflows/test-runner-pr-integration.yml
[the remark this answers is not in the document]
jaycarlton: That makes sense. Fine to leave it then.

.github/workflows/test-runner-pr-integration.yml
jaycarlton: It could take the timeout as a parameter. Plus, some sneaky maintainer might come along and add more bash.

.github/workflows/test-runner-pr-integration.yml
jaycarlton: do we even need  to specify it then?

.github/workflows/test-runner-pr-integration.yml
jaycarlton: Thanks!

.github/workflows/test-runner-pr-integration.yml
jaycarlton: That checks out. THanks.

.github/workflows/test-runner-pr-integration.yml
jaycarlton: It just means we need to keep track of the current directory across steps.

.github/workflows/test-runner-pr-integration.yml
jaycarlton: Sounds good.

jaycarlton (approved):

.github/workflows/test-runner-nightly-integration.yml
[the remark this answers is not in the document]
ungwudik: Ultimately, we want to create a central dashboard for all services. This involves rethinking the different dashboards we currently have and possibly merging them. This will be a big effort and QA hasn't worked through all of it yet. In the meantime, we should post WSM nightly test results to #dsde-qa Slack channel.

ungwudik (approved):

PR #243, opened 2021-03-02
[PF-509] Multi-quality tool Gradle plugin

jaycarlton:
[PMD.pdf](https://github.com/DataBiosphere/terra-workspace-manager/files/6070103/PMD.pdf)
[SpotBugs Report.pdf](https://github.com/DataBiosphere/terra-workspace-manager/files/6070104/SpotBugs.Report.pdf)
[CHECKSTYLE.pdf](https://github.com/DataBiosphere/terra-workspace-manager/files/6070105/CHECKSTYLE.pdf)

Introduction and bare bones configuration for the [Gradle Quality Plugin](http://xvik.github.io/gradle-quality-plugin/4.5.0/). Sample html report is attached.

This plugin dumps errors to console rather than just a file, and can optionally make rules break the build. This is turned off currently, until we get the config right and clean up a backlog of findings.

Fixes to issues identified will come in a follow-on PR.


build.gradle
```diff
@@ -231,6 +232,11 @@ spotless {
 	}
 }
 
+quality {
```
gjuggler, line 235: See the pattern I put forth for quality-related overrides here: https://github.com/DataBiosphere/terra-common-lib/blob/b03c0c1f26607a4a5a9d290dc6b6af021115185f/gradle/quality.gradle

A few key points:
 - All suppressions or exclusions should have a comment with "Terra" in the line, and describing why we're making the override.
 - Omit any .xml files that aren't changed; it's extra noise and the quality plugin will by default use their managed version.

gjuggler (approved):
Cool – I think this will be a nice improvement over the spotbugs plugin.

yonghaoy (approved):

jaycarlton:
Simplified to use SpotBugs only, and removed other configs. Also set setrict = false, as this is the only way to get console output  apparentnly. I don't see anything in the docs about it, but

yonghaoy (approved):

gradle/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,278 @@
+<?xml version="1.0" ?>
```
gjuggler, line 1: General comment: where possible, I'd err towards excluding any files where we haven't made Terra-specific exclusions or changes, and also prominently flagging where those changes have been made (e.g. with a  comment above each exclusion)

gjuggler (approved):

gradle/config/checkstyle/checkstyle.xml
jaycarlton: Whoops, forgot to commit those deletions.

PR #244, opened 2021-03-02
[PF-306] Fix invalid description field in GHA workflow

zloery:
On merging #241, I discovered that `description` isn't actually a valid field for a workflow. This change moves the description to a comment instead. Sorry for the churn!


ungwudik (approved):

jaycarlton (approved):
Whoops, sorry about that.

PR #247, opened 2021-03-06
[PF-477] revise schema and object model

ddietterich:
This is a major revamping of the WSM resource implementation. I have to ask forgiveness for that. I think it provides a more solid foundation for moving forward. Here are the high points:
- New initial schema - history of changes so far are deleted. See [WSM Schema Revision](https://docs.google.com/document/d/10smi1Wy2fgaI9tv1X3g8415-1Z_D2Nf3jC77rXCeCeQ/edit#)
- The ResourceDao calls specific object constructors by resource type, providing it a DbResource object. Each layer of the object hierarchy picks off the data it is responsible for.
- The WsmResourceType enumeration holds associated class information. That allows us to convert a super class into the appropriate subclass by consulting the type.
- Removed the DataReference code. The WorkspaceApiController holds the data reference controller methods until they can be deleted. The controller methods map those endpoint into the common internal ReferenceResource classes and methods.
- Separated input validation (using class methods) from reference verification (done in the CreateReferenceResourceFlight).
- Factored StageService out to isolate the assertMcWorkspace methods. That eliminated a circular autowire dependency loop.
- Made existence check methods in the CrlService so that I could unit test resources without configuring CRL.
- 
See [WSM Object Hierarchy](https://docs.google.com/document/d/1yOW5oYeNctijQghEA4PWFMHxHc4p5I5EpUMojDZq14Q/edit#) for more discussion of the implementation.

I did a few drive-bys that I should point out:
- Added a way to use ssh instead of http in setup_local_env.sh so I could use my ssh key and not have to provide a password
- Removed some of the gcp hierarchy. We probably want some of this back. I did it for my own sanity...
- Changed how parameters were passed into the GCS bucket create flight. Instead of putting the API creation parameters onto the ControlledGcsBucketResource object, I pass them in separately. That eliminates the need for generic classes there.

Things on my "to do later" list:
- implement "locking" on the resource row in the database to simplify logic around re-running of steps. It will also help avoid concurrent update issues.
- review where exceptions live. I think they are not consistently placed. I'm not sure we need as many specific packages as we have.
- rename Google to GCP everywhere. This should be done in the swagger and in the code.
- add an Api prefix to everything in the swagger so we can easily tell the difference between the API classes and our internal ones.
- resolve what a snapshot name is. Half the code thinks it is a name. The integration test thinks it is a snapshot id. I littered some TODOs around.
- make rules for what is allowed in a resource name.
- integration testing of referenced resources with actual resources to verify
- expose properties (key-value map) for workspace and resource


src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -373,25 +378,26 @@ private String getAsyncResultEndpoint(String jobId) {
...
-        .googleContext(googleContext);
+        .gcpContext(gcpContext);
   }
 
   @Override
   public ResponseEntity<Void> deleteCloudContext(UUID id, CloudContext cloudContext) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
-    ControllerValidationUtils.validateCloudContext(cloudContext);
-    workspaceService.deleteGoogleContext(id, userReq);
+    ControllerValidationUtils.validateCloudType(cloudContext);
```
jaycarlton, line 399: Should this be `validateCloudPlatform` for consistency?

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -0,0 +1,429 @@
...
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+
+  // -- Common Resource Methods -- //
+
+  @Autowired
+  public ResourceDao(NamedParameterJdbcTemplate jdbcTemplate) {
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  public boolean deleteResource(UUID workspaceId, UUID resourceId) {
```
jaycarlton, line 85: Why do we need the workspace ID if the resource ID is the PK? Belt and suspenders approach?

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -0,0 +1,429 @@
...
+    storeResource(
+        resource.getWorkspaceId(),
+        resource.getResourceId(),
+        resource.getName(),
+        resource.getDescription(),
+        REFERENCE,
+        resource.getResourceType(),
+        resource.getCloningInstructions(),
+        resource.attributesToJson(),
+        Optional.empty(),
```
jaycarlton, line 182: in the style guide it says not to use Optional for function parameters.

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    if (description != null) {
+      params.addValue("description", description);
+    }
+
+    return updateResource(workspaceId, referenceId, params);
+  }
+
+  // -- Private Methods -- //
```
jaycarlton, line 209: This divider needs to be moved down, or just deleted. I don't see us maintaining things like this.

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -0,0 +1,429 @@
...
+      Optional<UUID> associatedApp,
+      Optional<String> assignedUser) {
+
+    // TODO: add resource locking to fix this
+    //  We create resources in flights, so we have steps that call resource creation that may
+    //  get run more than once. The safe solution is to "lock" the resource by writing the flight id
+    //  into the row at creation. Then it is possible on a re-insert to know whether the error is
+    //  because this flight step is re-running or because some other flight used the same resource
+    // id.
+    //  The small risk we have here is that a duplicate resource id of will appear to be
```
jaycarlton, line 273: nit: remove "of"

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -0,0 +1,429 @@
...
+            .addValue("associated_app", associatedApp.map(UUID::toString).orElse(null))
+            .addValue("assigned_user", assignedUser.orElse(null));
+
+    try {
+      jdbcTemplate.update(sql, params);
+      logger.info("Inserted record for resource {} for workspace {}", resourceId, workspaceId);
+    } catch (DuplicateKeyException e) {
+      throw new DuplicateResourceException(
+          String.format(
+              "A resource already exists in the workspace that has the same name (%s) or the same id (%s)",
```
jaycarlton, line 313: might as well include the workspace ID in the message.

jaycarlton (commented):
first batch of comments

src/main/java/bio/terra/workspace/service/resource/reference/ReferenceGcsBucketAttributes.java
```diff
@@ -0,0 +1,17 @@
+package bio.terra.workspace.service.resource.reference;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class ReferenceGcsBucketAttributes {
```
jaycarlton, line 6: can we combine this with the controlled counterpart?

src/main/java/bio/terra/workspace/service/resource/reference/ReferenceResourceService.java
```diff
@@ -0,0 +1,138 @@
...
+    return getReferenceResource(resource.getWorkspaceId(), resourceIdResult, userReq);
+  }
+
+  /**
+   * At this moment, the only updates on references we are doing are to name and description so this
+   * is a common implementation. If we get more complicated, we can break it out.
+   *
+   * @param workspaceId workspace of interest
+   * @param resourceId resource to update
+   * @param name name to change - may be null
```
jaycarlton, line 76: use `@Nullable` on the nullable parameters (either in javadoc or inline).

src/main/java/bio/terra/workspace/service/resource/reference/ReferenceResourceService.java
```diff
@@ -0,0 +1,138 @@
...
+   * Delete a reference. The only state we hold for a reference is in the metadata database so we
+   * directly delete that.
+   *
+   * @param workspaceId workspace of interest
+   * @param resourceId resource to delete
+   * @param userReq authenticated user
+   */
+  public void deleteReferenceResource(
+      UUID workspaceId, UUID resourceId, AuthenticatedUserRequest userReq) {
+    // TODO: Fix this based on resolution of permission model issue.
```
jaycarlton, line 100: is there a ticket for this issue?

src/main/java/bio/terra/workspace/service/resource/reference/ReferenceResourceService.java
```diff
@@ -0,0 +1,138 @@
...
+
+  public List<ReferenceResource> enumerateReferences(
+      UUID workspaceId, int offset, int limit, AuthenticatedUserRequest userReq) {
+    // TODO: Fix this based on resolution of permission model issue.
+    workspaceService.validateWorkspaceAndAction(
+        userReq, workspaceId, SamConstants.SAM_WORKSPACE_READ_ACTION);
+    return resourceDao.enumerateReferences(workspaceId, offset, limit);
+  }
+
+  private ReferenceResource castReferenceResource(WsmResource wsmResource) {
```
jaycarlton, line 132: The other castTo methods don't live in services I thought, but on the hierarchy.

src/main/java/bio/terra/workspace/service/resource/reference/flight/create/CreateReferenceMetadataStep.java
```diff
@@ -0,0 +1,52 @@
...
+    return StepResult.getStepResultSuccess();
+  }
+
+  private ReferenceResource getReferenceResource(FlightContext flightContext) {
+    FlightMap inputMap = flightContext.getInputParameters();
+    WsmResourceType resourceType =
+        inputMap.get(WorkspaceFlightMapKeys.ResourceKeys.RESOURCE_TYPE, WsmResourceType.class);
+
+    // Use the resource type to deserialize the right class
+    return inputMap.get(JobMapKeys.REQUEST.getKeyName(), resourceType.getReferenceClass());
```
jaycarlton, line 50: I'm surprised this works. If it does, I want to see if it can't work with the flight keys enum 🤔

jaycarlton (commented):
batch 2

src/main/java/bio/terra/workspace/service/workspace/flight/DeleteProjectStep.java
```diff
@@ -20,44 +23,48 @@
  * <p>Undo always fails for this step.
  */
 public class DeleteProjectStep implements Step {
```
jaycarlton, line 25: We need a way to express that this step should precede the delete context step. That knowledge is embedded in the undo logic but not really locked in.

src/main/java/bio/terra/workspace/service/workspace/flight/DeleteProjectStep.java
```diff
@@ -20,44 +23,48 @@
...
 
   @Override
   public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
-    UUID workspaceId =
-        flightContext.getInputParameters().get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
-    String projectId = workspaceDao.getCloudContext(workspaceId).googleProjectId();
-    if (projectId == null) {
+    Optional<GcpCloudContext> cloudContext = getContext(flightContext);
+    if (cloudContext.isEmpty()) {
       // Nothing to delete, so nothing to undo.
```
jaycarlton, line 55: "nothing was deleted"

src/main/java/bio/terra/workspace/service/workspace/flight/GenerateProjectIdStep.java
```diff
@@ -28,6 +28,7 @@ public StepResult undoStep(FlightContext flightContext) {
   }
 
   /** Generate a pseudo-random resource id. */
+  // TODO: Why not just use the UUID? Or use a short UUID (once I put that into common-lib)
```
jaycarlton, line 31: please make a ticket for short UUID support.

src/main/java/bio/terra/workspace/service/workspace/model/CloudPlatform.java
```diff
@@ -0,0 +1,29 @@
...
+
+  CloudPlatform(String dbString) {
+    this.dbString = dbString;
+  }
+
+  public String toSql() {
+    return dbString;
+  }
+
+  public static CloudPlatform fromSql(String dbString) {
```
jaycarlton, line 20: you can make a generic utility method to do this.

src/main/java/bio/terra/workspace/service/workspace/model/Workspace.java
```diff
@@ -25,23 +67,123 @@
...
+      return new Workspace(
+          workspaceId,
+          displayName,
+          description,
+          spendProfileId,
+          properties,
+          workspaceStage,
+          gcpCloudContext);
+    }
+    ;
```
jaycarlton, line 187: What's this semicolon for?

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
```diff
@@ -0,0 +1,246 @@
...
+    author: dd
+    changes:
+    - createTable:
+        tableName: workspace
+        remarks: |
+          There is one row in the workspace table for each workspace.
+        columns:
+          - column:
+              name: workspace_id
+              type: text
```
jaycarlton, line 14: Is there not a UUID type?

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
```diff
@@ -0,0 +1,246 @@
...
+              remarks: |
+                A user-friendly name for the workspace. There is no guarantee of uniqueness across workspaces.
+                Can be null. Omitted for Rawls workspaces, since Rawls maintains its own name
+          - column:
+              name: description
+              type: text
+              remarks: |
+                Free-form text describing the workspace. Can be null. Omitted for Rawls workspaces, since Rawls
+                maintains its own name.
```
jaycarlton, line 29: it's own description?

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
```diff
@@ -0,0 +1,246 @@
...
+              name: spend_profile
+              type: text
+              constraints:
+                nullable: true
+                unique: false
+          - column:
+              name: properties
+              type: jsonb
+              remarks: |
+                Key-value map to allow applications to store their data about the workspace
```
jaycarlton, line 40: what happens if two applications use the same key?

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
```diff
@@ -0,0 +1,246 @@
...
+          - column:
+              name: workspace_stage
+              type: text
```
jaycarlton, line 43: why not use an enum type? Or is that not a thing?

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
```diff
@@ -0,0 +1,246 @@
...
+          current Terra deployment.
+        columns:
+          - column:
+              name: application_id
+              type: text
+              constraints:
+                primaryKey: true
+                nullable: false
+          - column:
+              name: name
```
jaycarlton, line 96: name is nullable?

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
```diff
@@ -0,0 +1,246 @@
...
+                REFERENCE - the resource is a reference to a cloud object not managed by WSM
+                CONTROLLED - the resource is controlled by WSM for lifecycle, IAM, and policy
+          - column:
+              name: resource_type
+              type: text
+              constraints:
+                nullable: false
+              remarks: |
+                Enumeration of the resource type. Resource types are specific to the cloud
+                and resource; e.g., GCS_BUCKET, BIGQUERY_DATASET
```
jaycarlton, line 191: not BIG_QUERY_DATASET?

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
```diff
@@ -0,0 +1,246 @@
...
+                between referenced and controlled resource data allows us to avoid a
+                separate referenced_resource table that would have a single column for its
+                metadata.
+          - column:
+              name: access
+              type: text
+              remarks: |
+                Must be valid, non-null for controlled resources. Must be null for reference resources.
+                Enum of the possible access types of the controlled resource. Values are
+                USER_SHARED, USER_PRIVATE, APP_SHARED, APP_PRIVATE
```
jaycarlton, line 216: why not store the two bits separately in different columns for managedBy and access?

jaycarlton (commented):
batch 3

src/test/java/bio/terra/workspace/service/resource/reference/ReferenceResourceServiceTest.java
```diff
@@ -0,0 +1,565 @@
...
+      assertThat(referenceResource.getStewardshipType(), equalTo(StewardshipType.REFERENCE));
+
+      ReferenceBigQueryDatasetResource resource = referenceResource.castToBigQueryDatasetResource();
+      assertThat(resource.getResourceType(), equalTo(WsmResourceType.BIG_QUERY_DATASET));
+
+      ReferenceResource resultReferenceResource =
+          referenceResourceService.createReferenceResource(referenceResource, USER_REQUEST);
+      ReferenceBigQueryDatasetResource resultResource =
+          resultReferenceResource.castToBigQueryDatasetResource();
+      assertThat(resource, equalTo(resultResource));
```
jaycarlton, line 436: swap args

src/test/java/bio/terra/workspace/service/resource/reference/ReferenceResourceServiceTest.java
```diff
@@ -0,0 +1,565 @@
...
+      assertThrows(
+          ResourceNotFoundException.class,
+          () ->
+              referenceResourceService.getReferenceResourceByName(
+                  workspaceId, UUID.randomUUID().toString(), USER_REQUEST));
+    }
+
+    @Test
+    void deleteResource() {
+      referenceResourceService.deleteReferenceResource(
```
jaycarlton, line 532: This test has no assertions. Maybe comment that there's no harm in deleting a not-found resource.

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -47,25 +47,23 @@
 import org.springframework.boot.test.mock.mockito.MockBean;
 
 class WorkspaceServiceTest extends BaseConnectedTest {
+  /** A fake authenticated user request. */
+  private static final AuthenticatedUserRequest USER_REQUEST =
+      new AuthenticatedUserRequest()
+          .token(Optional.of("fake-token"))
```
jaycarlton, line 53: this should just take in a nullable string

workspace-manager-clienttests/src/main/resources/servers/workspace-dd.json
```diff
@@ -0,0 +1,15 @@
+{
```
jaycarlton, line 1: Interesting. I don't have one of these.

src/main/java/bio/terra/workspace/app/controller/ControlledGoogleResourceApiController.java
```diff
@@ -73,7 +82,12 @@ public ControlledGoogleResourceApiController(
 
   @Override
   public ResponseEntity<GoogleBucketStoredAttributes> getBucket(UUID id, UUID resourceId) {
-    return new ResponseEntity<>(controlledResourceService.getBucket(id, resourceId), HttpStatus.OK);
+    final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    ControlledResource controlledResource =
+        controlledResourceService.getControlledResource(id, resourceId, userRequest);
+    GoogleBucketStoredAttributes response =
+        controlledResource.castToGcsBucketResource().toApiModel();
```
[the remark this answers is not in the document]
ddietterich: I'm open to suggestions. We could cast directly. I went with the belt'n'suspender of having common checking of the resource type along with the cast.

Another alternative would be for each specialized class to implement a static `fromControlledResource` so this would be `ControlledGcsBucketResource.fromControlledResource(controlledResource).toApiModel();`

It couldn't be an abstract method, but could be in all of the specializations by convention.

zloery (commented):
First pass, I'm still doing a deeper review on the DB schema and DAOs plus tests but overall this looks good so far.

src/main/java/bio/terra/workspace/app/controller/ReferenceController.java
```diff
@@ -0,0 +1,213 @@
...
+
+  @Override
+  public ResponseEntity<GoogleBucketReference> createBucketReference(
+      UUID id, @Valid CreateGoogleBucketReferenceRequestBody body) {
+
+    // Construct a ReferenceGcsBucketResource object from the API input
+    var resource =
+        new ReferenceGcsBucketResource(
+            id,
+            UUID.randomUUID(), // mint a resource id for this bucket
```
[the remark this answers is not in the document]
ddietterich: We could. The advantage to this is that the validator for the bucket resource can test that there is a valid id.
I made a builder for the specializations that generates the id of needed.

src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -74,7 +74,7 @@ public static void validateEmail(String email) {
   }
 
   /** Validate that a user is requesting a valid cloud for adding workspace context. */
-  public static void validateCloudContext(CloudContext context) {
+  public static void validateCloudType(CloudContext context) {
```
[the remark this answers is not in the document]
ddietterich: I did not make changes to the API. There are many changes that would make it more consistent and clearer. I plan to do that in a follow-on PR

src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -97,6 +103,28 @@ public BigQueryCow createBigQueryCow(AuthenticatedUserRequest userReq) {
...
+  /**
+   * Wrap the BigQuery existence check in its own method. That allows unit tests to mock this
+   * service and generate an answer without actually touching BigQuery.
+   *
+   * @param projectId Google project id where the dataset is
+   * @param datasetId name of the dataset
+   * @param userRequest auth info
+   * @return true if the dataset exists
+   */
+  public boolean bigQueryDatasetExists(
```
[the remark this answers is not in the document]
ddietterich: The whole reason for this method was to able to mock CRL.

src/main/java/bio/terra/workspace/service/iam/CustomGcpIamRoleMapping.java
```diff
@@ -89,13 +89,15 @@
           new CustomGcpIamRole(
               WsmResourceType.GCS_BUCKET, IamRole.OWNER, GCS_BUCKET_OWNER_PERMISSIONS),
           new CustomGcpIamRole(
-              WsmResourceType.BIGQUERY_DATASET,
+              WsmResourceType.BIG_QUERY_DATASET,
```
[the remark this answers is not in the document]
ddietterich: I did the rename to `BIG_QUERY`

src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -0,0 +1,64 @@
...
+      Pattern.compile("^[a-zA-Z0-9][_a-zA-Z0-9]{0,62}$");
+
+  /**
+   * GCS bucket name validation is somewhat complex due to rules about usage of "." and restricted
+   * names like "goog", but as a baseline they must be 3-222 characters long using lowercase
+   * letters, numbers, dashes, underscores, and dots, and must start and end with a letter or
+   * number. Matching this pattern is necessary but not sufficient for a valid bucket name.
+   */
+  public static final Pattern BUCKET_NAME_VALIDATION_PATTERN =
+      Pattern.compile("^[a-z0-9][-_.a-z0-9]{1,220}[a-z0-9]$");
```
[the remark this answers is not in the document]
ddietterich: For a reference, we would want to take any valid name. For a controlled resource, I am not sure. Is that a rule we should enforce in WSM? Or is that a rule the CLI should enforce for its own purposes?

src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * BigQuery datasets must be 1-1024 characters, using letters (upper or lowercase), numbers, and
+   * underscores.
+   */
+  public static final Pattern BQ_DATASET_NAME_VALIDATION_PATTERN =
```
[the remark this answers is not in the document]
ddietterich: Again, for references we need to take the full name.

src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -0,0 +1,64 @@
...
+      Pattern.compile("^[_a-zA-Z0-9]{1,1024}$");
+
+  public static void validateDataRepoName(String name) {
+    if (StringUtils.isEmpty(name) || !DATAREPO_NAME_VALIDATION_PATTERN.matcher(name).matches()) {
```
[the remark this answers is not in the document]
ddietterich: Turns out this name check is a bug. There is no name. It is an identifier under TDR control. We should not have an opinion about what characters are in it.

src/main/java/bio/terra/workspace/service/resource/WsmResource.java
```diff
@@ -1,27 +1,54 @@
...
-import bio.terra.workspace.service.datareference.model.CloningInstructions;
-import bio.terra.workspace.service.datareference.model.ReferenceObject;
+import bio.terra.workspace.common.exception.MissingRequiredFieldException;
+import bio.terra.workspace.db.model.DbResource;
+import bio.terra.workspace.service.resource.model.CloningInstructions;
+import bio.terra.workspace.service.resource.model.StewardshipType;
 import com.google.common.base.Strings;
-import java.util.Objects;
 import java.util.UUID;
+import org.apache.commons.lang3.builder.EqualsBuilder;
```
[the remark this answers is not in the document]
ddietterich: I must have different defaults set. I'll change it to use IJ.

src/main/java/bio/terra/workspace/service/resource/WsmResourceType.java
```diff
@@ -0,0 +1,64 @@
...
+import bio.terra.workspace.service.workspace.model.CloudPlatform;
+import org.apache.commons.lang3.SerializationException;
+import org.apache.commons.lang3.StringUtils;
+
+public enum WsmResourceType {
+  DATA_REPO_SNAPSHOT(
+      CloudPlatform.GCP,
+      "DATA_REPO_SNAPSHOT",
+      ReferenceDataRepoSnapshotResource.class,
+      ControlledGcsBucketResource.class),
```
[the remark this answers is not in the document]
ddietterich: Doh!

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAccessType.java
```diff
@@ -0,0 +1,57 @@
...
+import static bio.terra.workspace.generated.model.ControlledResourceCommonFields.ManagedByEnum.APPLICATION;
+import static bio.terra.workspace.generated.model.ControlledResourceCommonFields.ManagedByEnum.USER;
+
+import bio.terra.workspace.generated.model.ControlledResourceCommonFields;
+import bio.terra.workspace.generated.model.ControlledResourceCommonFields.ManagedByEnum;
+import bio.terra.workspace.service.workspace.exceptions.InternalLogicException;
+import org.apache.commons.lang3.SerializationException;
+import org.apache.commons.lang3.StringUtils;
+
+public enum ControlledAccessType {
```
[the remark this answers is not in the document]
ddietterich: I'm not sure. What I had in mind is that it maps to the four Sam object types and their rules. As notes below, we would want isPrivate() and isApplication() methods to tease them apart.

I am leaning toward thinking this is not a good merging and I should separate them out.
Do either of you want to vote? @zloery @jaycarlton

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledGcsBucketResource.java
```diff
@@ -0,0 +1,102 @@
...
+        description,
+        cloningInstructions,
+        assignedUser,
+        controlledAccessType);
+    this.bucketName = bucketName;
+    validate();
+  }
+
+  public ControlledGcsBucketResource(DbResource dbResource) {
+    super(
```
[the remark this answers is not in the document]
ddietterich: Honestly, I don't know what I was thinking!

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAccessType.java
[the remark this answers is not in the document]
zloery: At least for now, we need to store 4 mappings of workspace roles to resource roles (one for each ControlledAccessType) inside WSM. Having this as an enum is convenient for organizing that but not necessary, and I'm not sure where else we'd use it, so I'm pretty neutral on this one.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Autowired
+  public ControlledResourceService(
+      JobService jobService, WorkspaceService workspaceService, ResourceDao resourceDao) {
+    this.jobService = jobService;
+    this.workspaceService = workspaceService;
+    this.resourceDao = resourceDao;
+  }
+
+  public String createGcsBucket(
```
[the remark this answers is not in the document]
ddietterich: Good catch. I switched it as you suggested.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -0,0 +1,77 @@
...
+    // TODO: Fix this based on resolution of permission model issue.
+    workspaceService.validateWorkspaceAndAction(
+        userReq, workspaceId, SamConstants.SAM_WORKSPACE_READ_ACTION);
+    WsmResource wsmResource = resourceDao.getResource(workspaceId, resourceId);
+    return castControlledResource(wsmResource);
+  }
+
+  private ControlledResource castControlledResource(WsmResource wsmResource) {
+    if (!(wsmResource instanceof ControlledResource)) {
+      throw new InvalidMetadataException("Returned resource is not a controlled resource");
```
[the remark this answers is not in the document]
ddietterich: True. I changed it to check the stewardship type explicitly

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateControlledResourceFlight.java
```diff
@@ -20,15 +18,8 @@ public CreateControlledResourceFlight(FlightMap inputParameters, Object applicat
     super(inputParameters, applicationContext);
     final FlightBeanBag flightBeanBag = FlightBeanBag.getFromObject(applicationContext);
 
-    // Generate a new resource UUID
```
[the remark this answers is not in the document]
ddietterich: Since I am doing it before the flight starts and the input parameters are serialized to the database before the start, it has the same effect.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/StoreMetadataStep.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    final FlightMap inputMap = flightContext.getInputParameters();
+    final ControlledResource resource =
+        inputMap.get(JobMapKeys.REQUEST.getKeyName(), ControlledResource.class);
+
+    resourceDao.createControlledResource(resource);
```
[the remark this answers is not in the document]
ddietterich: From my POV, the DAO takes the model and stored it as database data and it takes the database data and returns the model. And, yeah, the model is helping via the DbResource, but that class is not really different from your DTO is it?

src/main/java/bio/terra/workspace/service/resource/model/StewardshipType.java
```diff
@@ -0,0 +1,30 @@
+package bio.terra.workspace.service.resource.model;
+
+import org.apache.commons.lang3.SerializationException;
+import org.apache.commons.lang3.StringUtils;
+
+public enum StewardshipType {
+  REFERENCE("REFERENCE"),
```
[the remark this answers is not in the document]
ddietterich: Yes, that was bugging me, too. As long as I have the world opened up, I might as well fix it.

src/main/java/bio/terra/workspace/service/resource/reference/ReferenceBigQueryDatasetAttributes.java
```diff
@@ -0,0 +1,25 @@
+package bio.terra.workspace.service.resource.reference;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class ReferenceBigQueryDatasetAttributes {
+  private final String projectId;
```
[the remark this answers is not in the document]
ddietterich: My guess is that we will have more attributes for controlled resources than for referenced ones, but that is a guess. What we we code differently if they were the same?

src/main/java/bio/terra/workspace/service/resource/reference/ReferenceGcsBucketAttributes.java
ddietterich: It is a good question. If it turns out that the attributes are always alike for referenced and controlled, then we should make an attributes package and just share them. OTOH, if we change one and not the other, it would be nice to have different classes serialized in Stairway. Hmm...

src/main/java/bio/terra/workspace/service/resource/reference/ReferenceResourceService.java
```diff
@@ -0,0 +1,138 @@
...
+            resource.getResourceType(), resource.getResourceId(), resource.getName());
+    JobBuilder createJob =
+        jobService
+            .newJob(
+                jobDescription,
+                UUID.randomUUID().toString(),
+                CreateReferenceResourceFlight.class,
+                resource,
+                userReq)
+            .addParameter(
```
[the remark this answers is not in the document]
ddietterich: I added a comment to explain. By having the request object, we can provide the right target class for pulling the request object out of the input parameter flight map.

src/main/java/bio/terra/workspace/service/resource/reference/ReferenceResourceService.java
ddietterich: Not yet. We are discussing on Thursday in the WSM sync meeting with the AS team. I have a list of about 7 items that I will make into tickets based on how this work comes out.

src/main/java/bio/terra/workspace/service/resource/reference/ReferenceResourceService.java
ddietterich: Good catch. I'll move them

src/main/java/bio/terra/workspace/service/resource/reference/flight/create/CreateReferenceMetadataStep.java
ddietterich: It works. It would be nifty to wrap the map key with the class type for pulling things out of the flight maps.

src/main/java/bio/terra/workspace/service/resource/reference/flight/create/CreateReferenceVerifyAccessBigQueryStep.java
```diff
@@ -0,0 +1,47 @@
...
+          String.format(
+              "Could not access BigQuery dataset %s in project %s. Ensure the name and GCP project are correct and that you have access.",
+              datasetName, projectId));
+    }
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
```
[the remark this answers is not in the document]
ddietterich: I'll add that on my Stairway list.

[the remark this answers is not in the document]
ddietterich: PF-554

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -124,41 +131,41 @@ public void deleteWorkspace(UUID id, AuthenticatedUserRequest userReq) {
...
     if (!bufferServiceConfiguration.getEnabled()) {
       throw new BufferServiceDisabledException(
-          "Cannot create a google context in an environment where buffer service is disabled or not configured.");
+          "Cannot create a GCP context in an environment where buffer service is disabled or not configured.");
     }
+
     Workspace workspace =
         validateWorkspaceAndAction(userReq, workspaceId, SamConstants.SAM_WORKSPACE_WRITE_ACTION);
-    workspaceDao.assertMcWorkspace(workspace, "createGoogleContext");
-    if (workspaceDao.getCloudContext(workspaceId).googleProjectId() != null) {
```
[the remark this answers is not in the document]
ddietterich: The approach I am thinking of is to take a lock on the workspace (where lock is storing the flight id on the workspace row) during the create. Then we can test the existence of the context and know that another flight will not be creating one concurrently. All of that would happen before the project allocation. I made PF-555.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -124,41 +131,41 @@ public void deleteWorkspace(UUID id, AuthenticatedUserRequest userReq) {
...
-      throw new DuplicateGoogleContextException(workspaceId);
-    }
+    stageService.assertMcWorkspace(workspace, "createCloudContext");
+
+    // TODO: We should probably do this in a step of the job. It will be talking to another
```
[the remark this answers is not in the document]
ddietterich: PF-556

src/main/java/bio/terra/workspace/service/workspace/flight/DeleteGoogleContextFlight.java
```diff
@@ -17,18 +16,13 @@ public DeleteGoogleContextFlight(FlightMap inputParameters, Object applicationCo
...
             INITIAL_INTERVALS_SECONDS, MAX_INTERVAL_SECONDS, MAX_OPERATION_TIME_SECONDS);
 
     addStep(
-        new DeleteProjectStep(crl.getCloudResourceManagerCow(), appContext.getWorkspaceDao()),
-        retryRule);
-    addStep(
-        new DeleteGoogleContextStep(
-            appContext.getWorkspaceDao(), appContext.getTransactionTemplate()),
-        retryRule);
+        new DeleteProjectStep(appContext.getCrlService(), appContext.getWorkspaceDao()), retryRule);
```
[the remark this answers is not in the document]
ddietterich: No it can't. The retry rule is stateful and needs to be specific instance for this flight. It can be shared within the flight, since the flight is sequential and the rule is initialized each time, but it cannot be static.

src/main/java/bio/terra/workspace/service/workspace/flight/GenerateProjectIdStep.java
ddietterich: I have coded up a TerraUUID class that has it in and is nicely deserialized by Stairway. So it is on the way. The ticket is PF-546

src/main/java/bio/terra/workspace/service/workspace/model/Workspace.java
ddietterich: Sometimes a semicolon just needs to stop always trying to do something and just practice being :)

src/main/resources/api/service_openapi.yaml
```diff
@@ -213,13 +213,15 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  #####
```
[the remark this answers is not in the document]
ddietterich: It is a comment on a comment on a comment on a comment... very meta.
I'll get rid of it.

src/main/java/bio/terra/workspace/service/resource/reference/flight/create/CreateReferenceResourceFlight.java
```diff
@@ -0,0 +1,42 @@
...
+        inputParameters.get(
+            WorkspaceFlightMapKeys.ResourceKeys.RESOURCE_TYPE, WsmResourceType.class);
+    switch (resourceType) {
+      case BIG_QUERY_DATASET:
+        addStep(new CreateReferenceVerifyAccessBigQueryStep(appContext.getCrlService()));
+        break;
+
+      case DATA_REPO_SNAPSHOT:
+        addStep(
+            new CreateReferenceVerifyAccessDataRepoSnapshotStep(appContext.getDataRepoService()));
```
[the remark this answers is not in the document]
jaycarlton: Having steps that can't be shared scoped to their flights seems useful to me. Maybe they should be inner classes of the flight?

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
ddietterich: I looked and there is a UUID type. I wonder if there are any downsides to using it.

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
ddietterich: Postgres doesn't have enums

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
```diff
@@ -0,0 +1,246 @@
...
+                nullable: false
+                deleteCascade: true
+              remarks: |
+                First part of the foreign key to cloud_context. Foreign key will cause
+                resource to be deleted when its associated cloud_context is deleted.
+          - column:
+              name: cloud_platform
+              type: text
+              constraints:
+                nullable: false
```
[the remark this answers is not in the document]
ddietterich: This is an interesting issue. Right now, data repo snapshots are set as GCP references.
I think I would not make this null, but maybe we need a CloudPlatform.NONE for things that are inherently multi-cloud.

src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
```diff
@@ -61,31 +59,31 @@ void AddedReaderCanRead() {
...
     samService.grantWorkspaceRole(
         workspaceId, defaultUserRequest(), IamRole.WRITER, userAccessUtils.getSecondUserEmail());
-    DataReference reference =
-        dataReferenceService.createDataReference(referenceRequest, secondaryUserRequest());
-    assertEquals(referenceRequest.name(), reference.name());
+
+    ReferenceResource ref =
+        referenceResourceService.createReferenceResource(referenceResource, secondaryUserRequest());
+    ReferenceDataRepoSnapshotResource resultResource = ref.castToDataRepoSnapshotResource();
+    assertEquals(referenceResource, resultResource);
```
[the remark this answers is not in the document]
ddietterich: I added a set of cast failure tests.

src/main/java/bio/terra/workspace/app/controller/ControlledGoogleResourceApiController.java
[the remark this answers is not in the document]
ddietterich: I'll leave it as is for now. It is not hard to switch later if we want to.

src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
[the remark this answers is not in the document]
ddietterich: PF-562

src/main/java/bio/terra/workspace/service/resource/WsmResource.java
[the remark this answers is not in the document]
ddietterich: I notice guava has some support for generating these also. We should just pick one.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAccessType.java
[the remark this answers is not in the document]
ddietterich: I will split these apart.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -0,0 +1,77 @@
...
+    this.resourceDao = resourceDao;
+  }
+
+  public String createGcsBucket(
+      ControlledGcsBucketResource resource,
+      GoogleBucketCreationParameters creationParameters,
+      IamRole creationIamRole,
+      JobControl jobControl,
+      AuthenticatedUserRequest userRequest) {
+
```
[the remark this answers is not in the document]
ddietterich: I plan to leave this for another pass. I think the existing checks are wrong, so I'd like to do one pass and put the proper permission checks in.

src/main/java/bio/terra/workspace/service/resource/model/StewardshipType.java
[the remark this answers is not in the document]
ddietterich: I added it to the list for a big rename pass. PF-562

src/main/java/bio/terra/workspace/service/resource/reference/flight/create/CreateReferenceResourceFlight.java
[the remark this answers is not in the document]
ddietterich: For simple flights that would be very tidy. You could see all the code in one place. No reason not to do that.

src/main/java/bio/terra/workspace/service/resource/reference/flight/create/CreateReferenceResourceFlight.java
```diff
@@ -0,0 +1,42 @@
...
+        addStep(new CreateReferenceVerifyAccessGcsBucketStep(appContext.getCrlService()));
+        break;
+
+      default:
+        throw new InternalLogicException("Unknown resource type");
+    }
+
+    // If all is well, then store the reference metadata
+    addStep(new CreateReferenceMetadataStep(appContext.getResourceDao()));
+  }
```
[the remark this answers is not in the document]
ddietterich: Done as far as I know

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
ddietterich: I looked at both the uuid and the enum types. When you use them, you are at the mercy of middle layers of JDBC and the postgres driver to get the transform right. I think we are better off keeping it vanilla.

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
ddietterich: OK, it does have enums.

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
ddietterich: I think we can come up with a key naming convention if the need arises.

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
ddietterich: I updated it to display_name and added remarks for it and the description.

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
ddietterich: I plan to make that change

src/main/resources/db/changesets/20210301_revised_initial_schema.yaml
```diff
@@ -0,0 +1,246 @@
...
+              remarks: |
+                Must be valid, non-null for controlled resources. Must be null for reference resources.
+                Enum of the possible access types of the controlled resource. Values are
+                USER_SHARED, USER_PRIVATE, APP_SHARED, APP_PRIVATE
+          - column:
+              name: associated_app
+              type: text
+              constraints:
+                references: application(application_id)
+                foreignKeyName: fk_associated_app
```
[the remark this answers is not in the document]
ddietterich: That is the default

src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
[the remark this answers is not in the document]
ddietterich: I'm not going to resolve this as part of this PR

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAccessType.java
```diff
@@ -0,0 +1,57 @@
...
+  private final String dbString;
+
+  ControlledAccessType(String dbString) {
+    this.dbString = dbString;
+  }
+
+  public String toSql() {
+    return dbString;
+  }
+
```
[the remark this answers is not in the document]
ddietterich: I'm going to split this into parts instead

src/main/java/bio/terra/workspace/service/resource/reference/ReferenceBigQueryDatasetAttributes.java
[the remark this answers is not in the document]
ddietterich: No. For the controlled case, the project will be from the cloud context - it is the workspace project.

src/main/java/bio/terra/workspace/service/resource/controlled/AccessScopeType.java
```diff
@@ -0,0 +1,47 @@
+package bio.terra.workspace.service.resource.controlled;
+
+import bio.terra.workspace.generated.model.ControlledResourceCommonFields;
+import bio.terra.workspace.service.workspace.exceptions.InternalLogicException;
+import org.apache.commons.lang3.SerializationException;
+import org.apache.commons.lang3.StringUtils;
+
+/** Controlled resources can be shared by all users in a workspace or be private to a single user */
+public enum AccessScopeType {
+    ACCESS_SCOPE_SHARED("ACCESS_SHARED"),
```
jaycarlton: I don't know that we need the `ACCESS_` prefix, since that's going to be the same for the whole enum.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResource.java
```diff
@@ -87,24 +93,22 @@ private void validateSubclass(WsmResourceType expectedType) {
...
   }
 
   @Override
   public int hashCode() {
-    return new HashCodeBuilder(17, 37)
-        .appendSuper(super.hashCode())
-        .append(assignedUser)
-        .append(controlledAccessType)
-        .toHashCode();
+    int result = super.hashCode();
```
jaycarlton, line 108: What's wrong with `HashCodeBuilder`? It looked pretty nice.

src/main/java/bio/terra/workspace/service/resource/controlled/ManagedByType.java
```diff
@@ -0,0 +1,46 @@
+package bio.terra.workspace.service.resource.controlled;
+
+import bio.terra.workspace.generated.model.ControlledResourceCommonFields;
+import bio.terra.workspace.service.workspace.exceptions.InternalLogicException;
+import org.apache.commons.lang3.SerializationException;
+import org.apache.commons.lang3.StringUtils;
+
+/** Controlled resources can be managed by a user or by an application */
+public enum ManagedByType {
+    MANAGED_BY_USER("USER"),
```
jaycarlton: prefix is redundant. `ManagedByType.USER` is  descriptive enough I think.

src/main/java/bio/terra/workspace/service/resource/WsmResource.java
```diff
@@ -133,10 +133,13 @@ public boolean equals(Object o) {
 
     WsmResource that = (WsmResource) o;
 
-    if (workspaceId != null ? !workspaceId.equals(that.workspaceId) : that.workspaceId != null) return false;
-    if (resourceId != null ? !resourceId.equals(that.resourceId) : that.resourceId != null) return false;
+    if (workspaceId != null ? !workspaceId.equals(that.workspaceId) : that.workspaceId != null)
```
jaycarlton, line 136: Is there a way to invert this so we don't have so many negations?

src/main/java/bio/terra/workspace/service/resource/controlled/AccessScopeType.java
```diff
@@ -7,41 +7,38 @@
...
-    public static AccessScopeType fromSql(String dbString) {
-        for (AccessScopeType value : values()) {
-            if (StringUtils.equals(value.dbString, dbString)) {
-                return value;
-            }
-        }
-        throw new SerializationException(
-                "Deeserialization failed: no matching access scope type for " + dbString);
+  public static AccessScopeType fromSql(String dbString) {
+    for (AccessScopeType value : values()) {
```
jaycarlton, line 36: We can write a generic method that does this for any enum and call that.

jaycarlton (approved):

zloery (approved):

src/main/java/bio/terra/workspace/service/resource/WsmResource.java
ddietterich: This is auto-generated by IntelliJ. We should talk about what code gen style we want to use and then use it everywhere. I agree this is ugly.

src/main/java/bio/terra/workspace/service/resource/controlled/AccessScopeType.java
ddietterich: I tried writing that. It is quite messy. Maybe you can do a better job.
I'm going to leave this in the interest of getting it done.
There is a shorter expression using streams like:
```
Arrays.stream(values())
            .filter(e -> StringUtils.equals(e, dbString)
            .findFirst()
            .orElseThrow(() -> SerializationException(...));
```

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResource.java
```diff
@@ -13,91 +16,99 @@
...
-
-    if (getResourceType() == null || getJsonAttributes() == null) {
-      throw new IllegalStateException("Missing required field for ControlledResource.");
+    if (getResourceType() == null
+        || attributesToJson() == null
+        || getAccessScope() == null
+        || getManagedBy() == null) {
+      throw new MissingRequiredFieldException("Missing required field for ControlledResource.");
+    }
+    if (getAssignedUser().isPresent() && getAccessScope() == AccessScopeType.ACCESS_SCOPE_SHARED) {
```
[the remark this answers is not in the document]
ddietterich: OK, it is kind of wrong one way or the other. If the data is coming from the database, it should be a 500.
But your scenario is the more likely case. I'll change it to a BadRequest.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -33,8 +33,8 @@ public ControlledResourceService(
     this.resourceDao = resourceDao;
   }
 
-  public String createGcsBucket(
-      ControlledGcsBucketResource resource,
+  public String createControlledResource(
+      ControlledResource resource,
       GoogleBucketCreationParameters creationParameters,
```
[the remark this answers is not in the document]
ddietterich: True that. I'll try Object. If that doesn't work, then we'll need individual create methods I guess.

PR #248, opened 2021-03-08
PF-520 Migrate WM to use bio.terra.common.tracing. 

wchamber:
Replace the workspace manager's tracing set up with the bio.terra.common.tracing package.
As currently configured, this will only trace `/api/*` paths, notably not tracing the `/status` path that floods the current trace logs.

Example trace from my personal environment: https://console.cloud.google.com/traces/list?tid=e3423c85586219bc1ea6808f20dc7dd2
Current trace from dev: https://console.cloud.google.com/traces/list?tid=dd6c42da21329bb790788b77dff2faf4


build.gradle
```diff
@@ -87,12 +87,6 @@ dependencies {
 	implementation group: "org.webjars", name: "webjars-locator-core", version: "0.46"
 	runtimeOnly group: "org.postgresql", name: "postgresql", version: "42.2.18"
 
-	// Tracing deps
```
zloery, line 90: at the top of this file we also declare the versions for these libraries, can we remove that block as well?

src/main/resources/application.yml
```diff
@@ -13,9 +13,8 @@ env:
       pass: ${DATABASE_USER_PASSWORD:wmpwd}
       user: ${DATABASE_USER:wmuser}
   tracing:
-    enabled: ${CLOUD_TRACE_ENABLED:false}
-    probability: ${SAMPLING_PROBABILITY:0}
-    project-id: ${SERVICE_GOOGLE_PROJECT:broad-dsde-dev}
```
zloery, line 18: out of scope, but it looks like this is the only place we're using this var so we can probably clean it out of helm + helmfile. can you ticket that so it doesn't get lost?

zloery (approved):
LGTM.
double checking, is filtering traced paths to `/api/*` a default in the common lib? I don't see it specified here

wchamber:
> LGTM.
> double checking, is filtering traced paths to `/api/*` a default in the common lib? I don't see it specified here

Yep. https://github.com/DataBiosphere/terra-common-lib/blob/887f36a798d8636d8b5b4f3482680d24cdd0f1cb/src/main/java/bio/terra/common/tracing/TracingProperties.java#L25

src/main/resources/application.yml
wchamber: I'll do it as a part of this ticket since I'm deprecating it.
Do I need to wait some amount of time after this is merged before updating the helm chart, or will any use of the new helm chart also pick up the latest WM binary?

build.gradle
wchamber: Good call, yes.

jmczerk (approved):

src/main/resources/application.yml
zloery: You'll have to wait until this version of the helm chart is rolled out everywhere before removing the helmfile value, you can see the current state of things in the [helmfile/versions dir](https://github.com/broadinstitute/terra-helmfile/tree/master/versions). I think it's usually ~1 week for releases to work their way to "prod"
The binary should be rolled out together with the chart though

PR #207, opened 2021-01-29
QA-1331: Run resiliency test through test runner.

ichengchang:
This Pull Request demonstrates writing a workspace manager resiliency tests and running the tests against a workspace manager deployment in k8s namespace (`ichang`). The code employs the latest Test Runner enhancements (v 0.0.3) and in-cluster Service Account with special RBAC roles described in the [Test Runner Service Accounts](https://docs.google.com/document/d/1-fGZqtwEUVRMmfeZfUVrn2V3M2D_eonNLepuX0ve6nM/edit?usp=sharing) document.

PR #250, opened 2021-03-11
QA-1398: GitHub Action Workflow for automating resiliency tests against a namespace.

ichengchang:
This workflow demonstrates the application of Test Runner test framework to run resiliency tests against a target namespace. After test run completes, the results are uploaded to Google bucket.


.github/workflows/test-runner-resiliency-test.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      # We can also set the env var in the GH action script to specify which server we are testing against.
+      # This will be particularly helpful once we switch over to testing against any test env -- then it's just a change to the env var.
+      # Please refer to https://github.com/DataBiosphere/terra-test-runner#override-the-server-from-the-command-line for more information.
+      - name: Run a basic Resiliency test suite against a namespace
+        run: |
+          cd workspace-manager-clienttests
+          echo "Test Runner config"
+          ./render-config.sh ${{ steps.vault-token-step.outputs.vault-token }}
+          ./render-k8s-config.sh ichang ${{ steps.vault-token-step.outputs.vault-token }}
```
marikomedlock, line 81: I don't think this should be running against your personal namespace.

.github/workflows/test-runner-resiliency-test.yml
```diff
@@ -0,0 +1,90 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Test Runner Resiliency Test
+
+on:
+  push
```
marikomedlock, line 7: This might need some discussion with the WSM group, RE which environment to test against and frequency of testing. If tests are running against a shared environment that is not enforcing single-file execution, then we may want to run this in the middle of the night to avoid conflicts.

.github/workflows/test-runner-resiliency-test.yml
```diff
@@ -0,0 +1,90 @@
...
+            vault write -field token \
+              auth/approle/login role_id=${{ secrets.VAULT_APPROLE_ROLE_ID }} \
+              secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+          echo ::set-output name=vault-token::$VAULT_TOKEN
+          echo ::add-mask::$VAULT_TOKEN
+      - name: Set permissions and render test configurations
+        run: |
+          chmod +x render_config.sh
+          ./render_config.sh ${{ steps.vault-token-step.outputs.vault-token }} dev
+          chmod +r rendered
```
marikomedlock, line 63: Do we need to run the main render_config.sh for Test Runner tests?

.github/workflows/test-runner-resiliency-test.yml
ichengchang: The `ichang` namespace is the only namespace with the `k8s` SA and `RBAC` setup. We can change this once we know which environment will be used to run `WSM` resiliency tests.

.github/workflows/test-runner-resiliency-test.yml
ichengchang: That's correct. This frequency can be changed once we know which environment `WSM` wants to run the resiliency tests and frequency of those tests.

.github/workflows/test-runner-resiliency-test.yml
ichengchang: This was a copy-paste thing. It's removed from the GH Action.

.github/workflows/test-runner-resiliency-test.yml
```diff
@@ -0,0 +1,85 @@
...
+  push
+
+jobs:
+  test-runner-resiliency:
+
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:12.3
```
jaycarlton, line 16: I think we just upgraded or are about to upgrade to 13

.github/workflows/test-runner-resiliency-test.yml
```diff
@@ -0,0 +1,85 @@
...
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
+    steps:
+      - uses: actions/checkout@v2
```
jaycarlton, line 28: nit: give this step a name too

.github/workflows/test-runner-resiliency-test.yml
```diff
@@ -0,0 +1,85 @@
...
+
+      - name: Build the WSM client
+        run: |
+          echo "Building WSM client library"
+          ./gradlew :workspace-manager-client:clean :workspace-manager-client:assemble
+          cd workspace-manager-clienttests
+          export ORG_GRADLE_PROJECT_workspacemanagerclientjar=$(find .. -type f -name "workspace-manager-client*.jar")
+          echo "ORG_GRADLE_PROJECT_workspacemanagerclientjar = ${ORG_GRADLE_PROJECT_workspacemanagerclientjar}"
+
+      # We can also set the env var in the GH action script to specify which server we are testing against.
```
jaycarlton, line 68: which env var?

workspace-manager-clienttests/build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
     // Apply the application plugin to add support for building a CLI application in Java.
     id 'application'
     // Terra Test Runner Plugin
-    id 'bio.terra.test-runner-plugin' version '0.0.3-20210311.184005-30'
+    id 'bio.terra.test-runner-plugin' version '0.0.3-20210316.131233-31'
```
jaycarlton, line 9: why not just take the latest?

jaycarlton (approved):

.github/workflows/test-runner-resiliency-test.yml
davidangb: See #258

workspace-manager-clienttests/src/main/resources/uploadlists/CompressDirectoryToTerraKernelK8S.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "CompressDirectoryToBucket",
+  "description": "Compress results directory to the terra-kernel-k8s-testrunner-results bucket",
+  "uploaderServiceAccountFile": "testrunner-iam-sa.json",
+  "uploadScripts": [
+    {
+      "name": "CompressDirectoryToBucket",
+      "description": "Save the compressed results directory to a bucket.",
+      "parameters": ["gs://terra-kernel-k8s-testrunner-results"]
```
davidangb, line 9: does this overwrite pre-existing files in the bucket, or is there any guarantee of uniqueness?

.github/workflows/test-runner-resiliency-test.yml
```diff
@@ -0,0 +1,85 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Test Runner Resiliency Test
+
+on:
+  push
```
davidangb, line 7: if you commit this as-is, it will run a resiliency test against the ichang k8s namespace on *every* push. If multiple people push to their individual PRs simultaneously, it will cause concurrent resiliency tests against the same namespace. This feels like it will cause big problems.

.github/workflows/test-runner-resiliency-test.yml
```diff
@@ -0,0 +1,85 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
```
davidangb, line 2: nit: comment is incorrect

.github/workflows/test-runner-resiliency-test.yml
```diff
@@ -0,0 +1,85 @@
...
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+
+      # These steps aren't needed for unit tests
```
davidangb, line 45: nit: is this comment relevant?

.github/workflows/test-runner-resiliency-test.yml
```diff
@@ -0,0 +1,85 @@
...
+      # This will be particularly helpful once we switch over to testing against any test env -- then it's just a change to the env var.
+      # Please refer to https://github.com/DataBiosphere/terra-test-runner#override-the-server-from-the-command-line for more information.
+      - name: Run a basic Resiliency test suite against a namespace
+        run: |
+          cd workspace-manager-clienttests
+          echo "Test Runner config"
+          ./render-config.sh ${{ steps.vault-token-step.outputs.vault-token }}
+          ./render-k8s-config.sh ichang ${{ steps.vault-token-step.outputs.vault-token }}
+          echo "Running test suite"
+          ./gradlew runTest --args="suites/BasicResiliency.json /tmp/TestRunnerResults"
```
davidangb, line 78: this requires #207, doesn't it?

davidangb (changes requested):
I'm very concerned about running potentially many concurrent resiliency tests against the  same k8s namespace.

.github/workflows/test-runner-resiliency-test.yml
marikomedlock: Okay makes sense. I think we should talk with them before this PR goes in though.

ichengchang: Updated workflow trigger to match integration test workflow.

.github/workflows/test-runner-resiliency-test.yml
ichengchang: Updated postgres image version.

.github/workflows/test-runner-resiliency-test.yml
ichengchang: Added env var name. The http link has all the information.

workspace-manager-clienttests/build.gradle
ichengchang: Updated version to `0.0.3-SNAPSHOT` which should pull latest version.

workspace-manager-clienttests/src/main/resources/uploadlists/CompressDirectoryToTerraKernelK8S.json
ichengchang: No it won't overwrite preexisting file. Each test run generates a unique id which is part of the test results archive that pushed over to GCS.

.github/workflows/test-runner-resiliency-test.yml
ichengchang: Updated the test trigger to match integration test workflow. Also changed namespace to `terra-wsmtest` as well.

.github/workflows/test-runner-resiliency-test.yml
ichengchang: Updated comment.

.github/workflows/test-runner-resiliency-test.yml
ichengchang: Yes it requires #207.

.github/workflows/test-runner-resiliency-test.yml
ichengchang: Removed this comment.

.github/workflows/test-runner-resiliency-test.yml
ichengchang: Added step name.

.github/workflows/test-runner-resiliency-test.yml
davidangb: But, the pull_request trigger you have in place will run a resiliency test every time a pull request is opened, and every time a commit is pushed to an open pull request. This still feels like too much, and with a significant risk of multiple simultaneous resiliency tests colliding.

ichengchang: Adjusted the trigger to run resiliency test once a day after midnight.

PR #251, opened 2021-03-12
[PF-530] Add an endpoint for listing workspaces

zloery:
This change adds an endpoint to list all workspaces a user has read access to (at the moment, all workspace IAM roles include read permission). This calls Sam to get a list of all workspaces a user can access (which includes Rawls workspaces), then joins that with WSM's database to get the final list of WSM workspaces a user can access.

In the future, I think we'll want to add display_name and description to `WorkspaceDescription` and include them as output here, but that also requires setting them in the create workspace call, which is out of scope for this change.


src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -273,7 +298,7 @@ String serializeGcpCloudContext(GcpCloudContext gcpCloudContext) {
   }
 
   @VisibleForTesting
-  GcpCloudContext deserializeGcpCloudContext(String json) {
+  static GcpCloudContext deserializeGcpCloudContext(String json) {
```
ddietterich, line 301: Since the Dao is a singleton bean, is the static needed?

src/main/resources/api/service_openapi.yaml
```diff
@@ -60,6 +60,19 @@ paths:
           $ref: '#/components/responses/PermissionDenied'
         '500':
           $ref: '#/components/responses/NotFound'
+    get:
```
ddietterich, line 63: I think this should have offset and limit for paging, and use a sorted query in the DAO to keep the list close to consistent. There is an example in the file already in the deprecated `enumerateReferences` endpoint.

ddietterich (changes requested):
This looks good. The one change I'd request is to put paging on the endpoint with an offset and length parameter.

jaycarlton (approved):

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -123,44 +133,39 @@ public Workspace getWorkspace(UUID id) {
       throw new MissingRequiredFieldException("Valid workspace id is required");
     }
     String sql =
-        "SELECT W.workspace_id, W.display_name, W.description, W.spend_profile,"
-            + " W.properties, W.workspace_stage, C.context"
-            + " FROM workspace W LEFT JOIN cloud_context C"
-            + " ON W.workspace_id = C.workspace_id"
+        WORKSPACE_SELECT_SQL
             + " WHERE W.workspace_id = :id AND (C.cloud_platform = 'GCP' OR C.cloud_platform IS NULL)";
```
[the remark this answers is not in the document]
ddietterich: It isn't, but this is an outer join, so for workspaces that do not have a cloud context, the database returns null.

src/main/java/bio/terra/workspace/db/WorkspaceDao.java
zloery: yes, but only so I can access this from the static RowMapper above. I could also change `serializeGcpCloudContext` for consistency, or just leave that as-is for now.

src/main/resources/api/service_openapi.yaml
zloery: done. I originally skipped this because the underlying Sam call doesn't support any pagination/filtering, but it's probably useful for WSM callers

ddietterich (approved):
lgtm

marikomedlock (approved):
LGTM from an API consumer perspective. Once this goes in, I'll make the corresponding change to support it in the CLI. Thank you!

PR #255, opened 2021-03-15
mark deprecated endpoints as deprecated

zarsky-broad:


ddietterich (commented):
Can we do this post rename merge?

zarsky-broad:
I don't think it really matters _when_ we do it, the primary effects are in the swagger UI and on the documentation on the client methods.

zarsky-broad:
Alternately @ddietterich I think it might be reasonable to roll these changes into that?

jaycarlton:
Where does the deprecated annotation show up? Just Swagger UI?

ddietterich:
OK, I added those marks into the rename PR.

zarsky-broad:
@jaycarlton I think it also adds an `@` annotation on the generated client code.

@ddietterich cool, I'll go ahead and close this!

PR #256, opened 2021-03-15
add google cloud profiler client [AS-640]

zarsky-broad:
Copied from [janitor](https://github.com/DataBiosphere/terra-resource-janitor/blob/603db78f0a985e7f64d6feaefa6896e3f1c4dce7/build.gradle#L128) and [buffer](https://github.com/DataBiosphere/terra-resource-buffer/blob/eb26012e4f745ead2c749f6ddc4acc86e953b265/build.gradle#L138)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -135,14 +135,14 @@ task unzipCloudResourceSchema(type: Copy) {
 // Download and extract the Cloud Profiler Java Agent
 ext {
 	// where to place the Cloud Profiler agent in the container
-	cloudProfilerLocation = '/opt/cprof'
+	cloudProfilerLocation = "/opt/cprof"
 
 	// location for jib extras, including the Java agent
 	jibExtraDirectory = "${buildDir}/jib-agents"
 }
 task downloadProfilerAgent(type: Download) {
 	// where to download the Cloud Profiler agent https://cloud.google.com/profiler/docs/profiling-java
-	src 'https://storage.googleapis.com/cloud-profiler/java/latest/profiler_java_agent.tar.gz'
+	src "https://storage.googleapis.com/cloud-profiler/java/latest/profiler_java_agent.tar.gz"
 	dest "${buildDir}/cprof_java_agent_gce.tar.gz"
 }
 task extractProfilerAgent(dependsOn: downloadProfilerAgent, type: Copy) {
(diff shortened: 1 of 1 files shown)


zloery (approved):

davidangb (approved):
:+1: for code.

What happens in prod if we run this but the APIs are not enabled on the Google project side? Should this wait on DDO-1057 to merge?

davidangb:
also ... I know github actions are stalled right now, but make sure everything passes before merging, of course

zarsky-broad:
TIL: https://coderwall.com/p/vkdekq/git-commit-allow-empty

zarsky-broad:
> What happens in prod if we run this but the APIs are not enabled on the Google project side? Should this wait on DDO-1057 to merge?

I'm honestly not sure. Since prod's not yet used, we could always try it and roll it back if there's an issue.

Also, I was looking at the janitor and buffer jib setups again, I just realized we were using alpine but they weren't, which is [an issue](https://cloud.google.com/profiler/docs/profiling-java#running_with_linux_alpine) since we don't want to have to customize the image, so I've switched us to a Debian base here. Also, neither was building with a blessed image, I'm not sure if there's a reason for that...?

ddietterich:
I'd like more information about the change to the base image before this goes in.

zarsky-broad:
@ddietterich We only changed off of it because we could get away with Alpine and it was lighter, we were running on this for a bit before that, so there have previously been live wsm instances running on Debian.

ddietterich:
I would also like to know why we are using cloud profiler. That comes from security, but it is not a security feature. I would like to understand what features it provides that we do not already get using prometheus.

Perhaps this is an item for our Thursday sync

zarsky-broad:
@davidangb @zloery could you re-thumb with the new base image?

zloery (approved):

davidangb:
I'd like to see what the resolution to Dan's questions are before approving, especially if something might change

ddietterich:
Resolution to my questions are:
1. Cloud Profiler is not a security requirement. It is lightweight and useful to have.
2. It is running in Janitor and Buffer services and they are also using debian, so this brings WSM into alignment rather than making it unique.
Thanks you for your patience!

davidangb (approved):
:+1: please make sure to follow this closely to prod to ensure that there are no issues with the relevant API not being enabled!

zarsky-broad:
Just verified, this appears to already be enabled on the prod google project, so I think we don't actually have a problem here.

PR #259, opened 2021-03-16
[PF-504] Add controlled resource descendentPermissions to Sam policies

zloery:
Part 2 of 4 for controlled resource IAM

This change modifies the policies WSM passes to Sam at workspace creation so that workspace IAM policies will also grant permissions on controlled resources in that workspace. The exact mapping of workspace roles to resource roles is stored in a static map in WSM, as this same mapping will be used later for cloud sync. Per design, WSM has a different mapping for each combination of accessScope/managedBy (i.e. user-shared, user-private, application-shared, application-private)


dvoet:
Why were the descendant roles removed from the sam role definition and added at the policy level? From the sam perspective we are not 100% we want to keep support for policy level descendant permissions.

zloery:
Mostly for convenience - we need to store the workspace role -> resource role inheritance mapping inside WSM for cloud sync, and using policy-based inheritance meant we could use the same mapping instead of duplicating it across WSM and Sam

dvoet:
If you are 100% sure about this mapping be static then that may be ok. Otherwise it could be burden later to update sam policies for preexisting resources when the mappings do change. I would also like to see a design for cloud IAM sync before we make decisions. Last [the sam config api](https://sam.dsde-prod.broadinstitute.org/#/Config/listResourceTypes) can be used to get the mappings if you want to use that (although the swagger looks like it needs updating to make the client lib right).

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,96 @@
...
+                  ControlledResourceIamRole.WRITER,
+                  ControlledResourceIamRole.READER),
+              // Applications are granted permissions individually, not as a role.
+              WsmIamRole.APPLICATION,
+              ImmutableList.of(),
+              WsmIamRole.READER,
+              ImmutableList.of(ControlledResourceIamRole.READER));
+
+  public static ImmutableMap<WsmIamRole, ImmutableList<ControlledResourceIamRole>>
+      USER_SHARED_MAPPING = SHARED_RESOURCE_MAPPING;
```
jaycarlton, line 44: I'd inline `SHARED_RESOURCE_MAPPING` and just make that `USER_SHARED_MAPPING`'s initialization.

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,96 @@
...
+ * roles or policies to groups. We need this functionality for WSM cloud sync, so this mapping must
+ * live in WSM instead of in Sam's resource definition.
+ *
+ * <p>This mapping is slightly different for each category of controlled resource (user-shared,
+ * user-private, app-shared, app-private), so this class holds four separate maps.
+ */
+public class ControlledResourceInheritanceMapping {
+  // Currently, shared resources have the same permission inheritance regardless of whether they
+  // are user- or application-controlled. This list is pulled out separately as a convenience.
+  private static ImmutableMap<WsmIamRole, ImmutableList<ControlledResourceIamRole>>
```
jaycarlton, line 24: Seems like you could use sets instead of lists as the value container type.

jaycarlton (approved):
Looks good, though I didn't track down the mappings to the source doc.

zloery:
> If you are 100% sure about this mapping be static then that may be ok. Otherwise it could be burden later to update sam policies for preexisting resources when the mappings do change. I would also like to see a design for cloud IAM sync before we make decisions. Last [the sam config api](https://sam.dsde-prod.broadinstitute.org/#/Config/listResourceTypes) can be used to get the mappings if you want to use that (although the swagger looks like it needs updating to make the client lib right).

The most likely change to this mapping would be adding a new workspace role, but that will require backfilling a new policy regardless of where inheritance happens. The mapping might change in the immediate future as we finish building out controlled resources, but I expect it to settle before we start caring about MC Terra workspace persistence.

We don't have a larger design for a cloud IAM sync service/library yet and afaik there are still some pretty large open design questions. For now, we're extending the existing pattern of granting cloud permissions (in this case, custom IAM roles) to Sam policy groups. We need the "workspace role -> resource role" mapping in WSM because Sam only creates groups for the workspace policies, and we need to know what resource-level GCP roles to grant each workspace-level Sam group.

ddietterich:
I thought that if we used the hierarchical resources, we could sync a Google group from the controlled resource object that would either be the same group or be the equivalent to the group on the parent. We would not have to maintain that mapping in WSM. (I may have the terminology wrong.) Is that not the case?

zloery:
IIUC when we use hierarchical resources (either role- or policy-based inheritance), the only Google groups created are for workspace policies. Sam will not create a group for e.g. "resource readers" because we don't have an explicit reader policy on the resource, it's just granted by inheritance.

Instead, we use the workspace-policy google groups plus the "workspace policy -> resource policy" mapping to grant appropriate resource roles to workspace policy groups.

dvoet:
Dan is correct. The workspace level policy (google) groups will contain the correct members. Then it is up to WSM to apply the correct google cloud roles for those policy groups depending on what kind of controlled resource it is.

There are 2 mappings at play, workspace policy -> sam role and workspace policy -> cloud IAM. I did not review this PR in detail but from the comments it appears like there is one mapping implemented that is interpreted to generate the 2 mappings. Sam already knows how to do the sam mapping. It is anticipated that cloud IAM sync will know how to do the other but the future is murky.

Sam's design is flexible so what you are doing here will work. However it is contrary to how we envisioned its use. Repeating the same descendant permissions across every workspace seems like a bad idea. I understand that expressing the mapping in 2 different places also seems like a bad idea. However, they really are 2 different mappings even if they are related.

I think I have typed all I care to at this point. I am happy to meet to clarify.

zloery:
Sounds good. I decided to use policy-based inheritance to reduce the risk of the two mappings drifting unexpectedly, but if this isn't an expected usecase on the Sam side, I think that's a much stronger argument for role-based inheritance instead, as is the risk of problematic migrations in the future.

Closing this PR as it's not the direction we want to go for inheritance. The next steps will be re-adding role-based inheritance in the Sam config and creating the Sam resources from WSM. Thanks all for the helpful input!

PR #260, opened 2021-03-16
jobid should not be empty string/whitespace [AS-679]

davidangb:
The `jobId` field, part of the create-workspace request, is too easy for an end user to get wrong by specifying the empty string.

to use a system-generated jobid, the caller must omit the jobId entirely:
```
{
  "id": "123e4567-e89b-12d3-a456-426614174000",
  "stage": "RAWLS_WORKSPACE"
}
```

but a common mistake is to leave the jobId key in the payload but blank out its value:
```
{
  "id": "123e4567-e89b-12d3-a456-426614174000",
  "jobId": "",
  "stage": "RAWLS_WORKSPACE"
}
```

which could result in many jobs attempting to have the same jobId of "", leading to conflicts.

This PR changes the behavior to validate that iif a caller specifies a jobId, its value is not the empty string nor is whitespace-only. This validation happens fairly low in `JobService`, so it affects any code using `JobService`, not just create-workspace.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/job/JobService.java
+++ b/src/main/java/bio/terra/workspace/service/job/JobService.java
@@ -24,7 +24,14 @@ import bio.terra.workspace.generated.model.ApiErrorReport;
 import bio.terra.workspace.generated.model.ApiJobReport;
 import bio.terra.workspace.generated.model.ApiJobReport.StatusEnum;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
-import bio.terra.workspace.service.job.exception.*;
+import bio.terra.workspace.service.job.exception.DuplicateJobIdException;
+import bio.terra.workspace.service.job.exception.InternalStairwayException;
+import bio.terra.workspace.service.job.exception.InvalidJobIdException;
+import bio.terra.workspace.service.job.exception.InvalidResultStateException;
+import bio.terra.workspace.service.job.exception.JobNotCompleteException;
+import bio.terra.workspace.service.job.exception.JobNotFoundException;
+import bio.terra.workspace.service.job.exception.JobResponseException;
+import bio.terra.workspace.service.job.exception.JobUnauthorizedException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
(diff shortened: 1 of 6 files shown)


ddietterich:
This looks fine to me. I don't understand why the jobId is even there. I think we should get rid of it.
First of all, CreateWorkspace is synchronous, so no polling by jobId is ever needed.
Second, the caller passes in the workspaceId which is sufficient for idempotency.

So an alternative would be to remove the property or deprecate it, and ignore it altogether.

davidangb:
@ddietterich hm, good point - what do you think of leaving the validation in at the `JobService` level, whether or not we change the behavior of create-workspace?

ddietterich:
Yes, you are right. And we do take jobids within JobControl objects and those should certainly not be whitespace.
I'll made a ticket to deprecate the jobId on create workspace. https://broadworkbench.atlassian.net/browse/PF-576

src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -158,6 +152,11 @@ public JobBuilder newJob(
       Class<? extends Flight> flightClass,
       Object request,
       AuthenticatedUserRequest userReq) {
+
+    // If clients provide a non-null job ID, it cannot be whitespace-only
+    if (StringUtils.isWhitespace(jobId))
+      throw new InvalidJobIdException("jobId cannot be whitespace-only.");
```
ddietterich, line 158: The if should have braces

src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -24,13 +24,7 @@
...
 import bio.terra.workspace.generated.model.ApiJobReport.StatusEnum;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
-import bio.terra.workspace.service.job.exception.DuplicateJobIdException;
-import bio.terra.workspace.service.job.exception.InternalStairwayException;
-import bio.terra.workspace.service.job.exception.InvalidResultStateException;
-import bio.terra.workspace.service.job.exception.JobNotCompleteException;
-import bio.terra.workspace.service.job.exception.JobNotFoundException;
-import bio.terra.workspace.service.job.exception.JobResponseException;
-import bio.terra.workspace.service.job.exception.JobUnauthorizedException;
+import bio.terra.workspace.service.job.exception.*;
```
ddietterich, line 27: Standard practice is no `*`. You have to make settings in IntelliJ for it to not do that. I can give you the info if you need it.

ddietterich (approved):
Two nits - otherwise looks good

MatthewBemis (approved):
LGTM

PR #263, opened 2021-03-18
[PF-579] Create controlled resources in Sam

zloery:
This change fills in the step to create controlled resources in Sam. This assumes we're using role-based inheritance defined in Sam's configuration. To support private users, I've also added empty policies to private resources for everything we might modify via re-assignment. We'll likely need to add similar empty policies for application-managed resources in the future, but I've disallowed that in validation until we have a clearer use case.

This also adds an owner policy on all controlled resources whose sole member is the WSM service account. This is intended as an admin control for lost resources, though we plan to switch to Sam's admin controls as those are built. As part of this, I've also added registering the WSM service account in Sam as an initialization step.

This also includes a change which allows us to specify a project for GCS bucket creation. Previously it was unspecified and would default to the same project the WSM service was running in instead of the user's workspace context. This isn't strictly required for this PR, but it seemed relevant to other threads at the moment so I included it here.


src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
```diff
@@ -61,14 +74,22 @@ public StepResult doStep(FlightContext flightContext)
...
     return StepResult.getStepResultSuccess();
   }
 
   @Override
   public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
-    final StorageCow storageCow = crlService.createStorageCow(userRequest);
+    String gcpProjectId =
+        workspaceDao
+            .getGcpCloudContext(resource.getWorkspaceId())
+            .orElseThrow(
```
jaycarlton, line 87: Usually the style is to `.map(GcpCloudContext::getGcpProjectId).orElseThrow(...)`. I've always seen the throw come last.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1301,14 +1301,29 @@ components:
...
+    PrivateResourceIamRole:
+      description: Description of role(s) granted to the user of a private resource
+      type: object
+      required: [role, editor]
+      properties:
+        role:
+          description: What access role (reader or writer) this user has
+          type: string
+          enum: ['READER', 'WRITER']
+        editor:
```
jaycarlton, line 1322: nit: This seems a bit ad hoc, but it should work. The business rule that says readers can't be writers but they can be editors seems like something I wouldn't bake into the API. So I guess  it's fine if this is a fundamental abstraction/rule.

src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
```diff
@@ -231,6 +241,99 @@ void ListWorkspacesIncludesWsmWorkspace() {
     assertTrue(samWorkspaceIdList.contains(workspaceId));
   }
 
+  @Test
+  void WorkspaceReaderIsSharedResourceReader() {
```
jaycarlton, line 245: Why capitalize the test method names?

jaycarlton (approved):

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -69,6 +83,50 @@ private UsersApi samUsersApi(String accessToken) {
...
+  private String getWsmServiceAccountToken() throws IOException {
+    GoogleCredentials creds =
+        GoogleCredentials.getApplicationDefault().createScoped(SAM_OAUTH_SCOPES);
+    creds.refreshIfExpired();
+    return creds.getAccessToken().getTokenValue();
+  }
+
+  /**
+   * Register WSM's service account as a user in Sam if it isn't already. This is called on every
+   * WSM startup, but should only need to register with Sam once per environment.
```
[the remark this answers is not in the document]
zloery: I think the "check registration, register if it isn't already" works well as a once-per-environment pattern. 
With multi-instance there's a risk of a small race condition here, but the first pod restart would fix it permanently for the env. I'm not sure we need to get into anything more complicated

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -69,6 +83,50 @@ private UsersApi samUsersApi(String accessToken) {
...
+   */
+  public void initialize() {
+    String wsmAccessToken;
+    try {
+      wsmAccessToken = getWsmServiceAccountToken();
+    } catch (IOException e) {
+      // In cases where WSM is not running as a service account (e.g. unit tests), the above call
```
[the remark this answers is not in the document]
zloery: not that I know of - the same call that reads the token is the one throwing

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -69,6 +83,50 @@ private UsersApi samUsersApi(String accessToken) {
...
+      return;
+    }
+
+    UsersApi usersApi = samUsersApi(wsmAccessToken);
+    try {
+      // getUserStatusInfo throws a 404 if the calling user is not registered, which will happen
+      // the first time WSM is run in each environment.
+      usersApi.getUserStatusInfo();
+      logger.info("WSM service account already registered in Sam");
+    } catch (ApiException userStatusException) {
```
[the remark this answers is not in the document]
zloery: nice, I like that structure a little better.

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -69,6 +83,50 @@ private UsersApi samUsersApi(String accessToken) {
...
+      if (userStatusException.getCode() == HttpStatus.NOT_FOUND.value()) {
+        registerWsmServiceAccount(usersApi);
+      }
```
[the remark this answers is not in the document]
zloery: whoops

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -273,6 +331,56 @@ public String syncWorkspacePolicy(
     }
   }
 
+  @Traced
+  public void createControlledResource(
+      ControlledResource resource,
+      Optional<ApiPrivateResourceIamRole> privateIamRole,
```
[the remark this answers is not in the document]
zloery: fixed and added javadoc

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -313,25 +421,81 @@ public ApiSystemStatusSystems status() {
...
+      AccessPolicyMembershipV2 ownerPolicy =
+          new AccessPolicyMembershipV2()
+              .addRolesItem(ControlledResourceIamRole.OWNER.toSamRole())
+              .addMemberEmailsItem(wsmSaEmail);
+      request.putPoliciesItem(ControlledResourceIamRole.OWNER.toSamRole(), ownerPolicy);
+    } catch (IOException e) {
+      // In cases where WSM is not running as a service account (e.g. unit tests), the above call to
+      // get application default credentials will fail. This is fine, as those cases don't create
+      // real resources.
+      logger.warn(
```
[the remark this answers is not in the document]
zloery: not that I know of. The only call that can fail here is reading local credentials though, so my gut feeling is that a logged warning is enough mitigation here.

src/main/java/bio/terra/workspace/service/iam/model/ControlledResourceIamRole.java
```diff
@@ -0,0 +1,25 @@
+package bio.terra.workspace.service.iam.model;
+
+/**
+ * Internal representation of resource-level IAM roles. See {@Code
+ * ControlledResourceInheritanceMapping} for the canonical mapping of workspace IamRoles to
+ * equivalent ControlledResourceIamRoles.
+ */
+public enum ControlledResourceIamRole {
+  OWNER("owner"),
```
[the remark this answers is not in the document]
zloery: They're duplicates here, but that matches the way they're defined in Sam. 
The name of the workspace reader/writer/owner role is independent of the name of the resource reader/writer/owner role, and I think this makes that clearer.

src/main/java/bio/terra/workspace/service/iam/model/SamConstants.java
```diff
@@ -9,4 +13,64 @@
   public static final String SAM_WORKSPACE_READ_IAM_ACTION = "read_policies";
   public static final String SPEND_PROFILE_RESOURCE = "spend-profile";
   public static final String SPEND_PROFILE_LINK_ACTION = "link";
+  public static final String SAM_CONTROLLED_USER_SHARED_RESOURCE =
```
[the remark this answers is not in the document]
zloery: I think that's reasonable given the context.
This also made me rethink including the `SAM_` prefix given that these are all defined in `SamConstants`

src/main/java/bio/terra/workspace/service/iam/model/SamConstants.java
```diff
@@ -9,4 +13,64 @@
...
+      "controlled-user-shared-workspace-resource";
+  public static final String SAM_CREATE_CONTROLLED_USER_SHARED_ACTION =
```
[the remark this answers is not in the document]
zloery: It is kind of messy at the moment. I switched to something like `WorkspaceFlightMapKeys` to organize things a little better, ptal

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
zloery: you're right, that feels neater

src/main/resources/api/service_openapi.yaml
zloery: At least currently this is how we've defined private resource roles - either reader or writer (which also includes `read` permission), and independently possibly editor.

src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
zloery: ...good question

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -41,12 +42,14 @@ public ControlledResourceService(
   public String createControlledResource(
       ControlledResource resource,
       ApiGcsBucketCreationParameters creationParameters,
-      ApiIamRole creationApiIamRole,
+      ApiPrivateResourceIamRole privateResourceIamRole,
```
ddietterich, line 45: In my branch, I translated the API IAM into a WSM IAM type. I thought we wanted to avoid using the API types and convert them to WSM internal types in the controller. If you agree, then I can fix this up when I resolve conflicts on my PR.

src/main/java/bio/terra/workspace/service/iam/SamService.java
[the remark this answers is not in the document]
jaycarlton: Oh so the same SAM instance is talking to multiple Terra environments?

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
```diff
@@ -53,6 +59,14 @@ public StepResult doStep(FlightContext flightContext)
     FlightMap inputMap = flightContext.getInputParameters();
     ApiGcsBucketCreationParameters creationParameters =
         inputMap.get(CREATION_PARAMETERS, ApiGcsBucketCreationParameters.class);
+    String gcpProjectId =
```
ddietterich, line 62: It doesn't hurt to have this here, but the check has already been done in the metadata step in ResourceDao around line 240.

src/main/resources/api/service_openapi.yaml
ddietterich: Another alternative is to accept a list of roles. If both READER and WRITER are specified, we choose writer. That would leave us a plain enumeration of possible roles. Seems more extensible to me.

src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
```diff
@@ -214,15 +225,15 @@ void GrantRoleInMissingWorkspaceThrows() {
...
+  void readRolesInMissingWorkspaceThrows() {
     UUID fakeId = UUID.randomUUID();
     assertThrows(
         WorkspaceNotFoundException.class,
         () -> samService.listRoleBindings(fakeId, defaultUserRequest()));
   }
 
   @Test
-  void ListWorkspacesIncludesWsmWorkspace() {
+  void listWorkspacesIncludesWsmWorkspace() {
```
ddietterich, line 236: Thanks for the cleanup

src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
```diff
@@ -231,6 +242,97 @@ void ListWorkspacesIncludesWsmWorkspace() {
     assertTrue(samWorkspaceIdList.contains(workspaceId));
   }
 
+  @Test
```
ddietterich, line 245: Do we need tests for the auto-registration code or is it just exercised all of the time?

ddietterich (approved):
A few minor things. LGTM

src/main/java/bio/terra/workspace/service/iam/model/ControlledResourceIamRole.java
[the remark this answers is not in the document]
jaycarlton: Sounds good. We don't want to externalize too much Sam logic.

src/main/java/bio/terra/workspace/service/iam/model/SamConstants.java
```diff
@@ -9,4 +13,73 @@
...
+              "Sam resource name not specified for access scope %s and ManagedByType %s",
+              accessScope.toString(), managedBy.toString()));
+    }
+
+    private SamControlledResourceNames() {}
+  }
+
+  public static class SamControlledResourceCreateActions {
+
+    public static final String CREATE_CONTROLLED_USER_SHARED_ACTION =
```
jaycarlton, line 53: You don't need the `_ACTION` suffix anymore since it's in the class name.

jaycarlton (approved):

src/main/java/bio/terra/workspace/service/iam/SamService.java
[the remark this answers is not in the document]
zloery: I was thinking about the case of multiple WSM replicas in a single env, where each replica independently runs its own initialization. There's a 1:1 of WSM to Sam deployments in each env, though a WSM deployment might have multiple instances. I think you have more context on the multi-instance work though - let me know if something feels off here

src/main/resources/api/service_openapi.yaml
zloery: Went with the list of roles approach here for extensibility

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
zloery: Yeah, this should be translated to internal types. The existing implementation is slightly off since this should be resource IAM roles instead of workspace IAM roles - I'll switch the interface + wiring up

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
zloery: Yep - I had to pull the value of of optional here and it seemed useful to have an informative error message, but this could also be a "this should never happen" error.

src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
zloery: It gets exercised every time WSM starts, added a test to verify registration status

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -68,13 +74,26 @@ public ControlledGcpResourceApiController(
             .bucketName(body.getGcsBucket().getName())
             .build();
 
+    // This value is serialized in Stairway, so we avoid making a generic list.
+    ControlledResourceIamRoleList privateRoles =
```
zloery, line 78: Is there a standard pattern for serializing generics like Lists in stairway? Deserializing as `List.class` works, but gives a lot of warnings about type safety. This seemed like a reasonable alternative (if a little heavy on boilerplate), but open to other approaches

zloery:
Re-adding you @ddietterich  as the changes to the private IAM role representation were nontrivial and might cause conflict in your PR

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
ddietterich: It is fine. If someone goes are re-orders the steps, the check will still be enforced.

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
ddietterich: I have always used List.class, but this seems cleaner to me

ddietterich (approved):
LGTM - thanks

zloery:
Publishing version `0.18.0` of the client library from this as it changed the interface for private resources.

PR #264, opened 2021-03-21
slack notifications for alpha/staging tests [AS-694]

davidangb:
add Slack notifications to #dsde-qa, using https://github.com/broadinstitute/action-slack. Code shamelessly copied from data repo.

Tested by manually changing the `if:` clause in the job to `if: always()` and seeing the notifications arrive in #dsde-qa. The [Jenkins job](https://fc-jenkins.dsp-techops.broadinstitute.org/job/workspacemanager-integration-test-trigger/configure) that triggers the alpha and staging tests is hardcoded to use the `dev` branch of this repo, so I can't test the real thing easily; I don't want to edit the Jenkins job.

if you don't like the :mariostar: emoji I can change that :)
Screen Shot 2021-03-21 at 5 58 34 PM

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -155,5 +155,4 @@ jobs:
         status: ${{ job.status }}
         channel: "#dsde-qa"
         username: "Workspace Manager tests"
-        text: "Integration tests"
         fields: repo,job,workflow,commit,eventName,author,took
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -144,12 +144,10 @@ jobs:
         path: build/reports/tests
 
     - name: "Notify QA Slack"
-#      if: steps.set-env-step.outputs.test-env == 'alpha' || steps.set-env-step.outputs.test-env == 'staging'
-      if: ${{ always() }}
+      if: steps.set-env-step.outputs.test-env == 'alpha' || steps.set-env-step.outputs.test-env == 'staging'
       uses: broadinstitute/action-slack@v3.8.0
       # see https://github.com/broadinstitute/action-slack
       env:
-#        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -142,3 +142,18 @@ jobs:
...
+      # see https://github.com/broadinstitute/action-slack
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        MATRIX_CONTEXT: ${{ toJson(matrix) }}
+      with:
+        status: ${{ job.status }}
+        channel: "#dsde-qa"
+        username: "Workspace Manager tests"
+        text: "Integration tests"
```
ansingh7115, line 158: This says "integration tests" for unit tests, connected tests and integration tests. Perhaps `text` should be specific to the type of tests that are running? We could also just remove `text `or make it something more generic?

davidangb: @ansingh7115 good call, I removed the `text` parameter altogether!

ansingh7115 (approved):

.github/workflows/test.yml
```diff
@@ -142,3 +142,17 @@ jobs:
       with:
         name: Test Reports
         path: build/reports/tests
+
+    - name: "Notify QA Slack"
+      if: steps.set-env-step.outputs.test-env == 'alpha' || steps.set-env-step.outputs.test-env == 'staging'
+      uses: broadinstitute/action-slack@v3.8.0
+      # see https://github.com/broadinstitute/action-slack
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
zarsky-broad, line 151: Do we still need this, or is it just left over from the old version? See https://action-slack.netlify.app/migration/01-migrate-from-v2-to-v3

zarsky-broad (approved):
Seems legit, 1 question

.github/workflows/test.yml
davidangb: @zarsky-broad good call, removed `GITHUB_TOKEN`

PR #265, opened 2021-03-22
[PF-529] Add functionality to update a workspace.

tnadler:
- Define new API for UpdateWorkspace.
- Add displayName and description fields as needed.
- Implement UpdateWorkspace in the controller and the service.

### DIFF
--- a/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -19,6 +19,7 @@ import bio.terra.workspace.generated.model.ApiReferenceTypeEnum;
 import bio.terra.workspace.generated.model.ApiRoleBinding;
 import bio.terra.workspace.generated.model.ApiRoleBindingList;
 import bio.terra.workspace.generated.model.ApiUpdateDataReferenceRequestBody;
+import bio.terra.workspace.generated.model.ApiUpdateWorkspaceRequestBody;
 import bio.terra.workspace.generated.model.ApiWorkspaceDescription;
 import bio.terra.workspace.generated.model.ApiWorkspaceDescriptionList;
 import bio.terra.workspace.generated.model.ApiWorkspaceStageModel;
@@ -120,6 +121,8 @@ public class WorkspaceApiController implements WorkspaceApi {
             .jobId(jobId)
             .spendProfileId(spendProfileId)
             .workspaceStage(internalStage)
+            .displayName(Optional.ofNullable(body.getDisplayName()))
+            .description(Optional.ofNullable(body.getDescription()))
             .build();
     UUID createdId = workspaceService.createWorkspace(internalRequest, userReq);
 
@@ -152,7 +155,9 @@ public class WorkspaceApiController implements WorkspaceApi {
         .id(workspace.getWorkspaceId())
         .spendProfile(workspace.getSpendProfileId().map(SpendProfileId::id).orElse(null))
         .stage(workspace.getWorkspaceStage().toApiModel())
-        .gcpContext(gcpContext);
+        .gcpContext(gcpContext)
+        .displayName(workspace.getDisplayName().orElse(null))
+        .description(workspace.getDescription().orElse(null));
   }
 
   @Override
@@ -161,19 +166,22 @@ public class WorkspaceApiController implements WorkspaceApi {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     logger.info("Getting workspace {} for {}", id, userReq.getEmail());
     Workspace workspace = workspaceService.getWorkspace(id, userReq);
-    ApiGcpContext gcpContext =
-        workspace.getGcpCloudContext().map(GcpCloudContext::toApi).orElse(null);
-
-    // Note projectId will be null here if no GCP cloud context exists.
-    // When we have another cloud context, we will need to do a similar retrieval for it.
-    ApiWorkspaceDescription desc =
-        new ApiWorkspaceDescription()
-            .id(workspace.getWorkspaceId())
-            .spendProfile(workspace.getSpendProfileId().map(SpendProfileId::id).orElse(null))
-            .stage(workspace.getWorkspaceStage().toApiModel())
-            .gcpContext(gcpContext);
-
+    ApiWorkspaceDescription desc = buildWorkspaceDescription(workspace);
     logger.info("Got workspace {} for {}", desc, userReq.getEmail());
+
+    return new ResponseEntity<>(desc, HttpStatus.OK);
+  }
+
+  @Override
+  public ResponseEntity<ApiWorkspaceDescription> updateWorkspace(
+      @PathVariable("workspaceId") UUID id, @RequestBody ApiUpdateWorkspaceRequestBody body) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    logger.info("Updating workspace {} for {}", id, userReq.getEmail());
+    Workspace workspace =
+        workspaceService.updateWorkspace(userReq, id, body.getDisplayName(), body.getDescription());
+
+    ApiWorkspaceDescription desc = buildWorkspaceDescription(workspace);
+    logger.info("Updated workspace {} for {}", desc, userReq.getEmail());
 
     return new ResponseEntity<>(desc, HttpStatus.OK);
   }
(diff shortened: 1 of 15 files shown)


workspace-manager-clienttests/src/main/resources/configs/integration/WorkspaceLifecycle.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "CreateGetDeleteWorkspace",
```
zloery, line 2: A couple extra changes to make with this:

- This file is the integration test case, but we have a similar test for performance with different parameters at workspace-manager-clienttests/src/main/resources/configs/perf/CreateGetDeleteWorkspace.json. could you update that as well?
- This file (and the perf test) is referenced by a test suite, please update the name there as well (workspace-manager-clienttests/src/main/resources/suites/BasicIntegration.json and BasicPerf.json)

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -161,19 +166,22 @@ private ApiWorkspaceDescription buildWorkspaceDescription(Workspace workspace) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     logger.info("Getting workspace {} for {}", id, userReq.getEmail());
     Workspace workspace = workspaceService.getWorkspace(id, userReq);
-    ApiGcpContext gcpContext =
-        workspace.getGcpCloudContext().map(GcpCloudContext::toApi).orElse(null);
+    ApiWorkspaceDescription desc = buildWorkspaceDescription(workspace);
```
zloery, line 169: Whoops - not sure how I missed this earlier. good catch!

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -161,19 +166,22 @@ private ApiWorkspaceDescription buildWorkspaceDescription(Workspace workspace) {
...
-            .spendProfile(workspace.getSpendProfileId().map(SpendProfileId::id).orElse(null))
-            .stage(workspace.getWorkspaceStage().toApiModel())
-            .gcpContext(gcpContext);
+    return new ResponseEntity<>(desc, HttpStatus.OK);
+  }
 
-    logger.info("Got workspace {} for {}", desc, userReq.getEmail());
+  @Override
+  public ResponseEntity<ApiWorkspaceDescription> updateWorkspace(
+      @PathVariable("workspaceId") UUID id, @RequestBody ApiUpdateWorkspaceRequestBody body) {
```
zloery, line 177: nit: first parameter should be `workspaceId` to match the underlying interface. 
I think the parameter name was changed without updating existing controller methods, so there are probably inconsistencies elsewhere in this file

zloery (commented):
Looks good, a few extra changes needed for the client tests.
The new connected tests are also failing, though I didn't dig into causes to see how real those were

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
ddietterich: Yes, I think I did the openApi change, but didn't change the code.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -126,6 +127,22 @@ public Workspace getWorkspace(UUID id, AuthenticatedUserRequest userReq) {
...
+   * @param workspaceId workspace of interest
+   * @param name name to change - may be null
+   * @param description description to change - may be null
+   */
+  public Workspace updateWorkspace(
+      AuthenticatedUserRequest userReq,
+      UUID workspaceId,
+      @Nullable String name,
+      @Nullable String description) {
+    validateWorkspaceAndAction(userReq, workspaceId, SamConstants.SAM_WORKSPACE_WRITE_ACTION);
```
ddietterich, line 141: I think we want to get the workspace and report any 404 before we do the validate. This is an extra step, but is nicer to the caller and is the convention.
@zloery check me on this, please

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -209,6 +209,58 @@ void testWithSpendProfile() {
...
+
+    String otherDescription = "The deprecated workspace";
+
+    Workspace secondUpdatedWorkspace =
+        workspaceService.updateWorkspace(USER_REQUEST, workspaceId, null, otherDescription);
+
+    // Since name is null, leave it alone. Description should be updated.
+    assertEquals(name, secondUpdatedWorkspace.getDisplayName());
+    assertEquals(otherDescription, secondUpdatedWorkspace.getDescription());
+
```
ddietterich, line 256: Maybe test the both passed as null should result in an error.

workspace-manager-clienttests/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
```diff
@@ -31,6 +32,12 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
     assertThat(workspaceDescription.getId(), equalTo(workspaceId));
     assertThat(workspaceDescription.getStage(), equalTo(WorkspaceStageModel.MC_WORKSPACE));
 
+    UpdateWorkspaceRequestBody updateBody = new UpdateWorkspaceRequestBody().displayName("name").description("desc");
+    WorkspaceDescription updatedDescription = workspaceApi.updateWorkspace(updateBody, workspaceId);
+    ClientTestUtils.assertHttpSuccess(workspaceApi, "PATCH workspace");
+    assertThat(updatedDescription.getDisplayName(), equalTo("name"));
```
ddietterich, line 38: Maybe make the strings static?

ddietterich (commented):
Mostly nits; one change to check existence before access in the update.

workspace-manager-clienttests/src/main/resources/configs/integration/WorkspaceLifecycle.json
tnadler: Done (I think)

src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
tnadler: Done

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
tnadler: I am not sure I understand... This is the function that getWorkspace calls?

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
tnadler: Done

workspace-manager-clienttests/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
tnadler: Done

tnadler:
Addressed comments and actually needed a bit more logic in the create. Current connected test failure seems spurious. Would love your thoughts on the Optional handling in the create.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -78,6 +79,10 @@ public UUID createWorkspace(WorkspaceRequest workspaceRequest, AuthenticatedUser
     createJob.addParameter(
         WorkspaceFlightMapKeys.WORKSPACE_STAGE, workspaceRequest.workspaceStage());
 
+    createJob.addParameter(
```
ddietterich, line 82: Instead of passing each element of the request into the flight, we should pass the whole `workspaceRequest` as the request - instead of the `null` that is floating there in the `newJob()` call.

If there are serdes issues, they should be addressed in the WorkspaceRequest object. This simplifies request changes going forward. It can be done as a separate PR. If you choose that, can you make a ticket?

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
ddietterich: Nevermind The validate call does the right thing.

ddietterich (approved):
One comment - probably not for this PR. Up to you.
LGTM

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
tnadler: Sure. I will file a follow up ticket so I can get this submitted.

zloery (approved):

PR #266, opened 2021-03-22
AS-695: Document hotfix procedure

MatthewBemis:
I think the original ticket suggests to copy the HOTFIX.md file into this repo but I would rather defer to the common documentation in the terra-helmfile repo since I have a feeling that it's more likely to be kept up to date. I don't think the WSM process is different than what's in terra-helmfile, anyway.

I'm open to discussing disagreements :)

### DIFF
--- a/HOTFIX.md
+++ b/HOTFIX.md
@@ -1,5 +1,13 @@
 # Hotfix Process
 
-See the [HOTFIX.md](https://github.com/broadinstitute/terra-helmfile/blob/master/docs/HOTFIX.md) documentation in [terra-helmfile](https://github.com/broadinstitute/terra-helmfile) for the general MC-Terra hotfix procedure, which applies to Workspace Manager
+1) Build the image that you would like to use for your hotfix:
+
+`./gradlew jibDockerBuild --image=gcr.io/terra-kernel-k8s/terra-workspace-manager:hotfix-<DATE> -Djib.console=plain`
+
+2) Push the image that you've created to GCR:
+
+`docker push gcr.io/terra-kernel-k8s/terra-workspace-manager:hotfix-<DATE>`
+
+3) To apply the hotfix, see the [HOTFIX.md](https://github.com/broadinstitute/terra-helmfile/blob/master/docs/HOTFIX.md) documentation in [terra-helmfile](https://github.com/broadinstitute/terra-helmfile) for the general MC-Terra hotfix procedure, which applies to Workspace Manager. Use the image that you built above when applying the hotfix.
 
 Should the hotfix process deviate from the general procedure outlined in terra-helmfile, this document should be updated to highlight those differences.
--- a/HOTFIX.md
+++ b/HOTFIX.md
@@ -15,7 +15,7 @@ docker push gcr.io/terra-kernel-k8s/terra-workspace-manager:hotfix-<DATE>
 3) Follow Step 1 in terra-helmfile/[HOTFIX.md](https://github.com/broadinstitute/terra-helmfile/blob/master/docs/HOTFIX.md)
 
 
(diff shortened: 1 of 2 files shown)


zarsky-broad:
Perhaps creating a hotfix.md that just contains a link, instead of just having it in the readme? A file stands out more to me.

MatthewBemis:
@zarsky-broad sure, I buy that

davidangb:
@ansingh7115 you wrote the original ticket, which I believe is a compliance/QA issue, do you have input on what approach we need to take here?

davidangb:
From the terra-helmfile hotfix doc: "Build new container image(s) for your hotfix following your team's documented procedures." ... what is the procedure to build the container image for WSM? Our hotfix.md probably needs that info.

MatthewBemis:
I've updated  the doc with build instructions, which I've extracted from [build.yml](https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/.github/workflows/build.yml)

HOTFIX.md
```diff
@@ -0,0 +1,13 @@
+# Hotfix Process
+
+1) Build the image that you would like to use for your hotfix:
+
+`./gradlew jibDockerBuild --image=gcr.io/terra-kernel-k8s/terra-workspace-manager:hotfix-<DATE> -Djib.console=plain`
```
zarsky-broad, line 5: Don't need to override jib.console, that's just to stop it from trying to show a progress bar in CI. You might also put these in proper code blocks, like
\`\`\`sh
./gradlew jibDockerBuild --image=gcr.io/terra-kernel-k8s/terra-workspace-manager:hotfix-<DATE>
\`\`\`
so they look like
```sh
./gradlew jibDockerBuild --image=gcr.io/terra-kernel-k8s/terra-workspace-manager:hotfix-<DATE>
```

davidangb (approved):
lgtm, modulo Isaac's comments ... I think you need @ansingh7115 's review or at least her deferral of a review

zarsky-broad (commented):
Should we mention `gcloud auth configure-docker`?

ansingh7115:
I think just having the link in our README is probably better - but I'm going to check with QA first!

ansingh7115:
Conclusion from talking to QA: this approach should be fine.

However, there are two steps in the original HOTFIX.md that don't apply to us. The steps that say to Sync the ArgoCD deployment would trigger a deploy but not test runs against alpha and staging. Instead, it would be better to trigger this Jenkins job: https://fc-jenkins.dsp-techops.broadinstitute.org/job/gke-deploy/  This will trigger the deploy as well as the tests.

choover-broad (approved):
Deployment steps are fine!

I'd have QA review this process and documentation, too, since they know more about our compliance requirements.

Not loving the manual image build. Would prefer a flow like:
* developer creates PR branch that starts with `hotfix-<date>`, adds `DO NOT MERGE` label
* image is automatically built and pushed to GCR with a tag like `hotfix-<date>-<commithash>`

If you want I can create a DDO ticket for that, although squeezing it in before WSM's prod launch will probably be tough... 🙃

davidangb:
Could this rely on the `build.yml` action to generate a hotfix docker image?

MatthewBemis:
@choover-broad agreed and definitely no need to get that working before WSM launch. As long as we have a workable process now I think we're good. If you want to make a DDO ticket for that, that would be awesome. We can always move that ticket to our board too if that feels appropriate.

@davidangb we could probably modify that job to detect the things that Chelsea mentioned. I'd prefer to do that either as a DDO ticket or as an App Services ticket, as a follow-on

zarsky-broad (approved):

PR #268, opened 2021-03-23
PF-522: Use StairwayComponent in WSM

jaycarlton:
Split the stairway initialization out of JobService and gave it its own StairwayComponent.

I'm initializing stairway in the startup initializer now, the same way I'm doing for buffer service.


src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -260,23 +249,17 @@ public FlightState call() throws Exception {
...
+        flightBeanBag,
+        ImmutableList.of(mdcHook, new TracingHook()));
   }
 
   @Traced
   public void releaseJob(String jobId, AuthenticatedUserRequest userReq) {
     try {
       verifyUserAccess(jobId, userReq); // jobId=flightId
-      stairway.deleteFlight(jobId, false);
+      stairwayComponent.get().deleteFlight(jobId, false);
```
jaycarlton, line 262: should the stairway component just expose these functions instead of exposing the stairway object with get()?

src/main/java/bio/terra/workspace/app/Main.java
```diff
@@ -14,12 +14,16 @@
...
+      // Dependencies for Stairway
+      "bio.terra.common.kubernetes",
       // Scan for logging-related components & configs
       "bio.terra.common.logging",
       // Scan for Liquibase migration components & configs
       "bio.terra.common.migrate",
       // Scan for tracing-related components & configs
       "bio.terra.common.tracing",
+      // Stairway
+      "bio.terra.common.stairway",
```
yonghaoy, line 26: nit:order by alphabetical

src/main/java/bio/terra/workspace/service/job/JobService.java
yonghaoy: I don't suggest that. Feels we are putting to much into stairwayComponent, and make it become part of 'Stairway library' which is unnecessary. And if Stairway method changes, we have to upgrade stairwayComponent correspondingly. 
Using stairwayComponent just for initialization and health status should be good enough.

src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -69,28 +69,16 @@ public JobService(
...
       MdcHook mdcHook,
-      ObjectMapper objectMapper) {
+      StairwayComponent stairwayComponent,
+      FlightBeanBag flightBeanBag) {
     this.jobConfig = jobConfig;
     this.ingressConfig = ingressConfig;
     this.stairwayDatabaseConfiguration = stairwayDatabaseConfiguration;
     this.executor = Executors.newScheduledThreadPool(jobConfig.getMaxThreads());
     this.mdcHook = mdcHook;
-    StairwayExceptionSerializer serializer = new StairwayExceptionSerializer(objectMapper);
```
yonghaoy, line 80: then you lose StairwayExceptionSerializer. 
You may have to change method common-lib's initialize method. (maybe passing Stairway.Builder() if we are seeing too many parameters)?
@ddietterich wdyt?

src/main/java/bio/terra/workspace/service/job/JobService.java
jaycarlton: 👍🏻

src/main/java/bio/terra/workspace/service/job/JobService.java
ddietterich: We definitely need a way to supply an exception serializer and override the Stairway built-in one.
If the param list gets too long, we can make an initializeStairway object with a builder and pass that in.

src/main/java/bio/terra/workspace/service/job/JobService.java
ddietterich: I agree with @yonghaoy.
We could capture the Stairway object into a `private final Stairway stairway` from the stairwayComponent in the JobService constructor and keep using the `stairway` variable, right?

src/main/resources/application.yml
```diff
@@ -73,7 +73,16 @@ workspace:
...
-    samplingRate: ${env.tracing.samplingRate}
+terra:
+  common:
+    tracing:
+        stackdriverExportEnabled: ${env.tracing.exportEnabled}
+        samplingRate: ${env.tracing.samplingRate}
+    stairway:
+      cluster-name-suffix: workspace-stairway
+      force-clean-start: false
+      max-parallel-flights: 500
```
ddietterich, line 84: Probably more than we need. I'd suggest 50. It is per node and we probably don't want more than 150 threads competing for access to the backend database server.

ddietterich (commented):
This is looking good. I think we need a solution for the exception serialization before we callit done.

src/main/java/bio/terra/workspace/service/job/JobService.java
jaycarlton: You can't initialize a final field except in the constructor, or at least it's a warning.

I'm concerned we don't have a clean abstraction. Either `StairwayComponent` owns `Stairway`, or it doesn't. I don't like `get()`, but I like saving the result of a `get()` even less. It doesn't _belong_ to `JobService`.

Should that method be removed from `StairwayComponent`, and we just return a `Stairway` from the initialize method so it's like a factory? That approach wouldn't make sense with the status, unless we assume `Stairway` really is a singleton.

src/main/java/bio/terra/workspace/service/job/JobService.java
jaycarlton: Do we need this before merging this (and/or before the release)?

ddietterich: The exception serializer improves the fidelity of exceptions caught with Stairway. But the default one in Stairway will deliver messages. So I vote not to let this hold up the merge. We can do it as a follow-on.

Let's discuss separately whether we can have one shared serializer for MC Terra based on the common exceptions or if we need to allow for individual ones per component.

src/test/resources/application-test.yml
```diff
@@ -30,7 +30,15 @@ workspace:
     password: dbpwd
     uri: jdbc:postgresql://127.0.0.1:5432/testdb
     username: dbuser
-
-terra.common:
+terra:
+  common:
+    stairway:
+      force-clean-start: true
```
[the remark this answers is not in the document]
yonghaoy: Buffer test does not have them. Those are from Buffer's in main's `application.yml`, not `test`.
For test, the only config Buffer uses is to reset stairway. And that is `force-clean-start: true` in test, but `force-clean-start: false` in prod. I do see WSM has similar configs here:
https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/app/configuration/external/StairwayDatabaseConfiguration.java#L18, and seems WSM does not reset Stairway when local/test server starts, but it does upgrade (see this file's line 19: migrate-upgrade: true), now it should be replaced by your new config dedpends on if you want to use this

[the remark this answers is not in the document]
ddietterich: Those setting look good to me. I think we should be doing `force-clean-start: true` on our test runs.

ddietterich (approved):
lgtm

src/main/java/bio/terra/workspace/service/job/JobService.java
jaycarlton: Sorry if that sounded aggresive.

src/test/resources/application-test.yml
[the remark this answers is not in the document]
yonghaoy: This does not look correct to me. At least we should remove the migrate-upgrade in line 19? I don't know which config you are reading.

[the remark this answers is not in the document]
yonghaoy: And you should remove all lin 38-42 from here and put them in application.yml not in test

yonghaoy (approved):
Thanks

src/main/resources/application.yml
```diff
@@ -73,7 +73,16 @@ workspace:
...
+terra:
+  common:
+    tracing:
+        stackdriverExportEnabled: ${env.tracing.exportEnabled}
+        samplingRate: ${env.tracing.samplingRate}
+    stairway:
+      cluster-name-suffix: workspace-stairway
+      force-clean-start: false
+      max-parallel-flights: 50
+      migrate-upgrade: false
```
yonghaoy, line 85: migrate-upgrade should be true in non-testing env. force-clean-start should be `${env.db.init}`
See line 59 and 60 above, and you should remove those two lines because they are not used anymore. 
Also those configs should be deleted as well:
https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/app/configuration/external/StairwayDatabaseConfiguration.java#L13-L42
Those two are replaced by your new code: https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/stairway/StairwayProperties.java#L19-L24

PR #269, opened 2021-03-23
PF-409 delete resource

ddietterich:
The delete code was not as hard as the testing code. Many bugs and interesting learnings along the way.

The current draft disables the `CreateGcsBucketStepTest` unit test where I was unable to figure out the mocks. I don't plan to chase that problem as part of this PR. Although if anybody has an idea how to solve it, I'm all ears.

One drive-by: I added stairway logging into the MDC logger to be able to see steps and states.


src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+      }
+      return StepResult.getStepResultSuccess();
+    }
+
+    /**
+     * Try deleting the bucket. It will fail if there are objects still in the bucket.
+     *
+     * @param bucket bucket we should try to delete
+     * @return bucket existence: true if the bucket still exists; false if we deleted it
```
jaycarlton, line 125: I think this is the opposite sense of a delete method or two I either wrote or saw somewhere. It was true for deleted, false for not deleted (still exists).

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
```diff
@@ -0,0 +1,174 @@
...
+    public DeleteMetadataStep(ResourceDao resourceDao) {
+      this.resourceDao = resourceDao;
+    }
+
+    @Override
+    public StepResult doStep(FlightContext flightContext)
+        throws InterruptedException, RetryException {
+      FlightMap inputParameters = flightContext.getInputParameters();
+      UUID workspaceId =
+          UUID.fromString(inputParameters.get(WorkspaceFlightMapKeys.WORKSPACE_ID, String.class));
```
jaycarlton, line 161: You could write a utility method `getUuidValue(String Key)` and use it in all these places.

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -212,4 +216,21 @@ public void deleteGcpCloudContext(UUID workspaceId, AuthenticatedUserRequest use
...
   }
+
+  /**
+   * Get the GCP project associated with a workspace. If none is defined, throw. This helper method
+   * is used by other classes that need to know the GCP project associated with the workspace.
+   *
+   * @param workspaceId unique workspace id
+   * @return GCP project id
+   */
+  public String getGcpProject(UUID workspaceId) {
```
jaycarlton, line 227: Why not make this return `Optional<String>` instead of throw? Then callers could throw whatever they want or handle it more quitely.

src/main/java/bio/terra/workspace/service/workspace/exceptions/CloudContextRequiredException.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.workspace.service.workspace.exceptions;
+
+import bio.terra.workspace.common.exception.BadRequestException;
+
+/** When you can't get there from here, but somehow end up there */
+public class CloudContextRequiredException extends BadRequestException {
+  public CloudContextRequiredException(String message) {
```
jaycarlton, line 7: fwiw, IJ will give you all 4-5 overloads of the constructor so you have flexibility in declaring these.

jaycarlton (commented):
first batch

src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -75,7 +75,7 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
   }
 
   @Override
-  public ResponseEntity<ApiGcsBucketReference> getBucketReference(UUID id, UUID referenceId) {
+  public ResponseEntity<ApiGcsBucketReference> getGcpBucketReference(UUID id, UUID referenceId) {
```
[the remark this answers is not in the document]
ddietterich: Well, the endpoint is
`resources/referenced/gcp/buckets/{resourceid}`
because we are talking about the cloud, not the resource type. In this case, I removed the gcs altogether.

I wish we hadn't used it anywhere, since GCS adds no information over GCP and will only be meaningful to GCP-savvy folks.

We should probably revisit the naming of resource components in the OpenAPI so they consistently include the cloud name. I added PF-605 to track this.

src/main/java/bio/terra/workspace/common/utils/MdcHook.java
```diff
@@ -48,6 +60,15 @@ public Object getSerializedCurrentContext() {
...
   public HookAction startStep(FlightContext flightContext) {
+    logger.info(
+        StepLogFormat,
+        "startStep",
+        flightContext.getFlightClassName(),
+        flightContext.getFlightId(),
+        flightContext.getStepClassName(),
+        flightContext.getStepIndex(),
+        flightContext.getDirection().name(),
+        Instant.now().atZone(ZoneId.of("Z")).format(DateTimeFormatter.ISO_INSTANT));
```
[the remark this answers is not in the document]
ddietterich: I removed all of the timestamps. I had just cut'n'pasted from TDR. I think they collected the timestamps for use in an alternative performance logging system.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
```diff
@@ -0,0 +1,174 @@
...
+    private boolean tryBucketDelete(BucketCow bucket) {
+      try {
+        logger.info("Attempting to delete bucket " + bucket.getBucketInfo().getName());
+        bucket.delete();
+        return false;
+      } catch (StorageException ex) {
+        logger.info(
+            "Attempt to delete bucket "
+                + bucket.getBucketInfo().getName()
+                + "failed, but that is OK we will try again",
```
zloery, line 136: We may or may not try again, depending

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
```diff
@@ -0,0 +1,174 @@
...
+        bucketCow.update();
+        bucketExists = tryBucketDelete(bucket);
+        if (bucketExists) {
+          TimeUnit.HOURS.sleep(1);
+        }
+        deleteTries++;
+        if (deleteTries >= MAX_DELETE_TRIES) {
+          // This will cause the flight to fail.
+          throw new BucketDeleteTimeoutException(
+              String.format("Failed to delete bucket after %d hours", MAX_DELETE_TRIES));
```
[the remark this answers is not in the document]
ddietterich: Good catch

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
ddietterich: Yes, but in this case, it is the right sense for the while loop and if. Otherwise, I get the complaint that all uses of the return value are negated.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
```diff
@@ -0,0 +1,174 @@
...
+      }
+    }
+
+    @Override
+    public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+      return StepResult.getStepResultSuccess();
+    }
+  }
+
+  // TODO: this looks generic, so factor it out and share it
```
zloery, line 148: +1, this feels like it could be a standalone `DeleteControlledResourceMetadataStep`

src/main/java/bio/terra/workspace/service/workspace/exceptions/CloudContextRequiredException.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.workspace.service.workspace.exceptions;
+
+import bio.terra.workspace.common.exception.BadRequestException;
+
+/** When you can't get there from here, but somehow end up there */
```
zloery, line 5: wrong comment, I think

src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
ddietterich: I renamed it `getRequiredGcpProject`. I believe the common case is where we are retrieving the project in order to use it for something.

src/main/resources/api/service_openapi.yaml
```diff
@@ -666,6 +705,25 @@ paths:
...
+        '200':
+          $ref: '#/components/responses/CreatedControlledGcsBucketResponse'
+        '202':
+          $ref: '#/components/responses/CreatedControlledGcsBucketResponse'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '500':
+          $ref: '#/components/responses/ServerError'
+
   /api/workspaces/v1/{workspaceId}/resources/controlled/gcp/buckets/{resourceId}:
```
zloery, line 727: Feels pretty odd that POST to this endpoint deletes, but I don't know a better way to structure that given you can't use delete requests with bodies

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -0,0 +1,211 @@
...
+
+    try {
+      resourceApi.getBucket(getWorkspaceId(), resourceId);
+      throw new IllegalStateException("Incorrectly found a deleted bucket!");
+    } catch (ApiException ex) {
+      assertThat(ex.getCode(), equalTo(HttpStatusCodes.STATUS_CODE_NOT_FOUND));
+    }
+    bucketName = null;
+
+    // Delete the cloud context
```
zloery, line 165: nit: this isn't required, the fixture will delete the workpace and that flight includes deleting the GCP project.

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -0,0 +1,211 @@
...
+                    .accessScope(ControlledResourceCommonFields.AccessScopeEnum.SHARED_ACCESS)
+                    .managedBy(ControlledResourceCommonFields.ManagedByEnum.USER)
+                    .jobControl(new JobControl().id(jobId));
+
+    var body =
+            new CreateControlledGcsBucketRequestBody()
+                    .gcsBucket(creationParameters)
+                    .common(commonParameters);
+
+    // Try and fail to create a bucket - no cloud context
```
zloery, line 192: nit: comment out of date I think

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceAllocateTestScriptBase.java
```diff
@@ -1,59 +1,91 @@
...
-   * @return
+   *
+   * @return workspace UUID
    */
   protected UUID getWorkspaceId() {
     return workspaceId;
   }
 
+  /**
+   * Allow inheriting classes to object the spend profile
```
zloery, line 46: nit: obtain*

workspace-manager-clienttests/src/main/java/scripts/utils/WorkspaceApiTestScriptBase.java
```diff
@@ -1,24 +1,24 @@
...
+import java.util.List;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+
 /**
  * Intermediate base class for testing Workspace Manager in the TestRunner framework. This class handles
  * obtaining the WorkspaceApi object and basic checks and debugging around its status code.
  *
- * Users of this class need to o verload doUserJourney() with efssentially the body of the userJourney() method,
+ * Users of this class need to o overload doUserJourney() with the body of the userJourney() method,
```
zloery, line 17: nit: "to o overload"

workspace-manager-clienttests/src/main/resources/suites/BasicIntegration.json
```diff
@@ -8,7 +8,8 @@
     "integration/CreateGetDeleteWorkspace.json",
     "integration/DataReferenceLifecycle.json",
     "integration/EnumerateDataReferences.json",
-    "integration/GetRoles.json"
+    "integration/GetRoles.json",
+    "integration/CreateGetDeleteControlledGcsBucket.json"
```
zloery, line 12: should any of these run as perf tests? I'm not exactly sure what our test strategy is there

zloery (commented):
Generally looks good but this does raise a potentially thorny question - currently, deleting a workspace deletes the underlying project (and it's synchronous). What happens if there are controlled resources in that workspace? I'm not sure if the call to delete the project will fail, or if GCP knows to clean up things in that workspace.

jaycarlton (approved):

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -74,4 +76,29 @@ public ControlledResource getControlledResource(
...
     return wsmResource.castControlledResource();
   }
+
+  public String deleteControlledGcsBucket(
+      ApiJobControl jobControl,
+      UUID workspaceId,
+      UUID resourceId,
+      AuthenticatedUserRequest userRequest) {
+    stageService.assertMcWorkspace(workspaceId, "deleteControlledGcsBucket");
+    workspaceService.validateWorkspaceAndAction(
```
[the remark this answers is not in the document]
ddietterich: Yes. When I coded this, the Sam stuff was still in flux. I have made PF-607 to track revisiting the resource permission checks.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -74,4 +76,29 @@ public ControlledResource getControlledResource(
...
+
+    final JobBuilder jobBuilder =
+        jobService
+            .newJob(
+                jobDescription,
+                jobControl.getId(),
+                DeleteControlledResourceGcsBucketFlight.class,
+                null,
+                userRequest)
+            .addParameter(WorkspaceFlightMapKeys.WORKSPACE_ID, workspaceId.toString())
```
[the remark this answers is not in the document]
ddietterich: Are you sure? I thought there was an issue serializing UUIDs.

src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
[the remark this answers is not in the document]
jaycarlton: It's the hexagon name, though. You include BigQuery with dataset, but no service name for bucket. Which is a small point, but they could always introduce another service that used buckets.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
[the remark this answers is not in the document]
jaycarlton: There's a UUID database type as well, but I don't think that's in play here.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -74,4 +76,29 @@ public ControlledResource getControlledResource(
...
+            .addParameter(WorkspaceFlightMapKeys.ResourceKeys.RESOURCE_ID, resourceId.toString());
+
```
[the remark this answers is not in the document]
ddietterich: Good catch. I will add that in. I wonder if we should make an internal version of the JobControl that includes what the caller provides as well as the result path. That would provide our API <-> WSM object separation and cause an error if the result path wasn't provided.

src/main/resources/api/service_openapi.yaml
ddietterich: Yes. We may be able to fix it, but I'm not sure it is worth the trouble for the few async delete cases we have.

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+public class CreateGetDeleteControlledGcsBucket extends WorkspaceAllocateTestScriptBase {
+  private static final Logger logger =
+      LoggerFactory.getLogger(CreateGetDeleteControlledGcsBucket.class);
+  private static final long CREATE_BUCKET_POLL_SECONDS = 5;
```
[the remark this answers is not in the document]
ddietterich: For a test I don't see the point of constructing an object converting seconds to some inner format just to use it in one place to deconstruct it into seconds.

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -0,0 +1,211 @@
...
+    // Create a user-shared controlled GCS bucket - should fail due to no cloud context
+    CreatedControlledGcsBucket bucket = createBucketAttempt(resourceApi);
+    assertThat(bucket.getJobReport().getStatus(), equalTo(JobReport.StatusEnum.FAILED));
+    assertThat(bucket.getErrorReport().getStatusCode(), equalTo(HttpStatusCodes.STATUS_CODE_BAD_REQUEST));
+
+    // Create the cloud context
+    logger.info("Creating cloud context");
+
+    String contextJobId = UUID.randomUUID().toString();
+    var createContext = new CreateCloudContextRequest()
```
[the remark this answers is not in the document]
ddietterich: I added PF-608 to track implementing spotless in clienttest.

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
ddietterich: Added a comment. Seems worth exercising deleting the cloud context.

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -0,0 +1,211 @@
...
+                    .managedBy(ControlledResourceCommonFields.ManagedByEnum.USER)
+                    .jobControl(new JobControl().id(jobId));
+
+    var body =
+            new CreateControlledGcsBucketRequestBody()
+                    .gcsBucket(creationParameters)
+                    .common(commonParameters);
+
+    // Try and fail to create a bucket - no cloud context
+    logger.info("Attempt to creating bucket {} jobId {} workspace {}", bucketName, jobId, getWorkspaceId());
```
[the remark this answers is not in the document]
ddietterich: The comment was wrong. It is not always expected to fail

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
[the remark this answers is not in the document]
zloery: It definitely works, we have a few steps which already deserialize UUIDs directly (e.g. [CreateWorkspaceStep](https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java#L33))

workspace-manager-clienttests/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -15,172 +13,184 @@
...
-      return new CreateDataReferenceRequestBody()
-          .name(getUniqueDataReferenceName())
-          .cloningInstructions(CloningInstructionsEnum.REFERENCE)
-          .referenceType(ReferenceTypeEnum.DATA_REPO_SNAPSHOT)
-          .reference(getTestDataRepoSnapshot());
-    }
+  public static void assertHttpSuccess(WorkspaceApi workspaceApi, String label) {
+    int httpCode = workspaceApi.getApiClient().getStatusCode();
+    logger.debug("{} HTTP code: {}", label, httpCode);
+    assertThat(HttpStatusCodes.isSuccess(httpCode), equalTo(true));
```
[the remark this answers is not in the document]
ddietterich: I tried. I believe that is present in JUnit, but doesn't seem to be available here.

workspace-manager-clienttests/src/main/resources/suites/BasicIntegration.json
ddietterich: We have not really worried performance yet. I don't think we have performance targets specified.
My focus is to make sure we have robust integration tests and worry performance later.

src/test/java/bio/terra/workspace/service/resource/controlled/flight/CreateGcsBucketStepTest.java
```diff
@@ -16,41 +18,43 @@
...
+  @Test
+  public void testCreatesBucket() throws RetryException, InterruptedException {
+    CreateGcsBucketStep createGcsBucketStep =
         new CreateGcsBucketStep(
-            mockCrlService, ControlledResourceFixtures.BUCKET_RESOURCE, mockUserRequest);
+            mockCrlService, ControlledResourceFixtures.BUCKET_RESOURCE, mockWorkspaceService);
 
-    doReturn(mockStorageCow).when(mockCrlService).createStorageCow(mockUserRequest);
-  }
+    when(mockCrlService.createStorageCow(mockUserRequest)).thenReturn(mockStorageCow);
```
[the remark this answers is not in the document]
ddietterich: The documentation suggests the opposite - that the thenReturn form is more type safe.

In any case, either way it is expressed the bucket create gets an NPE. I'd love your help sorting that out at some future time.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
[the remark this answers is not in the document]
ddietterich: OK, I changed it. Thanks

src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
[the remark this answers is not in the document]
ddietterich: "hexagon"?

I made ticket PF-605 to rename all of our GCP-specific api components with a Gcp prefix. I think that will help clarify what is cloud-specific.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
```diff
@@ -59,38 +55,27 @@ public StepResult doStep(FlightContext flightContext)
...
-        workspaceDao
-            .getGcpCloudContext(resource.getWorkspaceId())
-            .map(GcpCloudContext::getGcpProjectId)
-            .orElseThrow(
-                () ->
-                    new CloudContextRequiredException(
-                        "No cloud context found in which to create a controlled resource"));
-    final StorageCow storageCow = crlService.createStorageCow(gcpProjectId, userRequest);
+    FlightMap inputMap = flightContext.getInputParameters();
+    UUID workspaceId =
```
zloery, line 75: One more place to read UUID directly

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
```diff
@@ -0,0 +1,170 @@
...
+    public DeleteMetadataStep(ResourceDao resourceDao) {
+      this.resourceDao = resourceDao;
+    }
+
+    @Override
+    public StepResult doStep(FlightContext flightContext)
+        throws InterruptedException, RetryException {
+      FlightMap inputParameters = flightContext.getInputParameters();
+      UUID workspaceId =
+          UUID.fromString(inputParameters.get(WorkspaceFlightMapKeys.WORKSPACE_ID, String.class));
```
zloery, line 157: nit: deserialize uuid directly for workspace and resource IDs

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -84,21 +89,66 @@ public ControlledGcpResourceApiController(
                             .map(ControlledResourceIamRole::fromApiModel)
                             .collect(toCollection(ArrayList::new)))
                 .orElse(null);
-    if (privateRoles != null && privateRoles.isEmpty()) {
+
+    // REVIEWERS: I changed this from what was in Zach's recent PR. I think this is the proper
+    // check, but
+    // please check my work!
+    boolean privateRoleOmitted = (privateRoles == null || privateRoles.isEmpty());
```
zloery, line 96: lgtm, this should make API usage a little clearer

zloery (approved):

src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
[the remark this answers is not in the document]
jaycarlton: All the gcp services have hexagon logos. Or at least they did until about a year ago.

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -0,0 +1,211 @@
...
+    while (ClientTestUtils.jobIsRunning(result.getJobReport())) {
+      TimeUnit.SECONDS.sleep(DELETE_BUCKET_POLL_SECONDS);
+      result = resourceApi.getDeleteBucketResult(getWorkspaceId(), deleteJobId);
+    }
+    logger.info("Delete bucket status is {}", result.getJobReport().getStatus().toString());
+    assertThat(result.getJobReport().getStatus(), equalTo(JobReport.StatusEnum.SUCCEEDED));
+
+    // verify it's not there anymore
+    // - via metadata
+    // TODO: via GCP access
```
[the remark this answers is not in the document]
ddietterich: There is a testing ticket for further test development PF-512

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
ddietterich: Thank you!!

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
```diff
@@ -36,55 +42,115 @@ public DeleteControlledResourceGcsBucketFlight(FlightMap inputParameters, Object
...
+    public StepResult doStep(FlightContext flightContext)
+        throws InterruptedException, RetryException {
+
+      WsmResource wsmResource = resourceDao.getResource(workspaceId, resourceId);
+      ControlledResource resource = wsmResource.castToControlledResource();
+
+      try {
+        samService.deleteControlledResource(resource, userRequest);
+      } catch (SamApiException samApiException) {
+        if (samApiException.getApiExceptionStatus() == HttpStatus.NOT_FOUND.value()) {
```
zloery, line 107: nit: might be worth a comment explaining this is for step idempotency

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
```diff
@@ -36,55 +42,115 @@ public DeleteControlledResourceGcsBucketFlight(FlightMap inputParameters, Object
...
+          throw samApiException;
+        }
+      }
+      return StepResult.getStepResultSuccess();
+    }
+
+    @Override
+    public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+      // No undo for delete
+      return StepResult.getStepResultSuccess();
```
zloery, line 119: Should this be a fatal error? I think it's technically possible that we delete the Sam resource and then for whatever reason start undoing - "success" implies we've returned to the previous state (i.e. restore the deleted Sam resource), which we can't. otoh I don't want to raise a fatal stairway response every time the Sam call fails

zloery (approved):

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
ddietterich: It is a toss up. In TDR, we decided to just delete as much as we could and not error on the theory that it would help clean up corrupted state. The alternative is to cause a DISMAL_FAILURE (returning FATAL here) which, in production, should cause a human to take a look at what is going on.

Worth discussing further. Once we implement Sam retries, we should not be seeing transient Sam errors here.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceGcsBucketFlight.java
ddietterich: Good point. I wrote up the two cases that would cause not found.

PR #273, opened 2021-03-26
[PF-579] Remove ControlledResourceIamRoleList, fix enum fromApi method

zloery:
This fixes a couple of bugs that have surfaced from initial use of private controlled resources. Thanks @wchamber for finding the initial issues!

- The `fromApi` method of `ControlledResourceIamRole` throws a NPE when we call `apiRole.equals()` because not all internal resource roles have equivalent API roles. This is correct from a design perspective (users can't directly grant assigner/owner roles on private resources) but needs a fix
- In #263 I added `ControlledResourceIamRoleList` to avoid deserializing a parameterized List object from Stairway, but did so in a way that would throw casting exceptions. Fixed this
- We're hardcoding accessType and managedBy for resources as we store them in the database. This causes all kinds of problems for private resources


ddietterich (approved):
LGTM

jaycarlton (approved):
So the issue that necessitated `ControlledResourceIamRoleList` originally has been solved?

src/main/java/bio/terra/workspace/service/iam/model/ControlledResourceIamRole.java
```diff
@@ -28,7 +28,7 @@
   public static ControlledResourceIamRole fromApiModel(ApiControlledResourceIamRole apiModel) {
     Optional<ControlledResourceIamRole> result =
         Arrays.stream(ControlledResourceIamRole.values())
-            .filter(x -> x.apiRole.equals(apiModel))
+            .filter(x -> x.apiRole == apiModel)
```
wchamber, line 31: prefer `.equals` over `==`, even for enums. go/java-practices/enums#equals
You can invert the order to avoid the NPE, but worth a comment that `x.apiRole` can be null. `apiModel.equals(x.apiRole)`

src/main/java/bio/terra/workspace/service/iam/model/ControlledResourceIamRole.java
```diff
@@ -28,7 +28,7 @@
   public static ControlledResourceIamRole fromApiModel(ApiControlledResourceIamRole apiModel) {
```
wchamber, line 28: please add unit tests for this function.
What's the correct return value is if the argument `apiModel` is null?

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -250,12 +250,10 @@ public void createControlledResource(ControlledResource controlledResource)
...
-        // TODO: add this to ControlledResource
-        Optional.of(AccessScopeType.ACCESS_SCOPE_SHARED),
-        Optional.of(ManagedByType.MANAGED_BY_USER),
+        Optional.ofNullable(controlledResource.getAccessScope()),
+        Optional.ofNullable(controlledResource.getManagedBy()),
         // TODO: add associated app to ControlledResource
         Optional.empty(),
-        // TODO: rename this
         controlledResource.getAssignedUser());
   }
```
wchamber, line 258: very optional: consider changing the helper `storeResource` to take a `WsmResource` for the common fields to greatly cut down the arguments ane make clearer it's for storing WsmResource subclasses.

wchamber (commented):
How did you figure out that `ControlledResourceIamRoleList` didn't work? Can we add a unit test that would have caught that?

src/main/java/bio/terra/workspace/service/iam/model/ControlledResourceIamRole.java
zloery: done, with comment

src/main/java/bio/terra/workspace/service/iam/model/ControlledResourceIamRole.java
zloery: I think it's reasonable to throw in that case, added a ValidationException here

wchamber:
> `ControlledResourceIamRoleList` wasn't strictly necessary, it was a way to avoid Stairway's "deserialize as `List.class` and cast" pattern for type safety reasons. Since that solution didn't work, I went back to the regular Stairway pattern

@zloery  How did you figure out that ControlledResourceIamRoleList didn't work? Can we add a unit test that would have caught that?

zloery:
I found that through manual testing in a personal environment, so not the most repeatable. It's a little tricky to unit test because this is in the Controller which doesn't have much functionality.
We've generally tested Controller code via integration and client tests, and in this case any client test exercising private resource paths would have caught this. I'm adding some client testing as part of the last piece of controlled resource IAM, and I think we have some backlog tickets about other spots missing tests.

wchamber (approved):

zloery:
after more discussion it felt like the `ControlledResourceIamRoleList` approach should actually work, and I found the correct way to write it. Manually tested that this approach works, and client tests exercising this to follow in a separate PR. PT(another)L

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -79,17 +78,14 @@ public ControlledGcpResourceApiController(
...
-                            .map(ControlledResourceIamRole::fromApiModel)
-                            .collect(toCollection(ArrayList::new)))
-                .orElse(null);
-
+        Optional.ofNullable(body.getCommon().getPrivateResourceUser())
+            .map(
+                user ->
+                    user.getPrivateResourceIamRoles().stream()
+                        .map(ControlledResourceIamRole::fromApiModel)
+                        .collect(toCollection(ControlledResourceIamRoleList::new)))
```
wchamber, line 87: Not having to do a List cast in the `flightMap.get` is nice, but I'm a little worried about having a consistent way to serialize Lists in Stairway. This doesn't feel like a great long term solution for List serialization, and having only one kind of migration to do would be nice. Or is this the new preferred style to use everywhere?
If we do want to keep the ControlledResourceIamRoleList, consider adding a unit test for it that `flightMap.put -> toJson -> fromJson -> flightMap.get`, since that's the whole purpose of the class. But that might be just my paranoia about serialization.

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -79,17 +78,14 @@ public ControlledGcpResourceApiController(
...
+            .orElse(null);
```
wchamber, line 88: generally nullable collections is a bit of an anti-pattern. Is `null` semantically distinct from an empty list here? If not, consider an empty list.

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
zloery: Fair point - defining a new type every time we serialize a generic is pretty heavy. If this is something we want to look at more broadly in Stairway, I think it's better to be consistent here

PR #274, opened 2021-03-29
PF-431, PF-592, PF-605 - Resource enumeration

ddietterich:
- cleanup OpenAPI for consistent naming and contents for resource returns and enumerated resource list
 - remove stray uses of deprecated schemas
 - add enumeration endpoint
 - restructure/rename per-object converters from Wsm classes into Api classes
 - teach controlled bucket to return the whole resource on create and get
 - added enumerateResources testrunner test
 - added more logging and enabled retries for delete workspace authz step
 - there is a problem...

Something isn't working right with the permissions on controlled resources
Either the query in ResourceDao is broken or the permissions are not getting
setup properly. The test has a workaround to only create and enumerate
referenced resources.

Since the client is changed, the integration tests do not pass in the GHA. The FullIntegration suite (with the workaround) passes on my personal environment.


src/main/java/bio/terra/workspace/service/resource/WsmResource.java
```diff
@@ -97,6 +98,24 @@ public CloningInstructions getCloningInstructions() {
    */
   public abstract String attributesToJson();
 
+  /**
+   * Create the part of the common resource description that comes from this class. Subclasses
+   * should override this and fill in their specific info.
```
jaycarlton, line 103: Indicate in the comment that the info to be specified is on a field in the `ApiResourceMetadata` class, and that they shouldn't extend that class to hold their state.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResource.java
```diff
@@ -63,6 +66,20 @@ public ManagedByType getManagedBy() {
...
+  public ApiResourceMetadata toApiMetadata() {
+    ApiResourceMetadata metadata = super.toApiMetadata();
+    var controlled =
+        new ApiControlledResourceMetadata()
+            .accessScope(accessScope.toApiModel())
+            .managedBy(managedBy.toApiModel())
+            .privateResourceUser(
+                // TODO: PF-616 figure out how to supply the assigned user's role
+                new ApiPrivateResourceUser().userName(assignedUser));
+    metadata.controlledResourceMetadata(controlled);
```
jaycarlton, line 79: Do we stop here, or can we keep going with the subclass's metadata?

jaycarlton (commented):
quick pass

src/main/resources/api/service_openapi.yaml
```diff
@@ -1,3 +1,27 @@
+# DEVELOPER NOTES
+#
+# There are three sections of the file bracketed with ## ## that are deprecated. They are the
+# data references interface that we abandoned in favor of the unified resources interface.
+#
+# CLOUD RESOURCE STRUCTURE
+# Each cloud resource has a structure named '{resource-type}Resource';
+# for example, GcsBucketResource. It has two properties: 'metadata' and 'attributes'
+#
+# The 'metadata' property is a reference to the common metadata structure for all resource types:
```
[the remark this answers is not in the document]
ddietterich: It is not right now.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1,3 +1,27 @@
+# DEVELOPER NOTES
+#
+# There are three sections of the file bracketed with ## ## that are deprecated. They are the
+# data references interface that we abandoned in favor of the unified resources interface.
+#
+# CLOUD RESOURCE STRUCTURE
+# Each cloud resource has a structure named '{resource-type}Resource';
```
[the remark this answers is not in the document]
ddietterich: Thanks for the reminder. I updated the documentation and did the rename to add the cloud platform

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -134,6 +141,80 @@ public boolean deleteResource(UUID workspaceId, UUID resourceId) {
...
+      sb.append(" AND resource_type = :resource_type");
+      params.addValue("resource_type", resourceType.toSql());
+    }
+
+    // There are three cases for the stewardship type filter
+    // 1. If it is REFERENCED, then we ignore idList and just filter
+    //    for referenced resources.
+    // 2. If it is CONTROLLED, and the idList is not empty, then we filter for
+    //    CONTROLLED and require that the resources be in the idList.
+    // 3. If it is null, then we want the OR of 1 and 2
```
wchamber, line 183: This seems to me like we want different public enumerate methods on ResourceDao. When there are some arguments that are only valid with certain values of arguments, I think it's best to define a Java API where invalid argument combinations are less allowed. While our HTTP API can handle all the methods, and do the appropriate request validation on arguments, I don't think we should restrict our Java methods to be 1 to 1 because of that.
Maybe a enumerateControlledResources methods when we want to specify controlledResourceIds.

At least let's explain in the javadoc here and on ResourceService how these parameters interact together.

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -398,6 +398,25 @@ public void deleteControlledResource(
...
 
+  /**
+   * List all controlled resources ID in Sam this user has access to. This builds an aggregated list
+   * across the four types of controlled resources.
+   */
+  @Traced
+  public List<UUID> listControlledResourceIds(
+      AuthenticatedUserRequest userReq, String controlledResourceName) {
+    ResourcesApi resourceApi = samResourcesApi(userReq.getRequiredToken());
+    List<UUID> workspaceIds = new ArrayList<>();
```
wchamber, line 409: controlledResourceIds?

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -398,6 +398,25 @@ public void deleteControlledResource(
     }
   }
 
+  /**
+   * List all controlled resources ID in Sam this user has access to. This builds an aggregated list
+   * across the four types of controlled resources.
+   */
+  @Traced
+  public List<UUID> listControlledResourceIds(
+      AuthenticatedUserRequest userReq, String controlledResourceName) {
```
wchamber, line 407: I'm a little confused by the primitive String controlledResourceName and UUID return value. What kind of "name" are we storing in Sam? What kind of UUID, and how does it relate to the ResourceDao, does this return?
Bigger scope than this PR, but I think it would be very useful to strongly type our UUIDs and String "types" of data. For UUID, I know we have at least a "WorkspaceId" and a "ResourceId".  The Sam service also has "resource" as a part of its terminology and API, see SamService#isAuthorized: It's unclear to me what a `String resourceId` represent.

src/main/java/bio/terra/workspace/service/resource/WsmResource.java
wchamber: Should we have a protected abstract method for subclasses to override? Would make it harder to forget.

src/main/java/bio/terra/workspace/service/resource/WsmResourceType.java
```diff
@@ -50,6 +68,25 @@ public static WsmResourceType fromSql(String dbString) {
...
   }
 
+  /**
+   * Convert from api type to WsmResourceType. In some contexts, the API input can be null, so if
+   * the input is null we return null and leave it to the caller to raise any error.
+   *
+   * @param apiResourceType incoming resource type or null
+   * @return valid resource type; null if input is null
+   */
+  public static WsmResourceType fromApi(ApiResourceType apiResourceType) {
```
wchamber, line 78: nit: add `@Nullable` to the return type.

src/main/java/bio/terra/workspace/service/resource/WsmResourceType.java
```diff
@@ -50,6 +68,25 @@ public static WsmResourceType fromSql(String dbString) {
         "Deeserialization failed: no matching resource type for " + dbString);
   }
 
+  /**
+   * Convert from api type to WsmResourceType. In some contexts, the API input can be null, so if
+   * the input is null we return null and leave it to the caller to raise any error.
```
wchamber, line 73: "the null returns null, but invalid throws" is a little subtle to me. Do we really need 2 forms of "could not convert?" Can the clients do their own null checking?

src/main/java/bio/terra/workspace/service/resource/controlled/AccessScopeType.java
```diff
@@ -2,27 +2,24 @@
...
 
-  public String toSql() {
-    return dbString;
-  }
-
-  public static AccessScopeType fromApi(
-      ApiControlledResourceCommonFields.AccessScopeEnum apiAccessScope) {
+  public static AccessScopeType fromApi(ApiAccessScope apiAccessScope) {
     if (apiAccessScope == null) {
       throw new MissingRequiredFieldException("Valid accessScope is required");
```
wchamber, line 24: This enum conversion throws on null, but ApiResourceType doesn't? Consistency seems good.

src/main/java/bio/terra/workspace/service/workspace/flight/DeleteWorkspaceAuthzStep.java
```diff
@@ -31,8 +34,14 @@ public StepResult doStep(FlightContext flightContext) throws RetryException {
     } catch (SamApiException e) {
       // Stairway steps may run multiple times, so this may already have been deleted.
       // For all other errors we should always retry because there's no way to undo a Sam delete.
+      // TODO: there is no retry rule defined for this step, so it is never retried.
```
wchamber, line 37: It looks like you've now added one to WorkspaceDeleteFlight?

wchamber (commented):
It would be nice to have broken this into smaller PRs, e.g. rename all the models, rework where the toApi functions are (possibly in multiple PRs), the bucket changes, add the new enumerate endpoint + tests for it.
Here's some partial comments. I don't have any objections to the thrust of the PR, but it's possible I missed something.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1,3 +1,27 @@
...
+# for example, GcsBucketResource. It has two properties: 'metadata' and 'attributes'
+#
+# The 'metadata' property is a reference to the common metadata structure for all resource types:
+#  $ref: '#/components/schemas/ResourceMetadata'
+#
+# The 'attributes' property is a reference to a unique structure named '{resource-type}Attributes';
```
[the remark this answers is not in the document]
ddietterich: I don't think we want to assume that. Did you have something specific in mind?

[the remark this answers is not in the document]
jaycarlton: I don't know where I got that idea.

src/main/java/bio/terra/workspace/db/ResourceDao.java
ddietterich: I expanded the javadoc comment to try to clarify. The complexity here is due to a problem I found when passing in the id list. If the list is empty, the resulting SQL is invalid. So instead of simply adding one more filter, like resource type, I end up with this more complex query construction.

I have a bias to invest in the query generation and keep the database work to a single scan/single transaction.

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -134,6 +141,80 @@ public boolean deleteResource(UUID workspaceId, UUID resourceId) {
...
+    // Note that we supply the constant value as a parameter so that the SQL form
+    // is defined in exactly one place.
+    boolean includeReferenced = (stewardshipType == null || stewardshipType == REFERENCED);
+    boolean includeControlled =
+        (!idList.isEmpty() && (stewardshipType == null || (stewardshipType == CONTROLLED)));
```
jaycarlton, line 188: nit: you don't need the second open paren before `stewardshipType`

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -134,6 +141,80 @@ public boolean deleteResource(UUID workspaceId, UUID resourceId) {
...
+
+    final String referencedPhrase = "stewardship_type = :referenced_resource";
+    final String controlledPhrase =
+        "(stewardship_type = :controlled_resource AND resource_id IN (:id_list))";
+
+    sb.append(" AND ");
+    if (includeReferenced && includeControlled) {
```
jaycarlton, line 195: If we get any fancier, we should look into dedicated query building classes/DSLs.

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -134,6 +141,80 @@ public boolean deleteResource(UUID workspaceId, UUID resourceId) {
...
+    } else if (includeReferenced) {
+      sb.append(referencedPhrase);
+      params.addValue("referenced_resource", REFERENCED.toSql());
+    } else if (includeControlled) {
+      sb.append(controlledPhrase);
+      params.addValue("controlled_resource", CONTROLLED.toSql());
+      params.addValue("id_list", idList);
+    } else {
+      // Nothing is included, so we return an empty result
+      return new ArrayList<WsmResource>();
```
jaycarlton, line 209: `Collections.emptyList()` is cleaner I think

workspace-manager-clienttests/src/main/java/scripts/testscripts/EnumerateResources.java
```diff
@@ -0,0 +1,289 @@
...
+    logResult("controlledBucket", controlledBucketList);
+    long expectedControlledBuckets =
+        resourceList.stream()
+            .filter(
+                m ->
+                    (m.getStewardshipType() == StewardshipType.CONTROLLED
+                        && m.getResourceType() == ResourceType.GCS_BUCKET))
+            .count();
+    logger.info("Counted {} controlled buckets created", expectedControlledBuckets);
+    long actualControlledBucket = controlledBucketList.getResources().size();
```
jaycarlton, line 168: *actualControlledBuckets

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -398,6 +398,25 @@ public void deleteControlledResource(
...
+    ResourcesApi resourceApi = samResourcesApi(userReq.getRequiredToken());
+    List<UUID> workspaceIds = new ArrayList<>();
+    try {
+      for (var resourceAndPolicy : resourceApi.listResourcesAndPolicies(controlledResourceName)) {
```
zloery, line 411: this needs to call `listResourcesAndPoliciesV2`. Otherwise, the response won't include the resources a user has inherited permission on, which is most controlled resource

zloery (commented):
first pass of comments

workspace-manager-clienttests/src/main/java/scripts/utils/ResourceMaker.java
```diff
@@ -0,0 +1,172 @@
...
+
+  public static GcpGcsBucketResource makeGcsBucketReference(
+      ReferencedGcpResourceApi resourceApi, UUID workspaceId, String name) throws ApiException {
+
+    var body =
+        new CreateGcpGcsBucketReferenceRequestBody()
+            .metadata(
+                new ReferenceResourceCommonFields()
+                    .cloningInstructions(CloningInstructionsEnum.NOTHING)
+                    .description("Descriptiono of " + name)
```
jaycarlton, line 94: *description

workspace-manager-clienttests/src/main/java/scripts/utils/ResourceMaker.java
```diff
@@ -0,0 +1,172 @@
...
+                    .location("US-CENTRAL1"));
+
+    logger.info("Creating bucket {} jobId {} workspace {}", bucketName, jobId, workspaceId);
+    CreatedControlledGcpGcsBucket createdBucket = resourceApi.createBucket(body, workspaceId);
+    while (ClientTestUtils.jobIsRunning(createdBucket.getJobReport())) {
+      TimeUnit.SECONDS.sleep(CREATE_BUCKET_POLL_SECONDS);
+      createdBucket = resourceApi.getCreateBucketResult(workspaceId, jobId);
+    }
+    logger.info("Create bucket status is {}", createdBucket.getJobReport().getStatus().toString());
+    if (createdBucket.getJobReport().getStatus() != JobReport.StatusEnum.SUCCEEDED) {
```
jaycarlton, line 148: can we use `assert()` in this kind of test?

workspace-manager-clienttests/src/main/java/scripts/utils/ResourceMaker.java
```diff
@@ -0,0 +1,172 @@
...
+    logger.info("Deleting bucket resource id {} jobId {}", resourceId, deleteJobId);
+    DeleteControlledGcpGcsBucketResult result =
+        resourceApi.deleteBucket(deleteRequest, workspaceId, resourceId);
+    while (ClientTestUtils.jobIsRunning(result.getJobReport())) {
+      TimeUnit.SECONDS.sleep(DELETE_BUCKET_POLL_SECONDS);
+      result = resourceApi.getDeleteBucketResult(workspaceId, deleteJobId);
+    }
+    logger.info("Delete bucket status is {}", result.getJobReport().getStatus().toString());
+    if (result.getJobReport().getStatus() != JobReport.StatusEnum.SUCCEEDED) {
+      throw new RuntimeException("Delte bucket failed: " + result.getErrorReport().getMessage());
```
jaycarlton, line 169: *Delete

src/main/java/bio/terra/workspace/app/controller/ResourceController.java
```diff
@@ -0,0 +1,148 @@
...
+        ControlledResource controlledResource = wsmResource.castToControlledResource();
+        switch (wsmResource.getResourceType()) {
+          case GCS_BUCKET:
+            {
+              ControlledGcsBucketResource resource = controlledResource.castToGcsBucketResource();
+              union.gcpGcsBucket(resource.toApiAttributes());
+              break;
+            }
+
+          case BIG_QUERY_DATASET:
```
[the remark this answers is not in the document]
ddietterich: I added documentation

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -134,6 +141,80 @@ public boolean deleteResource(UUID workspaceId, UUID resourceId) {
...
+
+    // There are three cases for the stewardship type filter
+    // 1. If it is REFERENCED, then we ignore idList and just filter
+    //    for referenced resources.
+    // 2. If it is CONTROLLED, and the idList is not empty, then we filter for
+    //    CONTROLLED and require that the resources be in the idList.
+    // 3. If it is null, then we want the OR of 1 and 2
+    // Note that we supply the constant value as a parameter so that the SQL form
+    // is defined in exactly one place.
+    boolean includeReferenced = (stewardshipType == null || stewardshipType == REFERENCED);
```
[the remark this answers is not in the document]
jaycarlton: Oh I completely misread it. Never mind.

src/main/java/bio/terra/workspace/db/ResourceDao.java
jaycarlton: then why not also do it before the `== null` case?

src/main/java/bio/terra/workspace/service/resource/controlled/AccessScopeType.java
```diff
@@ -2,27 +2,24 @@
...
-import bio.terra.workspace.generated.model.ApiControlledResourceCommonFields;
+import bio.terra.workspace.generated.model.ApiAccessScope;
 import bio.terra.workspace.service.workspace.exceptions.InternalLogicException;
 import org.apache.commons.lang3.StringUtils;
 
 /** Controlled resources can be shared by all users in a workspace or be private to a single user */
 public enum AccessScopeType {
-  ACCESS_SCOPE_SHARED("ACCESS_SHARED"),
-  ACCESS_SCOPE_PRIVATE("ACCESS_PRIVATE");
+  ACCESS_SCOPE_SHARED("ACCESS_SHARED", ApiAccessScope.SHARED_ACCESS),
```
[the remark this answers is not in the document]
jaycarlton: Oof.

src/main/java/bio/terra/workspace/service/resource/controlled/AccessScopeType.java
```diff
@@ -46,4 +43,12 @@ public static AccessScopeType fromSql(String dbString) {
     throw new SerializationException(
         "Deeserialization failed: no matching access scope type for " + dbString);
   }
+
+  public String toSql() {
+    return dbString;
+  }
+
+  public ApiAccessScope toApiModel() {
```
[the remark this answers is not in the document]
jaycarlton: That's reasonable. FWIW I like this model; it was more of a devil's advocate question.

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -171,36 +144,35 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
     bucketName = null;
 
     // Delete the cloud context. This is not required. Just some exercise for deleteCloudContext
-    logger.info("Deleting the cloud context");
-    workspaceApi.deleteCloudContext(getWorkspaceId(), CloudPlatform.GCP);
+    CloudContextMaker.deleteGcpCloudContext(getWorkspaceId(), workspaceApi);
   }
 
-  private CreatedControlledGcsBucket createBucketAttempt(ControlledGcpResourceApi resourceApi)
+  private CreatedControlledGcpGcsBucket createBucketAttempt(ControlledGcpResourceApi resourceApi)
```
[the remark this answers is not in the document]
jaycarlton: Maybe `createBucketOrThrow` or something more descriptive?

workspace-manager-clienttests/src/main/java/scripts/testscripts/EnumerateResources.java
```diff
@@ -0,0 +1,289 @@
...
+            break;
+          }
+
+        case 3:
+          {
+            GcpBigQueryDatasetResource resource =
+                ResourceMaker.makeBigQueryReference(referencedGcpResourceApi, workspaceId, name);
+            resourceList.add(resource.getMetadata());
+            /*
+            TODO: REVIEWERS: something isn't working right with the permissions on controlled resources
```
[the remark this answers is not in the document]
jaycarlton: That's why we keep you around.

workspace-manager-clienttests/src/main/java/scripts/utils/CloudContextMaker.java
```diff
@@ -0,0 +1,47 @@
...
+import bio.terra.workspace.model.CreateCloudContextRequest;
+import bio.terra.workspace.model.CreateCloudContextResult;
+import bio.terra.workspace.model.JobControl;
+import bio.terra.workspace.model.JobReport;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CloudContextMaker {
```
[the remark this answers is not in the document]
jaycarlton: this may not technically be a factory in the sense of Factory Pattern, so maker might be more accurate. https://www.tutorialspoint.com/design_pattern/factory_pattern.htm

workspace-manager-clienttests/src/main/java/scripts/utils/ResourceMaker.java
jaycarlton: Some lightweight assertion library that would throw an assertion failure in the test seems helpful. I suspect others will have opinions here.

ddietterich:
I think I have addressed most all of the review comments or made tickets for things to do later.
I did add a unit test for makeApiDescription as suggested by Jay.

The testrunner integration tests will not pass in GHA, because I have changed the API and the client doesn't match.
I ran them myself and all is well.

Please take a look. There are two meaningful commits to look at. The first is the cleanup from the code review. The second is getting the new enumerate integration test to work and adding the new unit test.

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -403,18 +404,20 @@ public void deleteControlledResource(
    * across the four types of controlled resources.
```
zloery, line 404: nit: this method doesn't build a list across resource types. it returns the list for a single type, and the aggregation happens elsewhere

zloery (approved):
LGTM

src/test/java/bio/terra/workspace/service/resource/MakeApiResourceDescriptionTest.java
```diff
@@ -0,0 +1,175 @@
...
+                resourceId,
+                resourceName,
+                description,
+                cloning,
+                assignedUser,
+                accessScopeType,
+                managedByType,
+                bucketName);
+
+            ApiResourceDescription resourceDescription = resourceController.makeApiResourceDescription((WsmResource) resource);
```
jaycarlton, line 155: why is this cast necessary?

workspace-manager-clienttests/src/main/java/scripts/testscripts/EnumerateResources.java
```diff
@@ -0,0 +1,279 @@
...
+    super.doCleanup(testUsers, workspaceApi);
+  }
+
+  private void logResult(String tag, ResourceList resourceList) {
+    List<ResourceDescription> descList = resourceList.getResources();
+    logger.info("Enumeration results for {} - {} resources", tag, descList.size());
+    for (ResourceDescription desc : descList) {
+      ResourceMetadata metadata = desc.getMetadata();
+
+      String access = "<null>";
```
jaycarlton, line 192: nit: define this string

jaycarlton (approved):

src/test/java/bio/terra/workspace/service/resource/MakeApiResourceDescriptionTest.java
ddietterich: It isn't. Even IntelliJ knows :)

PR #275, opened 2021-03-29
PF-618 Only log errors and fatal from the cloud profiler agent.

wchamber:
Change the logging level for the profiler agent to be less spammy. Ref https://cloud.google.com/profiler/docs/profiling-java#agent_logging


ddietterich (approved):
LGTM, thank you!

zloery (approved):

PR #276, opened 2021-03-29
PF-469 Add AI Notebook Instance Controlled Resource Classes

wchamber:


wchamber:
dev Sam is having a bad morning, with lots of socket timeouts. I'll make sure tests pass before submitting, but would like to go ahead with code review.

ddietterich (approved):
LGTM

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
```diff
@@ -0,0 +1,196 @@
...
+/** A {@link ControlledResource} for a Google AI Platform Notebook instance. */
+public class ControlledAiNotebookInstanceResource extends ControlledResource {
+  private final String instanceId;
+  private final String location;
+
+  @JsonCreator
+  public ControlledAiNotebookInstanceResource(
+      @JsonProperty("workspaceId") UUID workspaceId,
+      @JsonProperty("resourceId") UUID resourceId,
+      @JsonProperty("name") String name,
```
jaycarlton, line 24: it's nice to have a prefix to distinguish the "resource name" in the resource table from the, er, name of the resource (in the cloud).

src/test/java/bio/terra/workspace/common/fixtures/ControlledResourceFixtures.java
```diff
@@ -92,4 +93,24 @@ public static ControlledGcsBucketResource makeControlledGcsBucketResource(UUID w
         ManagedByType.MANAGED_BY_USER,
         BUCKET_NAME);
   }
+
+  /**
+   * Returns a {@link ControlledAiNotebookInstanceResource.Builder} that is ready to be built.
+   *
+   * <p>Tests should not rely on any particular value for the fields returned by this function and
```
jaycarlton, line 100: Interesting. The other methods in this class are being relied on, but I think I made public static constants for some of the values as well.

src/test/java/bio/terra/workspace/service/resource/ValidationUtilsTest.java
```diff
@@ -0,0 +1,31 @@
...
+    ValidationUtils.validateAiNotebookInstanceId("valid-instance-id-0");
+    assertThrows(
+        InvalidReferenceException.class,
+        () -> ValidationUtils.validateAiNotebookInstanceId("1number-first"));
+    assertThrows(
+        InvalidReferenceException.class,
+        () -> ValidationUtils.validateAiNotebookInstanceId("-dash-first"));
+    assertThrows(
+        InvalidReferenceException.class,
+        () -> ValidationUtils.validateAiNotebookInstanceId("no whitespace"));
```
jaycarlton, line 21: nit: all the other names are describing themselves, but this one should be something like `white space` so it's still positive.

src/test/java/bio/terra/workspace/service/resource/ValidationUtilsTest.java
```diff
@@ -0,0 +1,31 @@
...
+    assertThrows(
+        InvalidReferenceException.class,
+        () -> ValidationUtils.validateAiNotebookInstanceId("other-symbols^&)"));
+    assertThrows(
+        InvalidReferenceException.class,
+        () ->
+            ValidationUtils.validateAiNotebookInstanceId(
+                "more-than-62-chars111111111111111111111111111111111111111111111111111"));
+  }
```
jaycarlton, line 30: MIght as well test with a unicode character/emoji.

jaycarlton (commented):
Only potential issue is the enum.

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -61,6 +61,7 @@ public ControlledGcpResourceApiController(
     ControlledGcsBucketResource resource =
         ControlledGcsBucketResource.builder()
             .workspaceId(workspaceId)
+            .resourceId(UUID.randomUUID())
```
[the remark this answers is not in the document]
wchamber: While I don't know anything about that, I think it's moot for this PR since the `.build()` calls `UUID.randomUUID()` before this change.

src/main/java/bio/terra/workspace/service/resource/WsmResourceType.java
```diff
@@ -19,7 +19,9 @@
       ReferencedGcsBucketResource.class,
       ControlledGcsBucketResource.class),
   BIG_QUERY_DATASET(
-      CloudPlatform.GCP, "BIG_QUERY_DATASET", ReferencedBigQueryDatasetResource.class, null);
+      CloudPlatform.GCP, "BIG_QUERY_DATASET", ReferencedBigQueryDatasetResource.class, null),
+  AI_NOTEBOOK_INSTANCE(
```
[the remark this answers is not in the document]
wchamber: There's no GCP resource called an 'AI Notebook', only 'AI Notebook Instances' which are the resource wrapping a GCE VM Instance.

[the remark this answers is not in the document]
jaycarlton: OK, then maybe we could drop the Instance suffix, as these names are stacking up.

src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -26,6 +26,13 @@
   public static final Pattern BQ_DATASET_NAME_VALIDATION_PATTERN =
       Pattern.compile("^[_a-zA-Z0-9]{1,1024}$");
 
+  /**
+   * AI Notebook instances must be 1-62 characters, using lower case letters, numbers, and dashes.
+   * The first character must be a lower case letter.
+   */
+  public static final Pattern AI_NOTEBOOK_INSTANCE_NAME_VALIDATION_PATTERN =
+      Pattern.compile("(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)");
```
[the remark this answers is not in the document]
wchamber: thanks, done.

src/test/java/bio/terra/workspace/common/fixtures/ControlledResourceFixtures.java
wchamber: Yep. I've found default builders to be a very helpful testing pattern, whereas many factory function fixtures gets verbose. It's also nice to asks tests to not rely on defaults unless you really want to commit to making them public and modifying all potential uses. Better to ask tests to explicitly set what they're relying on (not that developers always follow instructions.)

src/test/java/bio/terra/workspace/service/resource/ValidationUtilsTest.java
wchamber: done

src/main/java/bio/terra/workspace/service/resource/WsmResourceType.java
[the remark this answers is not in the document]
wchamber: `Instance` is really the part that most matches the GCP API, e.g.  https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1/projects.locations.instances We could one day have AiNotebookEnvironments.
I agree it's long. For my 2c `Controlled....Resource` prefix and suffix is a lot for every class that is extending `ControlledResource`. Some shorter terminology would help, or more package prefixing, e.g. put this under `resources.controlled.ainotebooks`, and be willing to have shorter variable names for when they are imported.

PR #277, opened 2021-03-30
PF-621 Make ErrorReportException status and causes always have non-null values.

wchamber:
While many of the ErrorReportException's constructors have default HttpStatus values, the constructors that explicitly take a  HttpStatus allow the value to be null. Change these constructors so that HttpStatus is never null.

When there's a non http-status Sam ApiException, Stairway serialization of the wrapping ErrorReportException fails because it expects a non-null statusCode. 
https://github.com/DataBiosphere/terra-workspace-manager/blob/894e24101da748254297a95fbb30cd29a24c3053/src/main/java/bio/terra/workspace/service/job/StairwayExceptionSerializer.java#L45

Do a similar change with causes so that the default value for the collection is an empty list instead of null so that clients can skip a null check.

I will also make the corresponding change in terra-common-libs.

### DIFF
--- a/src/main/java/bio/terra/workspace/common/exception/ErrorReportException.java
+++ b/src/main/java/bio/terra/workspace/common/exception/ErrorReportException.java
@@ -2,14 +2,15 @@ package bio.terra.workspace.common.exception;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import javax.annotation.Nullable;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.springframework.http.HttpStatus;
 
 /**
- * This base class has data that corresponds to the ApiErrorReport model generated from the
- * OpenAPI yaml. The global exception handler auto-magically converts exceptions of this base class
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/workspace/common/exception/ErrorReportException.java
```diff
@@ -1,46 +1,55 @@
...
+    this.statusCode = DEFAULT_STATUS;
   }
 
-  public ErrorReportException(String message, List<String> causes, HttpStatus statusCode) {
+  public ErrorReportException(
+      String message, @Nullable List<String> causes, @Nullable HttpStatus statusCode) {
     super(message);
-    this.causes = causes;
-    this.statusCode = statusCode;
+    this.causes = causes != null ? causes : Collections.emptyList();
```
jaycarlton, line 41: nit: I'd prefer `ofNullable/orElse`

jaycarlton (approved):
Definitely an improvement.

PR #278, opened 2021-03-31
[PF-473][PF-512][PF-332] Cloud permissions on controlled resources

zloery:
This change adds proper GCP permissions to controlled resources based on Workspace roles and private access. As part of this, I've also revisited the existing custom GCP role definitions to match our updated understanding of resource roles.

With this change, we have the full controlled resource create/use/delete flow available. I've added GCS client support to our client tests to verify that permissions are being granted appropriately.

A couple known issues I'm declaring out of scope for this change:

- We aren't verifying that private resource users are members of a workspace. Doing this requires a little rethinking of permissions for workspace writers, who can currently create private resources but not look up the permissions of other users (PF-640).
- As part of ^, `PrivateControlledGcsBucketLifecycle` grants a workspace reader the WRITER role on a private resource, which we'll probably disallow. For now, it's a nice way to test that users get write permission by being the private resource user, and not by being workspace writers (PF-640).
- There are a few changes to make to support individual access to application-managed resources called out in comments. I don't want to implement this until we actually have application-managed resources (PF-677)
- When granting a workspace role to a user, there is a delay of 10-15s between WSM responding and that user actually receiving cloud permissions. We should consider hiding this delay inside the WSM call (PF-643).


workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -0,0 +1,272 @@
...
+    logger.info(
+        "Added {} as a reader to workspace {}", workspaceReader.userEmail, getWorkspaceId());
+    workspaceApi.grantRole(
+        new GrantRoleRequestBody().memberEmail(privateResourceUser.userEmail),
+        getWorkspaceId(),
+        IamRole.READER);
+    logger.info(
+        "Added {} as a reader to workspace {}", privateResourceUser.userEmail, getWorkspaceId());
+
+    // TODO: expecting clients to do this feels bad. This should happen inside WSM.
```
jaycarlton, line 114: Can you file a story for this?

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -0,0 +1,272 @@
...
+    String jobId = UUID.randomUUID().toString();
+    var creationParameters =
+        new GcsBucketCreationParameters()
+            .name(bucketName)
+            .location(BUCKET_LOCATION)
+            .defaultStorageClass(GcsBucketDefaultStorageClass.STANDARD)
+            .lifecycle(new GcsBucketLifecycle().rules(Collections.emptyList()));
+
+    var privateUser = new PrivateResourceIamRoles();
+    privateUser.add(ControlledResourceIamRole.WRITER);
```
jaycarlton, line 217: It seems odd for this to happen here.

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -0,0 +1,272 @@
...
+    } catch (ApiException ex) {
+      assertThat(ex.getCode(), equalTo(HttpStatusCodes.STATUS_CODE_NOT_FOUND));
+    }
+
+    // also verify it was deleted from GCP
+    Bucket maybeBucket = ownerStorageClient.get(bucketName);
+    assertThat(maybeBucket, Matchers.is(Matchers.nullValue()));
+  }
+
+  private CreatedControlledGcsBucket createPrivateBucketAttempt(
```
jaycarlton, line 206: I'd spell it `attemptCreatePrivateBucket`

workspace-manager-clienttests/src/main/java/scripts/utils/GcpCloudContextTestScriptBase.java
```diff
@@ -0,0 +1,79 @@
...
+      throws Exception {
+    super.doSetup(testUsers, workspaceApi);
+    String createContextJobId = UUID.randomUUID().toString();
+    final var requestBody =
+        new CreateCloudContextRequest()
+            .jobControl(new JobControl().id(createContextJobId))
+            .cloudPlatform(CloudPlatform.GCP);
+    CreateCloudContextResult result =
+        workspaceApi.createCloudContext(requestBody, getWorkspaceId());
+    while (ClientTestUtils.jobIsRunning(result.getJobReport())) {
```
jaycarlton, line 59: we should be able to write a generic wrapper to do this polling

jaycarlton (commented):
I'm still struggling to follow all the mappings. Hopefully some of it is reusable for other resource types.

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,90 @@
...
+import bio.terra.workspace.service.resource.controlled.ManagedByType;
+import bio.terra.workspace.service.workspace.exceptions.InternalLogicException;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * A static mapping of workspace-level IAM roles to resource-level IAM roles.
+ *
+ * <p>This maps workspace roles to the resource roles, which are used for granting cloud
+ * permissions. Sam also has a similar mapping configured for granting Sam permissions. If you
```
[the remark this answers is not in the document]
zloery: Done

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,90 @@
...
+ * <p>This mapping is slightly different for each category of controlled resource (user-shared,
+ * user-private, app-shared, app-private), so this class holds four separate maps.
+ */
+public class ControlledResourceInheritanceMapping {
+  // Currently, shared resources have the same permission inheritance regardless of whether they
+  // are user- or application-controlled. This list is pulled out separately as a convenience.
+  private static ImmutableMap<WsmIamRole, ImmutableList<ControlledResourceIamRole>>
+      SHARED_RESOURCE_MAPPING =
+          ImmutableMap.of(
+              WsmIamRole.OWNER,
```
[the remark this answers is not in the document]
zloery: hm... true if the (key, value) fits on one line, otherwise it breaks lines as usual (but with indents too).

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * A static mapping of workspace-level IAM roles to resource-level IAM roles.
+ *
+ * <p>This maps workspace roles to the resource roles, which are used for granting cloud
+ * permissions. Sam also has a similar mapping configured for granting Sam permissions. If you
+ * change this map, also check that mapping.
+ *
+ * <p>This mapping is slightly different for each category of controlled resource (user-shared,
+ * user-private, app-shared, app-private), so this class holds four separate maps.
```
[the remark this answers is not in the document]
zloery: Right now this lives in the controlled resources design doc

src/main/java/bio/terra/workspace/service/iam/CustomGcpIamRoleMapping.java
```diff
@@ -1,24 +1,22 @@
...
+import java.util.Collections;
 
 /**
  * This class specifies a static list of all GCP custom IAM roles that will be created in Workspace
- * contexts. To modify a role's permission, edit the appropriate list here. Note that currently
- * OWNER roles also receive all WRITER permissions, and WRITER roles also receive all READER
- * permissions.
+ * contexts. To modify a role's permission, edit the appropriate list here. Unlike workspace roles,
+ * resource roles are not strictly hierarchical. The EDITOR role has a distinct set of permissions
+ * from WRITER, it is not a superset.
```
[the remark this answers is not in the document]
zloery: done

src/main/java/bio/terra/workspace/service/iam/CustomGcpIamRoleMapping.java
```diff
@@ -33,11 +31,8 @@
           .build();
 
   @VisibleForTesting
-  public static final ImmutableList<String> GCS_BUCKET_OWNER_PERMISSIONS =
-      new ImmutableList.Builder<String>()
-          .addAll(GCS_BUCKET_WRITER_PERMISSIONS)
-          .add("storage.buckets.get")
-          .build();
+  public static final ImmutableList<String> GCS_BUCKET_EDITOR_PERMISSIONS =
```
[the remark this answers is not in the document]
zloery: The original custom roles had workspace-level roles and resource-level roles mixed up and overpermissioned here as a result

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+public class ControlledResourceInheritanceMapping {
+  // Currently, shared resources have the same permission inheritance regardless of whether they
+  // are user- or application-controlled. This list is pulled out separately as a convenience.
+  private static ImmutableMap<WsmIamRole, ImmutableList<ControlledResourceIamRole>>
```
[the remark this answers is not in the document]
zloery: I played around with it for a bit but creating static maps ended being up messier than the current state. I'm inclined to leave this

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/GrantGcsBucketIamRolesStep.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A step for granting cloud permissions on resources to workspace members. This follows the read-
+ * modify-write pattern of modifying permissions on cloud objects to avoid clobbering existing IAM
+ * bindings.
+ */
+public class GrantGcsBucketIamRolesStep implements Step {
```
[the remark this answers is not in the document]
zloery: good point, almost none of it is. Broke this out into `ControlledResourceCloudSyncUtils`, which should work for any resource that uses GCP's `getIamPolicy`/`setIamPolicy` api pattern and Policy objects. Most of them do, BQ datasets don't :(

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/GrantGcsBucketIamRolesStep.java
```diff
@@ -0,0 +1,172 @@
...
+        "Syncing workspace roles to GCP permissions on bucket {}", resource.getBucketName());
+
+    crlService.createStorageCow(projectId).setIamPolicy(resource.getBucketName(), newPolicy);
+    return StepResult.getStepResultSuccess();
+  }
+
+  /**
+   * GCP expects all groups to be prepended with the literal "group:" in IAM permissions bindings.
+   */
+  private String gcpGroupNameFromSamEmail(String samEmail) {
```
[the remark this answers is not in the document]
zloery: done

src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
```diff
@@ -147,9 +148,16 @@ private void assertRolesExist(Project project) throws IOException {
       String fullRoleName =
           "projects/" + project.getProjectId() + "/roles/" + customRole.getRoleName();
       Role gcpRole = crl.getIamCow().projects().roles().get(fullRoleName).execute();
-      assertThat(
-          gcpRole.getIncludedPermissions(),
-          containsInAnyOrder(customRole.getIncludedPermissions().toArray()));
+
+      List<String> expectedPermissions = customRole.getIncludedPermissions();
+      if (expectedPermissions.isEmpty()) {
```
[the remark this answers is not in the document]
zloery: nice, thanks!

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -136,15 +149,96 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
...
+    Storage readerStorageClient = ClientTestUtils.getGcpStorageClient(reader, projectId);
+
+    // Second user cannot read the object yet.
+    try {
+      readerStorageClient.get(blobId);
+      throw new IllegalStateException(
+          "User accessed a controlled workspace bucket without being a workspace member");
+    } catch (StorageException storageException) {
+      assertThat(storageException.getCode(), equalTo(HttpStatusCodes.STATUS_CODE_FORBIDDEN));
+      logger.info("User {} outside of workspace could not access bucket", reader.userEmail);
```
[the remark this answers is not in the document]
zloery: Done throughout both tests

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -136,15 +149,96 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
...
+    } catch (StorageException storageEx) {
+      logger.info("Failed to delete bucket {} directly as reader", bucketName);
+      assertThat(storageEx.getCode(), equalTo(HttpStatusCodes.STATUS_CODE_FORBIDDEN));
+    }
+
+    // TODO(PF-624): Owners and writers can actually delete buckets due to workspace-level roles
+    //  included as a temporary workaround. This needs to be removed as we transition onto WSM
+    //  controlled resources.
+
+    // Owner also cannot delete the bucket directly
```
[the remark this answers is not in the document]
zloery: I'd prefer to leave it in - this is ready to uncomment once the change is made for PF-633, which I expect to do in the next ~week.

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
zloery: Done

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
zloery: `PrivateResourceIamRoles` is really just a name for `ArrayList<ControlledResourceIamRole>`, so it doesn't get the fluent builders like other generated objects. This seemed like the best option

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
zloery: done

workspace-manager-clienttests/src/main/java/scripts/utils/GcpCloudContextTestScriptBase.java
zloery: agreed, but that seems like something we might want as a util inside the TestRunner library

zloery:
`ControlledResourceInheritanceMapping` (mapping from workspace roles to resource roles) is common to all controlled resource types.
`CustomGcpIamRoleMapping` (mappings from resource roles -> GCP permissions) isn't, since the GCP permissions we grant  are resource-type specific.

src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
```diff
@@ -147,9 +149,10 @@ private void assertRolesExist(Project project) throws IOException {
       String fullRoleName =
           "projects/" + project.getProjectId() + "/roles/" + customRole.getRoleName();
       Role gcpRole = crl.getIamCow().projects().roles().get(fullRoleName).execute();
-      assertThat(
-          gcpRole.getIncludedPermissions(),
-          containsInAnyOrder(customRole.getIncludedPermissions().toArray()));
+
+      List<String> gcpPermissions =
+          Optional.ofNullable(gcpRole.getIncludedPermissions()).orElse(Collections.emptyList());
```
wchamber, line 154: prefer empty collections instead of null collections in objects, or else all clients have to do this kind of error-prone checking.

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -136,15 +149,96 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
...
+    Storage readerStorageClient = ClientTestUtils.getGcpStorageClient(reader, projectId);
+
+    // Second user cannot read the object yet.
+    try {
+      readerStorageClient.get(blobId);
+      throw new IllegalStateException(
+          "User accessed a controlled workspace bucket without being a workspace member");
+    } catch (StorageException storageException) {
+      assertThat(storageException.getCode(), equalTo(HttpStatusCodes.STATUS_CODE_FORBIDDEN));
+      logger.info("User {} outside of workspace could not access bucket as expected", reader.userEmail);
```
wchamber, line 177: can we use `assertThrows` instead? Ditto other try catch assert blocks

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -0,0 +1,273 @@
...
+
+  private CreatedControlledGcsBucket attemptCreatePrivateBucket(
+      ControlledGcpResourceApi resourceApi) throws Exception {
+    String jobId = UUID.randomUUID().toString();
+    var creationParameters =
+        new GcsBucketCreationParameters()
+            .name(bucketName)
+            .location(BUCKET_LOCATION)
+            .defaultStorageClass(GcsBucketDefaultStorageClass.STANDARD)
+            .lifecycle(new GcsBucketLifecycle().rules(Collections.emptyList()));
```
wchamber, line 214: are all of these required fields?

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -0,0 +1,273 @@
...
+    } catch (ApiException ex) {
+      assertThat(ex.getCode(), equalTo(HttpStatusCodes.STATUS_CODE_NOT_FOUND));
+    }
+
+    // also verify it was deleted from GCP
+    Bucket maybeBucket = ownerStorageClient.get(bucketName);
+    assertThat(maybeBucket, Matchers.is(Matchers.nullValue()));
+  }
+
+  private CreatedControlledGcsBucket attemptCreatePrivateBucket(
```
wchamber, line 206: nit: What's more "attempted" about this than anything else?

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -0,0 +1,273 @@
...
+    try {
+      resourceApi.getBucket(getWorkspaceId(), resourceId);
+      throw new IllegalStateException("Incorrectly found a deleted bucket!");
+    } catch (ApiException ex) {
+      assertThat(ex.getCode(), equalTo(HttpStatusCodes.STATUS_CODE_NOT_FOUND));
+    }
+
+    // also verify it was deleted from GCP
+    Bucket maybeBucket = ownerStorageClient.get(bucketName);
+    assertThat(maybeBucket, Matchers.is(Matchers.nullValue()));
```
wchamber, line 203: assertNull?

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -0,0 +1,273 @@
...
+  @Override
+  public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspaceApi)
+      throws Exception {
+
+    ControlledGcpResourceApi resourceApi =
+        ClientTestUtils.getControlledGcpResourceClient(testUser, server);
+
+    // Create a private bucket
+    CreatedControlledGcsBucket bucket = attemptCreatePrivateBucket(resourceApi);
+    assertThat(bucket.getJobReport().getStatus(), equalTo(JobReport.StatusEnum.SUCCEEDED));
```
wchamber, line 86: assertEquals?

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+    return result;
+  }
+
+  @Override
+  protected void doCleanup(List<TestUserSpecification> testUsers, WorkspaceApi workspaceApi)
+      throws ApiException {
+    super.doCleanup(testUsers, workspaceApi);
+    if (bucketName != null) {
+      logger.warn("Test failed to cleanup bucket " + bucketName);
```
wchamber, line 270: why doesn't this rely on janitor? When are we going to do anything about this warning?

workspace-manager-clienttests/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -87,24 +93,32 @@ public static ApiClient getClientForTestUser(
       logger.debug(
           "Fetching credentials and building Workspace Manager ApiClient object for test user: {}",
           testUser.name);
-
-      // refresh the user token
+      // TODO: TestRunner caches delegated credentials by TestUser, ignoring scopes. This should
```
wchamber, line 96: todo a ticket?

workspace-manager-clienttests/src/main/java/scripts/utils/GcpCloudContextTestScriptBase.java
```diff
@@ -0,0 +1,79 @@
...
+ * setup() and cleanup(), although doSetup() and doCleanup() can still be overridden, provided the
+ * caller calls super().
+ *
+ * <p>Per WorkspaceAllocateTestScriptBase, you must specify the spend profile to use as the first
+ * parameter in the test configuration.
+ *
+ * <p>No doUserJourney() implementation is provided, and this must be overridden by inheriting
+ * classes.
+ */
+public abstract class GcpCloudContextTestScriptBase extends WorkspaceAllocateTestScriptBase {
```
wchamber, line 30: Using inheritance to do test setup worries me. Java doesn't support multiple inheritance, so if we have two setup cases, and want to combine them, we have to start making multiple classes. Can we use composition instead of inheritance here to do various bits of setup? The composable setup bits can have input parameters and accessor functions for the things they create, like the project id. 
Not necessarily this PR to change everything, but maybe at least start a better way for this new setup wrinkle.

workspace-manager-clienttests/src/main/java/scripts/utils/GcpCloudContextTestScriptBase.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+    assertThat(result.getJobReport().getStatus(), equalTo(StatusEnum.SUCCEEDED));
+    projectId = result.getGcpContext().getProjectId();
+  }
+
+  /**
+   * Clean up the context first, then delete the workspace via super cleanup method.
+   *
+   * <p>This is not strictly necessary as deleting a workspace also deletes the context, but it
+   * ensures we exercise the deleteContext endpoint.
```
wchamber, line 71: If we want to test the endpoint, we should have an explicit test for it.

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,87 @@
...
+              WsmIamRole.WRITER, ImmutableList.of(ControlledResourceIamRole.EDITOR),
+              // Applications and readers have no permissions on private resources.
+              WsmIamRole.APPLICATION, ImmutableList.of(),
+              WsmIamRole.READER, ImmutableList.of());
+
+  public static ImmutableMap<WsmIamRole, ImmutableList<ControlledResourceIamRole>>
+      getInheritanceMapping(AccessScopeType accessScope, ManagedByType managedBy) {
+    if (accessScope == AccessScopeType.ACCESS_SCOPE_SHARED) {
+      if (managedBy == ManagedByType.MANAGED_BY_USER) {
+        return USER_SHARED_MAPPING;
```
wchamber, line 71: do these still need to be public if this is the right way to access them? Should they just be in a Table<AccessScopeType, ManagedByType, Map> instead?

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,87 @@
...
+    } else if (accessScope == AccessScopeType.ACCESS_SCOPE_PRIVATE) {
+      if (managedBy == ManagedByType.MANAGED_BY_USER) {
+        return USER_PRIVATE_MAPPING;
+      } else if (managedBy == ManagedByType.MANAGED_BY_APPLICATION) {
+        return APPLICATION_PRIVATE_MAPPING;
+      }
+    }
+    throw new InternalLogicException(
+        String.format(
+            "Inheritance map not specified for access scope %s and ManagedByType %s",
```
wchamber, line 84: nit: inconsistent between access scope and ManagedByType capitlalization

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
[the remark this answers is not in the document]
wchamber: Even with the builder? I'd think putAll would be ok. https://guava.dev/releases/17.0/api/docs/com/google/common/collect/ImmutableMultimap.Builder.html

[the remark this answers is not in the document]
wchamber: Even with the builder? I'd think `putAll` would work ok. https://guava.dev/releases/17.0/api/docs/com/google/common/collect/ImmutableMultimap.Builder.html

wchamber:
Thinking a little more, I think this is a little hard to follow because there's a large chain of maps that's spread across many enum types and static mappings. Maybe we can make the mapping chain more explicit with more behavior on enums, or dedicated classes for walking the map? Not sure.
What's the full list of maps? Is there some hierarchy?
(WsmResourceType, ControlledResourceIamRole) -> CustomGcpRole
(ManagedByTpe, AccessType, WsmIamRole) -> ControlledResourceIamRole
Maybe not having the intermediate maps is helpful? Is it helpful to have a `ResourceKind` type that's the tuple of ManagedByType and AccessType?

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/SyncResourceSamGroupsStep.java
```diff
@@ -0,0 +1,52 @@
...
+        samService.syncResourcePolicy(resource, ControlledResourceIamRole.WRITER, userReq));
+    workingMap.put(
+        ControlledResourceKeys.IAM_RESOURCE_READER_GROUP_EMAIL,
+        samService.syncResourcePolicy(resource, ControlledResourceIamRole.READER, userReq));
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    // This step only modifies the flight working map, nothing to undo.
```
wchamber, line 49: Isn't this also doing a Sam thing? I don't think there's something to undo, but good to be accurate.

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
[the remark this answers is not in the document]
zloery: ah, that's only in ImmutableMultimap and I missed that. Thanks!

build.gradle
```diff
@@ -77,7 +77,7 @@ dependencies {
 	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-bigquery', version: '0.4.0'
 	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-cloudresourcemanager', version: '0.4.0'
 	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-serviceusage', version: '0.4.0'
-	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-storage', version: '0.4.0'
+	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-storage', version: '0.5.0'
```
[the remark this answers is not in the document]
zloery: bumped to 0.7.0 since that's out now.
in general, I prefer only changing dependencies relevant to PRs for the sake of (relatively) smaller and more focused changes. separately, it's probably worth a pass updating our dependencies semi-frequently

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * A static mapping of workspace-level IAM roles to resource-level IAM roles.
+ *
+ * <p>This maps workspace roles to the resource roles, which are used for granting cloud
+ * permissions. Sam also has a similar mapping configured for granting Sam permissions. If you
+ * change this map, also check that mapping:
+ * https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf
+ *
+ * <p>This mapping is slightly different for each category of controlled resource (user-shared,
+ * user-private, app-shared, app-private), so this class holds four separate maps.
```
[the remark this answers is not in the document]
zloery: done, slightly more than a sentence

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,87 @@
...
+      APPLICATION_PRIVATE_MAPPING =
+          ImmutableMap.of(
+              WsmIamRole.OWNER, ImmutableList.of(ControlledResourceIamRole.EDITOR),
+              WsmIamRole.WRITER, ImmutableList.of(ControlledResourceIamRole.EDITOR),
+              // Applications and readers have no permissions on private resources.
+              WsmIamRole.APPLICATION, ImmutableList.of(),
+              WsmIamRole.READER, ImmutableList.of());
+
+  public static ImmutableMap<WsmIamRole, ImmutableList<ControlledResourceIamRole>>
+      getInheritanceMapping(AccessScopeType accessScope, ManagedByType managedBy) {
```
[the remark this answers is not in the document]
zloery: done

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/ControlledResourceCloudSyncUtils.java
```diff
@@ -0,0 +1,147 @@
...
+   *
+   * @param resource The controlled resource bindings are being build for
+   * @param projectId The ID of the GCP project permissions are applied to
+   * @param currentPolicy The current IAM policy on the cloud object
+   * @param workingMap The Stairway working flight map. This method expects Sam group emails to be
+   *     provided in this list using the appropriate WorkspaceFlightMapKeys
+   * @return
+   */
+  public static Policy updatePolicyWithSamGroups(
+      ControlledResource resource, String projectId, Policy currentPolicy, FlightMap workingMap) {
```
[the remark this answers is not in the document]
zloery: Yeah - I think part of the issue is that this isn't really a "utility" function, it's a common implementation for a couple of closely-related steps. Implementing that as a static function means a weird interface and heavy assumptions about the flight map you pass in, but implementing it as something like an abstract class felt like a misuse of types. I'm inclined to use this structure for now with the updated comments, but open to other ideas.

src/main/java/bio/terra/workspace/service/iam/CustomGcpIamRole.java
```diff
@@ -31,8 +33,9 @@ public CustomGcpIamRole(
    * Builds the name of a custom GCP IAM role from the resource type and workspace-level IAM
    * combination it applies to.
    */
-  public static String customGcpRoleName(WsmResourceType resourceType, WsmIamRole iamRole) {
-    return resourceType.name() + "_" + iamRole.name();
+  public static String customGcpRoleName(
```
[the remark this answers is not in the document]
zloery: SG. I was originally on the fence between keeping a static collection of them or just using names - switched to using the full objects everywhere and modified some of the methods here

src/main/java/bio/terra/workspace/service/iam/CustomGcpIamRoleMapping.java
```diff
@@ -70,34 +65,49 @@
...
-          .build();
+  public static final ImmutableList<String> BIG_QUERY_DATASET_EDITOR_PERMISSIONS =
+      ImmutableList.of(
+          "bigquery.datasets.getIamPolicy",
+          "bigquery.tables.create",
+          "bigquery.tables.delete",
+          "bigquery.tables.update");
 
+  // The ASSIGNER resource does not grant GCP permissions, only Sam permissions. It's included in
+  // this map for completeness.
```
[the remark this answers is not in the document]
zloery: It's helpful for keeping the model of "one GCP custom role for each (resource type, resource role) pair", and not needing to add special-case logic in other places. This does create the role in google with empty permissions

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/GrantGcsBucketIamRolesStep.java
```diff
@@ -0,0 +1,59 @@
...
+    this.workspaceService = workspaceService;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    final FlightMap workingMap = flightContext.getWorkingMap();
+    String projectId = workspaceService.getRequiredGcpProject(resource.getWorkspaceId());
+    // Users do not have read or write access to IAM policies, so this request is executed via
+    // WSM's service account.
```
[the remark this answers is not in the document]
zloery: storage and bq cows are slightly different than the other ones - this actually creates the cow with WSM SA credentials. There's an overloaded method which takes an `AuthenticatedUserRequest` for making requests on a user's behalf (currently used for verifying access to references), and this comment is to distinguish between the two

src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlight.java
```diff
@@ -45,8 +52,8 @@ public CreateGcpContextFlight(FlightMap inputParameters, Object applicationConte
             INITIAL_INTERVAL_SECONDS, MAX_INTERVAL_SECONDS, MAX_OPERATION_TIME_SECONDS);
     addStep(new SetProjectBillingStep(crl.getCloudBillingClientCow()));
     addStep(new CreateCustomGcpRolesStep(crl.getIamCow()), retryRule);
-    addStep(new StoreGcpContextStep(appContext.getWorkspaceDao()), retryRule);
-    addStep(new SyncSamGroupsStep(appContext.getSamService()), retryRule);
+    addStep(new StoreGcpContextStep(appContext.getWorkspaceDao(), workspaceId), retryRule);
+    addStep(new SyncSamGroupsStep(appContext.getSamService(), workspaceId, userReq), retryRule);
```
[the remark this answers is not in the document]
zloery: I'm not sure about a general preferred style. In this case, moving workspaceId to the constructor meant I could reuse the step without adding a workspace ID flightmap key - and once it was read in the constructor, I passed it into other steps for consistency.
There's probably a broader question about cross-package step use and whether we need a common package here

src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
zloery: Agreed, but Role is a GCP client library object, so we're stuck with this approach here

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
zloery: It looks like we previously didn't have JUnit assertions. I've added them to the WSM TestRunner project here, which gives more useful assertions

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
zloery: yes, these are all required

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
zloery: The client tests never directly create cloud resources, they do so through WSM (which should have Janitor configured in the environments we test against). Removed the warning, since it's not actionable

workspace-manager-clienttests/src/main/java/scripts/utils/GcpCloudContextTestScriptBase.java
zloery: At least so far, the setup cases have been pretty distinct and broad (workspace fixture -> cloud context fixture) - I don't anticipate building more of these fixtures in the near future. I think we probably should move to composition if we expand the set more, but as-is I'm not sure when we'd do that.

workspace-manager-clienttests/src/main/java/scripts/utils/GcpCloudContextTestScriptBase.java
zloery: we should (and we have PF-330 in the backlog for that), but making the extra call here seems reasonable

src/main/java/bio/terra/workspace/service/iam/ControlledResourceInheritanceMapping.java
zloery: done. building a table for 4 elements seems a little overkill, but it could reasonably go either way

zloery:
Worth calling out: I moved `ControlledResourceInheritanceMapping` and `CustomGcpIamRoleMapping` to a new package so they're at least a little closer together. These are two of the three static mappings we're using - the third is `CloudSyncRoleMapping`, which grants project-level roles and which we're slowly transitioning off. These are the only maps used here, I think having them in one place helps a bit with clarity

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,91 @@
...
+  /**
+   * Returns the resource-level roles that each workspace-level role inherits on the specified
+   * category of resource (user-shared, user-private, app-shared, or app-private).
+   *
+   * @return A map whose keys are workspace-level roles and values are the corresponding inherited
+   *     resource-level roles for the specified resource category.
+   */
+  public static Multimap<WsmIamRole, ControlledResourceIamRole> getInheritanceMapping(
+      AccessScopeType accessScope, ManagedByType managedBy) {
+    if (accessScope == AccessScopeType.ACCESS_SCOPE_SHARED) {
```
jaycarlton, line 73: This could be stored as a `Table` as well.

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/CustomGcpIamRole.java
```diff
@@ -0,0 +1,52 @@
...
+    this.iamRole = iamRole;
+    this.includedPermissions = includedPermissions;
+  }
+
+  /**
+   * Our custom role names are a combination of resource type and IAM role name, e.g.
+   * GCS_BUCKET_READER. This is an arbitrary standard designed to be human-readable.
+   */
+  public String getRoleName() {
+    return resourceType.name().toUpperCase() + "_" + iamRole.name().toUpperCase();
```
jaycarlton, line 38: nit: you might want to use two underscores as the separator, as single underscores can occur within types and roles (potentially).

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/CustomGcpIamRoleMapping.java
```diff
@@ -0,0 +1,140 @@
...
+                      WsmResourceType.BIG_QUERY_DATASET,
+                      ControlledResourceIamRole.EDITOR,
+                      BIG_QUERY_DATASET_EDITOR_PERMISSIONS))
+              .put(
+                  WsmResourceType.BIG_QUERY_DATASET,
+                  ControlledResourceIamRole.ASSIGNER,
+                  new CustomGcpIamRole(
+                      WsmResourceType.BIG_QUERY_DATASET,
+                      ControlledResourceIamRole.ASSIGNER,
+                      Collections.emptyList()))
```
jaycarlton, line 138: You could use the same convention as the multimaps and omit this entry from the table.

src/main/resources/application.yml
```diff
@@ -56,7 +56,7 @@ workspace:
   sam.base-path: ${env.urls.sam}
 
   stairway-database:
-    force-clean: ${env.db.init}
+    force-clean: true
```
jaycarlton, line 59: You don't want to check this in if it does what I think it does. ⚠️

jaycarlton (approved):

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/ControlledResourceInheritanceMapping.java
zloery: It could, though a 4 element table seems a little overkill - I could go either way on this

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/CustomGcpIamRole.java
zloery: It can, e.g. `GCS_BUCKET_READER`. I think that's fine - this should be a human-readable name, making it extra clear which part comes from where isn't that necessary.

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/CustomGcpIamRoleMapping.java
zloery: I think tables and multimaps treat this differently. In a multimap, a key with no value is the same as the key being there. In a table, having the empty list here is different than not having it

src/main/resources/application.yml
zloery: !! it does, forgot to clean that up from personal env testing and I definitely don't want to check this in. thanks!

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/ControlledResourceCloudSyncUtils.java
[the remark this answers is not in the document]
wchamber: Yeah, struggling still. I feel like we're missing some class(es) that brings together all the stuff involved in setting up the permissions for a controlled resource - the resource type, its corresponding gcp permissions, the sam resource and the policies on it, but I'm having a hard time seeing the relationship, or putting my finger on it. As-is, there's a bunch of interconnected static functions that need to be called together. Some 'policy' or 'binding' class, maybe the pair of the ControlledResource (or the (WsmResourceType, AccessScope, ManagedBy) tuple)  and the ControlledResourceIamRole that pop up together in creating the sam policy and the gcp bindings. Oh well.

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -320,23 +321,75 @@ public void removeWorkspaceRole(
...
+        group,
+        resource.getResourceId());
+    return group;
+  }
+
+  /**
+   * Common implementation for syncing a policy to a Google group on an object in Sam.
+   *
+   * @param resourceTypeName The type of the Sam object, as configured with Sam.
+   * @param resourceId The ID of the resource to sync a policy for
```
wchamber, line 377: nit: The Sam ID of the Sam resource to sync a policy for.

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -320,23 +321,75 @@ public void removeWorkspaceRole(
...
+        role.toSamRole(),
+        group,
+        resource.getResourceId());
+    return group;
+  }
+
+  /**
+   * Common implementation for syncing a policy to a Google group on an object in Sam.
+   *
+   * @param resourceTypeName The type of the Sam object, as configured with Sam.
```
wchamber, line 376: nit: 'Sam object' or 'Sam resource'?

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -320,23 +321,75 @@ public void removeWorkspaceRole(
...
+   * <p>This operation in Sam is idempotent, so we don't worry about calling this multiple times.
+   *
+   * @param resource The resource to sync a binding for
+   * @param role The policy to sync in Sam
+   * @param userReq User authentication
+   * @return
+   */
+  @Traced
+  public String syncPrivateResourcePolicy(
+      ControlledResource resource,
```
wchamber, line 350: little funny that we only need the workspaceId for `syncWorkspacePolicy` but we need the entire ControlledResource class here. I think we only use the resource "category" access/managed by and the resource id.

wchamber (commented):
last few comments, sorry for the slowness. Still trying to see some missing class of resource category and ControlledResourceIamRole binding.

workspace-manager-clienttests/src/main/java/scripts/utils/GcpCloudContextTestScriptBase.java
zloery: looks like I might not be the one starting anyways - #274 introduced a static method way of doing it, pulled that in here to hopefully minimize merge conflicts

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -333,7 +334,9 @@ public String syncWorkspacePolicy(
    * return the email of that group.
    *
    * <p>This should only be called for controlled resources which require permissions granted to
-   * individual users, i.e. private or application-controlled resources.
+   * individual users, i.e. private or application-controlled resources. All other cases are handled
+   * by the permissions that workspace-level roles inherit on resources, and do not use the policies
+   * synced by this function.
```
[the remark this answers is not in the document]
zloery: Added a brief note here, this is inheritance using hierarchical resources in Sam. I'm trying to hit the right balance of detail in the javadoc here, at some point this works better in the design doc

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -343,12 +346,18 @@ public String syncWorkspacePolicy(
...
    */
   @Traced
-  public String syncResourcePolicy(
+  public String syncPrivateResourcePolicy(
       ControlledResource resource,
       ControlledResourceIamRole role,
       AuthenticatedUserRequest userReq) {
+    // TODO: in the future, this function will also be called for application managed resources,
+    //  including app-shared. This check should be modified appropriately.
+    if (resource.getAccessScope() != AccessScopeType.ACCESS_SCOPE_PRIVATE) {
```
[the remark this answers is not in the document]
zloery: Right now we just throw at creation time if someone creates an application-managed resource, so it wouldn't execute either way. I'm fairly sure we could make this check "private or application-managed", but I've been wrong on applications before. I'm inclined to leave it out until we actually support application-managed resources

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,91 @@
...
+import bio.terra.workspace.service.workspace.exceptions.InternalLogicException;
+import com.google.common.collect.ImmutableMultimap;
+import com.google.common.collect.Multimap;
+
+/**
+ * Static mappings of workspace-level IAM roles to resource-level IAM roles.
+ *
+ * <p>This mapping represents which resource-level roles are granted to workspace members based on
+ * their workspace-level role. This mapping is slightly different for each category of controlled
+ * resource (user-shared, user-private, app-shared, app-private), so this class holds four separate
```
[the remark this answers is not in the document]
zloery: Yeah - this is something we went back and forth on earlier on with controlled resources. We ultimately decided against it because it would only be used in a few spots for IAM, but I still use that notion of a category a lot while describing things.

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * <p>For example, all workspace readers should be readers of a user-shared bucket in that
+ * workspace, but the same is not true for a user-private bucket. This is represented in the maps
+ * below, where WsmIamRole.READER maps to ControlledResourceIamRole.READER for shared resources but
+ * that entry is not present for private resources.
+ *
+ * <p>This maps workspace roles to the resource roles, which are used for granting cloud
+ * permissions. Sam also has a similar mapping configured for granting Sam permissions. If you
+ * change this map, also check that mapping:
+ * https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf
```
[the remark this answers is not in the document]
zloery: I don't like linking to line numbers in modifiable files, but added a pointer to the right section

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/ControlledResourceInheritanceMapping.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+public class ControlledResourceInheritanceMapping {
```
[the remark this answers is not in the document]
zloery: fair point - I picked up the "inherited" language from the hierarchical resource design in Sam, where policies on "parent" resources are inherited by "child" resources. Since most of the resource IAM inside WSM is based on that feature in Sam

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/CustomGcpIamRole.java
```diff
@@ -0,0 +1,52 @@
+package bio.terra.workspace.service.resource.controlled.mappings;
+
+import bio.terra.workspace.service.iam.model.ControlledResourceIamRole;
+import bio.terra.workspace.service.resource.WsmResourceType;
+import java.util.List;
+
+/**
+ * A CustomGcpIamRole is a POJO value class representing a GCP custom IAM role. These GCP custom
```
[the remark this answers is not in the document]
zloery: I don't think that adds much, and I wonder if that just makes it more complex to take in a `WsmResourceType` and `ControlledResourceIamRole` but only store a name.

src/test/java/bio/terra/workspace/service/iam/CustomGcpIamRoleMappingTest.java
```diff
@@ -20,24 +21,10 @@ void bucketWriterContainsReader() {
         everyItem(in(CustomGcpIamRoleMapping.GCS_BUCKET_WRITER_PERMISSIONS)));
   }
 
-  @Test
-  void bucketOwnerContainsWriter() {
-    assertThat(
-        CustomGcpIamRoleMapping.GCS_BUCKET_WRITER_PERMISSIONS,
-        everyItem(in(CustomGcpIamRoleMapping.GCS_BUCKET_OWNER_PERMISSIONS)));
-  }
```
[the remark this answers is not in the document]
zloery: whoops - I updated the comment on the non-test class, but not the one here.

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/CustomGcpIamRole.java
```diff
@@ -0,0 +1,52 @@
...
+ * A CustomRole object holds a name made by combining the IAM role and resource type, and also holds
+ * a set of GCP cloud permissions that are granted. See the full list of these permissions at
+ * https://cloud.google.com/iam/docs/permissions-reference
+ *
+ * <p>The role name is the combination of resource type + IAM role name with an underscore
+ * separator, e.g. GCS_BUCKET_READER.
+ */
+public class CustomGcpIamRole {
+  private final WsmResourceType resourceType;
+  private final ControlledResourceIamRole iamRole;
```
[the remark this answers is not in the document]
zloery: fair point, removed the resource type and IAM role fields and replaced with a `roleName` field. also added documentation explaining this to the constructor that still takes those inputs

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/CustomGcpIamRole.java
```diff
@@ -0,0 +1,52 @@
...
+
+import bio.terra.workspace.service.iam.model.ControlledResourceIamRole;
+import bio.terra.workspace.service.resource.WsmResourceType;
+import java.util.List;
+
+/**
+ * A CustomGcpIamRole is a POJO value class representing a GCP custom IAM role. These GCP custom
+ * roles are used to narrowly scope the sets of cloud permissions we grant to users.
+ *
+ * <p>Each custom role is the application of a resource-level IAM role to a specific resource type.
```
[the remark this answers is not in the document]
zloery: At the GCP level that's true, but the "specific resource type" bit is an additional restriction imposed by WSM to enforce narrowly scoped sets of permissions.
Clarified a bit, but I think this should be `IAM role`. It's a little overloaded between "resource-level IAM role" and "GCP role" elsewhere, though

src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -5,6 +5,14 @@
 import com.google.common.collect.ImmutableMap;
 import java.util.List;
 
+/**
+ * This mapping describes the project-level GCP roles granted to members of a workspace.
+ *
+ * <p>Granting these roles at the project level was implemented as a temporary workaround to support
+ * objects in a cloud context before controlled resources were built. As controlled resources become
+ * available, roles should be granted directly on controlled resources instead (see {@code
+ * CustomGcpIamRoleMapping}), and should be removed from this list.
```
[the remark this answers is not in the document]
zloery: Permissions granted at the project level apply to all the resources in that project - so, anyone granted "bucket reader" at the project level gets read access to all buckets in the project, which breaks the concept of private resources

src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: In Sam terms, "resource". inside WSM, that's overloaded so I went with `syncPolicyOnObject` for the method - cleaned up params/javadocs for consistency though.

src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: Yeah - I think I'd rather pass the whole object than 3 separate fields, but it's not fantastic.

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
wchamber: bump

workspace-manager-clienttests/src/main/java/scripts/utils/CloudContextMaker.java
```diff
@@ -0,0 +1,50 @@
...
+import bio.terra.workspace.model.CreateCloudContextRequest;
+import bio.terra.workspace.model.CreateCloudContextResult;
+import bio.terra.workspace.model.JobControl;
+import bio.terra.workspace.model.JobReport;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CloudContextMaker {
```
wchamber, line 17: add some java doc on what this is useful for.

workspace-manager-clienttests/src/main/java/scripts/utils/CloudContextMaker.java
```diff
@@ -0,0 +1,50 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(CloudContextMaker.class);
+  private static final long CREATE_CONTEXT_POLL_SECONDS = 10;
```
wchamber, line 19: nit: even when only used in sleep, prefer `Duration` to primitives for time durations.

wchamber (approved):
Thanks! Sorry for the slow turnaround time.
Couple nits, and one pending question on required fields for the API, but not really this PR's concern anyway.

workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
zloery: whoops, missed this one - digging in a little more, these shouldn't actually be required, but we do throw a 500 atm because we aren't handling nulls properly somewhere. PF-635 to fix

workspace-manager-clienttests/src/main/java/scripts/utils/CloudContextMaker.java
zloery: done

workspace-manager-clienttests/src/main/java/scripts/utils/CloudContextMaker.java
zloery: makes sense, this is really storing both a value and unit.

src/main/java/bio/terra/workspace/service/resource/controlled/mappings/ControlledResourceInheritanceMapping.java
jaycarlton: I like smart data structures and dumb code, and I penalize `if` statements.

src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlight.java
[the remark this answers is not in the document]
ddietterich: I don't object to pulling input params out of the map in the flight constructor.

ddietterich (approved):
I have several suggestions. Overall, it looks fine.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateControlledResourceFlight.java
```diff
@@ -20,40 +22,53 @@ public CreateControlledResourceFlight(FlightMap inputParameters, Object beanBag)
...
-    final List<ControlledResourceIamRole> privateResourceIamRole =
+    final List<ControlledResourceIamRole> privateResourceIamRoles =
         inputParameters.get(ControlledResourceKeys.PRIVATE_RESOURCE_IAM_ROLES, List.class);
 
+    // store the resource metadata in the WSM database
+    addStep(new StoreMetadataStep(flightBeanBag.getResourceDao()));
+
+    // create the Sam resource associated with the resource
+    addStep(
+        new CreateSamResourceStep(
```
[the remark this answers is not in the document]
zloery: Moved to serializing maps instead of individual group names - that also saves a bunch of duplicated code, thanks!

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/SyncResourceSamGroupsStep.java
```diff
@@ -0,0 +1,52 @@
...
+    this.samService = samService;
+    this.resource = resource;
+    this.userReq = userReq;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    FlightMap workingMap = flightContext.getWorkingMap();
+    workingMap.put(
```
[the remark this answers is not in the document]
zloery: switched to a map here as well, and yep `syncPrivateResourcePolicy` is idempotent

src/main/java/bio/terra/workspace/service/iam/SamService.java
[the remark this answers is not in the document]
zloery: I think it still would be called for application shared, because we need to grant one app SA extra (i.e. resource editor) permission on the resource. Either way, I'd rather not implement this until we have apps so we can say for sure.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/ControlledResourceCloudSyncUtils.java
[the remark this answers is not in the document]
zloery: That's definitely a cleaner interface than the static method, switched to this

PR #279, opened 2021-04-01
PF-522: roll back Multi-instance changes

jaycarlton:
Decided to roll this back ahead of the release Monday.

revert 75a8c0e5f7d623fdce4a7d71beca7bd29e8cf71a

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ plugins {
 }
 
 sourceCompatibility = JavaVersion.VERSION_11
-def useMavenLocal = false
+
 allprojects {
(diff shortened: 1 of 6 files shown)


zloery (approved):
LGTM as a clean revert of #268. I haven't reviewed this thoroughly, but this was the most recent change to WSM.

yonghaoy (approved):
Seem slack discussion on how to rollback. Rolling back in server code is harder and slower. But seems you already choose this approach, I think this would also work. To fix alpha: You need to make a new release with the new image version after version is bumped

PR #280, opened 2021-04-01
PF-522: delay hook

jaycarlton:
Delay hook for manual testing and debugging. Makes a flight step take arbitrarily long.


zloery (commented):
The hook itself LGTM, but how will this get registered in WSM's stairway instance?

jaycarlton:
> The hook itself LGTM, but how will this get registered in WSM's stairway instance?

Same as the other hoooks, in JobService. You just don't check that in.

zloery (approved):
Ah, got it. I missed that this was for manual testing and assumed it would be included in automated tests.

tnadler (approved):

PR #287, opened 2021-04-08
[PF-607] Fix Sam authz check actions

zloery:
Currently, a number of authz checks to Sam check for permission to `write` to a workspace instead of more specific actions. This change corrects those places, and calls to `get` or `delete` a controlled resource will now look for the correct permission directly on the resource instead of on the workspace.

Functionally, this should not change any behavior, as every workspace-level IAM role that grants `write` permission also grants these more specific permissions.


src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -31,17 +36,59 @@
...
+   * @param resourceId id of the resource in question
+   * @param action the action to authorize against the resource
+   * @return the resource, if it exists and the user is permitted to perform the specified action.
+   */
+  @Traced
+  public ControlledResource validateControlledResourceAndAction(
+      AuthenticatedUserRequest userReq, UUID workspaceId, UUID resourceId, String action) {
+    WsmResource resource = resourceDao.getResource(workspaceId, resourceId);
+    if (resource.getStewardshipType() != StewardshipType.CONTROLLED) {
+      throw new InvalidControlledResourceException(
```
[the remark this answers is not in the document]
ddietterich: We know the name of the resource, too, which might be more useful to humans?
Should this always be combined with a "do you have any access to the workspace?" check?

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -31,17 +36,59 @@
...
+   * <p>Throws ResourceNotFound from getResource if the workspace does not exist in the specified
+   * workspace, regardless of the user's permission.
+   *
+   * <p>Throws InvalidControlledResourceException if the given resource is not controlled.
+   *
+   * <p>Throws SamUnauthorizedException if the user is not permitted to perform the specified action
+   * on the resource in question.
+   *
+   * <p>Returns the controlled resource object if it exists and the user is permitted to perform the
+   * specified action.
```
ddietterich, line 68: I appreciate the comment. Very clear

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -104,7 +104,7 @@ void testGetExistingWorkspace() {
   void testGetForbiddenMissingWorkspace() {
     doThrow(new SamUnauthorizedException("forbid!"))
         .when(mockSamService)
-        .workspaceAuthzOnly(any(), any(), any());
+        .checkAuthz(any(), any(), any(), any());
```
ddietterich, line 107: nit: You could put these tests in an `@Nested` and have a second `@BeforeEach` that sets up the mock.

ddietterich (approved):
LGTM - one question about workspace-level permission checks

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -208,29 +208,39 @@ public boolean isAuthorized(
...
+   * @param resourceId The ID of the resource being checked
+   * @param resourceType The Sam type of the resource being checked
+   * @param action The action being checked on the resource
+   * @param userReq Credentials of the user whose permissions are being checked
+   */
   @Traced
-  public void workspaceAuthzOnly(
-      AuthenticatedUserRequest userReq, UUID workspaceId, String action) {
+  public void checkAuthz(
+      String resourceId, String resourceType, String action, AuthenticatedUserRequest userReq) {
```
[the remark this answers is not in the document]
zloery: done. really glad intellij has refactoring support for that...

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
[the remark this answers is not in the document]
zloery: Added resource ID to the error here. If the resource is not found in the given workspace, `resourceDao.getResource` will throw a `ResourceNotFoundException`

The name is probably more useful, but I think names for references aren't validated atm. Since they're user-provided, we probably shouldn't echo them back

Added TODO for PF-640. I think we probably want that check for workspace read access, but I want to be consistent about fixing private users / workspace access

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
zloery: credit to @davidangb, this is "heavily inspired" by `validateWorkspaceAndAction`

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
zloery: I kind of like having the non-standard mock setup right next to the test, even though this is duplicated in a few places

PR #288, opened 2021-04-09
PF-522: restore WSM multi-instance code 

jaycarlton:
Unrevert code from earlier.


src/main/resources/application.yml
```diff
@@ -74,6 +72,16 @@ workspace:
     username: ${env.db.ws.user}
 
 terra.common:
+  kubernetes:
+    in-kubernetes: true
+  stairway:
+    cluster-name-suffix: workspace-stairway
+    force-clean-start: false # ${env.db.init}
+    max-parallel-flights: 50
+    migrate-upgrade: false # true
```
yonghaoy, line 81: I feel this should be true. But I can not make the call for WSM. @ddietterich

src/main/resources/application.yml
```diff
@@ -74,6 +72,16 @@ workspace:
     username: ${env.db.ws.user}
 
 terra.common:
+  kubernetes:
+    in-kubernetes: true
+  stairway:
+    cluster-name-suffix: workspace-stairway
+    force-clean-start: false # ${env.db.init}
+    max-parallel-flights: 50
```
yonghaoy, line 80: is 50 enough? It a bigger number in `buffer`. @ddietterich

src/main/resources/application.yml
```diff
@@ -74,6 +72,16 @@ workspace:
     username: ${env.db.ws.user}
 
 terra.common:
+  kubernetes:
+    in-kubernetes: true
```
yonghaoy, line 76: I remember staring local WSM also read from this file? If yes, we need to fix for it.

jaycarlton: Hmm. There's no application-local.yml?

src/main/resources/application.yml
ddietterich: Yes, we should upgrade the schema by default.

src/main/resources/application.yml
ddietterich: I have no reason to think we need more than 50. We run with 3 pods, so that is a net of 150 flights running in parallel. We are likely to hit the connection limit on the underlying database before we run out of threads.

I have raised concerns with you before @yonghaoy about the very high number of threads you specify and the memory consequences. But so long as you have nodes able to support your memory use, it should work.

src/main/resources/application.yml
ddietterich: I would like to get rid of using the configuration that runs a local WSM server on the laptop to run "integration" tests. It will not work for new testing, including controlled buckets and K8s disruptive tests. So I am not motivated to do the work to support it. I think we should be deploying to our personal clusters and running integration tests against those.

However, we have this limitation that the only way to run integration tests via GHA is to run the server on the GHA node. So we have to make that work - maybe by making an application-local.yml and setting the profile to include local.

yonghaoy: I guess people does run local server that offten. RBS has this little script: https://github.com/DataBiosphere/terra-resource-buffer/blob/master/local-dev/run_local.sh
But that is not urgent, and barely used.

yonghaoy (approved):

src/main/java/bio/terra/workspace/common/utils/MdcHook.java
```diff
@@ -128,4 +128,11 @@ private String serializeMdc(@Nullable Map<String, String> mdcMap) {
     }
     return mdcContext;
   }
+
+  @Override
+  public HookAction stateTransition(FlightContext context) throws InterruptedException {
+    logger.info(
```
jaycarlton, line 134: The log level might benefit from being set per-hook or even per-flight class instead of per-statement. I could see this becoming noisy in high traffic flights.

src/test/resources/application-integration-test.yml
```diff
@@ -1,3 +1,4 @@
+
```
jaycarlton, line 1: Would it be better to merge `application-integration-test-gha` into this file and set it as the profile? Do I need to specify a value for `in-kubernetes` in this file?

ddietterich: Here is my thought about what the end state should be:
- we have a way to set in-kubernetes in the helm values file. It is set to true by default in workspacemanager helm, so that it is true for dev, alpha, staging, production, and our personal environments. This is missing right now.
- we explicitly set it to false in all application-*test yaml files
- we do not set profiles in the testrunner code. It should not be opinionated about whether it is being run as a local test or in a K8s cluster.
What do you think?

yonghaoy: 1. This is what WSM is doing: https://github.com/broadinstitute/terra-helm/blob/master/charts/workspacemanager/templates/deployment.yaml#L185
2. Everywhere can all be false if helm file seems to be the only place where we deploy WSM into kubernetes?

yonghaoy (approved):

src/test/resources/application-integration-test.yml
jaycarlton: Good ideas. I filed https://broadworkbench.atlassian.net/browse/PF-676 for those.

yonghaoy: > we have a way to set in-kubernetes in the helm values file. It is set to true by default in workspacemanager helm, so that it is true for dev, alpha, staging, production, and our personal environments. This is missing right now.
This is not missing, this is what WSM is doing now.  https://github.com/broadinstitute/terra-helm/blob/master/charts/workspacemanager/templates/deployment.yaml#L185

ddietterich: Oh, great! So then we could set it false in application.yml and be done.

ddietterich: @jaycarlton I will leave it to you to either set false there and remove the -integration-test-gha or not.

ddietterich (approved):

PR #291, opened 2021-04-12
[PF-589] validate non-existence of Controlled Resources (buckets and AI Notebook instances)

jaycarlton:
In the controlled resource metadata step, query to make sure there's no other resource with the same unique attributes (that we know of). Inside the bucket create  step, simply do nothing if the bucket exists (since that's the step that created it most likely). Since we share a namespace with the whole cloud, there are race conditions not accounted for, and those will manifest in less pleasant error states.


src/main/java/bio/terra/workspace/service/resource/controlled/flight/ValidateNoExistingGcsBucketStep.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    final String projectId = workspaceService.getRequiredGcpProject(resource.getWorkspaceId());
+    final StorageCow storageCow = crlService.createStorageCow(projectId);
+    final BucketCow existingBucket = storageCow.get(resource.getBucketName());
+    if (existingBucket != null) {
+      throw new DuplicateResourceException(
+          String.format("GCS Bucket with name %s already exists.", resource.getBucketName()));
```
wchamber, line 41: drive-by: Isn't there a race condition here where two create bucket flights can both get past this step and then try to create the same GcsBucket? I think we need to do the existence/creation as a transaction.
I was wondering if we should have the "cloud uri" column for controlled resources in the WM database and use that to ensure exclusive controlled resources.

jaycarlton: So doing it as a transaction would mean that we don't do it as two steps, right?

ddietterich: We already know the bucket attributes, so we don't need to add anything in the database. We could do this:
Step 1: in a serialized DAO transaction, verify that no other controlled GCS bucket exists with that bucket name and, if not, write the metadata into the database. That prevents parallel bucket creates within WSM.
Step 2: check that the GCS bucket does not exist in GCP. This leaves a race between us and other bucket creators, but I don't see how to close that.
Step 3: create the bucket and just use it if already exists

wchamber: That sounds good to me.
I think Step 1 is generic for all controlled resources. Do we want to have a uniform cloud URI column, or write some switch based on the controlled resource type and look at the attributes json?

ddietterich: Each controlled resource has a different comparison to do. The database column is jsonb, so we can construct SQL queries to do this, but the queries would be per-resource-type.

wchamber: Doing per-resource type queries sounds fine to me (what I was referring to "some switch based on the controlled resource type and look at the attributes json").
However, I'm not sure that I'm being clear in the alternative to that I'm suggesting: make a new text column "controlled_uri" in the resource table. Add a field to the `ControlledResource` class for the uri. As long as each controlled resource type can represent its underlying cloud resource uniqueness as a canonical string, we can put a uniqueness constraint on the "controlled_uri" text column and not write per-type SQL jsonb comparison switches. For buckets, the "controlled_uri" is the GCS bucket uri, e.g. `gs://my-bucket-id`. For BigQuery datasets, is the project id + dataset id. - e.g. 'my_project.my_dataset'. That does duplicate information in the attributes class.

I'm ok with either way, just want to consider both.

jaycarlton: Yeah, I like having the unique constraint live outside the JSON, so it becomes the database's job to check. That may be out of scope for this little ticket though.

jaycarlton:
Since `CreateGcsBucketStepTest` is disabled, I'm not sure I can write a good test for the database check 🤔 . I think I can test just that step.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
```diff
@@ -79,6 +79,13 @@ public String attributesToJson() {
         new ControlledAiNotebookInstanceAttributes(getInstanceId(), getLocation()));
   }
 
+  @Override
+  public boolean matchesUniqueAttributes(String attributesJson) {
+    final ControlledAiNotebookInstanceAttributes attributes =
+        DbSerDes.fromJson(attributesJson, ControlledAiNotebookInstanceAttributes.class);
+    return getInstanceId().equals(attributes.getInstanceId());
```
jaycarlton, line 86: Is this unique enough without the location too?

wchamber: No, it's unique by (projectId, location, instanceId).

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -324,6 +324,21 @@ public void createControlledResource(ControlledResource controlledResource)
           "No cloud context found in which to create a controlled resource");
     }
 
+    // scan entire attributes column for existing controlled resource
+    // TODO: use a dedicated URI column to make this more efficient PF-690
+    final String attributesSql =
+        String.format(
+            "SELECT attributes FROM resource "
+                + "WHERE resource_type = '%s' "
+                + "AND stewardship_type = 'CONTROLLED'",
```
ddietterich, line 333: In general, we have been using parameters to substitute values in queries.

Suggest populating the stewardship_type with StewardshipType.CONTROLLED.toSql() so that you are not relying on the specific string in the query.

ddietterich (approved):
Suggestion on the query construction
Otherwise, LGTM

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
jaycarlton: Thanks. 

Unfortunately, I don't see an easy way to get the projectId  from here and the projectID isn't in the attributes structure, so my scheme only generalizes to one controlled resource type.

jaycarlton: Workspace ID would be a pretty good proxy for project ID I would think, if that helps. I don't like adding stuff to the attributes structure just to make this work, especially since every GCP resource will have a project ID.

wchamber: I think workspace id as a proxy for the project id is fine with a comment about what's going on. But it's also not crazy to add the project id to the attributes - it is an attribute to identify the notebook instance that isn't going anywhere and its frequently useful. But it does duplicate information.
Every GCP resource doesn't need to have the project id, just most - buckets are globally namespaced.

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -324,6 +324,21 @@ public void createControlledResource(ControlledResource controlledResource)
...
+            controlledResource.getResourceType().toSql());
+    final List<String> attributesList =
+        jdbcTemplate.queryForList(attributesSql, Collections.emptyMap(), String.class);
+    final boolean exists =
```
wchamber, line 337: nit: seems like excessive use of `final` on local variables. (ditto above). I'm not sure we need extra characters to signal we're not changing these variables on a short function where we have no reason to think otherwise. go/java-practices/final#params_locals

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -324,6 +324,21 @@ public void createControlledResource(ControlledResource controlledResource)
           "No cloud context found in which to create a controlled resource");
     }
 
+    // scan entire attributes column for existing controlled resource
+    // TODO: use a dedicated URI column to make this more efficient PF-690
```
wchamber, line 328: Scanning/parsing all the controlled resource attributes into memory to do an equality check seems like a lot of work to do on every resource creation. Even if we don't want to make a special URI column, we can still write per-resource type SQL query to check the resource/attributes jsonb for equality.

For buckets that's something like
```
SELECT COUNT(1) FROM resource
WHERE attributes->>bucketName = :bucket_name
AND stewardship_type = :controlled
```

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
jaycarlton: If I leave it as just a check on instance ID and location, the risk is that we fail to allow a duplicate ID in another project when that should otherwise be allowed. It wouldn't be obvious to the user what happened. Which means maybe it's a lesser evil to just return false unconditionally for AI notebooks for now until we do the uri column or something similar.

src/main/java/bio/terra/workspace/db/ResourceDao.java
jaycarlton: I thought about that but didn't want that layer of the stack to know about what kind of SQL we're using or to have that step know anything about the different subclasses (like in a switch statement), but it's certainly more efficient. It is hard to defend a full table scan, though.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
wchamber: Why don't we do the right thing now? We can pass the workspace id or even the whole existing ControlledResource to this functions instead of just the attributes text. I'm not seeing the value in having an incorrect solution merged to dev.
If this always returns false, we're allowing controlled resource entries to refer to the same resource. Then we have to protect against granting inappropriate access on "private" notebook instances that are controlled by different users.

ddietterich: I think we should give up on making this generic. For most resources, we only care about collisions within the WSM project. But buckets are in a global namespace and need the full table scan.

We already have a switch statement to put the right steps into the create, so I don't think we are making things particularly worse by having a separate checking step.

If most of the steps turn out to be resource specific, we can rethink the single create flight.

src/test/java/bio/terra/workspace/db/ResourceDaoTest.java
```diff
@@ -59,5 +63,87 @@ public void createGetControlledAiNotebookInstance() {
...
+    final UUID workspaceId1 = createGcpWorkspace();
+    final ControlledResource resource1 =
+        ControlledResourceFixtures.makeDefaultAiNotebookInstance()
+            .workspaceId(workspaceId1)
+            .build();
+    resourceDao.createControlledResource(resource1);
+    assertEquals(
+        resource1, resourceDao.getResource(resource1.getWorkspaceId(), resource1.getResourceId()));
+
+    final ControlledResource resource2 =
```
wchamber, line 111: consider more meaningful names since we have a few ControlledResources here. maybeinstead of resource1-4  `initialResource`, `duplicatingResource`, `differentWorkspaceId`, and `differentLocation`,

wchamber (approved):
Update the PR name and description as well.
Thanks!

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
```diff
@@ -62,7 +65,15 @@ public StepResult doStep(FlightContext flightContext)
...
     StorageCow storageCow = crlService.createStorageCow(projectId);
-    storageCow.create(bucketInfo);
+
+    // Don't try to create it if it already exists. At this point the assumption is
+    // this is a redo and this step created it already.
+    BucketCow existingBucket = storageCow.get(resource.getBucketName());
+    if (existingBucket == null) {
+      storageCow.create(bucketInfo);
+    } else {
+      logger.info("Bucket {} already exists. Continuing.", resource.getBucketName());
```
ddietterich, line 75: Is it worth cross-checking that the bucket is in the workspace project?
I think the project id of the bucket is retrievable here and the project can be retrieved using the workspaceId.

ddietterich (approved):
LGTM - one belt-and-suspenders suggestion

wchamber (approved):

jaycarlton:
> Is it worth cross-checking that the bucket is in the workspace project?
I don't see how to get the project ID from the existing bucket object's `bucketInfo`.

PR #292, opened 2021-04-12
[PF-599] Make controlled bucket creation synchronous

zloery:
Currently, controlled GCS bucket creation is asynchronous. However, the call completes very quickly and this can be made synchronous. While this pattern works for GCS buckets, I don't know that all controlled resource creation can be synchronous. To support this, I've left the asynchronous creation method in the service and moved JobReport out of the common + required request fields of the API object.

Controlled GCS bucket deletion still needs to be asynchronous, as that call can potentially take much longer


zloery:
Client tests pass with a locally published client, I'll publish the new version once this is approved

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
```diff
@@ -258,23 +254,15 @@ private CreatedControlledGcpGcsBucket createBucketAttempt(ControlledGcpResourceA
...
-    logger.info(
-        "Attempt to creating bucket {} jobId {} workspace {}", bucketName, jobId, getWorkspaceId());
-    CreatedControlledGcpGcsBucket bucket = resourceApi.createBucket(body, getWorkspaceId());
-    while (ClientTestUtils.jobIsRunning(bucket.getJobReport())) {
-      TimeUnit.SECONDS.sleep(CREATE_BUCKET_POLL_SECONDS);
-      bucket = resourceApi.getCreateBucketResult(getWorkspaceId(), jobId);
-    }
-    logger.info("Create bucket status is {}", bucket.getJobReport().getStatus().toString());
-    return bucket;
+    logger.info("Attempt to creating bucket {} workspace {}", bucketName, getWorkspaceId());
```
jaycarlton, line 264: *create

jaycarlton (approved):

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -43,6 +44,29 @@ public ControlledResourceService(
     this.stageService = stageService;
   }
 
+  public ControlledResource syncCreateControlledResource(
```
ddietterich, line 47: Ah, good idea to leave the async code. We are likely to need it.

workspace-manager-clienttests/src/main/java/scripts/testscripts/CreateGetDeleteControlledGcsBucket.java
ddietterich: My typo, but thanks for fixing it :)

ddietterich (approved):
LGTM

zloery:
published `0.21.0-SNAPSHOT` from this branch

PR #295, opened 2021-04-14
PF-469 Give controlled bucket creation its own method on ControlledResourceService.

wchamber:
Creating controlled buckets requires its own set of parameters (`ApiGcpGcsBucketCreationParameters`). Other controlled resource types will have their own creation parameters. Rather than trying to make a single "create controlled resource" function on ControlledResourceService, make creating buckets its own method, paving the way for adding other resource overloads.

This matches the bucket specific method for deletion.

I'd like to make this tiny change to avoid further merge conflicts on AI notebook instances & bigQuery data set controlled resources.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
+++ b/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
@@ -102,14 +102,7 @@ public class ControlledResourceService {
         commonCreationJobBuilder(
                 resource, privateResourceIamRoles, UUID.randomUUID().toString(), null, userRequest)
             .addParameter(ControlledResourceKeys.CREATION_PARAMETERS, creationParameters);
-    String jobId = jobBuilder.submit();
-    jobService.waitForJob(jobId);
-    JobResultOrException<ControlledGcsBucketResource> jobResult =
-        jobService.retrieveJobResult(jobId, ControlledGcsBucketResource.class, userRequest);
-    if (jobResult.getException() != null) {
-      throw jobResult.getException();
-    }
-    return jobResult.getResult();
+    return jobBuilder.submitAndWait(ControlledGcsBucketResource.class);
   }
 
   /** Create a JobBuilder for creating controlled resources with the common parameters populated. */
--- a/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
+++ b/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
@@ -10,7 +10,6 @@ import bio.terra.workspace.service.iam.model.SamConstants.SamControlledResourceA
 import bio.terra.workspace.service.job.JobBuilder;
 import bio.terra.workspace.service.job.JobMapKeys;
 import bio.terra.workspace.service.job.JobService;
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -92,34 +92,31 @@ public ControlledResource validateControlledResourceAndAction(
...
-            creationParameters,
-            privateResourceIamRoles,
-            syncJobControl,
-            null,
-            userRequest);
+    JobBuilder jobBuilder =
+        commonCreationJobBuilder(
+                resource, privateResourceIamRoles, UUID.randomUUID().toString(), null, userRequest)
+            .addParameter(ControlledResourceKeys.CREATION_PARAMETERS, creationParameters);
+    String jobId = jobBuilder.submit();
```
zloery, line 105: this can just be `jobBuilder.submitAndWait()` without all the code below

zloery (approved):
This approach works, but I think we'll end up with a lot of duplicated code.

One alternative we've talked about before is passing `creationParameters` as an Object instead, since the common create method will just add it as a job parameter with the same key every time. (Object is kind of unfortunate, but there's no common base class for creation parameters as they're generated types). I'm not actually sure why we didn't write that in the first place, I need to go dig through PRs for that

wchamber:
> This approach works, but I think we'll end up with a lot of duplicated code.
> 
> One alternative we've talked about before is passing `creationParameters` as an Object instead, since the common create method will just add it as a job parameter with the same key every time. (Object is kind of unfortunate, but there's no common base class for creation parameters as they're generated types). I'm not actually sure why we didn't write that in the first place, I need to go dig through PRs for that

I'd start by trying to avoid Object casts or class hierarchies. Storing different classes in the same flight map key seems a little error prone. I don't think there's too much duplication (we can move job waiting and catching into a helper method for more sync methods).

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
wchamber: thanks, done.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -92,34 +92,24 @@ public ControlledResource validateControlledResourceAndAction(
...
   }
 
-  public String createControlledResource(
+  /** Create a JobBuilder for creating controlled resources with the common parameters populated. */
+  private JobBuilder commonCreationJobBuilder(
       ControlledResource resource,
-      ApiGcpGcsBucketCreationParameters creationParameters,
       List<ControlledResourceIamRole> privateResourceIamRoles,
-      ApiJobControl jobControl,
+      String jobId,
```
ddietterich, line 112: This is fine for now, but we will want the full JobControl object for the JobService when we implement pub/sub completions.

ddietterich (approved):
LGTM - One suggestion

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
wchamber: done.

PR #296, opened 2021-04-16
[PF-666] Use common lib exceptions, other than Sam exceptions.

tnadler:
This change is a mechanical rename to use the common lib exceptions.


ddietterich (approved):
LGTM

src/main/java/bio/terra/workspace/app/controller/GlobalExceptionHandler.java
```diff
@@ -1,6 +1,6 @@
 package bio.terra.workspace.app.controller;
 
-import bio.terra.workspace.common.exception.ErrorReportException;
+import bio.terra.common.exception.ErrorReportException;
```
wchamber, line 3: Can we also change to subclassing the TCL's AbstractGlobalExceptionHandler? Doesn't have to be this PR, but a quick follow up if not would be good.
https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/exception/AbstractGlobalExceptionHandler.java

wchamber (approved):

src/main/java/bio/terra/workspace/app/controller/GlobalExceptionHandler.java
tnadler: Yes, will follow up.

PR #301, opened 2021-04-20
PF-669 Minimal fix for status endpoint

ddietterich:
This is a small PR to try to address the likely issues with the stuck /status endpoint. I made two changes:
1. Put a timeout on the database connection check
2. Allocate a single thread scheduled executor just for the status check.

In thinking about the problem, I realized I want to understand why we are bothering doing all of this. I wrote [Status Checking Reconsidered](https://docs.google.com/document/d/1UqEZ0UsnSvv0E_bWfMuSYAZiCQFVIgNUDmWIGHNXxG0/edit#heading=h.jl83h28nja15l)

When MC Terra Arch Council has figured out the system level view, we can come back to elaborating or eliminating this endpoint. In the meantime, I am hoping this will get us by!

### DIFF
--- a/src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
+++ b/src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
@@ -42,7 +42,7 @@ public class BaseStatusService {
 
   @PostConstruct
   public void startStatusChecking() {
-    scheduler.scheduleAtFixedRate(this::checkSubsystems, 1, 1, TimeUnit.MINUTES);
+    scheduler.scheduleAtFixedRate(this::checkSubsystems, 5, 60, TimeUnit.SECONDS);
   }
 
   protected void registerSubsystem(String name, StatusSubsystem subsystem) {
--- a/src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
+++ b/src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
@@ -4,10 +4,13 @@ import bio.terra.workspace.generated.model.ApiSystemStatus;
 import bio.terra.workspace.generated.model.ApiSystemStatusSystems;
 import java.util.Date;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import javax.annotation.PostConstruct;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.scheduling.annotation.Scheduled;
 
 /*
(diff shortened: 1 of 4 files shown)


zloery (approved):
LGTM, I think the DB connection is the most likely cause of stuck statuses. If the issue continues to come up, we should look at timeouts on the `/status` calls to other services as well.

Since we're moving off `@Scheduled`, can you also remove:

- `@EnableScheduled` annotation from [main](https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/app/Main.java#L30)
- `workspace.status-check.cron` config from `application.yml` and `application-test.yml`

src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
```diff
@@ -33,13 +37,18 @@ public BaseStatusService(long staleThresholdMillis) {
     currentStatus = new ApiSystemStatus().ok(false);
     lastUpdatedTimestampMillis = 0;
     this.staleThresholdMillis = staleThresholdMillis;
+    this.scheduler = Executors.newScheduledThreadPool(1);
+  }
+
+  @PostConstruct
+  public void startStatusChecking() {
+    scheduler.scheduleAtFixedRate(this::checkSubsystems, 1, 1, TimeUnit.MINUTES);
```
davidangb, line 45: this has a full 1-minute delay before starting checks, is that what we want? You could do something like
```
scheduler.scheduleAtFixedRate(this::checkSubsystems, 5, 60, TimeUnit.SECONDS)
```
to make the first status check happen sooner.

src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
```diff
@@ -33,13 +37,18 @@ public BaseStatusService(long staleThresholdMillis) {
...
   }
 
   protected void registerSubsystem(String name, StatusSubsystem subsystem) {
     subsystems.put(name, subsystem);
   }
 
-  @Scheduled(cron = "${workspace.status-check.cron}")
```
davidangb, line 42: should we remove the corresponding line from `application.yml`?

davidangb (commented):
I'm curious to see how this fares. I think it's an improvement over what we had before, but if it doesn't solve the problem we may need to add more logging to the individual checks in `WorkspaceManagerStatusService` or elsewhere.

Two small comments inline, I will approve once you've responded to those!

src/main/java/bio/terra/workspace/common/utils/BaseStatusService.java
ddietterich: Good point. That would actually delay readiness, the way things are setup.

ddietterich: I considered plumbing these in as parameters, but didn't want to bother if we end up ditching this style of checking. I made a PR to track the work-not-done: PF-700

davidangb (approved):

PR #302, opened 2021-04-21
PF-469 Add initial controlled notebook instance resource creation flight support

wchamber:
Add the initial steps and support for creating a controlled AI Notebook Instance with the Controlled Resource Service.

- Adds CRL google-notebooks dependency.
- Updates Stairway version to get newest FlightDebugInfo fields for testing.
- Adds connected tests for creating controlled AI notebook instance with some Step idempotency checking. This is unfortunately slow as creating an AI notebook instance takes ~3-7 minutes.
- Adds an initial OpenApi model definition for the AI notebook instance creation parameters.

Points of interest:
This follows the Bucket creation in storing the raw API creation parameters in the FlightMap. This locks us into changing the API model in ways supported by Stairway serialization. An alternative would be to make a Java POJO for JSON serialization. Or we can wait for serialization to shake out.

Right now the notebook instance network is hard coded. We could either expose the network as a part of the WM API, or retrieve and use any network/subnetwork in the appropriate GCE region. Probably a follow up PR.

TODOs for follow up PRs
- Setting up permissions for AI Notebook Instances and the service account on it
- Not hardcoding the network/subnetwork to use on the project
- More instance creation parameters.
- Add the actual OpenAPI creation endpoint.


wchamber:
@rtitle FYI: First chunk of AI Notebook Instance as a controlled resource in WM.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateControlledResourceFlight.java
```diff
@@ -35,6 +49,7 @@ public CreateControlledResourceFlight(FlightMap inputParameters, Object beanBag)
     addStep(new StoreMetadataStep(flightBeanBag.getResourceDao()));
 
     // create the Sam resource associated with the resource
+    // TODO consider adding retry rules for Sam steps.
```
zloery, line 52: may not be needed with Tova's work on Sam exceptions

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateControlledResourceFlight.java
```diff
@@ -17,6 +22,15 @@
...
 public class CreateControlledResourceFlight extends Flight {
+  /**
+   * Retry rule for Notebook GCP steps. If GCP is down, we don't know when it will be back, so don't
+   * wait forever.
+   */
+  private static final RetryRule NOTEBOOK_GCP_RETRY_RULE =
+      new RetryRuleExponentialBackoff(
+          /* initialIntervalSeconds= */ 1,
+          /* maxIntervalSeconds= */ 8,
+          /* maxOperationTimeSeconds= */ 30);
```
zloery, line 33: is this a long enough maximum timeout? from the PR description I'd expect notebook description to take O(minutes)

zloery (commented):
generally LGTM, some comments

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
```diff
@@ -0,0 +1,200 @@
...
+    // Assume the zone is related to the location like 'us-west1' is to 'us-west1-b'.
+    Preconditions.checkArgument(location.length() > 2, "Invalid location '%s'", location);
+    String zone = location.substring(0, location.length() - 2);
+    // Like 'network', 'subnetwork' is the name of the subnetwork created by the Buffer Service in
+    // each zone.
+    instance.setSubnet("projects/" + projectId + "/regions/" + zone + "/subnetworks/subnetwork");
+  }
+
+  private InstanceName createInstanceName(String projectId) {
+    return InstanceName.builder()
```
ddietterich, line 157: What if we have already created an instance of this name in this workspace?
Do we need another step that looks in the workspace metadata to make sure that is not the case?

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
```diff
@@ -0,0 +1,200 @@
...
+    AIPlatformNotebooksCow notebooks = crlService.getAIPlatformNotebooksCow();
+    try {
+      OperationCow<Operation> creationOperation;
+      try {
+        creationOperation =
+            notebooks
+                .operations()
+                .operationCow(notebooks.instances().create(instanceName, instance).execute());
+      } catch (GoogleJsonResponseException e) {
+        // If the instance already exists, this step must have already run successfully. Otherwise
```
ddietterich, line 70: How can you tell the difference between the instance already existing from this flight and the instance existing from a previous flight creating it?

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
```diff
@@ -0,0 +1,200 @@
...
+        .setMachineType(creationParameters.getMachineType());
+    ApiGcpAiNotebookInstanceVmImage vmImageParameters = creationParameters.getVmImage();
+    if (vmImageParameters != null) {
+      instance.setVmImage(
+          new VmImage()
+              .setProject(vmImageParameters.getProjectId())
+              .setImageFamily(vmImageParameters.getImageFamily())
+              .setImageName(vmImageParameters.getImageName()));
+    }
+    ApiGcpAiNotebookInstanceContainerImage containerImageParameters =
```
ddietterich, line 134: From the comments in the open_api, I got the idea the vm image and the container image were mutually exclusive. Is that true? This could would allow both or neither to be set.

ddietterich (commented):
Looks like there might be a duplicate instance issue.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1640,6 +1640,77 @@ components:
...
+    # TODO(PF-469): Add the creation endpoint.
+    GcpAiNotebookInstanceCreationParameters:
+      description: >-
+        AI Platform Notebook instance specific properties to be set on creation. These are a subset of the
+        values accepted by the GCP AI Platforms API. See https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1/projects.locations.instances/create
+      type: object
+      properties:
+        instanceId:
+          description: >-
+            Required. A instanceId unique to this project and location. Must be 1-63 characters, using lower
```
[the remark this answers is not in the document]
wchamber: 63 AFAICT. `(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)` 1 + max 61 + 1.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1640,6 +1640,77 @@ components:
...
+          description: Required. The canonical GCP id for this location. For example, \'us-east1\'.
+          type: string
+        machineType:
+          description: Required. The Compute Engine machine type of this instance, see
+            https://cloud.google.com/compute/docs/machine-types
+          type: string
+        postStartupScript:
+          description: >-
+            Path to a Bash script that automatically runs after a notebook instance fully boots up.
+            The path must be a URL or Cloud Storage path (gs://path-to-file/file-name).
```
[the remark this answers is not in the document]
wchamber: The service account on the VM. Which in this case has no special privileges, so we'll have to consider how to handle this when we have a use case.
Added a comment.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -108,6 +109,21 @@ public ControlledGcsBucketResource syncCreateBucket(
...
+      ApiJobControl jobControl,
+      String resultPath,
+      AuthenticatedUserRequest userRequest) {
+    JobBuilder jobBuilder =
+        commonCreationJobBuilder(
+            resource, privateResourceIamRoles, jobControl, resultPath, userRequest);
+    jobBuilder.addParameter(ControlledResourceKeys.CREATE_NOTEBOOK_PARAMETERS, creationParameters);
+    return jobBuilder.submit();
+  }
+
```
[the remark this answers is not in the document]
wchamber: I wasn't even thinking it was necessary for the Service. The [WorkspaceApiController](https://github.com/DataBiosphere/terra-workspace-manager/blob/a1fd416f69c3c032abb33060c56753fbb88db283/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java#L419) and the [ControlledGcpResourceApiController](https://github.com/DataBiosphere/terra-workspace-manager/blob/a1fd416f69c3c032abb33060c56753fbb88db283/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java#L135) use the job service directly to get their job results. If it's preferable to wrap the job service lookup in `ControlledResourceService` I'd like to do that in a follow up PR once I get to actually having endpoints for the controller.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
wchamber: I was going to rely on Jay's PR #291 to do a transaction on the ResourceDao to ensure uniqueness of the underlying cloud resource. As-is, it's broken as it will assume success.

I'll add a TODO to add a test to verify that multiple creations for the same instance fail once Jay's PR is merged.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
```diff
@@ -0,0 +1,200 @@
...
+      OperationCow<Operation> creationOperation;
+      try {
+        creationOperation =
+            notebooks
+                .operations()
+                .operationCow(notebooks.instances().create(instanceName, instance).execute());
+      } catch (GoogleJsonResponseException e) {
+        // If the instance already exists, this step must have already run successfully. Otherwise
+        // retry.
+        if (HttpStatus.CONFLICT.value() == e.getStatusCode()) {
```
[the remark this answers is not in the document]
wchamber: AFAICT it's guaranteed to be on the initial create call. I can't find any documentation stating either way.
But if it can show up on the poll, we have an issue.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
```diff
@@ -0,0 +1,200 @@
...
+          return StepResult.getStepResultSuccess();
+        }
+        return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+      }
+
+      creationOperation =
+          OperationUtils.pollUntilComplete(
+              creationOperation, Duration.ofSeconds(20), Duration.ofMinutes(12));
+      if (creationOperation.getOperation().getError() != null) {
+        throw new RetryException(
```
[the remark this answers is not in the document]
wchamber: Yes. I don't know if there's a preferred style here. Undo's can't throw a RetryException, so maybe it's better to always return a StepResult?
This doesn't have an explicit exception, so without throwing a RetryException I wasn't sure how to encode any message for Stairway debugging what happened. A `StepResult` doesn't have its own error message field, other than an `Exception`. Rather than invent an exception, I went with throwing a new RetryException with my own message.
Happy to do something else here.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
```diff
@@ -0,0 +1,200 @@
...
+                instanceName.formatName(),
+                creationOperation.getOperation().getError()));
+      }
+    } catch (IOException e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  private static Instance createInstanceModel(FlightContext flightContext, String projectId) {
```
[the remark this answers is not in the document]
wchamber: pure (tiny) style preference. I'm in the camp that `static` is an easy signal to the reader that this is a pure method that doesn't mess with or rely on instance state. [go/java-practices/static#already_static](go/java-practices/static#already_static)

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
wchamber: Yes. The GCP API will throw exceptions if both or neither are set. I was planning on doing creation parameter validation in the controller in a follow up PR.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
wchamber: per other comment, I was relying on Jay's PR #291 to do the resource uniqueness check at ResourceDao insertion time (the earlier `StoreMetadataStep` in this flight). I agree that without that this is a problem. I was not planning on blocking this PR on that, but if that doesn't get merged we'll have to come up with another solution here.
This is a generic problem for controlled resources and their underlying cloud resource.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
```diff
@@ -0,0 +1,198 @@
...
+              .setTag(containerImageParameters.getTag()));
+    }
+  }
+
+  private static void setNetworks(Instance instance, String projectId, String location) {
+    // 'network' is the name of the VPC network instance created by the Buffer Service.
+    // Instead of hard coding this, we could try to look up the name of the network on the project.
+    instance.setNetwork("projects/" + projectId + "/global/networks/network");
+    // Assume the zone is related to the location like 'us-west1' is to 'us-west1-b'.
+    String zone = location.substring(0, location.length() - 2);
```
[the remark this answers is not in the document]
zloery: ah, I was comparing to BQ Datasets where you can specify location at the region level - e.g. `us-central1` is a valid location to create a Dataset. If GCP requires you specify a zone (e.g. `us-central1-a`) for a notebook instance, this is fine.

nit: can you rename `zone` to `region` here? after digging into this more it looks like `zone` is actually the more specific area like `us-central1-a` while `region` is the broader area `us-central1`: https://cloud.google.com/compute/docs/regions-zones. I don't know how consistent that is throughout GCP documentation, though

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateControlledResourceFlight.java
```diff
@@ -27,10 +27,7 @@
    * wait forever.
    */
   private static final RetryRule NOTEBOOK_GCP_RETRY_RULE =
-      new RetryRuleExponentialBackoff(
-          /* initialIntervalSeconds= */ 1,
-          /* maxIntervalSeconds= */ 8,
-          /* maxOperationTimeSeconds= */ 30);
+      new RetryRuleFixedInterval(/* intervalSeconds= */ 10, /* maxCount=  */ 10);
```
zloery, line 30: RetryRule objects are stateful, I think you need a different object for each step here. Otherwise, you're sharing 10 total retries across both steps.

(We've definitely mixed this up elsewhere, PF-557 for other ones)

zloery (approved):
LGTM, acknowledging the concurrency issue that (I think) will be fixed by a DB check in Jay's PR.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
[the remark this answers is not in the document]
ddietterich: No need to change it. It just struck me odd looking at it. I don't think I've seen a case before where requesting Stairway retry is not in response to an exception. If this turns up more frequently, it might be worth adding a StepResult that takes a String instead of Exception

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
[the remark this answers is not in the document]
ddietterich: OK, I'll keep that in mind in my code.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
ddietterich: OK, that sounds fine to me

ddietterich (approved):
LGTM

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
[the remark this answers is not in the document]
wchamber: thanks, s/zone/region done.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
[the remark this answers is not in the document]
wchamber: I don't think so. I think that's code that we're very unlikely to need. If our create AI notebook flights are failing, we should debugging why that's happening, not trying to predict what the issue is.
It also seems like it's not the style for the GCP operations API to use the operations.get endpoint HTTP status code for the result of the logical operation. The `Operation` response has an `error` field for describing errors about the operation. 
Reminiscent of our JobReport HTTP status code conversations.
https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1/projects.locations.operations
Here, we're throwing the RetryException with the error retrieved from the operation response body.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateControlledResourceFlight.java
wchamber: I did not know that. For my 2c, PF-557 SG. I don't know when I'd want to share a retry rule, so changing the `addStep` api to take a factory for retry rules to prevent that from happening seems less error prone.

zloery: whoops, Dan's right here. I didn't see that stairway re-initialized RetryRules for each step

PR #304, opened 2021-04-21
[PF-388] This PR is the first in a chain of PRs to add retrying around Sam API calls. 

tnadler:
This PR cleans up the way that Sam exceptions are created and handled:

* Use the SamExceptionFactory class for creating Sam-specific exceptions. These exceptions derive from ErrorReportException and are based on specific HTTP Status codes.
* Move all control flow that is based on HTTP Status into the SamService class itself. In particular, in both Create and Delete functionality, there are certain Sam error statuses that are not truly errors: In the context of deletion, a NOT_FOUND error means that the resource is already deleted and in the context of creation, a CONFLICT error means the resource has already been created. By moving these checks into SamService, we no longer need to catch the SamApiException type in flights themselves. Any error that is raised by SamService is a true unexpected error and hsould be bubbled up.
* Retrying will be happening within the SamService itself. That means we do not need to use the Flight's retry capabilities.


src/main/java/bio/terra/workspace/service/workspace/flight/DeleteWorkspaceAuthzStep.java
```diff
@@ -29,21 +27,10 @@ public StepResult doStep(FlightContext flightContext) throws RetryException {
...
-            String.format(
-                "Signaling retry for unexpected Sam error: code (%d) error: %s",
-                e.getApiExceptionStatus(), e.toString()),
-            e);
-        return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
-      }
-    }
-    return StepResult.getStepResultSuccess();
+    // Sam Service will take care of retrying any retry-able errors. Failures will be bubbled up
+    // as derivatives of ErrorReportException.
```
wchamber, line 31: nit: funny this step but only this step gets a comment. Maybe just remove?

src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
```diff
@@ -309,40 +308,6 @@ void workspaceReaderIsNotPrivateResourceReader() {
...
-    UUID workspaceId = createWorkspaceDefaultUser();
-
-    ControlledResource bucketResource = defaultBucket(workspaceId).build();
-    samService.createControlledResource(bucketResource, null, defaultUserRequest());
-
-    SamApiException exception =
-        assertThrows(
-            SamApiException.class,
-            () -> samService.createControlledResource(bucketResource, null, defaultUserRequest()));
-    assertEquals(HttpStatus.CONFLICT.value(), exception.getApiExceptionStatus());
```
wchamber, line 325: should we test that nothing happens on the second create/delete in these tests instead of removing the tests? I think that's also behavior we want and should therefore have a test for. ditto on delete test.

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -269,18 +272,17 @@ void testUpdateWorkspace() {
...
     String errorMsg = "fake SAM error message";
-
-    doThrow(new SamApiException(errorMsg))
-        .when(mockSamService)
-        .createWorkspaceWithDefaults(any(), any());
-
-    SamApiException exception =
+    String apiErrorMsg = "test";
+    ErrorReportException testex =
+        SamExceptionFactory.create(errorMsg, new ApiException(apiErrorMsg));
```
wchamber, line 277: nit: Should we explicitly `SamInternalServerErrorException exception = new SamInternalServerErrorException(apiErrorMessage)` instead of using the factory and relying on what class it returns?

wchamber (approved):
lgtm

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
ddietterich: Or assert that the factory returns what you expect?

ddietterich (approved):
LGTM

src/main/java/bio/terra/workspace/service/workspace/flight/DeleteWorkspaceAuthzStep.java
tnadler: Done

src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
tnadler: Debated that. Putting them back

src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
tnadler: Going with the former as the factory has it's own tests.

PR #305, opened 2021-04-23
PF-710 Restore exception serializer

ddietterich:
Companion PR to common lib [PF-671](https://github.com/DataBiosphere/terra-common-lib/pull/51)

This restores setting up the WSM exception serializer.

NOTE: This won't build here until I merge PF-671 and reset the version number in the build.gradle.


src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -62,9 +63,9 @@
   private final MdcHook mdcHook;
   private final StairwayComponent stairwayComponent;
   private final FlightBeanBag flightBeanBag;
-  private FlightDebugInfo flightDebugInfo;
-
   private final Logger logger = LoggerFactory.getLogger(JobService.class);
```
zloery, line 66: tiny nit: put `logger` back in its own block, since it's initialized differently than the other vars here?

zloery (approved):

wchamber (approved):

PR #306, opened 2021-04-23
PF-469 Fix gcp notebook retry rule to not be re-used across flights.

wchamber:
I read the comments and then forgot to actually change the RetryRule to not be a static instance...


ddietterich (approved):
lgtm

zloery (approved):

PR #308, opened 2021-04-26
[PF-388] Use SamRetry in WSM.

tnadler:
SamRetry intentionally throws an InterruptedException. This exception is needed in the context of Flights, but is not needed outside of Flights. This PR uses SamRetry in all contexts, but for functions that are used outside of Flights it catches the InterruptedException and rethrows as an unchecked exception type - in those contexts there is no known reason for InterruptedExceptions and so they are truly unexpected.


ddietterich (approved):
LGTM

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -47,6 +49,14 @@
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Component;
 
+/**
+ * SamService encapsulates logic for interacting with Sam. HTTP Statuses returned by Sam are
+ * interpreted by the functions in this class.
+ *
+ * <p>This class is used both by Flights and outside of Flights. Within flights,
+ * InterruptedExceptions need to be raised and handled by Stairway. Outside of flights, they are
+ * unexpected and will be converted into SamUnavailableExceptions which are unchecked.
```
wchamber, line 58: I don't think there are `SamUnavailableException`s. Using `{@link SamInterruptedException}` will help detect when class names change.

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -216,11 +239,25 @@ public void deleteWorkspace(String authToken, UUID id) {
   }
 
   @Traced
-  public boolean isAuthorized(
+  public boolean isAuthorizedWrapped(
```
wchamber, line 242: nit: can we develop some more precise terminology? `Wrapped` isn't very informative.
Alternatively, consider a helper function on SamService to catch and rethrow interrupted exceptions instead of doubling the API space. Little more verbose at call sites, but more explicit. Or just suffer the try-catch everywhere.
Or just use the static as a private helper in SamService.
```
public static <T> rethrowIfSamInterrupted(InterruptedSupplier<T> function, String operation) {
  try {
    return function.apply();
  } catch (InterruptedException e) {
    throw SamExceptionFactory.create("Interrupted during Sam operation " + operation, e);
  }
}

// at call site instead of
samService.isAuthorizedWrapped(accessToken, iamResourceType, resourceId, action);
// wrap
rethrowIfSamInterrupted(
  () -> samService.isAuthorized(accessToken, iamResourceType, resourceId, action), "isAuthorized");
```
I'm not convinced this is a good idea, but the double API method seems easy to use in the wrong place.

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -278,12 +315,20 @@ public void grantWorkspaceRole(
...
+                  SamConstants.SAM_WORKSPACE_RESOURCE,
+                  workspaceId.toString(),
+                  role.toSamRole(),
+                  email));
       logger.info(
           "Granted role {} to user {} in workspace {}", role.toSamRole(), email, workspaceId);
     } catch (ApiException apiException) {
       throw SamExceptionFactory.create("Error granting workspace role in Sam", apiException);
+    } catch (InterruptedException interruptedException) {
+      throw SamExceptionFactory.create(
```
wchamber, line 330: But these other methods aren't `Wrapped`? Seems error prone that we may start to use this in a step. I think if we really want to avoid rethrowing InterruptedExceptions in Steps, we need to make it default to throw the InterruptedException. Or if it doesn't matter when we don't throw the InterruptedException, then we don't bother.

wchamber (approved):

src/main/java/bio/terra/workspace/service/iam/SamService.java
tnadler: Done

src/main/java/bio/terra/workspace/service/iam/SamService.java
tnadler: I like this approach, adopted.

tnadler:
@wchamber do you want to take another look?

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -47,6 +49,14 @@
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Component;
 
+/**
+ * SamService encapsulates logic for interacting with Sam. HTTP Statuses returned by Sam are
+ * interpreted by the functions in this class.
+ *
+ * <p>This class is used both by Flights and outside of Flights. Within flights,
+ * InterruptedExceptions need to be raised and handled by Stairway. Outside of flights, they are
+ * unexpected and will be converted into {@link SamInterruptedException} which are unchecked.
```
wchamber, line 58: nit: update the comment on what the right way to do this is?

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -92,11 +102,35 @@ public String getWsmServiceAccountToken() throws IOException {
     return creds.getAccessToken().getTokenValue();
   }
 
+  @FunctionalInterface
+  public interface InterruptedSupplier<R> {
+    R apply() throws InterruptedException;
+  }
+
+  public interface VoidInterruptedSupplier {
```
wchamber, line 110: worth some javadoc tying these to the rethrowIfSamInterrupted

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -92,11 +102,35 @@ public String getWsmServiceAccountToken() throws IOException {
...
+
+  public static <T> T rethrowIfSamInterrupted(InterruptedSupplier<T> function, String operation) {
+    try {
+      return function.apply();
+    } catch (InterruptedException e) {
+      throw SamExceptionFactory.create("Interrupted during Sam operation " + operation, e);
+    }
+  }
+
+  public static void rethrowIfSamInterrupted(VoidInterruptedSupplier function, String operation) {
```
wchamber, line 122: worth a little comment on how and when to use them. Ok to refer to comments in other places

wchamber (approved):
LGTM, but I suggested the approach, so maybe @ddietterich  take a look too.
couple nits asking for more comments :)

src/main/java/bio/terra/workspace/service/iam/SamService.java
tnadler: Done.

src/main/java/bio/terra/workspace/service/iam/SamService.java
tnadler: Done.

src/main/java/bio/terra/workspace/service/iam/SamService.java
tnadler: Done.

tnadler:
Good point. LMK @ddietterich

ddietterich:
It looks fine to me. I like the rethrow thingy.

PR #310, opened 2021-04-27
PF-469 Add a test that a duplicate controlled AI notebook instance cannot be created.

wchamber:


src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -52,6 +45,12 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import org.apache.commons.lang3.StringUtils;
```
ddietterich, line 48: I wonder if we can get the IntelliJ plugin and the spotless to agree on where the imports should go :(

ddietterich (approved):
LGTM

zloery (approved):

PR #312, opened 2021-04-27
[PF-419] Add endpoint for deleting controlled BQ dataset

zloery:
This change adds an endpoint to delete a controlled BigQuery dataset. As part of this change, I've refactored the existing delete methods/flight a bit to less specific to GCS buckets, including moving steps out into their own files. However, there's no strong requirement that we move the common steps to their own files, so we could put them back if needed.


src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteSamResourceStep.java
```diff
@@ -0,0 +1,52 @@
...
+    this.resourceId = resourceId;
+    this.userRequest = userRequest;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    WsmResource wsmResource = resourceDao.getResource(workspaceId, resourceId);
+    ControlledResource resource = wsmResource.castToControlledResource();
+    samService.deleteControlledResource(resource, userRequest);
```
wchamber, line 43: I think it'd be worth testing that this can be retried. I think if we delete the Sam resource, but fail the flight, no one will have permission to delete the controlled resource anymore.

src/main/resources/api/service_openapi.yaml
```diff
@@ -862,6 +862,24 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/workspaces/v1/{workspaceId}/resources/controlled/gcp/bqdatasets/{resourceId}:
```
wchamber, line 865: nit: we should probably be consistent on if we're '-' separating words in our path. `bqdatasets` vs `ai-notebook-instances`

src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -351,6 +352,60 @@ public void createBqDatasetUndo() throws Exception {
...
+            creationParameters,
+            Collections.emptyList(),
+            userAccessUtils.defaultUserAuthRequest());
+    assertEquals(resource, createdDataset);
+
+    // Test idempotency of delete by retrying BQ specific steps once.
+    Map<String, StepStatus> retrySteps = new HashMap<>();
+    retrySteps.put(DeleteBigQueryDatasetStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_RETRY);
+    jobService.setFlightDebugInfoForTest(
+        FlightDebugInfo.newBuilder().doStepFailures(retrySteps).build());
```
wchamber, line 386: I think it's always good to have the ERROR undo case tested as well with `lastStepFailure`. Otherwise our tests never run the undo path.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -182,30 +182,56 @@ public ControlledResource getControlledResource(
     return wsmResource.castToControlledResource();
   }
 
-  public String deleteControlledGcsBucket(
+  /** Synchronously delete a controlled resource. */
+  public void deleteControlledResourceSync(
+      UUID workspaceId, UUID resourceId, AuthenticatedUserRequest userRequest) {
+    stageService.assertMcWorkspace(workspaceId, "deleteControlledResourceSync");
+    validateControlledResourceAndAction(
+        userRequest, workspaceId, resourceId, SamControlledResourceActions.DELETE_ACTION);
```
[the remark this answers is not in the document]
zloery: sgtm, also renamed the `deleteControlledresourceJobBuilder` for consistency

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteGcsBucketStep.java
```diff
@@ -0,0 +1,117 @@
...
+                      new BucketInfo.LifecycleRule(
+                          BucketInfo.LifecycleRule.LifecycleAction.newDeleteAction(),
+                          BucketInfo.LifecycleRule.LifecycleCondition.newBuilder()
+                              .setAge(0)
+                              .build())))
+              .build();
+      bucket = bucketCow.update();
+      bucketExists = tryBucketDelete(bucket);
+      if (bucketExists) {
+        TimeUnit.HOURS.sleep(1);
```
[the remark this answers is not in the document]
zloery: I think so, yes. in practice the bucket deletions don't usually require this wait so I'm not as concerned, but worst case yes that's an issue.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteSamResourceStep.java
```diff
@@ -0,0 +1,52 @@
...
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    // No undo for delete. There is no way to put it back.
+    return StepResult.getStepResultSuccess();
```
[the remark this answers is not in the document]
zloery: We don't have a best practice for this, and we aren't consistent across deletes for buckets vs. contexts vs. workspaces. This is a risk for a dismal failure (which would require manual intervention), so the best option here may really be logging an error and failing the flight. Updated other in this flight to use this pattern as well.
We have general error monitoring, but I don't believe we have alerting for this, or fatal flight failures in general.

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteSamResourceStep.java
zloery: added comment, retry rule, and retry debugInfo inside the test

src/main/resources/api/service_openapi.yaml
zloery: I think we've generally used no separator. elsewhere, we have `cloudcontexts` and `bigquerydatasets` (which is a separate issue, I need to make controlled BQ datasets consistent with referenced BQ datasets).
I see `-` separators in a couple async result endpoints though, PR-726 to fix since that will (potentially) be a breaking API change.

src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
zloery: I'd generally agree, but in this case there is no undo case, only dismal failure. I don't think it's worth testing that

src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteGcsBucketStep.java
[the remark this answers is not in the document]
ddietterich: Yup

ddietterich (approved):
LGTM

src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
wchamber: I'll push back a little on that. Whether the undos should succeed and the flight end in error or whether it should be a dismal failure is behavior we want to rely on and so should have a test for. That the involved client code (undo return stepresult successful) is simple isn't the same as unimportant.
In this, is it a dismal failure or just an ERROR failure?

wchamber (approved):

src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
zloery: All 3 steps now return a failure from undo, so this will be a dismal failure. I think that's what we want here because none of these steps can be undone, so we can't guarantee atomicity.

Fair point that we're relying on the lack of undo though, so maybe we should be enforcing this in a test, even if it won't be a very interesting one.

PR #313, opened 2021-04-27
PF-469 Retrieve the network and subnetwork instead of hardcoding them.

wchamber:
We currently don't have a reason to care which subnetwork the AI Platform Notebook VM instance connects to, so arbitrarily pick one. We can revisit this if we change our approach to GCE networks.
Introduce CRL google-compute package and associated cloud object wrapper.
Also use a CRL helper for generating a service account email.


wchamber:
@rtitle fyi grabbing some GCE networks for AI notebook instance creation

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/RetrieveNetworkNameStep.java
```diff
@@ -0,0 +1,110 @@
...
+    workingMap.put(CREATE_NOTEBOOK_NETWORK_NAME, extractNameFromUrl(subnetwork.getNetwork()));
+    workingMap.put(CREATE_NOTEBOOK_SUBNETWORK_NAME, subnetwork.getName());
+  }
+
+  /**
+   * Extract the name from a network URL like
+   * "https://www.googleapis.com/compute/v1/projects/{PROJECT_ID}/global/networks/{NAME}" or route
+   * URL like ""https://www.googleapis.com/compute/v1/projects/{PROJECT_ID}/regions/{REGION_NAME}"
+   */
+  private String extractNameFromUrl(String url) {
```
zloery, line 96: nit: static?

zloery (approved):
LGTM for WSM.
I'm not as familiar with the GCE/notebook side of things, so no input on the network/subnetwork logic being correct

src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/RetrieveNetworkNameStep.java
wchamber: thanks, done.

ddietterich (approved):
The code looks fine. I don't see a consumer of the network data you have retrieved. Is that following on, or did I miss something?

wchamber:
> The code looks fine. I don't see a consumer of the network data you have retrieved. Is that following on, or did I miss something?

The retrieved network data is being used in this PR to set the network and subnetwork field on `Instance` in `CreateAiNotebookInstanceStep#setNetworks`. There's no more follow up work for networks atm.

ddietterich:
OK, got it. Thanks

PR #315, opened 2021-04-28
[PF-418] Update version and publish new client

zloery:
Small change to update WSM version and publish an updated client library with the "create controlled BQ dataset" endpoint available (#307). I had originally planned to publish once delete/update/get were also available, but making this available now is easier for CLI development.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ sourceCompatibility = JavaVersion.VERSION_11
 def useMavenLocal = false
 allprojects {
 	group = "bio.terra"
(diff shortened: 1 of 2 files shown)


zloery:
Client tests will fail to build until `0.22.0-SNAPSHOT` is actually published, but they still pass otherwise

marikomedlock (approved):
Thank you!

wchamber (approved):

zloery:
Published 0.22.0-SNAPSHOT from this

PR #316, opened 2021-04-28
PF-715 Add more notebook creation fields.

wchamber:
Expand the create AI notebook endpoint with more fields mirroring the underlying GCP AI Platform Notebook instance creation https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1/projects.locations.instances#Instance. This will allow clients more control in how to configure their AI Notebook instance.
This is the last set of fields I'm planning on adding for now. The other fields don't seem likely to be useful for us.


src/main/resources/api/service_openapi.yaml
```diff
@@ -1792,23 +1792,86 @@ components:
...
+            Input only. The type of the data disk attached to this instance, defaults to standard
+            persistent disk (PD_STANDARD). See https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1/projects.locations.instances#DiskType
+          type: string
+        dataDiskSizeGb:
+          description: >-
+            Input only. The size of the data disk in GB attached to this instance, up to a maximum
+            of 64000 GB (64 TB). If not specified, this defaults to 100.
+          type: integer
+          format: int64
+        labels:
```
zloery, line 1843: Previously, we've talked about using labels for spend tracking within Terra, which would mean users couldn't arbitrarily set these. Is there a defined usecase for exposing this?

zloery (commented):
Generally LGTM, but I'm not sure we should expose labels on notebooks

src/main/resources/api/service_openapi.yaml
ddietterich: Yes, I would suggest leaving this off for now.

ddietterich (approved):
LGTM assuming removing labels

src/main/resources/api/service_openapi.yaml
wchamber: SG, removed.

zloery (approved):

PR #317, opened 2021-04-29
[PF-640] Restrict AssignedUser to self

jaycarlton:
The chief complaint of this PR can be satisfied by restricting all assignments to oneself. We don't yet have a solid use case for assigning resources to other users, and some of the questions around how that will work can wait until we know more.


src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -208,4 +213,61 @@ public String deleteControlledGcsBucket(
...
+  private void validateUserWorkspaceRole() {
+
+  }
+
+  private void validateOnlySelfAssignmentForWorkspaceWriter(ControlledResource controlledResource, AuthenticatedUserRequest userRequest) {
+    if (!controlledResource.getAccessScope().equals(AccessScopeType.ACCESS_SCOPE_PRIVATE)) {
+      // No need to handle SHARED resources
+      return;
+    }
+    final boolean userIsWriterOnWorkspace = samService.isAuthorized(
```
zloery, line 230: owners also have "write" permission on a workspace. It's probably easier to check `userIsWorkspaceOwner` and flip some conditions here.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -208,4 +213,61 @@ public String deleteControlledGcsBucket(
...
+        controlledResource.getWorkspaceId().toString(),
+        SamConstants.SAM_WORKSPACE_WRITE_ACTION);
+    if (userIsWriterOnWorkspace) {
+      // UserEmail field isn't always populated for requests going through Swagger, so get it the hard way
+      final String requestUserEmail = samService.getEmailFromToken(userRequest.getRequiredToken());
+
+      final boolean isAllowed = controlledResource
+                  .getAssignedUser()
+                  .map(requestUserEmail::equals)
+                  .orElseThrow(() -> new IllegalStateException("Assigned User field is required for private controlled resources."));
```
zloery, line 242: nit: we should throw an ErrorReport with a 400 code, like `BadRequestException`. `IllegalStateException` will cause a 500 status code, but this is really a problem with the user's request.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -208,4 +213,61 @@ public String deleteControlledGcsBucket(
...
+      return;
+    }
+    final boolean userIsWriterOnWorkspace = samService.isAuthorized(
+        userRequest.getRequiredToken(),
+        SamConstants.SAM_WORKSPACE_RESOURCE,
+        controlledResource.getWorkspaceId().toString(),
+        SamConstants.SAM_WORKSPACE_WRITE_ACTION);
+    if (userIsWriterOnWorkspace) {
+      // UserEmail field isn't always populated for requests going through Swagger, so get it the hard way
+      final String requestUserEmail = samService.getEmailFromToken(userRequest.getRequiredToken());
```
zloery, line 237: if you want to speed this up, you could check if the userRequest has a non-null/empty email first and use that if it's available.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -208,4 +213,61 @@ public String deleteControlledGcsBucket(
...
+    }
+  }
+
+  /**
+   * Workspace READERs may only have READER permission on controlled resources to which they are
+   * assigned. WRITERs may have either READER or WRITER permission, and OWNERS are unrestricted.
+   * @param controlledResource - controlled resource to check permission match on
+   * @param userRequest - current request
+   */
+  private void validateResourceAssigneeHasSameWorkspacePermissionAsResource(ControlledResource controlledResource, AuthenticatedUserRequest userRequest) {
```
zloery, line 256: I don't think we should be checking permissions on the resource, we're doing validation before a resource is created here

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
jaycarlton: zl: check self-assignment 1st. For now, that's the only thing allowed.

In future, letting owners reassign can be done, at cost of some complexity.

jaycarlton: TODO: do we remove the assignedUser field from the API? We can deprecate it for now.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -208,4 +225,109 @@ public String deleteControlledGcsBucket(
...
+  }
+
+  /**
+   * Workspace READERs may only have READER permission on private controlled resources to which they are
+   * assigned. WRITERs may have either READER or WRITER permission, and OWNERS are unrestricted.
+   *
+   * @param controlledResource - controlled resource to check permission match on
+   * @param userRequest - current request
+   */
+  private void validateWorkspaceReadersOnlyHaveReaderRoleOnResources(
```
jaycarlton, line 268: Delete this check, because only writers and owners are creating resources anyway. (self-assignment only).

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -208,4 +225,109 @@ public String deleteControlledGcsBucket(
...
+            String.format(
+                "Unrecognized role(s) for private resource: %s",
+                privateResourceIamRoles.toString()));
+      }
+    }
+    // All roles on the resource are permissible
+  }
+
+  /** Check to see that the assigned user is a member of the workspace of the controlled resource */
+  private void validateAssigneeForPrivateResourceIsWorkspaceMember(
```
jaycarlton, line 307: Delete this - same reason.

jaycarlton (commented):
zl: one client test will break (further), but I'll handle it.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
jaycarlton: 👍🏻

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResource.java
```diff
@@ -55,6 +55,11 @@ public StewardshipType getStewardshipType() {
     return StewardshipType.CONTROLLED;
   }
 
+  /**
+   * If specified, the assigned user must be equal to the user making the request.
+   * @return
```
ddietterich, line 60: nit: define the return?

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -208,4 +221,23 @@ public String deleteControlledGcsBucket(
...
+      final String requestUserEmail = samService.getEmailFromToken(userRequest.getRequiredToken());
+      // If there is no assigned user, this condition is satisfied.
+      final boolean isAllowed =
+          controlledResource
+              .getAssignedUser()
+              .map(requestUserEmail::equals)
+              .orElse(true);
+      if (!isAllowed) {
+        throw new BadRequestException(
+            "Workspace Writer User may only assign a private controlled resource to themselves.");
```
ddietterich, line 240: nit: message text: right now it is any controlled resource creator

ddietterich (approved):
LGTM - a couple of nits

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -208,4 +221,23 @@ public String deleteControlledGcsBucket(
...
   }
+
+  private void validateOnlySelfAssignment(
+      ControlledResource controlledResource, AuthenticatedUserRequest userRequest) {
+    if (!controlledResource.getAccessScope().equals(AccessScopeType.ACCESS_SCOPE_PRIVATE)) {
+      // No need to handle SHARED resources
+      return;
+    }
+      final String requestUserEmail = samService.getEmailFromToken(userRequest.getRequiredToken());
+      // If there is no assigned user, this condition is satisfied.
```
zloery, line 232: A private resource with no assigned user will currently pass this check, but we still rely on this field when creating a controlled resource. If you're marking this field deprecated in this PR, can you also change the place it's used: https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/service/iam/SamService.java#L545

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -173,6 +175,17 @@ private JobBuilder commonCreationJobBuilder(
     return jobBuilder;
   }
 
+  private void validateFlightPrerequisites(
```
zloery, line 178: nit: this is currently specific to `createControlledResource`, rename to `validateCreateFlightPrerequisites` or something similar?

src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -102,6 +103,28 @@ public String getWsmServiceAccountToken() throws IOException {
...
+   * @return - email address of user
+   */
+  public String getRequestUserEmailOrThrow(AuthenticatedUserRequest userRequest) {
+    return Optional.ofNullable(userRequest.getEmail())
+        .orElseGet(
+            () -> {
+              try {
+                return getEmailFromToken(userRequest.getRequiredToken());
+              } catch (InterruptedException e) {
+                // Treat as a 500 for now.
```
zloery, line 121: This call is ultimately run in a flight, so I don't think we want to convert this to a checked exception. Following the pattern of other methods here, I think `getRequestUserEmailOrThrow` should throw a checked InterruptedException.
`createControlledResource` already throws this as well, so this would only be a change to `getRequestUserEmailOrThrow`.

jaycarlton: Yes, but not the other call site at `validateOnlySelfAssignment`. I'll convert it there.

zloery: Good point about the call site inside `validateOnlySelfAssignment`. I think for non-Stairway calls to Sam methods that throw InterruptedException, the common pattern is to use rethrowIfSamInterrupted, e.g:
 `final String requestUserEmail = SamService.rethrowIfSamInterrupted(() -> samService.getRequestUserEmail(userRequest), "getRequestUserEmail");`
 which handles this conversion. This is all pretty new though, so I'm still getting used to that

jaycarlton: I don't love that pattern.

zloery (approved):

PR #318, opened 2021-04-29
[PF-420] Add get and update endpoints for controlled BQ datasets

zloery:
This change adds new endpoints to support "get" and "update" operations on BQ datasets. Unlike other controlled resources, we do not currently have cloud attributes to update, so this only supports modifying the "name" and "description" metadata properties in WSM.

I've also modified `getBucket` and `getAiNotebookInstance` to respond with a 400 error code when the user requests a resource of the wrong type, rather than a 500.


src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -182,6 +183,27 @@ public ControlledResource getControlledResource(
     return wsmResource.castToControlledResource();
   }
 
+  /**
+   * Update the name and description metadata fields of a controlled resource. These are only stored
+   * inside WSM, so this does not require any calls to clouds.
+   *
+   * @param workspaceId workspace of interest
+   * @param resourceId resource to update
+   * @param name name to change - may be null
```
jaycarlton, line 192: Indicate what null means: clear out the field or don't change the field.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1613,6 +1645,17 @@ components:
         cloningInstructions:
           $ref: '#/components/schemas/CloningInstructionsEnum'
 
+    UpdateControlledResourceRequestBody:
+      type: object
+      description: Common request body for updating a controlled resource
+      properties:
+        name:
+          description: Optional. New name to give to this resource. The resource name will not be updated if this is omitted.
```
jaycarlton, line 1653: Does `null` have different semantics from the key being omitted?

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -145,8 +149,60 @@ public ControlledGcpResourceApiController(
...
+  @Override
+  public ResponseEntity<ApiGcpBigQueryDatasetResource> getBigQueryDataset(
+      UUID workspaceId, UUID resourceId) {
+    final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    String projectId = workspaceService.getRequiredGcpProject(workspaceId);
+    ControlledResource controlledResource =
+        controlledResourceService.getControlledResource(workspaceId, resourceId, userRequest);
+    try {
+      ApiGcpBigQueryDatasetResource response =
+          controlledResource.castToBigQueryDatasetResource().toApiResource(projectId);
```
ddietterich, line 173: I think this casting approach is not working so well. I have a thought. What if each controlled resource subclass provided two methods. For example:
```
public static from(ControlledResource resource)
```
and
```
public static fromRequest(ControlledResource resource)
```
We would use `fromRequest()` when the inputs were from the user. It would throw BadRequestException if it isn't the right type. The `from()` we would use internally when we are sanity checking our own logic. That can throw InvalidMetadataException.

This idea is probably not for this PR, but I think it would avoid putting stuff in ControlledResource.java per-sub-class, so reduce the touch points when adding a new controlled resource. And it would remove all of the try/throw-something-different copies.

ddietterich (approved):
Looks good. I suggest a way to DRY it out, but it is larger more general fix.
If you like the idea, let me know and I'll make a ticket for it.

src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -278,6 +278,73 @@ public void createAiNotebookInstanceUndo() throws Exception {
...
+            .build();
+    ControlledBigQueryDatasetResource createdDataset =
+        controlledResourceService.createBqDataset(
+            resource,
+            creationParameters,
+            Collections.emptyList(),
+            userAccessUtils.defaultUserAuthRequest());
+    assertEquals(resource, createdDataset);
+
+    ControlledBigQueryDatasetResource getDataset =
```
jaycarlton, line 306: this name confuses me because it looks like a function name.

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
zloery: SGTM, I'm not a fan of catching unchecked exceptions here and we could avoid that by separating "cast and fail with 400" from "cast and fail with 500" into separate methods. We'd still have a type-specific casting method, but at least it would live inside the specific resource class, which seems better.

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -145,8 +149,60 @@ public ControlledGcpResourceApiController(
...
+      return new ResponseEntity<>(response, HttpStatus.OK);
+    } catch (InvalidMetadataException ex) {
```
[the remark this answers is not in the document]
zloery: True, but there are other places we use this for internal validation where it really is a 500 internal error. I think the right way to handle this is what Dan mentioned above (separate casting methods for "user's fault" vs "our fault")

src/main/resources/api/service_openapi.yaml
zloery: I think the problem is that when reading a request object we can't tell the difference between a key being present and `null` vs. being absent, so semantically they're the same here.

src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
zloery: good point, renamed for clarity

src/main/resources/api/service_openapi.yaml
jaycarlton: Oh, does Swagger insert all the missing keys and give them a null value? That would do it. I was thinking it was like a map where you could use `contiansKey` to distinguish the two cases.

zloery: It's more like we get an `UpdateControlledResourceRequestBody` object passed to our Controller method, and when `UpdateControlledResourceRequestBody.getName()` returns null, we can't tell whether that means the value was omitted, or included and set to null

jaycarlton: Yeah, it sounds like we need a free-form map if we want the patch semantics.

src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
jaycarlton: Or parameterize the cast method with the type you want thrown.

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -182,6 +184,32 @@ public ControlledResource getControlledResource(
...
+      UUID resourceId,
+      @Nullable String name,
+      @Nullable String description,
+      AuthenticatedUserRequest userReq) {
+    stageService.assertMcWorkspace(workspaceId, "updateControlledResource");
+    validateControlledResourceAndAction(
+        userReq, workspaceId, resourceId, SamControlledResourceActions.EDIT_ACTION);
+    // Name may be null if the user is not updating it in this request.
+    if (name != null) {
+      ValidationUtils.validateResourceName(name);
```
jaycarlton, line 208: If a name exists and I want to clear it, is there a way other than with an empty string?

jaycarlton (approved):

src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
zloery: No, you'd have to send an empty string.

zloery:
Published `0.23.0-SNAPSHOT` from this branch

PR #322, opened 2021-04-30
PF-712 Implement delete endpoints for controlled AI notebook instances.

wchamber:
Client tests for this and the other controlled AI Notebook instance endpoints to come in a follow up PR.

I'm matching what's already there for the other async endpoints, setting the wrong status status code for the result endpoint based on the job. I'll leave correction and standardization to PF-732.


ddietterich (approved):
LGTM - I recall you said tests to follow separately.

src/main/resources/api/service_openapi.yaml
```diff
@@ -1969,6 +2012,21 @@ components:
...
+      type: object
+      required: [ jobControl ]
+      properties:
+        jobControl:
+          $ref: '#/components/schemas/JobControl'
+
+    DeleteControlledGcpAiNotebookInstanceResult:
+      type: object
+      properties:
+        jobReport:
```
jaycarlton, line 2025: This is a little confusing, having both report objects in this "happy" result object. If you only return it for 200 or 202, why do we need an `ErrorReport`?

jaycarlton (approved):

src/main/resources/api/service_openapi.yaml
wchamber: You can have an ErrorReport in a happy result object is that the contained ErrorReport describes how the job failed, if present. The 20x happy response is that the client successfully polled the result endpoint and got a response from the service. Whether or not the job they're polling about finished successfully or had an error is encoded in the `jobReport` with more details in the `errorReport`. There's a difference between whether result endpoint worked and whether the job worked.

I'm following the [Async REST API spec](https://docs.google.com/document/d/1PTd4xvmV9xnEkWaIgFc6d3VUyJwwymQTFdhpKrdsUKw/edit), and the other APIs already implemented in WSM.  We can consider if there's another way we want to document its patterns for clients to read about.

PR #323, opened 2021-04-30
[PF-638] Fix broken client tests for controlled resources

zloery:
A number of controlled resource related client tests do not run regularly through GitHub actions, and they've been broken by other changes recently. This PR fixes the following:

- Adds a new user, `liam`. This is to replace `neville`, who is no longer listed in Sam.
- Adds a spendprofile parameter to the perf tests which need it. These tests have been failing since creation due to this missing parameter.
- Refactors `PrivateControlledGcsBucketLifecycle` to reflect the requirement that users assign private resources to themselves. Previously, this test had a workspace owner assign a resource to a workspace reader.

### DIFF
--- a/workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
+++ b/workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
@@ -83,26 +83,10 @@ public class PrivateControlledGcsBucketLifecycle extends WorkspaceAllocateTestSc
 
     String projectId = CloudContextMaker.createGcpCloudContext(getWorkspaceId(), workspaceApi);
 
-    ControlledGcpResourceApi resourceApi =
+    ControlledGcpResourceApi workspaceOwnerResourceApi =
         ClientTestUtils.getControlledGcpResourceClient(testUser, server);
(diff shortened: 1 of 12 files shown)


workspace-manager-clienttests/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -113,14 +97,32 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
...
         getWorkspaceId(),
-        IamRole.READER);
+        IamRole.WRITER);
     logger.info(
-        "Added {} as a reader to workspace {}", privateResourceUser.userEmail, getWorkspaceId());
+        "Added {} as a writer to workspace {}", privateResourceUser.userEmail, getWorkspaceId());
 
     // TODO(PF-643): this should happen inside WSM.
     logger.info("Waiting 15s for permissions to propagate");
     Thread.sleep(15000);
```
ddietterich, line 106: nit: a more self-documenting sleep is `TimeUnit.SECONDS.sleep(15)`.

ddietterich (approved):
LGTM - one nit

jaycarlton (approved):

PR #324, opened 2021-04-30
[PF-638] Add default RBS config to local servers, run FullIntegration from GHA

zloery:
This change adds a default spend profile and RBS configuration for local servers. This makes it possible to run operations requiring Cloud Contexts locally, which will be useful for development and testing. This default is overridden in any environment deployed using Helm, including all personal and live environments.

As local servers can now use Cloud Contexts, we can now run the FullIntegretion test suite, which also has client tests for controlled resources, on PR actions and nightly. This restriction was the reason we split the BasicIntegration and FullIntegration suites apart, so I've deleted the BasicIntegration suite now that the restriction no longer applies.


src/main/resources/application.yml
```diff
@@ -70,6 +70,21 @@ workspace:
...
+    client-credential-file-path: rendered/buffer-client-sa-account.json
+    instanceUrl: https://buffer.tools.integ.envs.broadinstitute.org
+    poolId: workspace_manager_v6
+
+  spend:
+    spend-profiles:
+    - # A default spend profile available in dev Sam.
+      id: wm-default-spend-profile
+      # The billing account workspace-dev has access to.
+      billing-account-id: 01A82E-CA8A14-367457
```
wchamber, line 86: Hmmm, this does mean the CWB deployment gets these values by default. Does our helm setup necessarily override them, or only if a helm value is specified.
I'm not convinced we shouldn't instead have a non-CWB spring profile to encode these kinds of things. Then make it easy from bootRun or a helper script to define whether we want a non-CWB version running. WDYT?
Either way, check the src/test resources for applications that are rendered obsolete by this.

wchamber (commented):
Is there any reason to keep the Basic Integration test suite, or should we just remove it? I think having multiple buckets of tests requires strong reasoning for the distinction.

src/main/resources/application.yml
wchamber: (Is there an accepted term for CWB vs non-CWB/cloud contexts? Would be nice to improve our terminology here and even record it in some documentation)

.github/workflows/test-runner-pr-integration.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         ./render-k8s-config.sh wsmtest ${{ steps.vault-token-step.outputs.vault-token }}
         echo "Running test suite"
         export TEST_RUNNER_SERVER_SPECIFICATION_FILE="workspace-local.json"
-        ./gradlew runTest --args="suites/BasicIntegration.json /tmp/TestRunnerResults"
+        ./gradlew runTest --args="suites/FullIntegration.json /tmp/TestRunnerResults"
```
jaycarlton, line 90: How much longer is this taking than the basic suite? I thought the plan was for nightly to be the long-running one and this one to be quicker, but it looks like they're doing the same suite now.

jaycarlton (approved):

.github/workflows/test-runner-pr-integration.yml
zloery: The basic suite is ~5m total, and the new suite is 8-10m. To me that's getting close to the limit of a presubmit test, but I think it's worth the additional coverage (especially since our connected tests are already pushing 15 minutes.

src/main/resources/application.yml
zloery: [Helm explicitly overrides](https://github.com/broadinstitute/terra-helm/blob/master/charts/workspacemanager/values.yaml#L192) `workspace.buffer.enabled` with a default value of `false` for all environments that don't set it otherwise. The other values will now be set in CWB, but they aren't used unless `enabled` is true.

My concern with splitting this into a separate spring profile is that we're introducing another configuration mechanism without really saving any complexity. We wouldn't just have one CWB and one non-CWB version, we would still need to pass through all the environment-specific configuration - except it would be across two application files instead of one, and we would need to set another env variable everywhere we call bootRun. Spend profiles and RBS are also currently the only difference between the two - a separate profile makes more sense to me if these diverge more, but I don't think it's at that threshold yet.

Good call, removed the duplicate buffer config from `application-test`. I'm leaving the spend profile config in `application-connected-test` since that uses a separate test-specific spend profile.

zloery: I don't think we have have a standard term for CWB vs. non-CWB. I've generally thought of it as "current CWB vs non-CWB" or "Rawls vs. MC Terra"

wchamber: Thanks, that sounds good to me.
Leaving the application-connected-test seems ok for now, but I think it would also be ok to move those tests using the same spend profile all the rest of the tests are using. I'll leave it up to you if that's easy to do now or we should wait on.

.github/workflows/test-runner-pr-integration.yml
jaycarlton: Yeah, we might want to consolidate at some point. I think the sweet spot is under 3 minutes if we can get back down.

src/main/resources/application.yml
zloery: Good point, we're already using `wm-default-spend-profile` for client tests. Switched connected-tests over

wchamber (approved):

PR #326, opened 2021-05-03
[DRAFT] PF-259 include clienttest in the build

ddietterich:
Draft PR that reorganizes the WSM build:
1. Moved to Gradle 7.0
2. Wrote `writeVersionProperties.sh` because the previous plugin does not support Gradle 7.0. It was way complicated given what we need.
2. Move the service code into `workspace-manager` directory at the same level as the client and clienttest. This accounts for almost all of the "Files changed" in the PR.
3. Top-level settings provide
    - resolution for test runner plugin
    - global variables providing version, swagger file location, common dependencies. We may want to put more dependencies or dependency versions here to share them.
4. Top-level common.gradle providing repository search path and dependency locking configuration. There may be opportunities for factoring out more stuff.
5. `workspace-manager-clienttest` depends on the `workspace-manager-client` project. So it will always run against the client it was built against. You don't have to build the service in order to build clienttest.
6. `workspace-manager` has a testImplementation dependency on the client project
7. All projects now use dependency version locking. It is a PITA and I'm not sure I like it, but there ya go.
8. Top-level `build.gradle` that is ONLY there so spotless can find its dependency.

I have more testing to do. This is working in my environment after a lot of messing around, so it may not actually work yet.

Files to look at:
- settings.gradle
- common.gradle
- build.gradle
- workspace-manager/build.gradle
- workspace-manager-client/build.gradle
- workspace-manager-clienttest/build.gradle


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.0-bin.zip
```
zarsky-broad, line 3: This should use the `-all` variant so IDE's can use the gradle sources.

workspace-manager-client/build.gradle
```diff
@@ -1,32 +1,36 @@
...
-	implementation group: "org.glassfish.jersey.core", name: "jersey-client"
-	implementation group: "org.glassfish.jersey.media", name: "jersey-media-json-jackson"
-	implementation group: "org.glassfish.jersey.media", name: "jersey-media-multipart"
-	implementation group: "com.fasterxml.jackson.datatype", name: "jackson-datatype-jsr310"
-	implementation group: "io.swagger.core.v3", name: "swagger-annotations"
-	swaggerCodegen group: "io.swagger.codegen.v3", name: "swagger-codegen-cli"
+	implementation group: "org.glassfish.jersey.core", name: "jersey-client", version: "2.32"
+	implementation group: "org.glassfish.jersey.media", name: "jersey-media-json-jackson", version: "2.32"
+	implementation group: "org.glassfish.jersey.media", name: "jersey-media-multipart", version: "2.32"
+	implementation group: "com.fasterxml.jackson.datatype", name: "jackson-datatype-jsr310", version: "2.11.3"
```
zarsky-broad, line 22: Why are we allowing these to get out of sync from the spring boot versions, rather than using `dependencyManagement` as before?

.gitignore
```diff
@@ -2,6 +2,7 @@ HELP.md
 
 ### Gradle ###
 .gradle
+../.gradle
```
jaycarlton, line 5: Why do you need to ignore something in a parent dir. I didn't think that was how `.gitignore` worked.

common.gradle
```diff
@@ -0,0 +1,26 @@
+ext {
```
jaycarlton, line 1: It would be good to have words here about what constitutes common and is safe to put here.

common.gradle
```diff
@@ -0,0 +1,26 @@
...
+    theVersion = '0.22.0-SNAPSHOT'
```
jaycarlton, line 2: maybe name this `workspaceManagerVersion` or `serviceVersion`?

workspace-manager-client/build.gradle
ddietterich: I do not want to have client as a subproject of the service. One of the goals of the work is that clienttest should not have to build the service to get itself built.

Is there a reason why the dependencies are required to be in sync?

From what I read, dependencyManagement is being replace by a new gradle feature that is in preview in 7.0.

settings.gradle
```diff
@@ -1,4 +1,32 @@
+pluginManagement {
+    repositories {
+        resolutionStrategy {
+            eachPlugin {
```
jaycarlton, line 4: please put a comment on why we're doing this. It looks suspicious, but I don't know enough gradle to put my finger on it.

workspace-manager-client/build.gradle
zarsky-broad: The dependency management provided by the plugin is focused on keeping versions of dependencies in sync between multiple projects or subprojects, it doesn't add a dependency between projects itself.

The reason to keep them in sync would be to make sure that the json parsing for the client and the server is following the same logic and matches.

settings.gradle
```diff
@@ -1,4 +1,32 @@
...
+include 'workspace-manager-clienttests'
+
+gradle.ext.openapiSourceFile = "${rootDir}/workspace-manager/src/main/resources/api/service_openapi.yaml"
 
-enableFeaturePreview('ONE_LOCKFILE_PER_PROJECT')
+// Single place to define the versions of dependencies shared between components.
+// This is just data. Each sub-project can pick and choose the dependencies they want.
+// To make the structure simple, we use the one-string version of the dependency expression.
+gradle.ext.librarySwaggerAnnotations = "io.swagger.core.v3:swagger-annotations:2.1.6"
+gradle.ext.librarySwaggerCli = "io.swagger.codegen.v3:swagger-codegen-cli:3.0.23"
```
jaycarlton, line 32: the .DS_Store files following probably shouldn't be included.

workspace-manager-client/build.gradle
```diff
@@ -1,32 +1,36 @@
...
+	implementation gradle.librarySwaggerAnnotations
+	swaggerCodegen gradle.librarySwaggerCli
 }
 
 // OpenAPI/Swagger Client Generation
+def artifactGroup = "${group}.workspace"
+def swaggerOutputDir = "${buildDir}/swagger-code"
```
jaycarlton, line 29: maybe `swagger-generated` would be clearer, since yaml is code too

workspace-manager-client/build.gradle
```diff
@@ -1,32 +1,36 @@
...
 
 swaggerSources {
 	client {
-		inputFile = file("${resourceDir}/api/service_openapi.yaml")
+		inputFile = file("${gradle.openapiSourceFile}")
```
jaycarlton, line 33: I don't think you need the `"${`. Does `file(gradle.openapiSourceFile)` not work?

workspace-manager-clienttests/build.gradle
```diff
@@ -1,23 +1,18 @@
+buildscript {
```
jaycarlton, line 1: Can we take the opportunity to rename this `workspace-manager-client-tests`?

workspace-manager-client/build.gradle
ddietterich: Where would you put the dependencyManagement declaration in this scheme?

I don't think it can go in `settings.gradle`.  Putting it in top-level `build.gradle` doesn't make sense to me, because that is not used in the component builds. Putting it in `common.gradle` wouldn't seem to work, since that is expanded per-project so wouldn't give the desired identical results.

workspace-manager-clienttests/build.gradle
```diff
@@ -93,4 +76,4 @@ spotless {
 }
 
 compileJava.dependsOn spotlessApply
-
+*/
```
jaycarlton, line 79: for POSIX compliance, set your editor to insert a newline at EOF.

workspace-manager-clienttests/build.gradle
```diff
@@ -37,54 +32,42 @@ dependencies {
...
         googleContainer = "v1-rev93-1.25.0"
 
-        workspacemanagerClient = "0.22.0-SNAPSHOT"
         datarepoClient = "1.0.155-SNAPSHOT"
         testRunner = "0.0.3-SNAPSHOT"
     }
     implementation group: 'org.apache.commons', name: 'commons-math3', version: "${apacheMath}"
     implementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: "${jUnit}"
     compileOnly "com.google.code.findbugs:annotations:${findbugsAnnotations}"
     implementation group: 'com.google.guava', name: 'guava', version: "${guava}"
```
jaycarlton, line 41: can these be alphabetized within each section?

.gitignore
ddietterich: No idea. I removed it.

common.gradle
ddietterich: I called it `wsmVersion` and am using it in all of the projects.

settings.gradle
ddietterich: Yeah, this is a funky thing for testrunner. I'll add a comment to that effect.

settings.gradle
ddietterich: Yes. Now I have to figure out how to get rid of them...

ddietterich:
Closing this draft. I will redraft a separate branch and PR of the current state of the repo

PR #327, opened 2021-05-04
PF-637 publish on merge

ddietterich:
This change removes the `build.yml` github action. I think it is redundant.

The `tag-publish.yml` is the workflow used to tag a branch, publish the client, and publish a docker to GCR. It runs automatically on push to dev (PR merge). It can be triggered manually to make hot fixes.

The `test.yml` is unchanged.

The automatic behavior should be:
- `test` runs on every PR and merge to dev. It is run manually for alpha, staging.
- `tag-publish` runs on merge to dev.

@zarsky-broad, I think you have worked on these before, so hoping you have time and context to review.


.github/workflows/tag-publish.yml
```diff
@@ -1,121 +1,80 @@
...
   workflow_dispatch:
     inputs:
       bump:
         description: 'Part of the version to bump: major, minor, patch'
         required: false
         default: 'patch'
+      branch:
+        description: 'Branch to run the workflow on'
+        required: false
+        default: 'main'
```
zarsky-broad, line 39: There isn't a branch with this name. Should it be dev?

.github/workflows/tag-publish.yml
```diff
@@ -130,20 +89,28 @@ jobs:
...
-          DOCKER_TAG=pr${{ github.event.number }}-$(echo ${{ github.event.pull_request.head.sha }} | cut -c1-7)
-        elif ${{github.event_name == 'push' }}; then
-          DOCKER_TAG=${{ steps.tag.outputs.tag }}
-        fi
+        DOCKER_TAG=${{ steps.tag.outputs.tag }}
         echo ::set-output name=name::gcr.io/${GOOGLE_PROJECT}/${SERVICE_NAME}:${DOCKER_TAG}
     - name: Build image locally with jib
       run: "./gradlew jibDockerBuild --image=${{ steps.image-name.outputs.name }} -Djib.console=plain"
     - name: Run Trivy vulnerability scanner
-      # https://github.com/broadinstitute/dsp-appsec-trivy-action
```
zarsky-broad, line 146: Is there a reason to remove this convenience link to the source of this action?

.github/workflows/tag-publish.yml
```diff
@@ -130,20 +89,28 @@ jobs:
...
+        service_account_email: ${{ secrets.GCR_PUBLISH_EMAIL }}
+        service_account_key: ${{ secrets.GCR_PUBLISH_KEY }}
+    - name: Explicitly auth Docker for GCR
+      run: gcloud auth configure-docker --quiet
     - name: Construct docker image name and tag
       id: image-name
       run: |
-        DOCKER_TAG=""
-        if ${{github.event_name == 'pull_request' }}; then
-          DOCKER_TAG=pr${{ github.event.number }}-$(echo ${{ github.event.pull_request.head.sha }} | cut -c1-7)
```
zarsky-broad, line 138: So we're no longer publishing PR images, then?

.github/workflows/tag-publish.yml
```diff
@@ -1,121 +1,80 @@
...
-        GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
-        DEFAULT_BUMP: patch${{ steps.semver.outputs.semver-part }}
+        GITHUB_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
+        DEFAULT_BUMP: patch
+        OVERRIDE_BUMP: ${{ steps.controls.outputs.semver-part }}
         RELEASE_BRANCHES: dev
+        HOTFIX_BRANCHES: hotfix.*
         VERSION_FILE_PATH: build.gradle
-        VERSION_LINE_MATCH: ^version
+        VERSION_LINE_MATCH: "^	version = \".*\""
```
zarsky-broad, line 76: Could we do something like `"^ *version = \".*\""` so that the number of spaces doesn't matter?

zarsky-broad (approved):
As long as devops and devsec are happy with the use of github secrets, this is ok for me. I think we've historically used vault because those values could rotate.

.github/workflows/tag-publish.yml
ddietterich: No. I thought it was old code. I'll put it back with more context.

.github/workflows/tag-publish.yml
ddietterich: We don't run this action on PR creation. Only on merge.
Is there a use case for publishing a docker for a PR?

.github/workflows/tag-publish.yml
ddietterich: Good idea. I'll see if there is a good matcher for whitespace. The character there is actually a tab in this case.

.github/workflows/tag-publish.yml
ddietterich: yup

.github/workflows/tag-publish.yml
zarsky-broad: There's always `\s`

.github/workflows/tag-publish.yml
zarsky-broad: I don't know, I just know we were doing it before.

.github/workflows/tag-publish.yml
ddietterich: Works like a charm. Thanks

ddietterich:
A note on using vault: what dev ops explained to me is they want to get away from GHA logging into vault and use secrets instead. Mike Flinn setup the new secrets this action depends on.

.github/workflows/tag-publish.yml
```diff
@@ -1,121 +1,80 @@
...
-      uses: databiosphere/github-actions/actions/bumper@v0.0.3
+        ref: ${{ steps.controls.outputs.checkout-branch }}
+        token: ${{ secrets.REPO_ACCESS_TOKEN }}
+    - name: Bump the tag to a new version
+      uses: databiosphere/github-actions/actions/bumper@-0.0.3
       id: tag
       env:
-        GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
-        DEFAULT_BUMP: patch${{ steps.semver.outputs.semver-part }}
+        GITHUB_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
```
jaycarlton, line 70: nit: alphabetize this list

.github/workflows/tag-publish.yml
```diff
@@ -1,121 +1,80 @@
...
+        DEFAULT_BUMP: patch
+        OVERRIDE_BUMP: ${{ steps.controls.outputs.semver-part }}
         RELEASE_BRANCHES: dev
+        HOTFIX_BRANCHES: hotfix.*
         VERSION_FILE_PATH: build.gradle
-        VERSION_LINE_MATCH: ^version
+        VERSION_LINE_MATCH: "^\\s*version = \".*\""
```
jaycarlton, line 76: you should use `\\s*` around the equals sign as extra or fewer spaces would break this match

.github/workflows/tag-publish.yml
```diff
@@ -1,121 +1,80 @@
...
-        ARTIFACTORY_REPO_KEY: "libs-snapshot-local"
-    - name: Auth to GCR
-      uses: google-github-actions/setup-gcloud@master
-      with:
-        version: '270.0.0'
-        service_account_email: ${{ steps.vault-secret-step.outputs.gcr-email }}
-        service_account_key: ${{ steps.vault-secret-step.outputs.gcr-key }}
-    - name: Explicitly auth Docker for GCR
-      run: gcloud auth configure-docker --quiet
     - name: Set up AdoptOpenJDK 11
```
jaycarlton, line 78: I'd leave out the version number in the step name since it's clear in the `with` statement.

.github/workflows/tag-publish.yml
```diff
@@ -130,20 +89,29 @@ jobs:
         restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
+    - name: "Publish to Artifactory"
```
jaycarlton, line 92: nit: this might make a good reusable action

jaycarlton (approved):

.github/workflows/tag-publish.yml
```diff
@@ -1,121 +1,80 @@
...
-#  1. Bump the version number
-#  2. Build and publish the build to Artifactory
+#  1. Bump the version number and tag the release
+#  2. Build and publish the client to Artifactory
 #  3. Build docker image and publish to GCR
 #  4. Trigger deployment to the dev environment
 #
-# When run on merge to dev, it bumps the patch version
+# When run on merge to dev, it tags and bumps the patch version by default. You can
+# bump other parts of the version by putting #major, #minor, or #patch in your commit
```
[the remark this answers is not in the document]
ddietterich: No, that would require monkeying with yet another tool. But I don't think it is important either. If we are hotfixing, we are doing a manual release anyway.

.github/workflows/tag-publish.yml
```diff
@@ -1,121 +1,80 @@
-name: Tag, Build, and Push Image
+name: Tag, Publish, and Deploy
```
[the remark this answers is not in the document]
ddietterich: I added a bit, but the Action tab truncates the name so not worth making it too long.

.github/workflows/tag-publish.yml
```diff
@@ -1,122 +1,81 @@
...
-        echo ::set-output name=artifactory-username::$ARTIFACTORY_USERNAME
-        echo ::add-mask::$ARTIFACTORY_USERNAME
-        echo ::set-output name=artifactory-password::$ARTIFACTORY_PASSWORD
-        echo ::add-mask::$ARTIFACTORY_PASSWORD
-    - name: "Bump the tag to a new version"
-      uses: databiosphere/github-actions/actions/bumper@v0.0.3
+        ref: ${{ steps.controls.outputs.checkout-branch }}
+        token: ${{ secrets.REPO_ACCESS_TOKEN }}
+    - name: Bump the tag to a new version
+      uses: databiosphere/github-actions/actions/bumper@-0.0.3
```
mflinn-broad, line 67: I believe this needs to be updated to `databiosphere/github-actions/actions/bumper@bumper-0.0.3` to pick up your recent changes. The tag also includes the action name since there are multiple actions in that repo.

PR #329, opened 2021-05-04
PF-734 Avoid " around 'service_account' for git secrets false positives

wchamber:
For some reason git secrets has a false positive for the string 'service_account' when it is in double quotes, `"like_this"`. 

The only pattern that looks related is `secrets.patterns ("type": "service_account"|"privateKeyType": "TYPE_GOOGLE_CREDENTIALS_FILE",)` but this doesn't have the type.
This is mysterious, but not especially hard to avoid.

### DIFF
--- a/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
+++ b/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
@@ -49,6 +49,9 @@ import org.springframework.http.HttpStatus;
 public class CreateAiNotebookInstanceStep implements Step {
   /** The Notebook instance metadata key used to control proxy mode. */
   private static final String PROXY_MODE_METADATA_KEY = "proxy-mode";
+  /** The Notebook instance metadata value used to set the service account proxy mode. */
+  // git secrets gets a false positive if 'service_account' is double quoted.
+  private static final String PROXY_MODE_SA_VALUE = "service_" + "account";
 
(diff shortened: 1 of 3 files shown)


ddietterich (changes requested):
I'd suggest adding a `.gitallowed` file containing `service_account` in it.
I tested that and it seems to work. Then we don't have to mess up the code.

wchamber:
> I'd suggest adding a `.gitallowed` file containing `service_account` in it.
> I tested that and it seems to work. Then we don't have to mess up the code.

If we do that, we're explicitly disabling the `("type": "service_account"|"privateKeyType": "TYPE_GOOGLE_CREDENTIALS_FILE",)` pattern matching that's supposed to check against checking in service account files.
I tested adding a `.gitallowed` with `service_account` in it, and it no longer stops commits for `"type": "service_account"` as expected.
As that pattern's listed at https://dsp-security.broadinstitute.org/platform-security-categories/git/setup-git-secrets, I think we'd need to have a broader conversation about working around it.

ddietterich:
> > I'd suggest adding a `.gitallowed` file containing `service_account` in it.
> > I tested that and it seems to work. Then we don't have to mess up the code.
> 
> If we do that, we're explicitly disabling the `("type": "service_account"|"privateKeyType": "TYPE_GOOGLE_CREDENTIALS_FILE",)` pattern matching that's supposed to check against checking in service account files.
> I tested adding a `.gitallowed` with `service_account` in it, and it no longer stops commits for `"type": "service_account"` as expected.
> As that pattern's listed at https://dsp-security.broadinstitute.org/platform-security-categories/git/setup-git-secrets, I think we'd need to have a broader conversation about working around it.

OK, got it.

ddietterich (approved):
LGTM

zloery (approved):

PR #333, opened 2021-05-06
PF-635: relax creation parameters requirements for buckets

jaycarlton:
Previously we assumed the storage class and lifecycle rules were required, but they're actually not.


jaycarlton:
Unit test failures should be fixed by https://github.com/DataBiosphere/terra-workspace-manager/pull/335

src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -454,6 +454,7 @@ private void storeResource(WsmResource resource) {
             .addValue("attributes", resource.attributesToJson());
     if (resource.getStewardshipType().equals(CONTROLLED)) {
       ControlledResource controlledResource = resource.castToControlledResource();
+      //noinspection deprecation
```
ddietterich, line 457: What does that mean?

ddietterich (approved):
lgtm

src/main/java/bio/terra/workspace/db/ResourceDao.java
jaycarlton: It shushes the deprecation warning. My thinking ATM is that I want to warn on new uses and grandfather in the two existing ones.

PR #335, opened 2021-05-06
replace static fixtures with methods that generate unique bucket names

jaycarlton:
Reports of flakkiness when unit tests are run simultaneously in the same DB point to the bucket name uniqueness check. Hence, using unique bucket names everywhere.


src/test/java/bio/terra/workspace/service/resource/controlled/flight/CreateGcsBucketStepTest.java
```diff
@@ -49,9 +51,12 @@
 
   @Test
   public void testCreatesBucket() throws RetryException, InterruptedException {
+    final ApiGcpGcsBucketCreationParameters creationParameters =
```
jaycarlton, line 54: The tests in this file are still disableld. Fixed in https://github.com/DataBiosphere/terra-workspace-manager/pull/333

zloery (approved):
LGTM

ddietterich (approved):
lgtm, thanks!

PR #339, opened 2021-05-11
[no ticket] Improve test failure messaging for AI notebook creation 

wchamber:
Test only change to improve the assertion errors when we fail to create an AI notebook instance.
Before, the test would NPE when trying to use the null, not create AI notebook instance. Now, the causing exception should show up as not-null.
This happens at least when Sam is down and we socket timeout repeatedly.

Also unrelated spotlessApply changes.


zloery (approved):

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -435,10 +437,15 @@ private ControlledAiNotebookInstanceResource createDefaultPrivateAiNotebookInsta
...
-            createJobId, ControlledAiNotebookInstanceResource.class, user.getAuthenticatedRequest())
-        .getResult();
+    JobService.JobResultOrException<ControlledAiNotebookInstanceResource> creationResult =
+        jobService.retrieveJobResult(
+            createJobId,
+            ControlledAiNotebookInstanceResource.class,
+            user.getAuthenticatedRequest());
+    assertNull(creationResult.getException(), "Error creating controlled AI notebook instance");
+    assertNotNull(
+        creationResult.getResult(), "Unexpected null created controlled AI notebook instance");
```
jaycarlton, line 447: Maybe add in your notes from the PR description about known causes for this.

jaycarlton (approved):

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
wchamber: I think that's a comment that'll get stale and not be especially helpful. If someone's looking at this test, it's because it failed and they have the cause in front of them.

PR #340, opened 2021-05-11
PF-750 fix up nightly tests

ddietterich:
This PR is a refactoring of the WSM github actions:
- Created a local GHA `integration-test` that runs an integration test and uploads the results
- Derived `nightly-tests` from the `test-runner-*` tests. It runs all three (integration, perf, resiliency) in sequence using the `integration-test` GHA
- Derived `pr-integration` from the `test-runner-pr-integration` again, using the `integration-test` action.
- Added notification on the terra-workspace-manager channel with the status of the nightly test runs

We now have only these workflows:
1. `test` - runs the JUnit tests on pull request and merge to dev
2. `pr-integration` - runs the TestRunner-based integration tests from the GHA host VM on pull request and merge to dev
3. `nightly-tests` - runs integration, perf, and resiliency on the wsmtest personal environment
4. `tag-publish` - tags, version bumps, publishes client to artifactory, pushes image to GCR - on merge to dev

For @ichengchang and @zarsky-broad: I added you as reviewers so you'd be aware of these changes. Feel free to review, of course, but consider yourselves optional.


ichengchang (approved):
This looks great, the refactoring was well done!

.github/actions/integration-test/action.yml
```diff
@@ -0,0 +1,48 @@
...
+    description: 'Test suite or config to run'
+    required: true
+    default: 'suites/FullIntegration.json'
+  vault-token:
+    description: 'Vault token for rendering'
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Set permissions and render test configurations
```
zloery, line 19: Do we need to render configs for the service in order to run integration tests? I thought the rendering was only required for running the service, and integration tests only use the client.

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,102 @@
...
+          test: suites/BasicResiliency.json
+          vault-token: ${{ steps.vault-token-step.outputs.vault-token }}
+
+      - name: Compose status message
+        if: ${{ always() }}
+        id: status-message
+        run: |
+          L2="Integration: ${{ steps.integration-test.outcome }} "
+          L3="Perf: ${{ steps.perf-test.outcome }} "
+          L4="Resiliency : ${{ steps.resiliency-test.outcome }}"
```
zloery, line 83: nit: extra space after "Resiliency"

.github/workflows/pr-integration.yml
```diff
@@ -2,15 +2,22 @@
 # server changes but runs a published version of the client, so it will not pick
 # up local client changes.
 
-name: PR Client Library Integration Tests
+name: PR Integration Tests
```
jaycarlton, line 5: include author and description

.github/workflows/pr-integration.yml
```diff
@@ -29,16 +36,18 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
```
jaycarlton, line 38: maybe name this step for the UI

.github/workflows/test.yml
```diff
@@ -6,16 +6,24 @@ name: Run Service Tests
...
+      - 'service/local-dev/**'
   workflow_dispatch:
     inputs:
       testEnv:
         description: 'Environment in which tests should be run. Currently works on dev, alpha and staging'
         required: true
 
 jobs:
-  test:
+  test-job:
```
jaycarlton, line 26: I don't know why the suffix is necessary; this can't be used in any context that doesn't expect a job.

.github/workflows/pr-integration.yml
```diff
@@ -47,6 +56,7 @@ jobs:
           ~/.gradle/wrapper
         key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
         restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
```
jaycarlton, line 61: include shell

.github/workflows/test.yml
```diff
@@ -6,16 +6,24 @@ name: Run Service Tests
 on:
   push:
     branches: [ dev ]
+    paths-ignore:
+      - 'README.md'
+      - '.github/**'
+      - 'service/local-dev/**'
   pull_request:
     branches: [ '**' ]
```
jaycarlton, line 14: should this pattern exclude dev? Or can dev never be the right side of a pull request?

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,102 @@
...
+  schedule:
+    - cron: '0 7 * * *' # run at 7 AM UTC, 2 AM ET.
+
+jobs:
+  nightly-tests:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
```
jaycarlton, line 17: If there's no name, does it put it ini the UI or not? I'd name this Checkout Branch.

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,102 @@
...
+      - name: Set up AdoptOpenJDK 11
+        uses: joschi/setup-jdk@v2
+        with:
+          java-version: 11
+
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
```
jaycarlton, line 28: does this have to be newline separated, or can it be an array?

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,102 @@
...
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
```
jaycarlton, line 30: put these strings on their own line so they'e line up and you can see one is a prefix to the other.

Is there any way to squeeze that bit of duplication out of these two values?

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,102 @@
...
+        with:
+          test-server: workspace-wsmtest.json
+          test: suites/BasicResiliency.json
+          vault-token: ${{ steps.vault-token-step.outputs.vault-token }}
+
+      - name: Compose status message
+        if: ${{ always() }}
+        id: status-message
+        run: |
+          L2="Integration: ${{ steps.integration-test.outcome }} "
```
jaycarlton, line 81: `LINE_2` would be a bit clearer I think. But they're all joined into one line, right?

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,102 @@
...
+        uses: broadinstitute/action-slack@v3.8.0
+        # see https://github.com/broadinstitute/action-slack
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        with:
+          status: ${{ job.status }}
+          channel: "#terra-workspace-manager"
+          username: "WSM nightly test"
+          author_name: ${{ steps.status-message.outputs.status-bold }}
+          icon_emoji: ':vertical_traffic_light:'
```
jaycarlton, line 101: Can we use a different emoji depending on the test outcome?

.github/workflows/pr-integration.yml
```diff
@@ -29,16 +36,18 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    # This step is needed when running test using a local server.
+
+      # This step is needed when running test using a local server.
     - name: Initialize Postgres DB
```
jaycarlton, line 41: add "for local server test run" to the name

.github/workflows/pr-integration.yml
```diff
@@ -77,29 +81,11 @@ jobs:
       # With a timeout of 300 seconds, try to connect to localhost:8080 every 1 second.
```
jaycarlton, line 81: add shell

.github/workflows/pr-integration.yml
```diff
@@ -77,29 +81,11 @@ jobs:
...
-    - name: Upload test results to Google Bucket
-      if: ${{ always() }}
-      run: |
-        cd integration
-        echo "Upload test results to Google Bucket"
-        ../gradlew uploadResults --args="CompressDirectoryToTerraKernelK8S.json /tmp/TestRunnerResults"
-
-    - name: Upload test results to BigQuery Test Runner Dataset
+    - name: Run the integration test suite
+      id: integration-test
```
jaycarlton, line 85: add description, author

jaycarlton (approved):
This is awesome! Making a big dent in the duplication and complexity of these GHAs.

.github/actions/integration-test/action.yml
ddietterich: I believe it is required, but I am not sure. It was in the existing actions so I assumed it was needed.
I made PF-744 suggesting that we combine the render in `integration` and the render in `service` into a single tool that both use. When we do that, we won't care if this was needed or not.

.github/actions/integration-test/action.yml
```diff
@@ -0,0 +1,48 @@
...
+        chmod +x render_config.sh
+        ./render_config.sh dev ${{ inputs.vault-token }}
+        chmod +r rendered
+      shell: bash
+
+    # We run the test as one script, collecting the test result
+    - name: Run the test
+      id: test-run
+      run: |
+        cd integration
```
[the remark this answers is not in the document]
ddietterich: Yes. An alternative would have been to use a script instead of an action and invoke the script multiple times. I used the action form before I understood that it did not support calling other actions. In fact, I started with three actions and ended up with just this one. I still think this is a good form and will get better as GitHub adds use of actions.

I ended up merging the steps to be able to hold the test result for the return from the run. It started as separate steps, but it got a lot longer just passing around the testResult and resultDir. So something of a compromise.

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,102 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Nightly Tests
+
```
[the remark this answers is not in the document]
ddietterich: Those are not available in workflows. Only actions.

.github/actions/integration-test/action.yml
zloery: My understanding is that it's required if we're running a local server (which we do for PR tests) but shouldn't be otherwise. sgtm doing this as part of a larger config refactor

zloery (approved):

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,102 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Nightly Tests
```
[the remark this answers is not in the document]
ddietterich: It would be nice if the "workflow" part of the GHA was separate from the "on event" part of the workflow.

.github/workflows/nightly-tests.yml
ddietterich: Oh, you can do soooo much. If you make a custom format, you can make anything. I wasted a good chunk of time fussing with format. Then I slapped myself on the hand and said, "That is not the goal, and made it pretty simple."

Feel free to play. It is fun to mess with. One thing I realized afterwards is that I should have aimed the notification at my personal DM instead of the channel and spared everyone my tests.

.github/workflows/pr-integration.yml
ddietterich: The `shell` declaration is required in the action and not available in the workflow. I assume they will normalize this at some point.

.github/actions/integration-test/action.yml
ddietterich: I think we should assign the work to @jaycarlton to give him an excuse to play with the slack notification! :)

PR #342, opened 2021-05-13
PF-768 Don't retry notebook bad requests to fail faster.

wchamber:


zloery (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
```diff
@@ -89,6 +89,10 @@ public StepResult doStep(FlightContext flightContext)
           logger.debug("Notebook instance {} already created.", instanceName.formatName());
           return StepResult.getStepResultSuccess();
         }
+        if (HttpStatus.BAD_REQUEST.value() == e.getStatusCode()) {
```
jaycarlton, line 92: nit: could do `else if`

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
wchamber: done

PR #343, opened 2021-05-13
PF-743 reorg tidying

ddietterich:
Apply suggestions from the reorg code review.
Update the documentation.


client/build.gradle
```diff
@@ -1,105 +1,108 @@
...
+        contextUrl = "https://broadinstitute.jfrog.io/broadinstitute/"
+        repository {
+            repoKey = "libs-snapshot-local" // The Artifactory repository key to publish to
+            username = "${artifactory_username}" // The publisher user name
+            password = "${artifactory_password}" // The publisher password
+        }
+        defaults {
+            // This is how we tell the Artifactory Plugin which artifacts should be published to Artifactory.
+            // Reference to Gradle publications defined in the build script.
+            publications("workspaceManagerClientLibrary")
```
jaycarlton, line 99: Is there a way to publish source from here?

common.gradle
```diff
@@ -1,10 +1,6 @@
 // Common definitions and constraints to be applied to the projects.
 // These are applied in the buildscript block.
-def useMavenLocal = false
 repositories {
-    if (useMavenLocal) {
```
jaycarlton, line 5: did this not work here?

service/build.gradle
```diff
@@ -1,138 +1,144 @@
...
+    constraints {
+        implementation 'com.google.guava:guava:30.1.1-jre' // "-jre" for Java 8 or higher
+    }
+    implementation platform('com.google.cloud:libraries-bom:20.2.0')
+    implementation group: "com.google.auto.value", name: "auto-value-annotations"
+    implementation group: "com.google.guava", name:"guava"
+    
+    // Terra deps
+    implementation group: "bio.terra", name: "datarepo-client", version: "1.41.0-SNAPSHOT"
+    // hk2 is required for datarepo client, but not correctly included
```
jaycarlton, line 53: How is it not correctly included? Can it not be fixed from here?

jaycarlton (approved):

zloery (approved):
Nice, thanks for the documentation update! That's been drifting out of date for a while now

common.gradle
jmczerk: Same question...

jmczerk (approved):
Good cleanup on Gradle and doc updates look great!

client/build.gradle
ddietterich: Yes, I'll add that.

common.gradle
ddietterich: I removed it when it seemed that mavenLocal would be useless for testing the libraries. Now that I fixed the libraries to be mavenLocal-friendly, I'll put this back.

service/build.gradle
ddietterich: datarepo should have `hk2` as an `api` dependency, since it is needed by users of the client. They have it as an `implementation` dependency so it doesn't show through. We have to include it ourselves.

ddietterich: I improved the comment and added a ticket for the TDR project

PR #345, opened 2021-05-17
[PF-417] Update GCS Bucket Endpoint

jaycarlton:
Update supports the resource name and description, and the bucket's default storage class and lifecycle. Null entries are ignored by GCS, which means resetting an entry to its default requires either passing an empty array in the case of the lifecycle rules, or the default storage class (STANDARD I believe).


service/build.gradle
```diff
@@ -52,7 +52,7 @@ dependencies {
 	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-compute', version: '0.8.0'
 	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-notebooks', version: '0.4.0'
 	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-serviceusage', version: '0.7.0'
-	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-storage', version: '0.8.0'
+	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-storage', version: '0.9.0'
```
wchamber, line 55: debug: pointer no action needed

service/src/test/java/bio/terra/workspace/service/resource/controlled/GcsApiConversionsTest.java
```diff
@@ -0,0 +1,215 @@
...
+  @Test
+  public void testLifecycleRuleConversions() {
+    final ApiGcpGcsBucketLifecycleRule wsmRule1 = toWsmApi(GCS_LIFECYCLE_RULE_1);
+    assertEquals(ApiGcpGcsBucketLifecycleRuleActionType.DELETE, wsmRule1.getAction().getType());
+    assertNull(wsmRule1.getAction().getStorageClass());
+
+    assertEquals(42, wsmRule1.getCondition().getAge());
+    assertFalse(wsmRule1.getCondition().isLive());
+    assertEquals(2, wsmRule1.getCondition().getNumNewerVersions());
+    assertNull(wsmRule1.getCondition().getMatchesStorageClass()); // not empty list
```
jaycarlton, line 106: The `BucketInfo` builder converts null to an empty list for lifecycle rules, but the `LifecycleRule` builder does not do that for `matchtesStorageClass`. It's hard to come up with a clean scheme for handling these.

Here are two different conventions for null lists side by side:
```

    /** @deprecated Use {@code setLifecycleRules} method instead. * */
    @Override
    @Deprecated
    public Builder setDeleteRules(Iterable<? extends DeleteRule> rules) {
      this.deleteRules = rules != null ? ImmutableList.copyOf(rules) : null;
      return this;
    }

    @Override
    public Builder setLifecycleRules(Iterable<? extends LifecycleRule> rules) {
      this.lifecycleRules =
          rules != null ? ImmutableList.copyOf(rules) : ImmutableList.<LifecycleRule>of();
      return this;
    }
    ```

File: service/src/main/java/bio/terra/workspace/service/resource/controlled/GcsApiConversions.java [hunk 3]
```diff
@@ -0,0 +1,236 @@
...
+
+public class GcsApiConversions {
+
+  private GcsApiConversions() {}
+
+  private static final BiMap<ApiGcpGcsBucketDefaultStorageClass, StorageClass> STORAGE_CLASS_MAP =
+      HashBiMap.create();
+
+  static {
+    STORAGE_CLASS_MAP.put(ApiGcpGcsBucketDefaultStorageClass.STANDARD, StorageClass.STANDARD);
```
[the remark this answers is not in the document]
ddietterich: I don't understand. I think you are saying, "hey look, I added a layer that wasn't there before", right?

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1698,7 +1717,20 @@ components:
...
+      properties:
+        name:
+          description: Optional. New name to give to this resource. The resource name will not be updated if this is omitted.
+          type: string
+        description:
+          description: Optional. New description to give to this resource. The resource description will not be updated if this is omitted.
+          type: string
+        updateParameters:
+          $ref: '#/components/schemas/GcpGcsBucketUpdateParameters'
+        jobControl:
```
ddietterich, line 1732: Needed? I thought this was synchronous.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1698,7 +1717,20 @@ components:
...
+      description: Update a GCS bucket's metadata and lifecycle rules.
+      properties:
+        name:
+          description: Optional. New name to give to this resource. The resource name will not be updated if this is omitted.
+          type: string
+        description:
+          description: Optional. New description to give to this resource. The resource description will not be updated if this is omitted.
+          type: string
+        updateParameters:
+          $ref: '#/components/schemas/GcpGcsBucketUpdateParameters'
```
ddietterich, line 1731: Is this optional? Are the sub-parts optional? For example, can I only change the default storage class?

service/src/main/java/bio/terra/workspace/service/resource/controlled/GcsApiConversions.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  public static BucketInfo toBucketInfo(
+      String bucketName, ApiGcpGcsBucketUpdateParameters updateParameters) {
+    return BucketInfo.newBuilder(bucketName)
+        .setStorageClass(toGcsApi(updateParameters.getDefaultStorageClass()))
+        .setLifecycleRules(toGcsApiRulesList(updateParameters.getLifecycle()))
+        .build();
+  }
+
+  // If input is null, return null, as we need the null value preserved in case it's not there.
```
ddietterich, line 71: nit: make this javadoc?

service/src/main/java/bio/terra/workspace/service/resource/controlled/GcsApiConversions.java
```diff
@@ -0,0 +1,236 @@
...
+   * @return - populated parameters object
+   */
+  public static ApiGcpGcsBucketUpdateParameters toUpdateParameters(BucketInfo bucketInfo) {
+    return new ApiGcpGcsBucketUpdateParameters()
+        .lifecycle(new ApiGcpGcsBucketLifecycle().rules(toWsmApiRulesList(bucketInfo)))
+        .defaultStorageClass(toWsmApi(bucketInfo.getStorageClass()));
+  }
+
+  /**
+   * build an incomplete BucketInfo object corresponding to the fields to be updated TODO: add
```
ddietterich, line 56: Is there a Jira ticket for that?

service/src/main/java/bio/terra/workspace/service/resource/controlled/GcsApiConversions.java
```diff
@@ -0,0 +1,236 @@
...
+
+  public static LifecycleAction toGcsApi(ApiGcpGcsBucketLifecycleRuleAction lifecycleRuleAction) {
+    switch (lifecycleRuleAction.getType()) {
+      case DELETE:
+        return LifecycleAction.newDeleteAction();
+      case SET_STORAGE_CLASS:
+        return LifecycleAction.newSetStorageClassAction(
+            toGcsApi(lifecycleRuleAction.getStorageClass()));
+      default:
+        throw new IllegalStateException(
```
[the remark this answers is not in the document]
ddietterich: Good question. It is certainly possible to get sent bad enums. Not sure if the client auto-magically validates and rejects. It is certainly a bad request, as opposed to `IllegalStateException`. I would make it some kind of validation exception.

service/src/main/java/bio/terra/workspace/service/resource/controlled/GcsApiConversions.java
```diff
@@ -0,0 +1,236 @@
...
+      default:
+        throw new IllegalArgumentException(
+            String.format("GCS BucketLifecycle action type %s not recognized.", actionType));
+    }
+  }
+
+  public static LifecycleCondition toGcsApi(ApiGcpGcsBucketLifecycleRuleCondition condition) {
+    final var resultBuilder = LifecycleCondition.newBuilder();
+
+    /* TODO(PF-506): some conditions aren't in the version of the Google Storage API in the
```
ddietterich, line 189: Is that still the case after the CRL upgrade?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/RetrieveControlledResourceMetadataStep.java
```diff
@@ -0,0 +1,62 @@
...
+    this.workspaceId = workspaceId;
+    this.resourceId = resourceId;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    WsmResource resource = resourceDao.getResource(workspaceId, resourceId);
+    if (resource.getStewardshipType() != StewardshipType.CONTROLLED) {
+      throw new IllegalArgumentException(
```
ddietterich, line 36: Suggestion finding a better exception or make one. We shouldn't be throwing `IllegalArgument` or `IllegalState` in the mainline code as a general rule.

And this is yet another case where the throw from the cast call is the wrong exception. Sigh... I got that wrong big time...

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledResourceMetadataStep.java
```diff
@@ -0,0 +1,66 @@
...
+    this.resourceId = resourceId;
+    this.workspaceId = workspaceId;
+    this.controlledResourceMetadataManager = controlledResourceMetadataManager;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    final FlightMap inputParameters = flightContext.getInputParameters();
+    final ControlledResource resource =
```
ddietterich, line 38: I don't think you use this variable, do you?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledResourceMetadataStep.java
```diff
@@ -0,0 +1,66 @@
...
+   * @param flightContext
+   */
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    final FlightMap workingMap = flightContext.getWorkingMap();
+    final String previousName =
+        workingMap.get(ControlledResourceKeys.PREVIOUS_RESOURCE_NAME, String.class);
+    final String previousDescription =
+        workingMap.get(ControlledResourceKeys.PREVIOUS_RESOURCE_DESCRIPTION, String.class);
+    resourceDao.updateResource(workspaceId, resourceId, previousName, previousDescription);
```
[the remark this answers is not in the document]
ddietterich: What validation? I don't see any on the `do()` either. But it seems fine to me as written.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledResourceMetadataStep.java
```diff
@@ -0,0 +1,66 @@
...
```
ddietterich, line 63: Since we don't do any locking, it is possible that the resource was deleted between the time this flight started and when this executes this step. You would get a not-found raised here. I think that should be tolerated and not cause the flight to fail dismally.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateGcsBucketStep.java
```diff
@@ -0,0 +1,80 @@
...
+  private StepResult updateBucket(ApiGcpGcsBucketUpdateParameters updateParameters) {
+    final String projectId =
+        workspaceService.getRequiredGcpProject(bucketResource.getWorkspaceId());
+
+    final BucketInfo bucketInfo = toBucketInfo(bucketResource.getBucketName(), updateParameters);
+
+    final StorageCow storageCow = crlService.createStorageCow(projectId);
+
+    final BucketCow existingBucketCow = storageCow.get(bucketInfo.getName());
+    if (existingBucketCow == null) {
```
ddietterich, line 67: Same comment here. On `undo()`, I think we should tolerate a missing bucket.

service/src/test/java/bio/terra/workspace/service/resource/controlled/GcsApiConversionsTest.java
ddietterich: Well, that is kind of annoying. You have to read everything out and supply everything in, only changing the thing you want to change.

ddietterich (commented):
Bunch 'o' comments and suggestions. Looks like this is getting there. Too bad the SDK update semantics are setup the way that they are.

common.gradle
```diff
@@ -1,6 +1,6 @@
 // Common definitions and constraints to be applied to the projects.
 // These are applied in the buildscript block.
-def useMavenLocal = false
+def useMavenLocal = true
```
jaycarlton, line 3: Note to self: switch this back when https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/92 is merged and published.

zloery (approved):

integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
```diff
@@ -193,6 +213,18 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
...
+    // Update the bucket
+    logger.info("About to update the bucket {} resourceID {} workspace ID {}",
+        bucketName, bucket.getResourceId(), getWorkspaceId());
+
+    final GcpGcsBucketResource resource = updateBucketAttempt(resourceApi, resourceId);
+    logger.info("Updated resource name to {} and description to {}",
+        resource.getMetadata().getName(), resource.getMetadata().getDescription());
+    assertEquals(UPDATED_RESOURCE_NAME, resource.getMetadata().getName());
+    assertEquals(UPDATED_DESCRIPTION, resource.getMetadata().getDescription());
+
```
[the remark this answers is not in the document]
jaycarlton: Used `ownerStorageClient` to validate the cloud attributes.

service/src/main/java/bio/terra/workspace/service/resource/controlled/GcsApiConversions.java
```diff
@@ -0,0 +1,236 @@
...
+        .condition(toWsmApi(lifeCycleRule.getCondition()));
+  }
+
+  public static LifecycleAction toGcsApi(ApiGcpGcsBucketLifecycleRuleAction lifecycleRuleAction) {
+    switch (lifecycleRuleAction.getType()) {
+      case DELETE:
+        return LifecycleAction.newDeleteAction();
+      case SET_STORAGE_CLASS:
+        return LifecycleAction.newSetStorageClassAction(
+            toGcsApi(lifecycleRuleAction.getStorageClass()));
```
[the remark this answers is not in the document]
jaycarlton: Decided not to add validation at this time. Null is possible here.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateGcsBucketStep.java
jaycarlton: good point

integration/src/main/resources/configs/integration/ControlledGcsBucketLifecycle.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ControlledGcsBucketLifecycle",
   "description": "CRUD controlled GCS bucket tests.",
-  "serverSpecificationFile": "workspace-dev.json",
+  "serverSpecificationFile": "workspace-local.json",
```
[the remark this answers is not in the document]
jaycarlton: yes

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
```diff
@@ -0,0 +1,100 @@
...
+   * <p>Returns the controlled resource object if it exists and the user is permitted to perform the
+   * specified action.
+   *
+   * @param userReq the user's authenticated request
+   * @param workspaceId if of the workspace this resource exists in
+   * @param resourceId id of the resource in question
+   * @param action the action to authorize against the resource
+   */
+  @Traced
+  public void validateControlledResourceAndAction(
```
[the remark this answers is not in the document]
jaycarlton: You're right that it doesn't really fit here. I prefer to come back to it the next time I think about layering.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/RetrieveControlledResourceMetadataStep.java
ddietterich: The throw will be caught by Stairway and will be the return error for the flight. Since `Illegal*` isn't derived from one of the common exception base classes, it gets whatever default the global exception handler gives it.

I'd prefer to exercise more control at the point of exception. See `InvalidMetadataException` as an example of a "totally server side" error.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledResourceMetadataStep.java
ddietterich: Yes, you are right

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -116,6 +73,30 @@ public ControlledGcsBucketResource createBucket(
     return jobBuilder.submitAndWait(ControlledGcsBucketResource.class);
   }
 
+  public ControlledGcsBucketResource updateGcsBucket(
+      ControlledGcsBucketResource resource,
+      ApiGcpGcsBucketUpdateParameters updateParameters,
```
ddietterich, line 78: Should updateParameters be Nullable?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateGcsBucketStep.java
```diff
@@ -0,0 +1,110 @@
...
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    final FlightMap workingMap = flightContext.getWorkingMap();
+    final ApiGcpGcsBucketUpdateParameters previousUpdateParameters =
+        workingMap.get(PREVIOUS_UPDATE_PARAMETERS, ApiGcpGcsBucketUpdateParameters.class);
+
+    return updateBucket(previousUpdateParameters);
+  }
+
+  private StepResult updateBucket(ApiGcpGcsBucketUpdateParameters updateParameters) {
```
ddietterich, line 60: I think the `updateParameters` can be null. Have you tested the case where the update request does not contain any update parameters at all?

ddietterich (commented):
There is one case I'd like to double check before approval: can the update parameters be null from the REST API? That is, could I simply leave out the `UpdateControlledGcpGcsBucketRequestBody`? If so, I think this code is vulnerable to NPE.

ddietterich (approved):
lgtm, thanks!

PR #346, opened 2021-05-17
[PF-742] Add logging in DB when creating resources

zloery:
This is a small change to log the values we're inserting into the database every time we create a resource. This may be helpful for debugging the postgres crash we've seen occur on resource creation. Separately, we could also enable CloudSQL-side query logging, but that can be a fairly high volume of logging and I'm not sure it's necessary.

The gitignore changes are unrelated, but this is a small cleanup from earlier changes and I wanted to be extra sure I wasn't checking in rendered secrets.


service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -461,6 +472,12 @@ private void storeResource(WsmResource resource) {
           // TODO: add associatedApp to ControlledResource
           .addValue("associated_app", null)
           .addValue("assigned_user", controlledResource.getAssignedUser().orElse(null));
+      logger.info(
+          "Resource also has controlled-resource specific parameters access_scope: {}, managed_by: {}, associated_app: {}, assigned_user: {}",
+          controlledResource.getAccessScope().toSql(),
+          controlledResource.getManagedBy().toSql(),
+          null,
+          controlledResource.getAssignedUser().orElse(null));
```
davidangb, line 480: could you move the logging down into a new `catch (Throwable t)` clause of the update statement? That way, we could log success/failure of the update statement alongside the values that were used in that statement.

ddietterich: I had been thinking it would be useful to see that the same values succeed one time and fail another time. In any case, if we don't already log the success/failure, that would a good addition.

ddietterich (approved):
lgtm

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
zloery: sgtm if we only want to log the failure query instead of all queries.
I don't think we'll ever see the exact same values because we're generating random UUIDs as the resource ID. Even if a user retries the exact request, the resource ID will end up being different.

(I messed with that behavior locally to confirm you can rerun exact queries and see a success, but I don't think that will happen in a live deployment)

davidangb (approved):
:+1: this looks good to me, I also see Dan's point about wanting to log both successes and errors, but I think this is a good improvement!

PR #348, opened 2021-05-18
[PF-772] assertEquals argument order

jaycarlton:
Expected value should come first. In almost all cases it's clear which is which. It might  be good to have another set of eyes search for "assertEquals" and scan for anomalies.


yonghaoy (approved):

zloery (approved):

jaycarlton:
Integration job is broken currently (local server won't start). Merging with other tests passing.

PR #349, opened 2021-05-19
Add data reference info in perf config files

ungwudik:
This PR fixes the snapshot-related error in perf tests
`java.lang.IllegalArgumentException: Must provide Spend Profile ID, Data Repo snapshot ID, and Data Repo Instance Name in the parameters list`


ddietterich (approved):
LGTM, thank you

jaycarlton (approved):
lgtm. I didn't dig to find where those are used, but if the test passes now I'm happy.

PR #351, opened 2021-05-19
PF-783 Update to Resource Manager v3 via CRL

wchamber:
Update to the ResourceMangaer v3 GCP API.
Add a utility for polling a GCP OperationCow until it has completed successfully in a step and use that.


zloery (approved):

service/gradle.lockfile
```diff
@@ -44,7 +44,7 @@ com.google.api:gax-grpc:1.63.0=compileClasspath,productionRuntimeClasspath,runti
 com.google.api:gax-httpjson:0.80.0=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
 com.google.api:gax:1.63.0=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
 com.google.apis:google-api-services-bigquery:v2-rev20201030-1.30.10=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
-com.google.apis:google-api-services-cloudresourcemanager:v1-rev20200810-1.30.10=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+com.google.apis:google-api-services-cloudresourcemanager:v3-rev20210411-1.31.0=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
```
jaycarlton, line 47: so v3 == 1.31.0?

service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
```diff
@@ -46,4 +55,21 @@ public static void deleteProject(String projectId, CloudResourceManagerCow resou
       throw e;
     }
   }
+
+  /**
+   * Poll until the Google Service API operation has completed. Throws any error or timeouts as a
+   * {@link RetryException}.
+   */
+  public static void pollUntilSuccess(
```
jaycarlton, line 63: or: `pollDamnTheTorpedoes`

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
```diff
@@ -95,15 +95,7 @@ public StepResult doStep(FlightContext flightContext)
...
-      creationOperation =
-          OperationUtils.pollUntilComplete(
-              creationOperation, Duration.ofSeconds(20), Duration.ofMinutes(12));
-      if (creationOperation.getOperation().getError() != null) {
-        throw new RetryException(
-            String.format(
-                "Error creating notebook instance %s. %s",
-                instanceName.formatName(), creationOperation.getOperation().getError()));
-      }
+      GcpUtils.pollUntilSuccess(creationOperation, Duration.ofSeconds(20), Duration.ofMinutes(12));
```
jaycarlton, line 98: maybe make constants for the durations since they appear twice

jaycarlton (approved):

service/gradle.lockfile
wchamber: no, google has 1.31 for both v1 and v3. It's the `v3-rev...` that's changing it. See https://mvnrepository.com/artifact/com.google.apis/google-api-services-cloudresourcemanager and https://github.com/googleapis/google-api-java-client-services/tree/master/clients/google-api-services-cloudresourcemanager

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/CreateAiNotebookInstanceStep.java
wchamber: Create and delete are independent operations, so we might actually want to vary them independently. And these numbers don't matter that much as they're overly conservative.

PR #352, opened 2021-05-20
[PF-753] Handle workspace and context deletion with controlled resources

zloery:
Currently, deleting a cloud context that contains controlled resources leaves dangling information, and deleting a workspace that contains controlled resources will fail with a 400 error. 
This change makes both of those operations run without errors instead. This is done by deleting the controlled resources automatically as part of a Stairway flight before deleting the context/workspace.


service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledSamResourcesStep.java
```diff
@@ -0,0 +1,82 @@
...
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    FlightMap workingMap = flightContext.getWorkingMap();
+
+    @SuppressWarnings("unchecked")
+    List<ControlledResource> controlledResourceList =
+        workingMap.get(ControlledResourceKeys.CONTROLLED_RESOURCE_LIST, List.class);
+
+    try {
```
zloery, line 39: Reviewers:
This implementation is kind of hacky, but I think we need to run the Sam requests in parallel to avoid too long on this request. I'm not super familiar with streams - is there a cleaner/more canonical pattern for this?

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledDbResourcesStep.java
```diff
@@ -0,0 +1,36 @@
...
+
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import bio.terra.workspace.db.ResourceDao;
+import bio.terra.workspace.service.workspace.model.CloudPlatform;
+import java.util.UUID;
+
+/** A step to delete the controlled resources on a single cloud from a workspace. */
```
jaycarlton, line 11: a single cloud, like GCP?

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledDbResourcesStep.java
```diff
@@ -0,0 +1,36 @@
...
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    resourceDao.deleteAllControlledResources(workspaceId, cloudPlatform);
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    return flightContext.getResult();
```
jaycarlton, line 34: In other places, we log something about not being able to undo a delete step.

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledSamResourcesStep.java
```diff
@@ -0,0 +1,82 @@
...
+                  samService.deleteControlledResource(controlledResource, userRequest);
+                } catch (InterruptedException e) {
+                  // This function must handle the checked InterruptedException, even though the
+                  // Step as a whole is fine surfacing it. Here, we wrap the InterruptedException in
+                  // an unchecked exception, smuggle it out, then catch and throw it later.
+                  throw new UncheckedInterruptedException(e);
+                }
+              });
+    } catch (UncheckedInterruptedException e) {
+      throw e.getException();
```
jaycarlton, line 55: Can Sam implement a batch delete endpoint? This is a lot of round trips.

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledSamResourcesStep.java
```diff
@@ -0,0 +1,82 @@
...
+    @SuppressWarnings("unchecked")
+    List<ControlledResource> controlledResourceList =
+        workingMap.get(ControlledResourceKeys.CONTROLLED_RESOURCE_LIST, List.class);
+
+    try {
+      // Deleting each resource takes a call to Sam. As there may be many many resources and they
+      // are all independent, running these calls in parallel can save time.
+      controlledResourceList.parallelStream()
+          .forEach(
+              controlledResource -> {
```
jaycarlton, line 44: If your function returned `Optional<InterruptedException>` or an enum of succeeded/interrupted, then you could collect a map of controlled resource -> interruptedException and do something clever with that, like maybe an exception saying "3 of 99 controlled resources failed to delete".

Though, style-wise, I usually like to avoid side effects in `map` or other functional functions.

You could also try putting each delete call in a `CompletableFuture` and doing an `awaitAll()` on them. I haven't done that in a few years, and you still need to decide how to handle the exception. Why is it checked anyway, if it's almost always fatal?

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteGcpContextFlight.java
```diff
@@ -5,24 +5,42 @@
...
 import bio.terra.stairway.RetryRuleExponentialBackoff;
 import bio.terra.workspace.common.utils.FlightBeanBag;
+import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
+import bio.terra.workspace.service.job.JobMapKeys;
+import bio.terra.workspace.service.workspace.model.CloudPlatform;
+import java.util.UUID;
 
 /** A {@link Flight} for deleting a Google cloud context for a workspace. */
 public class DeleteGcpContextFlight extends Flight {
   private static final int INITIAL_INTERVALS_SECONDS = 1;
```
jaycarlton, line 15: I recommend `Duration` for this kind of data

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteGcpContextFlight.java
```diff
@@ -5,24 +5,42 @@
...
 
+    addStep(
+        new ListControlledResourceIdsStep(appContext.getResourceDao(), workspaceId, CLOUD_PLATFORM),
+        retryRule);
+    addStep(
+        new DeleteControlledSamResourcesStep(appContext.getSamService(), userRequest), retryRule);
+    addStep(
+        new DeleteControlledDbResourcesStep(
+            appContext.getResourceDao(), workspaceId, CLOUD_PLATFORM),
+        retryRule);
```
jaycarlton, line 41: maybe put a comment about how we don't need to delete them from GCP because the project is going away when we delete the cloud context. I got confused again on that point.

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteWorkspaceStateStep.java
```diff
@@ -15,22 +15,22 @@
 public class DeleteWorkspaceStateStep implements Step {
```
jaycarlton, line 15: I've been calling this DB state "metadata", not state. You could also say "dbState" I guess.

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
```diff
@@ -25,22 +26,30 @@ public WorkspaceDeleteFlight(FlightMap inputParameters, Object applicationContex
     WorkspaceStage workspaceStage =
         WorkspaceStage.valueOf(
             inputParameters.get(WorkspaceFlightMapKeys.WORKSPACE_STAGE, String.class));
+    UUID workspaceId =
+        UUID.fromString(inputParameters.get(WorkspaceFlightMapKeys.WORKSPACE_ID, String.class));
     // TODO: we still need the following steps once their features are supported:
```
jaycarlton, line 31: I don't know if we need to comment here about these systems that don't exist, though I suppose it's a logical place for it. It's not really a TODO because too many things have to happen first.

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
```diff
@@ -49,6 +58,6 @@ public WorkspaceDeleteFlight(FlightMap inputParameters, Object applicationContex
         throw new InternalLogicException(
             "Unknown workspace stage during creation: " + workspaceStage.name());
```
jaycarlton, line 59: this isn't creation

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledSamResourcesStep.java
ddietterich: Is this a premature optimization? My theory is that the typical number of controlled resources in a workspace will be maybe 10. So do we really have a performance problem here?

That said, this doesn't strike me as hacky. Seems like a perfect case for parallelStream() and the code is straightforward. My only concern us that Sam will not welcome the rapid fire requests. I'm not sure how things are queued or how they apply backpressure. It would be counterproductive if the result of parallelism here is failure/retries in the SamService. Might be worth checking with CI team.

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
```diff
@@ -25,22 +26,30 @@ public WorkspaceDeleteFlight(FlightMap inputParameters, Object applicationContex
...
+    // 1. Delete the cloud contexts from non-GCP cloud platforms
     // 2. Notify all registered applications of deletion, once applications are supported
     // 3. Delete policy objects in Policy Manager, once it exists.
 
     RetryRule retryRule =
         new RetryRuleExponentialBackoff(
             INITIAL_INTERVALS_SECONDS, MAX_INTERVAL_SECONDS, MAX_OPERATION_TIME_SECONDS);
 
+    addStep(
+        new ListControlledResourceIdsStep(appContext.getResourceDao(), workspaceId, null),
```
jaycarlton, line 41: in which step do we delete the controlled resources from the database?

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledSamResourcesStep.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  /**
+   * An unchecked exception wrapping a checked InterruptedException. This is a workaround for the
+   * stream above which cannot handle checked exceptions, even though that's really what we want.
+   */
+  private static class UncheckedInterruptedException extends RuntimeException {
+    private final InterruptedException interruptedException;
```
ddietterich, line 65: OK, now this is a little hacky :)
Do you know the semantics of InterruptedException across a paralleStream? I assume it just works.

I wonder if Stairway should expose an unchecked interrupted exception that it would catch and process just like the InterruptedException.

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteGoogleContextFlightTest.java
```diff
@@ -82,6 +82,12 @@ void deleteContextDo() throws Exception {
 
     // Force each step to be retried once to ensure proper behavior.
     Map<String, StepStatus> doFailures = new HashMap<>();
+    doFailures.put(
```
jaycarlton, line 85: aside: I wonder if we could use a helper method to retry all steps.

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteGcpContextFlight.java
```diff
@@ -5,24 +5,42 @@
...
+        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
 
     RetryRule retryRule =
         new RetryRuleExponentialBackoff(
             INITIAL_INTERVALS_SECONDS, MAX_INTERVAL_SECONDS, MAX_OPERATION_TIME_SECONDS);
 
+    addStep(
+        new ListControlledResourceIdsStep(appContext.getResourceDao(), workspaceId, CLOUD_PLATFORM),
+        retryRule);
+    addStep(
```
ddietterich, line 36: There is a potential Sam leak here if a new controlled resource is created after the enumeration.
The Sam step will miss it, but the metadata step will delete it. For now, would you make a comment to that effect. I am thinking that we could use locking to close the window. There is already a ticket PF-555 you can reference.

Is there a reason to split the List and the DeleteControlledSamResourceStep into two steps?
It would be idempotent to read the list in memory and apply it to Sam in one go. If the step failed and restarted, it would read the list again and apply it to Sam again. We would have to tolerate "not found" in the Sam call.

service/src/test/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlightTest.java
```diff
@@ -0,0 +1,155 @@
...
+            jobService.getStairway(),
+            WorkspaceDeleteFlight.class,
+            deleteParameters,
+            DELETION_FLIGHT_TIMEOUT,
+            debugInfo);
+    assertEquals(FlightStatus.FATAL, flightState.getFlightStatus());
+
+    // Verify the resource and workspace are still deleted, as delete steps have no undo.
+    assertThrows(
+        WorkspaceNotFoundException.class,
```
jaycarlton, line 147: shouldn't this be `ResourceNotFoundException`?

jaycarlton (approved):
This looks great!

ddietterich (commented):
A couple of comments worth considering

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledSamResourcesStep.java
zloery: I was working on the assumption that an average workspace might have 10s of resources, and that ~30 resources would not be unreasonable. This comment made check the traces on `SamService.deleteControlledResource` though, and it turns out I was overestimating how long the Sam calls took by 3-4x (each one is actually ~250ms in the traces I sampled). something something measure twice...

Removed the parallelism for now. If this becomes an issue in the future there's room for optimization, including discussion with CI about how Sam wants to handle many requests.

integration/src/main/java/scripts/utils/ResourceMaker.java
```diff
@@ -165,18 +165,17 @@ public static void deleteControlledGcsBucket(
...
         new CreateControlledGcpBigQueryDatasetRequestBody()
             .common(
                 new ControlledResourceCommonFields()
                     .accessScope(AccessScope.SHARED_ACCESS)
                     .managedBy(ManagedBy.USER)
                     .cloningInstructions(CloningInstructionsEnum.NOTHING)
-                    .description("Description of " + name)
-                    .name(name))
+                    .description("Description of " + datasetId)
+                    .name(datasetId))
```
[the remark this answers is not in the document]
zloery: You're right that they're distinct values - resource name is a WSM concept shared across resources, and datasetID is a BigQuery specific concept on GCP. For convenience I'm using the same value for both here, added docs to clarify

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledDbResourcesStep.java
zloery: Yep, exactly

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledDbResourcesStep.java
zloery: Good call, done

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledSamResourcesStep.java
zloery: it's definitely a possible future optimization. If performance becomes an issue here, it will be worth a discussion with CI on how we should send many requests.

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledSamResourcesStep.java
zloery: That's probably a question for Stairway that I don't know enough to answer - tabling this for now since I'm scrapping the parallelism

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteGcpContextFlight.java
zloery: added in both flights

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
zloery: Yeah - I found the old comment useful for context when I was making this change so I'm inclined to leave it, but it's not really something we can do yet

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
zloery: clarified in comment but we never explicitly do this when deleting a workspace, we have resources set up to cascade on workspace deletion in the DB.

service/src/test/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlightTest.java
zloery: If you ask for a resource that doesn't exist in a workspace that also doesn't exist, we throw `WorkspaceNotFound`. You'll only get `ResourceNotFound` if you ask for a resource that doesn't exist in a workspace that does

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteGcpContextFlight.java
zloery: Added TODO referencing PF-555

No strong reason to separate the two other than I thought of them as distinct operations. Given that this saves serializing/deserializing a list of complex objects, that seems like a better approach. 
`SamService.deleteControlledResource` internally handles not found errors, so we don't need to manage those in steps.

service/src/test/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlightTest.java
jaycarlton: that makes sense

ddietterich (approved):
lgtm

PR #353, opened 2021-05-21
PF-557 gather retry rules into a utility class

jaycarlton:
Consolidated the retry rules a little bit. This gives a central place for adjusting them and moves the burden of deciding the values out of individual flights.


service/src/main/java/bio/terra/workspace/common/utils/RetryRules.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public static RetryRule shortExponential() {
+    return new RetryRuleExponentialBackoff(1, 8, 16);
+  }
+
+  /**
+   * Retry rule for handling unexpected timeouts with Sam. Note that some errors from Sam (like
+   * NOT_FOUND responses to resources which were already deleted) are not handled by retries.
+   */
+  public static RetryRule sam() {
```
ddietterich, line 41: I don't think we need this. We have retries implemented at the SamService layer, so I don't think we should have them at the step level as well.

service/src/main/java/bio/terra/workspace/common/utils/RetryRules.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class RetryRules {
+
+  private RetryRules() {}
+
+  /**
+   * Retry rule for immediately retrying a failed step twice. This is useful for retrying operations
+   * where waiting is not necessary, e.g. retrying conflicted SQL transactions.
+   */
+  public static RetryRule databaseConflict() {
```
ddietterich, line 15: When @zloery finishes his database retry work, we won't want this one either.
I'd suggest removing it and its usage now.

ddietterich (commented):
I think we shouldn't have `sam` or `databaseConflict` in use, because we are handling those retries within the services. (Not DB yet, bug @zloery is working on that.)

You didn't put them in, but would you take them out, please?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceFlight.java
```diff
@@ -73,6 +54,7 @@ public DeleteControlledResourceFlight(FlightMap inputParameters, Object beanBag)
             userRequest),
         samRetryRule);
 
+    RetryRule gcpRetryRule = RetryRules.cloud();
```
marikomedlock, line 57: nit: `final`? ditto other classes (some use `final`, some don't). would be good to be consistent.

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlight.java
```diff
@@ -17,25 +18,14 @@
 public class CreateGcpContextFlight extends Flight {
   // Buffer Retry rule settings. For Buffer Service, allow for long wait times.
   // If the pool is empty, Buffer Service may need time to actually create a new project.
```
marikomedlock, line 20: nit: Move this comment to the `RetryRules.buffer()` method.

service/src/main/java/bio/terra/workspace/common/utils/RetryRules.java
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Retry rule for steps interacting with GCP. If GCP is down, we don't know when it will be back,
+   * so don't wait forever. Note that RetryRules can be re-used within but not across Flight
+   * instances.
+   */
+  public static RetryRule cloud() {
+    return new RetryRuleFixedInterval(10, 10);
+  }
+
+  public static RetryRule cloudLongRunning() {
```
marikomedlock, line 28: Include a comment about when to use each of these retry rules. ditto `shortExponential`, `buffer`.

service/src/main/java/bio/terra/workspace/common/utils/RetryRules.java
```diff
@@ -0,0 +1,51 @@
+package bio.terra.workspace.common.utils;
+
+import bio.terra.stairway.RetryRule;
+import bio.terra.stairway.RetryRuleExponentialBackoff;
+import bio.terra.stairway.RetryRuleFixedInterval;
+
+public class RetryRules {
```
marikomedlock, line 7: nit: public classes should have a description

marikomedlock (approved):
Just some nits, otherwise lgtm.

service/src/main/java/bio/terra/workspace/common/utils/RetryRules.java
jaycarlton: ok

service/src/main/java/bio/terra/workspace/common/utils/RetryRules.java
jaycarlton: 👍🏻

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceFlight.java
jaycarlton: In the style guide it's optional for local declarations. I like it everywhere because it calls out assignments and I don't have to worry about reassignment, but some other folks aren't as convinced.

ddietterich (approved):
lgtm

PR #356, opened 2021-05-21
PF-194 bump CRL dependencies for updated janitor client

wchamber:


jaycarlton (approved):
I'm starting to think having to do all this repeatedly is negating some of the benefit of having separate CRL dependencies.

wchamber:
> I'm starting to think having to do all this repeatedly is negating some of the benefit of having separate CRL dependencies.

Automated bumping would be nice. It's actually less frequent that we need to bump all the versions simultaneously, but we've done a few in a row now where we have to update one of their shared dependencies.

jmczerk:
> > I'm starting to think having to do all this repeatedly is negating some of the benefit of having separate CRL dependencies.
> 
> Automated bumping would be nice. It's actually less frequent that we need to bump all the versions simultaneously, but we've done a few in a row now where we have to update one of their shared dependencies.

We could probably get the best of both worlds by publishing a [Gradle platform](https://docs.gradle.org/current/userguide/java_platform_plugin.html)  in CRL, then we could move the versions independently inside CRL while only bumping the platform dependency in downstream repos.

jmczerk (approved):

PR #358, opened 2021-05-24
PF-730 prework - add workflow tester

ddietterich:
You cannot run a workflow that is on a branch unless the workflow already exists in the default branch.
So here is a test workflow that is manually triggered that we can use for testing workflows.
In particular, I will use it for testing the database cleaning.

### DIFF
--- a/.github/workflows/workflow-tester.yml
+++ b/.github/workflows/workflow-tester.yml
@@ -1,0 +1,36 @@
+# Workflow for testing new actions and workflows
+#
+# Github does not provide a way to run workflows that are not already merged to the
+# default branch. Once a workflow that is configured for manual running is merged,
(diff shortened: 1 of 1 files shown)


zloery (approved):

.github/workflows/workflow-tester.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  workflow-test-job:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Get Vault token
```
jaycarlton, line 23: I don't think it needs a vault token. It could just take in a parameter for whether it should pass or fail so you can also test output.

jaycarlton (approved):

PR #359, opened 2021-05-24
[PF-487] Add retries to DB operations

zloery:
This change pulls in the DB retry utilities from common lib to support re-running queries that fail with retryable exceptions. I set this up to retry queries up to 10 times with a 250ms delay. These are pretty arbitrary values, and we can change them in the future if we see more conflicts or need higher performance.

This also replaces our existing `DataSourceTransactionManager` with the newer and more JDBC-specific `JdbcTransactionManager`, which properly translates SQL errors during transaction commits.

Note that the previous implementation actually didn't work - when a SQL statement failed (e.g. with a conflict), the transaction it was a part of was immediately aborted. Further retries would then always fail, as they were running in a cancelled transaction, which Postgres forbids. Instead, I now retry the entire `Transactional` method, which creates a new transaction for each attempt.


service/build.gradle
```diff
@@ -169,7 +169,7 @@ jacocoTestReport {
         xml.enabled = true
         xml.destination = file("$buildDir/reports/jacoco/test/jacoco.xml")
     }
-    dependsOn test
+//    dependsOn test
```
jaycarlton, line 172: why is this commented out?

service/src/main/java/bio/terra/workspace/db/DbRetryUtils.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * This class contains wrappers around JDBC methods which run them with retries. Only DB exceptions
+ * considered retryable ({@link RecoverableDataAccessException} and {@link
+ * org.springframework.dao.TransientDataAccessException}) are retried. See Spring documentation of
+ * these exceptions for more information:
+ * https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/dao/DataAccessException.html
+ *
+ * <p>Retries are with no delay, up to a certain maximum number of retries.
+ */
+public class DbRetryUtils {
```
jaycarlton, line 20: Why not put all these in the common lib?

service/src/main/java/bio/terra/workspace/db/DbRetryUtils.java
```diff
@@ -0,0 +1,80 @@
...
+      return DatabaseRetryUtils.executeAndRetry(
+          () -> jdbcTemplate.update(sql, params), Duration.ZERO, MAX_RETRIES);
+    } catch (InterruptedException e) {
+      throw new IllegalStateException(
+          "Got an InterruptedException without any threads sleeping in updateWithRetries. This should never happen.",
+          e);
+    }
+  }
+
+  public static <T> List<T> queryWithRetries(
```
jaycarlton, line 36: The "WithRetries" is kinda redundant if you'll always import with the class name. `DbRetryUtils.query()` works for me.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -436,7 +439,8 @@ private void validateUniqueAiNotebookInstance(
             .addValue("workspace_id", notebookResource.getWorkspaceId().toString())
             .addValue("instance_id", notebookResource.getInstanceId())
             .addValue("location", notebookResource.getLocation());
-    Integer matchingCount = jdbcTemplate.queryForObject(sql, sqlParams, Integer.class);
+    Integer matchingCount =
+        DbRetryUtils.queryForObjectWithRetries(jdbcTemplate, sql, sqlParams, Integer.class);
```
jaycarlton, line 443: nit: you can make another overload like `queryForCountWithRetries` so you don't have to pass `Integer.class` in these places.

service/src/test/java/bio/terra/workspace/db/DbRetryUtilsTest.java
```diff
@@ -0,0 +1,85 @@
...
+  private static final CannotSerializeTransactionException RETRY_EXCEPTION =
+      new CannotSerializeTransactionException("test");
+
+  @Mock NamedParameterJdbcTemplate jdbcTemplate;
+
+  @Test
+  void updateSucceedsWithRetry() {
+    when(jdbcTemplate.update(any(), (SqlParameterSource) any()))
+        .thenThrow(RETRY_EXCEPTION)
+        .thenReturn(42);
```
jaycarlton, line 36: Oh neat, I didn't realize you could have multiple then statements like this.

jaycarlton (approved):
very nice

service/build.gradle
zloery: whoops, fixed. I think our Jacoco setup is the root cause of unit tests running twice and I was investigating, but that's not for this branch.

service/src/main/java/bio/terra/workspace/db/DbRetryUtils.java
zloery: I think we'll want to call `executeAndRetry` with different parameters in different services - I think the 0 time retries work here, but would cause problems for slow queries, where the conflict might not be immediately over. Having the base `executeAndRetry` in common lib and calling with different parameters from each service seemed like the right compromise to me

service/src/main/java/bio/terra/workspace/db/DbRetryUtils.java
zloery: Good call, I like that better.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
zloery: I like having the `DbRetryUtils` methods wrap the jdbcTemplate methods closely here - I personally think the clarity from that outweighs the extra verbosity. It's a pretty weak preference, though

service/src/test/java/bio/terra/workspace/db/DbRetryUtilsTest.java
zloery: it's pretty handy - I just learned about `thenThrow(array)` this PR too, which is pretty cool.

service/src/main/java/bio/terra/workspace/db/DbRetryUtils.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public class DbRetryUtils {
+
+  public static int MAX_RETRIES = 10;
+
+  public static int update(
+      NamedParameterJdbcTemplate jdbcTemplate, String sql, MapSqlParameterSource params) {
+    try {
+      return DatabaseRetryUtils.executeAndRetry(
+          () -> jdbcTemplate.update(sql, params), Duration.ZERO, MAX_RETRIES);
```
ddietterich, line 28: I zero duration is too short. I'd allow at least ~250ms to allow the conflicting transaction to complete. Besides, I don't think any of our endpoints are so performance sensitive that we need to be concerned at that amount of time.

What has been your experience in testing this?

service/src/main/java/bio/terra/workspace/db/DbRetryUtils.java
```diff
@@ -0,0 +1,80 @@
...
+    } catch (InterruptedException e) {
+      throw new IllegalStateException(
```
ddietterich, line 30: I don't think you should hide interrupted exception. Just because you are not sleeping does not mean that somewhere in the JDBC layers, there is not a wait. So I am not sure it can never happen.

ddietterich (commented):
This version of retry will not let us ride through a "CloudSQL down" situation. We know by the exception or SQL code whether it is a transaction conflict or a connection error. I had hoped you would add that logic into the common lib retry. Would you make a new ticket for adding that support?

I have questions about `InterruptedException` hiding and the retry wait time.

service/src/main/java/bio/terra/workspace/db/DbRetryUtils.java
zloery: sgtm. I hadn't seen any failures with the instant retries, but that seems like a risk for flakiness if we have longer running queries in the future.

service/src/main/java/bio/terra/workspace/db/DbRetryUtils.java
zloery: Done. I think throwing those as `InternalServerErrorException` from services (or letting Steps handle them) is the right level to surface them

zloery:
I revisited this implementation after testing showed the previous approach actually didn't work due to how we handle transactions - requesting another review on this

service/src/main/java/bio/terra/workspace/db/DbRetryUtils.java
ddietterich: This could be done in common lib by having a setter for default backoff and attempts, and then a variant of `executeAndRetry` that did not take those arguments.

Maybe collect these suggestions in a follow-on PR. Along with different timings for transaction vs connection failures.

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
```diff
@@ -83,7 +83,11 @@ public void updateReferenceResource(
       AuthenticatedUserRequest userReq) {
     workspaceService.validateWorkspaceAndAction(
         userReq, workspaceId, SamConstants.SAM_UPDATE_REFERENCED_RESOURCE);
-    resourceDao.updateResource(workspaceId, resourceId, name, description);
+    try {
```
ddietterich, line 86: Tova made a method `rethrowIfSamInterrupted` that packages this pattern up in a method that takes a function. You could add that to DbRetryUtils or elsewhere to DRY this up a bit.

ddietterich (approved):
Looking good. One suggestion to DRY up the code.

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
zloery: I saw that pattern and have mixed feelings. It does save a lot of repeated try/catch boilerplate, but it does also make the code a little harder for me to follow - that's a second layer of lambdas to actually execute a function. Maybe that's not a bad thing, but it does feel a little off. 
This is already what SamService does though, so I added methods to DbRetryUtils for more consistency and to cut down the repeated code.

service/src/main/java/bio/terra/workspace/db/DbRetryUtils.java
zloery: Sounds good, PF-805 to track these suggestions

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
ddietterich: Yes, I share your "feels a little off", but I am slowly learning to like it. What I notice is: if we decide we want to throw a different exception, it is a change in one place.

PR #360, opened 2021-05-24
AS-723: Migrate WSM integration tests from JUnit to TestRunner

ungwudik:
**TL;DR - This PR will run TestRunner integration tests in dev, alpha, and staging.**

**The Nitty-Gritty:**

- Define and organize config files (server specifications, test suite configs, test file configs) for the different environments accordingly. Note that there's a plan to consolidate at least the test file configs for easier maintenance. This will come after some required updates in TestRunner https://broadworkbench.atlassian.net/browse/QA-1460

- Update integration/render-config.sh to accommodate different environments.

- Create a new workflow for running TestRunner integration tests only, and run TestRunner integration tests on dev, alpha, and staging.

- William Thunderlord is having some issues. Reference ticket: https://broadworkbench.atlassian.net/browse/PF-746. Replace him with Hermione.
Future work - Hermione is probably used used in other non-WSM tests, but she has access to necessary Data Repo reference stuff. Consider configuring WSM-dedicated users (bella, elijah, or liam) to gain access to Data Repo stuff so we can use those across the board.

- The nightly WSM JUnit integration test that's part of the Jenkins release train will be replaced by the TestRunner tests. I have a stand-by PR in [dsp-jenkins](https://github.com/broadinstitute/dsp-jenkins/pull/528) that will update the [jenkins job](https://fc-jenkins.dsp-techops.broadinstitute.org/job/workspacemanager-integration-test-trigger/).

- Disable the JUnit integration tests (Actual deletion of files and README update will come later. Let's get everything running smoothly for a few days or so before proceeding).

- Continue to run nightly-tests on `wsmtest` environment, and pr-integration tests on `local` environment

- Continue to run Unit and Connected tests as usual

- Note: In the meantime, some tests may leave Stairway flights behind in the environment after a failure, and may require manual cleanup/intervention. An automated cleanup plan is being addressed in [PF-730](https://broadworkbench.atlassian.net/browse/PF-730)

- Finally - This PR has a lot of file changes, but about 80% of them are config updates including files that were simply moved/re-organized. But this PR is a big change, regardless.


ungwudik:
Nit: Note that all check have passed. The very last commit changed only a github workflow, and we don't run/re-run everything when the changes are in github workflow, READMEs, etc.

See the prior workflow run here https://github.com/DataBiosphere/terra-workspace-manager/pull/360/checks?sha=d8856913c26f741cf197f4290eb31591d1b0e678

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,110 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Integration Tests
```
ddietterich, line 4: This is too generic a name. We have several actions that are running integration tests.
The `pr-integration` workflow and the `nightly-tests` workflow both run them.
Let's come up with something more specific, like `release-promotion-tests`

 Does any process run this on `dev`? It is not part of our regular testing in WSM eng.

.github/workflows/nightly-tests.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         uses: ./.github/actions/integration-test
         with:
           test-server: workspace-wsmtest.json
-          test: suites/FullIntegration.json
+          test: suites/wsmtest/FullIntegration.json
```
ddietterich, line 56: Why do these need to be put into a subdirectory as opposed to being renamed in place?
We need a different name. `wsmtest` is a personal environment, not a kind of test suite. These tests are run in may environments. Not just wsmtest.

ddietterich: It is even stranger, because you have configs for dev, alpha, staging, but you add wsmtest for suites?!
I think this scheme is unmaintainable.

ddietterich (changes requested):
I would like a couple of name changes.

I need more context. If this is a short-term change until we get parameterMaps, then I would like to _not_ reorganize the configs and suites directories, because we won't want them when we have a proper solution.

If this is intended to be a long-term change, then I think we need to take a step back and reconsider. This layout is unmaintainable. A single new test might need to be added in 9 places.

ungwudik:
> I would like a couple of name changes.
> 
> Re-reading the summary, I see that this is a short-term change until we get parameterMaps. Given that, I would like to _not_ reorganize the configs and suites directories, because we won't want them when we have a proper solution.
> 
> If `parameterMap` is not enough to capture what we need for variability, then we need to add features to testrunner to capture that. It is not maintainable to have nearly duplicate definitions that need to be updated in sync.

Yes, it's a short-term change.
Also - Yes, `parameterMap` won't be enough to address parameterizing different configs for different environments. I had added that thought as a comment in [QA-1460](https://broadworkbench.atlassian.net/browse/QA-1460), but I've now updated the ticket description itself for better visibility.
The reason from breaking the files down into different subdirectories is that: Those files need different values, and there doesn't seem an apparent way to combine them in the meantime without overcomplicating things. Definitely plan to consolidate as soon as [QA-1460](https://broadworkbench.atlassian.net/browse/QA-1460) is done.

.github/workflows/integration-tests.yml
ungwudik: This [workspacemanager-integration-test-trigger](https://fc-jenkins.dsp-techops.broadinstitute.org/job/workspacemanager-integration-test-trigger/) jenkins job is kicked off after WSM is deployed to dev. Although worth noting that QA/release train mostly looks at only alpha and staging runs.

ddietterich: Just for my information, can you explain who runs that jenkins job on dev? Right now when we merge to the dev branch, the merge script writes to a repo and argo deploys that version to dev. I didn't think there was a tie in to Jenkins.

ungwudik: Sure. The `gke-deploy` job that apparently deploys WSM (and other apps) resides in Jenkins [here](https://fc-jenkins.dsp-techops.broadinstitute.org/job/gke-deploy/). DevOps can speak better to if/how `gke-deploy` is tied to argo, but there's a shell script in that job that runs the project's docker image.

After deploy, the `gke-deploy` then triggers WSM tests as part of release promotion.

I believe there are plans to move the `gke-deploy` job to GitHub actions in the future, though.

ungwudik:
> > I would like a couple of name changes.
> > Re-reading the summary, I see that this is a short-term change until we get parameterMaps. Given that, I would like to _not_ reorganize the configs and suites directories, because we won't want them when we have a proper solution.
> > If `parameterMap` is not enough to capture what we need for variability, then we need to add features to testrunner to capture that. It is not maintainable to have nearly duplicate definitions that need to be updated in sync.
> 
> Yes, it's a short-term change.
> 
> Also - Yes, `parameterMap` won't be enough to address parameterizing different configs for different environments. I had added that thought as a comment in [QA-1460](https://broadworkbench.atlassian.net/browse/QA-1460), but I've now updated the ticket description itself for better visibility.
> 
> The reason for breaking the files down into different subdirectories is that: Those files need different values, and there doesn't seem an apparent way to combine them in the meantime without overcomplicating things. Definitely plan to consolidate as soon as [QA-1460](https://broadworkbench.atlassian.net/browse/QA-1460) is done.


Overall, I think we can further raise the priority of QA-1460, and communicate further via TODOs/comments/sync-meeting so everyone is aware of this change and the parts that are for the short-term.
Another other option is to hold off on merging this PR until QA-1460 is complete, but I'll really like to proceed with this workaround, unless anyone feels strongly against it.
I'm also open to other ideas.

ungwudik:
> I suggest moving the contents of the `wsmtest` back to where it was, but leave the other subdirectories. That will minimize change to the existing GHAs. Also, when Test Runner is able to express per-environment parameters, that is where we will centralize the finished product.

That makes sense. Done!

ddietterich (approved):
LGTM, thanks!

ansingh7115 (approved):
LGTM!

PR #361, opened 2021-05-25
[PF-633][PF-735] Remove bucket and BQ project level permissions

zloery:
This change will make WSM stop granting most project level permissions for GCS buckets and BigQuery datasets in new projects. The major exception is `roles/bigquery.jobUser`. This grants permission to run BigQuery jobs, which was previously included with `roles/viewer`. This role must be granted at the project level, rather than the individual dataset level.

Removing `roles/viewer` will remove a ton of over-granted permissions across GCP. All controlled resources have appropriate READER roles and I think the remaining lifescience and serviceusage roles should cover existing usecases, but we should keep an eye out if this causes unexpected breakage for users.

As always there is no backfill mechanism, so users in existing projects will still be overpermissioned.

Leaving this as a draft until we have a green light on not breaking early users.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
@@ -202,18 +202,23 @@ public class ControlledGcsBucketLifecycle extends WorkspaceAllocateTestScriptBas
     assertEquals(HttpStatusCodes.STATUS_CODE_FORBIDDEN, readerCannotDeleteBucket.getCode());
     logger.info("Failed to delete bucket {} directly as reader as expected", bucketName);
 
-    // TODO(PF-633): Owners and writers can actually delete buckets due to workspace-level roles
-    //  included as a temporary workaround. This needs to be removed as we transition onto WSM
-    //  controlled resources.
+    // Writer also cannot delete the bucket directly
+    StorageException writerCannotDeleteBucket =
+        assertThrows(
+            StorageException.class,
+            () -> readerStorageClient.get(bucketName).delete(),
+            "Workspace writer was able to delete a bucket directly!");
+    assertEquals(HttpStatusCodes.STATUS_CODE_FORBIDDEN, writerCannotDeleteBucket.getCode());
+    logger.info("Failed to delete bucket {} directly as writer as expected", bucketName);
 
-    // // Owner also cannot delete the bucket directly
-    // StorageException ownerCannotDeleteBucket =
-    //     assertThrows(
-    //         StorageException.class,
-    //         () -> ownerStorageClient.get(bucketName).delete(),
-    //         "Workspace owner was able to delete a bucket directly!");
-    // assertEquals(HttpStatusCodes.STATUS_CODE_FORBIDDEN, ownerCannotDeleteBucket.getCode());
-    // logger.info("Failed to delete bucket {} directly as owner as expected", bucketName);
+    // Owner also cannot delete the bucket directly
+    StorageException ownerCannotDeleteBucket =
+        assertThrows(
+            StorageException.class,
+            () -> ownerStorageClient.get(bucketName).delete(),
+            "Workspace owner was able to delete a bucket directly!");
+    assertEquals(HttpStatusCodes.STATUS_CODE_FORBIDDEN, ownerCannotDeleteBucket.getCode());
+    logger.info("Failed to delete bucket {} directly as owner as expected", bucketName);
 
     // Owner can delete the bucket through WSM
     ResourceMaker.deleteControlledGcsBucket(resourceId, getWorkspaceId(), resourceApi);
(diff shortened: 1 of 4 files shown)


integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
```diff
@@ -202,18 +202,23 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
...
-    //         StorageException.class,
-    //         () -> ownerStorageClient.get(bucketName).delete(),
-    //         "Workspace owner was able to delete a bucket directly!");
-    // assertEquals(HttpStatusCodes.STATUS_CODE_FORBIDDEN, ownerCannotDeleteBucket.getCode());
-    // logger.info("Failed to delete bucket {} directly as owner as expected", bucketName);
+    // Writer also cannot delete the bucket directly
+    StorageException writerCannotDeleteBucket =
+        assertThrows(
+            StorageException.class,
+            () -> readerStorageClient.get(bucketName).delete(),
```
marikomedlock, line 209: Should this be `writerStorageClient`?

integration/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -124,15 +124,14 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
...
-    // TODO(PF-633): workspace owners can write to private buckets via "roles/storage.admin".
-    // // Owner cannot write object to bucket
-    // StorageException ownerCannotWritePrivateBucket =
-    //     assertThrows(
-    //         StorageException.class,
-    //         () -> ownerStorageClient.create(blobInfo, GCS_BLOB_CONTENT.getBytes()),
-    //         "Workspace owner was able to write to private bucket");
-    // assertEquals(HttpStatusCodes.STATUS_CODE_FORBIDDEN, ownerCannotWritePrivateBucket.getCode());
-    // logger.info("Workspace owner cannot write to private resource as expected");
+    // Owner cannot write object to bucket
```
marikomedlock, line 127: I thought owners (and other workspace users) should be able to write to their own private bucket. And so I think this is testing whether the owner can write to a bucket that is private to another user. Is that correct? If so, I think just spelling that out a bit more in the comment would be helpful.

integration/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -152,26 +151,23 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
...
+    // Owner cannot read the bucket contents
+    StorageException ownerCannotReadPrivateBucket =
+        assertThrows(
+            StorageException.class,
+            () -> ownerStorageClient.get(blobId),
+            "Workspace owner was able to read private bucket");
+    assertEquals(HttpStatusCodes.STATUS_CODE_FORBIDDEN, ownerCannotReadPrivateBucket.getCode());
+    logger.info("Workspace owner cannot read from private bucket as expected");
+
+    // Workspace reader also cannot read private bucket contents
```
marikomedlock, line 163: Just curious, in this class I see you're testing owners and readers -- how come not writers also?

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -14,20 +14,22 @@
  * CustomGcpIamRoleMapping}), and should be removed from this list.
  */
 public class CloudSyncRoleMapping {
-  private static final List<String> READER_PERMISSIONS = ImmutableList.of("roles/viewer");
```
marikomedlock, line 17: Checking my understanding here: We no longer need the generic `viewer` role for readers because they will be given read access on individual resources instead, yes?

integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
zloery: I'm not quite sure what I was thinking here - there is no writer in this test, just the workspace reader and owner. good catch!

integration/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
zloery: Yep, you've got it exactly. Clarified the comments a bit

integration/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
zloery: No strong reason. Workspace owners already have a superset of writer permission so I think it would be fairly low signal and the test is reasonably long already, but that's pretty much it.

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
zloery: Yep, exactly. We no longer need project level roles for buckets/datasets/notebooks, so I replaced the broad `roles/viewer` with a much narrower set of permissions covering the pieces we still need/haven't replaced.

marikomedlock (approved):

ddietterich (approved):
lgtm

PR #362, opened 2021-05-25
taskInfo plugin

jaycarlton:
[Plugin](https://gitlab.com/barfuin/gradle-taskinfo) to show task dependencies as a tree and ordering. Supports JSON output for downstream analysis (like a graphviz translator).

I added it to the top level gradle file, since there's no reason to prefer one of the sub-projects.

[Example output](https://gist.github.com/jaycarlton/fe1c15b6723ee22bab33f635ad384a20)
Screen Shot 2021-05-25 at 7 33 58 PM
Screen Shot 2021-05-25 at 7 42 29 PM

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,3 +3,8 @@
 buildscript {
     repositories { mavenCentral() }
 }
+
+plugins {
+    // https://gitlab.com/barfuin/gradle-taskinfo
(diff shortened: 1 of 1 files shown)


ddietterich (approved):
Handy

zloery (approved):

PR #363, opened 2021-05-26
split up gradle file

jaycarlton:
Split the `service` gradle file into functional sections. Added a few comments and some style cleanup. Tested with the unit tests task.

Prior art in the common lib is [here](https://github.com/DataBiosphere/terra-common-lib/blob/develop/build.gradle#L113).


service/build.gradle
```diff
@@ -20,249 +20,30 @@ plugins {
...
+def includeDir = "$projectDir/gradle"
+// include order matters, so don't alphabetize (yet)
+apply(from: "$includeDir/buildScan.gradle")
+apply(from: "$includeDir/profiler.gradle")
+apply(from: "$includeDir/testing.gradle")
+apply(from: "$includeDir/deploy.gradle")
+apply(from: "$includeDir/lint.gradle")
+apply(from: "$includeDir/openapi.gradle")
+apply(from: "$includeDir/versionProperties.gradle")
+apply(from: "$includeDir/dependencies.gradle")
```
ddietterich, line 42: I'm not loving this. Is the file really so big that it needs to be exploded into 8 pieces? That won't make it easier for me anyway.

I wouldn't mind factoring out parts that are common. Like the spotless invocation and the jacoco and the buildscan are boilerplate.

The dependencies and tests are not.

service/build.gradle
```diff
@@ -20,249 +20,30 @@ plugins {
...
-
-// Generate/clean the version properties file
-// This was being done by a plugin: https://github.com/palantir/gradle-git-version
-// but it does not work with Gradle 7.0. It was also very complicated for a
-// what we need, so easily replaced with a bash script
-def generatedVersionFile = "${resourceDir}/generated/version.properties"
-
-task generateVersionProperties(type: Exec) {
-    commandLine "${projectDir}/writeVersionProperties.sh", "${version}", "${generatedVersionFile}"
+project.ext {
```
ddietterich, line 27: Maybe you can explain when to prefer `ext` and when to prefer `def`.
You changed to `ext` here, but then introduced a new `def` later.

service/build.gradle
jaycarlton: There's lots of room to massage this. @gjuggler has done some similar grouping in the common lib.

It's not about the size of the file so much as the number of different concerns that are in it. It's bewildering to me to see so many different aspects of the build all in one file. If it were a Java class, we'd surely split it up based on the single responsibility principle.

service/build.gradle
jaycarlton: I belive `def` is local to the current translation unit (file), and `ext` puts an `extra property` on the project itself, so things are visible across cuts.

jaycarlton:
Will do this again on another branch. too much has changed to pull in main now.

PR #364, opened 2021-05-26
PF-730 - clean database ahead of nightly tests

ddietterich:
This PR adds the ability to clean out the database tables before each of the nightly test runs. (Ah, it sounds so simple...)

As part of the work, I needed to collect a bunch of additional information from vault, so I took the first step for PF-744 and created `scripts/write-config.sh` as a replacement for `service/render_config.sh`. I added a `write-config` GHA to run the script in workflows.

My original idea was to drop all of the tables in the databases including the liquibase ones and then restart the pods. I still think that would be the better approach. However, I don't have a straightforward way to restart the pods. There might be a way to do it from Test Runner. So something to look at down the road. I will make a separate ticket for that.

The approach I took is to TRUNCATE the data tables and leave the liquibase tables intact. Then I avoid needing to restart the pods. Postgres has special semantics for the TRUNCATE statement: if you mention all of the tables that have foreign key constraints, it does no scans and just immediately frees all of the storage.

One downside of this approach is that if WSM or Stairway adds tables, we will have to update the script to truncate those tables.

**Notes about the revamp:** UC pointed out the ugliness (my words) of the large repeated code in the `nightly-tests` workflow. Because github composite actions cannot call other actions, there is tension between making small, reusable actions and making larger, more custom actions. I realized that I could extract the action code into scripts. Those scripts would be handy to have for developer use as well as for the github workflows. So I made three scripts:
- `launch-sql-proxy` - reads the config info to launch the dockerized cloud sql proxy
- `run-sql-script` - launches the proxy and then runs psql to execute a script; shuts down the proxy on exit
- `connect-sql` - launches the proxy and then runs psql for interactive querying; shuts down the proxy on exit. This one means you can trivially connect to one of the CloudSQL databases using the info generated from the `write-config` script.

I made the `clean-databases` github action that runs the `run-sql-script` twice to clean both databases. That makes the `nightly-tests` workflow much tidier.


.github/workflows/nightly-tests.yml
```diff
@@ -47,6 +47,33 @@ jobs:
...
+
+      - name: truncate tables in the wsm db
+        uses: ./.github/actions/run-sql-script
+        with:
+          db-connection-name: ${{ steps.config.outputs.db-connection-name }}
+          db-name:     ${{ steps.config.outputs.db-name }}
+          db-password: ${{ steps.config.outputs.db-password }}
+          db-username: ${{ steps.config.outputs.db-username }}
+          cloudsql-sa-file: ./config/sqlproxy-sa.json
+          script-file: ./scripts/clean_wsm.sql
```
ungwudik, line 65: Looks like these lines are repeated two more times in this workflow. Could they be consolidated somewhere and reused? Same with `- name: truncate tables in the wsm stairway db` step

scripts/write-config.sh
```diff
@@ -0,0 +1,206 @@
...
+  <vaulttoken> defaults to the token found in ~/.vault-token.
+
+  <outputdir> defaults to "config/", so when run in the gradle rootdir, it will be
+  in the expected place for automation.
+EOF
+ exit 1
+}
+
+# Get the inputs with defaulting
+default_target=${WSM_WRITE_CONFIG:-local}
```
ungwudik, line 74: Nit: I think there's a mismatch between the envvar name here (`WSM_WRITE_CONFIG`) and the envvar name alluded to in the documentation (`WSM_WRITE_CONFIG_TARGET`)

.github/actions/run-sql-script/action.yml
```diff
@@ -0,0 +1,49 @@
...
+      creddir=$(cd "${safile%/*}" && pwd)
+      credfile="${safile##*/}"
+      # We can parameterize this if need be, but this should work
+      port=5434
+      echo "Launching the CloudSQL proxy"
+      dod=$(docker run --rm -d \
+        -v ${creddir}:/config \
+        -p 127.0.0.1:${port}:${port} \
+        gcr.io/cloudsql-docker/gce-proxy:latest \
+        /cloud_sql_proxy \
```
jaycarlton, line 39: how is `cloud_sql_proxy` downloaded/installed?

.github/actions/run-sql-script/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    run: |
+      safile="${{ inputs.cloudsql-sa-file }}"
+      creddir=$(cd "${safile%/*}" && pwd)
+      credfile="${safile##*/}"
+      # We can parameterize this if need be, but this should work
+      port=5434
+      echo "Launching the CloudSQL proxy"
+      dod=$(docker run --rm -d \
+        -v ${creddir}:/config \
+        -p 127.0.0.1:${port}:${port} \
```
jaycarlton, line 37: I'd make host ip and port parameters with these values as defaults. The  IP is in at least two places.

.github/actions/run-sql-script/action.yml
```diff
@@ -0,0 +1,49 @@
...
+  db-password:
+    description: 'password to login to the postgres database'
+    required: true
+  db-username:
+    description: 'username to login to the postgres database'
+    required: true
+  cloudsql-sa-file:
+    description: 'path to the CloudSQL SA json file somewhere'
+    required: true
+  script-file:
```
jaycarlton, line 20: maybe name this `sql-script-path`

.github/actions/run-sql-script/action.yml
```diff
@@ -0,0 +1,49 @@
...
+        gcr.io/cloudsql-docker/gce-proxy:latest \
+        /cloud_sql_proxy \
+        -instances=${{ inputs.db-connection-name }}=tcp:0.0.0.0:${port} \
+        -credential_file=/config/${credfile})
+      sleep 5
+      echo "Running the script via psql"
+      PGPASSWORD="${{ inputs.db-password }}" \
+        psql "host=127.0.0.1 port=$port sslmode=disable dbname=${{ inputs.db-name }} user=${{ inputs.db-username }}" \
+        -f ${{ inputs.script-file }}
+      sleep 2
```
jaycarlton, line 47: might need to parameterize this as other scripts could take longer. Is there not an option for `psql` and the container to self destruct upon completion of the script?

.github/actions/run-sql-script/action.yml
```diff
@@ -0,0 +1,49 @@
+name: 'run-sql-script'
+description: 'Run a script against a CloudSQL instance'
```
jaycarlton, line 2: document that this only works for postgres

.github/actions/write-config/action.yml
```diff
@@ -0,0 +1,43 @@
+name: 'write-config'
```
jaycarlton, line 1: maybe name it `populate-test-credentials` or something. Write config seems too generic.

.github/actions/write-config/action.yml
```diff
@@ -0,0 +1,43 @@
...
+description: 'Collect the vault information used for testing'
+author: 'dd'
+inputs:
+  vault-token:
+    description: 'Vault token'
+    required: true
+  target:
```
jaycarlton, line 8: maybe `target_namespace`

.github/actions/write-config/action.yml
```diff
@@ -0,0 +1,43 @@
...
+    description: 'Target environment: local, dev, alpha, staging, wsmtest or another personal namespace'
+    default: 'local'
+    required: true
+outputs:
+  db-connection-name:
+    value: ${{ steps.files-to-outputs.outputs.db-connection-name }}
+  db-name:
+    value: ${{ steps.files-to-outputs.outputs.db-name }}
+  db-password:
+    value: ${{ steps.files-to-outputs.outputs.db-password }}
```
jaycarlton, line 18: will this show up in the GitHub UI? I wonder if there's a way to mask it if so.

.github/actions/write-config/action.yml
```diff
@@ -0,0 +1,43 @@
...
+      id: files-to-outputs
+      run: |
+        echo "::set-output name=db-connection-name::$(<./config/db-connection-name.txt)"
+        echo "::set-output name=db-name::$(<./config/db-name.txt)"
+        echo "::set-output name=db-password::$(<./config/db-password.txt)"
+        echo "::set-output name=db-username::$(<./config/db-username.txt)"
+        echo "::set-output name=stairway-db-name::$(<./config/stairway-db-name.txt)"
+        echo "::set-output name=stairway-db-password::$(<./config/stairway-db-password.txt)"
+        echo "::set-output name=stairway-db-username::$(<./config/stairway-db-username.txt)"
+      shell: bash
```
jaycarlton, line 43: do we need to clean these sensitive files up afterwards?

.github/actions/run-sql-script/action.yml
```diff
@@ -0,0 +1,49 @@
...
+author: 'dd'
+inputs:
+  db-connection-name:
+    description: 'full CloudSQL connection name: <project>:<region>:<cloudsql-instance>'
+    required: true
+  db-name:
```
jaycarlton, line 8: Might be good to put a guard up so that it doesn't run against prod unless `damnTheTorpedoes` is also true, or something.

.github/workflows/nightly-tests.yml
jaycarlton: I think this is like a function call with a bunch of arguments. It's only squashable if we have default arguments, and I don't know of any that would make sense here. But yeah, definitely reuse if you can, even if it's with a yaml anchor.

.github/workflows/nightly-tests.yml
```diff
@@ -56,6 +83,28 @@ jobs:
           test: suites/FullIntegration.json
           vault-token: ${{ steps.vault-token-step.outputs.vault-token }}
 
+      - name: truncate tables in the wsm db
```
jaycarlton, line 86: So we're cleaning before and after? I guess add (post-test) to the name to avoid confusion.

It might be helpful to have a flag to skip the cleanup in case you need to debug DB state.

.github/workflows/nightly-tests.yml
```diff
@@ -78,9 +149,9 @@ jobs:
         if: always()
         id: status-message
         run: |
-          L2="Integration: ${{ steps.integration-test.outcome }} "
+          L2="Int: ${{ steps.integration-test.outcome }} "
```
jaycarlton, line 152: I think we need a separator like `  |  ` between each entry for readability in Slack.

.github/workflows/pr-integration.yml
```diff
@@ -73,10 +73,12 @@ jobs:
         echo ::set-output name=vault-token::$VAULT_TOKEN
         echo ::add-mask::$VAULT_TOKEN
 
-    - name: Render config
-      uses: ./.github/actions/render-config
+    - name: Write config
```
jaycarlton, line 76: "Write account config"? Something more specific would help.

.github/workflows/workflow-tester.yml
```diff
@@ -34,3 +34,30 @@ jobs:
         echo ::set-output name=vault-token::$VAULT_TOKEN
         echo ::add-mask::$VAULT_TOKEN
 
+    - name: Write config
```
jaycarlton, line 37: I thought this was a no-op workflow. I'd expect it to print hello and stop or something. Was this block from testing?

DEVELOPMENT.md
```diff
@@ -120,6 +120,43 @@ At some point, we will connect this to a CloudSQL instance but for local dev pur
...
+Usage: [<target>] [<vaulttoken>] [<outputdir>]"
+
+  <target> can be:
+    local - for testing against a local (bootRun) WSM
+    dev - uses secrets from the dev environment
+    alpha - alpha test environment
+    staging - release staging environment
+    help or ? - print this help
+    clean - removes all files from the output directory
+    * - anything else is assumed to be a personal environment using the terra-kernel-k8s
```
jaycarlton, line 147: nit: this doc is giving internal configuration info in a public repo

scripts/write-config.sh
```diff
@@ -0,0 +1,206 @@
...
+#
+# We want to use this in a gradle task, so it takes arguments both as command line
+# options and as envvars. For automations, like GHA, the command line can be specified.
+# For developer use, we can set our favorite envvars and let gradle ensure the directory is properly populated.
+#
+# The environment passed in is used to configure several other parameters, including the target for running
+# the integration tests.
+#
+# For personal environments, we assume that the target name is the same as the personal namespace name.
+# The output directory includes the following files:
```
jaycarlton, line 18: You could use a single JSON file and parse it with `jq` if it's available in the GHA environment.

scripts/write-config.sh
```diff
@@ -0,0 +1,206 @@
...
+#   --------------------------+-------------------------------------------------------------------------
+#   buffer-client-sa.json     | Creds to access Buffer service
+#   --------------------------+-------------------------------------------------------------------------
+#   db-connection-name.txt    | Connection string for CloudSQL proxy
+#   --------------------------+-------------------------------------------------------------------------
+#   db-name.txt               | Database name
```
jaycarlton, line 24: nit: to make it parallel with `stairway-db-name.txt`, you could make these `wm-db-name.txt`

scripts/write-config.sh
```diff
@@ -0,0 +1,206 @@
...
+# Get the inputs with defaulting
+default_target=${WSM_WRITE_CONFIG:-local}
+target=${1:-$default_target}
+vaulttoken=${2:-$(cat "$HOME"/.vault-token)}
+outputdir=${3:-config}
+
+# The vault paths are irregular, so we map the target into three variables:
+# k8senv    - the kubernetes environment: alpha, staging, or integration
+# namespace - the namespace in the k8s env: alpha, staging, dev, or the target for personal environments
+# fcenv     - the firecloud delegated service account environment: dev, alpha, stagin
```
jaycarlton, line 82: *staging?

.github/workflows/nightly-tests.yml
ddietterich: I want to have the re-usable run-sql-script action. Once I do that, I get stuck because I cannot call an action from inside an action so I don't have a good way to group this. (Although github says they are working on that feature).

I also didn't want to put the cleaning operation _inside_ the `integration-test` action, because we only want to clean in very specific circumstances. That left me with the way you see it here. Suggestions welcome...

scripts/write-config.sh
ddietterich: Good catch!

scripts/write-config.sh
```diff
@@ -0,0 +1,206 @@
+#!/bin/bash
```
jaycarlton, line 1: When I have to read more than 6 lines of bash, my rate doubles 😄

scripts/write-config.sh
```diff
@@ -0,0 +1,206 @@
...
+    fil=${outputdir}/dbtmp.json
+    docker run --rm -e VAULT_TOKEN=${vaulttoken} broadinstitute/dsde-toolbox:consul-0.20.0 \
+        vault read -format=json ${vaultpath} | jq -r .data > ${fil}
+    cat ${fil} | jq -r '.db' > ${outputdir}/${fileprefix}-name.txt
+    cat ${fil} | jq -r '.password' > ${outputdir}/${fileprefix}-password.txt
+    cat ${fil} | jq -r '.username' > ${outputdir}/${fileprefix}-username.txt
+    rm ${fil}
+}
+
+vaultget "secret/dsde/firecloud/${fcenv}/common/firecloud-account.json" "user-delegated-sa.json"
```
jaycarlton, line 167: Now we're publishing details about the organization of our vault setup. From an OpSec perspective, maybe this script should live in a private repo?

scripts/write-config.sh
```diff
@@ -0,0 +1,206 @@
...
+vaultgetb64 "secret/dsde/terra/kernel/${k8senv}/${namespace}/workspace/app-sa" "wsm-sa.json"
+
+# The magic tokens for access to Buffer Service, Janitor, and TestRunner are in fixed places
+vaultgetb64 "secret/dsde/terra/kernel/integration/tools/crl_janitor/client-sa" "janitor-client-sa.json"
+vaultgetb64 "secret/dsde/terra/kernel/integration/tools/buffer/client-sa" "buffer-client-sa.json"
+vaultgetb64 "secret/dsde/terra/kernel/integration/common/testrunner/testrunner-sa" "testrunner-sa.json"
+
+# Test runner K8s configuration
+# The setup for this is inconsistent. The usage is unknown. So we will skip it for now.
```
jaycarlton, line 176: ticket for this work?

service/src/test/resources/application-test.yml
```diff
@@ -1,6 +1,6 @@
 workspace:
   crl:
-    janitor-client-credential-file-path: rendered/janitor-client-sa-account.json
+    janitor-client-credential-file-path: ../config/janitor-client-sa.json
```
jaycarlton, line 3: I still think `config` is where I would look for all kinds of parameters and constants. This directory is more of a `credentials` or something.

jaycarlton (approved):
This is a big step forward. Thanks for all the actioning.

.github/workflows/nightly-tests.yml
ddietterich: OTOH, if I give up on the reusable action and assume I have the config directory, it is simpler. And maybe reusable in a different way.

ddietterich:
UC's comment got me to thinking. I've decided to restructure a bit, moving some the low-level run-sql-script action into a set of script files. Then will use the action to run the scripts.

The big advantage of this approach is that the scripts are usable outside of the GHA. Very convenient for debugging. Hoping to have the new version up for review tomorrow.

.github/actions/run-sql-script/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    description: 'name of the database to connect to'
+    required: true
+  db-password:
+    description: 'password to login to the postgres database'
+    required: true
+  db-username:
+    description: 'username to login to the postgres database'
+    required: true
+  cloudsql-sa-file:
```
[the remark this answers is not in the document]
ddietterich: You are right. The username/password is for logging into the CloudSQL database. That is separate from service accounts.

.github/actions/run-sql-script/action.yml
ddietterich: The docker application downloads the docker image. That particular docker image is maintained by Google and has the cloud_sql_proxy installed.

.github/actions/run-sql-script/action.yml
ddietterich: I need to clean up the running docker, not the psql.
This is done more reliably in the script rewrite I did.

scripts/write-config.sh
ddietterich: Tell me about it. I have to re-learn it every time I work on it. And then Codacy shows up and tells me I did it wrong :(

scripts/write-config.sh
ddietterich: This is not new behavior. If it is an issue, it is an issue across all of the repos.

jaycarlton: I mean...we're making public details that are of no use to anyone else except for an attacker. That's my threshold for keeping things private. But yes, easier said than done.

.github/workflows/pr-integration.yml
```diff
@@ -75,11 +75,13 @@ jobs:
...
+      id: config
+      uses: ./.github/actions/write-config
       with:
         test-env: dev
```
ungwudik, line 82: Looks like this line can be removed...since `./.github/actions/write-config` no longer uses `test-env` as input.

.github/workflows/release-promotion-tests.yml
```diff
@@ -79,8 +79,8 @@ jobs:
           echo ::set-output name=vault-token::$VAULT_TOKEN
           echo ::add-mask::$VAULT_TOKEN
 
-      - name: Render config
-        uses: ./.github/actions/render-config
+      - name: Write configuration
+        uses: ./.github/actions/write-config
         with:
           test-env: ${{ steps.set-env-step.outputs.test-env }}
```
ungwudik, line 85: Also need to update this step to supply `target` as input, instead of `test-env`

.github/workflows/test.yml
```diff
@@ -109,12 +109,14 @@ jobs:
...
         echo ::add-mask::$VAULT_TOKEN
 
-    - name: Render config
+    - name: Write config
       if: matrix.gradleTask != 'unitTest'
-      uses: ./.github/actions/render-config
+      id: config
+      uses: ./.github/actions/write-config
       with:
         test-env: ${{ steps.set-env-step.outputs.config-env }}
```
ungwudik, line 117: Also remove `test-env` here.

FWIW - It doesn't actually seem to break this workflow. Message in GHA logs is `Warning: Unexpected input(s) 'test-env', valid inputs are ['vault-token', 'target']`

.github/workflows/pr-integration.yml
ddietterich: Thanks for catching this!

ungwudik (approved):

scripts/write-config.sh
```diff
@@ -0,0 +1,207 @@
...
+        ;;
+
+    local)
+        k8senv=integration
+        namespace=wsmtest
+        fcenv=dev
+        ;;
+
+    dev)
+        k8senv=integration
```
ichengchang, line 101: Is `k8senv=integration` the intended kubernetes environment for `dev` environment? Isn't `dev` refers to the `DSDE Development` project?

scripts/write-config.sh
```diff
@@ -0,0 +1,207 @@
...
+vaultgetb64 "secret/dsde/terra/kernel/${k8senv}/${namespace}/workspace/app-sa" "wsm-sa.json"
+
+# The magic tokens for access to Buffer Service, Janitor, and TestRunner are in fixed places
+vaultgetb64 "secret/dsde/terra/kernel/integration/tools/crl_janitor/client-sa" "janitor-client-sa.json"
+vaultgetb64 "secret/dsde/terra/kernel/integration/tools/buffer/client-sa" "buffer-client-sa.json"
+vaultgetb64 "secret/dsde/terra/kernel/integration/common/testrunner/testrunner-sa" "testrunner-sa.json"
+
+# Test runner K8s configuration
+# TODO (PF-744): The setup for this is inconsistent. The usage is unknown. So we will skip it for now.
+#vaultgetb64 "secret/dsde/terra/kernel/integration/${namespace}/testrunner-k8s-sa" "trtmp.json"
```
ichengchang, line 177: Should `integration` be `${k8senv}`?

scripts/write-config.sh
ddietterich: Thanks! I didn't realize dev was in a different project. So it follows the alpha and staging pattern.

scripts/write-config.sh
ddietterich: As far as I can see, the only testrunner-k8s-sa definitions are in integration. It looks like ichang, zloery, and wsmtest are the only places it is defined.

Do you plan to define it in other environments?

PR #365, opened 2021-05-26
PF-751 lazy initilize Sam user

jaycarlton:
Waiting for this user to be initialized during startup caused a situation where we were blocked from bringing the service up on Sam. Now, we defer the initialize call until after startup, at the first place we need it. This occurs when making the first controlled resource.


zloery (approved):

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -170,23 +173,31 @@ public static void rethrowIfSamInterrupted(VoidInterruptedSupplier function, Str
...
+        wsmAccessToken = getWsmServiceAccountToken();
+      } catch (IOException e) {
+        // In cases where WSM is not running as a service account (e.g. unit tests), the above call
+        // will throw. This can be ignored now and later when the credentials are used again.
+        logger.warn(
+            "Failed to register WSM service account in Sam. This is expected for tests.", e);
+        return;
+      }
+      UsersApi usersApi = samUsersApi(wsmAccessToken);
+      // If registering the service account fails, all we can do is to keep trying.
```
ddietterich, line 194: I don't think we should wait forever. This is now happening due to a REST API request.  I'd suggest doing this operation using the register operation using the sam retry code that wraps other sam calls so it has the same failure behavior.

ddietterich (commented):
I think we should re-use the sam retry and not retry forever.

ddietterich (approved):
LGTM

PR #366, opened 2021-05-26
QA-1482: Fixing test runner nightly test issues. Bump test runner version to 0.0.5-SNAPSHOT.

ichengchang:
The Test Runner 0.0.5-SNAPSHOT fix address the following issues:

Caching of SA credentials which caused instabilities in WSM tests when running multiple tests with a single action workflow.

Refactored `componentLabel` and `apiComponentLabel` properties (the associated k8s metadata label key/value pair used to identify a deployed service) into Cluster Specification. These properties are removed from Application Specification due to this refactoring and hence can be applied across all test config.

As part of this ticket, added more debug log for `DeleteInitialPods`.


ddietterich (approved):
LGTM

marikomedlock (approved):

PR #369, opened 2021-05-28
[PF-702] Add endpoint for validating referenced resource

zloery:
This change adds a new endpoint to validate that a user has access to a referenced resource (i.e. that they can actually from read the relevant TDR snapshot/cloud object). This behavior previously existed as part of creating a new referenced resource, but was not exposed directly.

As part of this change, I've slightly reworked how reference validation happens. I added a new abstract `validateReference` method to `ReferencedResource`, with a type-specific implementation in each referenced resource class. This should not affect serialization, as it does not modify or store additional state.


service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedBigQueryDatasetResource.java
```diff
@@ -106,6 +109,12 @@ public void validate() {
     ValidationUtils.validateBqDatasetName(getDatasetName());
   }
 
+  @Override
+  public boolean validateReference(FlightBeanBag context, AuthenticatedUserRequest userReq) {
```
jaycarlton, line 113: nit: I think of functions starting with `validate` as void functions that throw on invalid. I wonder if there's a better name scheme for these boolean-valued ones.

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
```diff
@@ -123,4 +129,13 @@ public ReferencedResource getReferenceResourceByName(
         userReq, workspaceId, SamConstants.SAM_WORKSPACE_READ_ACTION);
     return resourceDao.enumerateReferences(workspaceId, offset, limit);
   }
+
+  public boolean validateReference(
+      UUID workspaceId, UUID resourceId, AuthenticatedUserRequest userReq) {
+    workspaceService.validateWorkspaceAndAction(
+        userReq, workspaceId, SamConstants.SAM_WORKSPACE_READ_ACTION);
+    ReferencedResource referencedResource =
+        resourceDao.getResource(workspaceId, resourceId).castToReferenceResource();
```
jaycarlton, line 138: nit: should rename this to `castToReferencedResource`

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -312,7 +312,26 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
-  ## END OF DEPRECATED PATHS ##
+
+  /api/workspaces/v1/{workspaceId}/resources/referenced/{resourceId}:
```
jaycarlton, line 316: I'm confused by this path. I thought it would have `validate` or something in it. Where do I go if I just want to get a referenced resource?

service/src/test/java/bio/terra/workspace/service/resource/referenced/ReferenceResourceServiceTest.java
```diff
@@ -228,6 +228,9 @@ void testDataRepoReference() {
           resultReferenceResource.castToDataRepoSnapshotResource();
       assertThat(resource, equalTo(resultResource));
 
+      referenceResourceService.validateReference(
```
jaycarlton, line 231: *referenced here and in filename

service/src/test/java/bio/terra/workspace/service/resource/referenced/ReferenceResourceServiceTest.java
```diff
@@ -350,6 +353,9 @@ void testGcsBucketReference() {
             resultReferenceResource.castToGcsBucketResource();
         assertThat(resource, equalTo(resultResource));
 
+        referenceResourceService.validateReference(
```
jaycarlton, line 356: do you not need to check the return value of these?

jaycarlton (approved):

service/src/main/resources/api/service_openapi.yaml
marikomedlock: I don't think you can just fetch a single resource without knowing its type beforehand. You can enumerate all the resources, or call the type-specific endpoint.

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
```diff
@@ -123,4 +129,13 @@ public ReferencedResource getReferenceResourceByName(
         userReq, workspaceId, SamConstants.SAM_WORKSPACE_READ_ACTION);
     return resourceDao.enumerateReferences(workspaceId, offset, limit);
   }
+
+  public boolean validateReference(
```
marikomedlock, line 133: nit: Do you need "Reference" in the method name, since it's already in the class name?

service/src/test/java/bio/terra/workspace/service/resource/referenced/ReferenceResourceServiceTest.java
marikomedlock: +1 here. I think you either need to check the return value or change `validateReference` to throw an exception if it doesn't exist instead.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -312,7 +312,26 @@ paths:
...
+      tags: [Resource]
+      responses:
+        '200':
+          $ref: '#/components/responses/ValidateReferenceResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '404':
+          $ref: '#/components/responses/NotFound'
```
marikomedlock, line 332: Can you point me where these 403 and 404 errors are thrown? In the `ResourceController` class, it looks like it's returning a boolean if it exists or not. I'm not clear on when to expect a `false` vs one of these codes.

service/src/main/resources/api/service_openapi.yaml
jaycarlton: Oh, interesting. Good point.

service/src/main/java/bio/terra/workspace/app/controller/ResourceController.java
```diff
@@ -95,6 +99,14 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     return new ResponseEntity<>(apiResourceList, HttpStatus.OK);
   }
 
+  @Override
+  public ResponseEntity<Boolean> validateReference(UUID workspaceId, UUID resourceId) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    boolean isValid =
+        referencedResourceService.validateReference(workspaceId, resourceId, userRequest);
+    return new ResponseEntity<>(isValid, HttpStatus.OK);
```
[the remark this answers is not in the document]
zloery: This is a JSON object - it's just a single boolean field instead of a complex object. I could wrap it in another object, but I'm not sure that really adds anything.

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedBigQueryDatasetResource.java
zloery: Yeah - at one point that's what this function did, but I thought it might be more useful to have an output. Renamed to `checkReferenceAccess` / `checkAccess`.

service/src/main/resources/api/service_openapi.yaml
zloery: ^ yep, this was part of the big discussion around generic vs. resource-specific return types from the API. That said, we might consider supporting a generic GET endpoint in the future, so it's probably prudent to add `/access` here. Does the new URL seem reasonable?

service/src/test/java/bio/terra/workspace/service/resource/referenced/ReferenceResourceServiceTest.java
zloery: `referenceResource` gets used in a lot of places (var names in this file, method names in ReferencedResourceService, etc). I filed PF-828 to fix this everywhere - doing this as a separate change to save this PR from too many unrelated changes

service/src/test/java/bio/terra/workspace/service/resource/referenced/ReferenceResourceServiceTest.java
zloery: done

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
zloery: renamed without `reference` in the name

service/src/main/resources/api/service_openapi.yaml
zloery: Yep, this comes from the call to `validateWorkspaceAndAction` ([link](https://github.com/DataBiosphere/terra-workspace-manager/blob/zl-pf702/service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java#L115)) inside `checkAccess`, which is common set of checks before a lot of WSM endpoints.

That will throw a 404 if the workspace doesn't exist, or a 403 if it does exist and the user doesn't have the right permission (in this case, at least read access). If the workspace does exist and the user does have read access, the response should be a `true/false` (or any unexpected error from the cloud/TDR)

service/src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -147,10 +148,10 @@ public BigQueryCow createWsmSaBigQueryCow() {
    * @param userRequest auth info
    * @return true if the dataset exists
```
marikomedlock, line 149: nit: true if the user can list the dataset tables

service/src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -207,10 +208,15 @@ private StorageCow createStorageCowWorker(
    * @param userRequest auth info
    * @return true if the bucket exists
```
marikomedlock, line 209: nit: true if the user has storage.objects get or list permissions

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Ah okay, so the error codes apply to the workspace and the response code applies to the resource. Makes sense.

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedDataRepoSnapshotResource.java
```diff
@@ -103,6 +106,12 @@ public void validate() {
     }
   }
 
+  @Override
+  public boolean checkAccess(FlightBeanBag context, AuthenticatedUserRequest userReq) {
+    DataRepoService dataRepoService = context.getDataRepoService();
+    return dataRepoService.snapshotExists(instanceName, snapshotId, userReq);
```
marikomedlock, line 112: It looks like this is doing something different than `checkAccess` for other resource types (exists vs canRead). Would it make sense for this to return false if the snapshot is not found or not accessible by the user, similar to the other resource types?

Note that the CLI does not currently support TDR snapshots, so this won't affect the usage of this endpoint right now.

service/src/test/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceServiceTest.java
```diff
@@ -320,7 +325,7 @@ void testEnumerate() {
       @BeforeEach
       void setup() throws Exception {
         // Make the Verify step always succeed
-        doReturn(true).when(mockCrlService).gcsBucketExists(any(), any());
+        doReturn(true).when(mockCrlService).canReadGcsBucket(any(), any());
```
marikomedlock, line 328: Are there existing tests that will hit the new `checkAccess` implementations instead of the mocks?

marikomedlock (approved):
Looks good, thanks! One question about whether it would make sense for TDR snapshots to check access the same way as other resources.

service/src/test/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceServiceTest.java
zloery: I assumed so but this made me double check - we do have integration tests which run the new `checkAccess` as part of creating resources, but I realized we don't have any tests checking the "no access" path. Added one to exercise that as well for all three reference types

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedDataRepoSnapshotResource.java
zloery: This ends up calling the `retrieveSnapshot` on TDR which checks for READ_DATA access, so as I understand this actually is checking for read access and not just existence. Renamed the method in DataRepoService for clarity

PR #372, opened 2021-06-03
AS-723 follow-up: Create integration suite for dev env

ungwudik:
Create dedicated test suite for dev environment. I had set it up earlier to use the [generic test suite](https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/integration/src/main/resources/suites/FullIntegration.json) which is more targeted towards wsmtest environment.
This provides consistency by running the same tests that we run on alpha and staging during release promotion.


davidangb (approved):
lgtm, I haven't tried to run it, but syntax looks fine

ddietterich (approved):
lgtm

jaycarlton (approved):

PR #374, opened 2021-06-07
[PF-576] Remove jobId field from create workspace request

zloery:
This change removes the jobId field from the request body for CreateWorkspace calls. This was originally intended to allow for idempotency between services, but we've generally just used random workspace IDs to manage this instead. In the time this feature has been available, no WSM clients (Rawls, CLI) have used it. This is technically a breaking change - however, given that it is unused, I'm comfortable taking this out without a deprecation/migration plan.


jaycarlton (approved):

davidangb (approved):
the code here looks fine - and I think this is transparent to Rawls. But, could you coordinate with App Services when this merges and/or make sure to smoke test via Rawls in the real dev env, once this merges?

PR #376, opened 2021-06-08
[PF-832] Change from NO_CONTENT to OK for /status success response

ddietterich:


jaycarlton (approved):
Is this temporary? 204 seems more correct, but it's probably worth a little white lie to get the service up.

zloery (approved):

PR #380, opened 2021-06-10
[PF-732] Fix async endpoints

zloery:
This change fixes issues with WSM's asynchronous endpoints:

- By design, the async endpoints (and their polling endpoints) were supposed to return 202 for "job still running" and 200 for "job finished", with a more informative code inside the JobReport of the response body. As written, the endpoint copied the status from the JobReport, and the JobReport status was incorrectly either 202 or 200 (even for failed jobs, when it should have been 4xx or 5xx).
- A utility for building result URLs had parameters switched and was missing the scheme, so the provided URLs were incorrect. This change fixes the utility, so that the `resultURL` field of all JobReports WSM returns now point to valid endpoints.
- Unrelated to async endpoints,  but I also fixed some findbugs warnings and updated the skaffold template to reflect the repo's new structure.

### DIFF
--- a/service/local-dev/skaffold.yaml.template
+++ b/service/local-dev/skaffold.yaml.template
@@ -4,7 +4,7 @@ kind: Config
 build:
   artifacts:
   - image: gcr.io/terra-kernel-k8s/terra-workspace-manager
-    context: ../
+    context: ../../
     jib:
       project: service
 deploy:
--- a/service/src/main/java/bio/terra/workspace/common/utils/ControllerUtils.java
+++ b/service/src/main/java/bio/terra/workspace/common/utils/ControllerUtils.java
@@ -1,6 +1,9 @@
 package bio.terra.workspace.common.utils;
 
+import bio.terra.workspace.generated.model.ApiJobReport;
+import bio.terra.workspace.generated.model.ApiJobReport.StatusEnum;
 import javax.servlet.http.HttpServletRequest;
+import org.springframework.http.HttpStatus;
 
 /** Class of static helper methods for controllers */
 public class ControllerUtils {
@@ -20,7 +23,7 @@ public class ControllerUtils {
    */
   public static String getAsyncResultEndpoint(
       HttpServletRequest request, String jobId, String resultWord) {
-    return String.format("%s/%s/%s", request.getServletPath(), jobId, resultWord);
+    return String.format("%s/%s/%s", request.getServletPath(), resultWord, jobId);
   }
 
   /**
@@ -33,6 +36,16 @@ public class ControllerUtils {
    */
   public static String getAsyncResultEndpoint(HttpServletRequest request, String jobId) {
     return getAsyncResultEndpoint(request, jobId, "result");
+  }
+
+  /**
+   * Return the appropriate response code for an endpoint, given an async job report. For a job
+   * that's still running, this is 202. For a job that's finished (either succeeded or failed), the
+   * endpoint should return 200. More informational status codes will be included in either the
+   * response or error report bodies.
+   */
+  public static HttpStatus getAsyncResponseCode(ApiJobReport jobReport) {
+    return jobReport.getStatus() == StatusEnum.RUNNING ? HttpStatus.ACCEPTED : HttpStatus.OK;
   }
 
   private ControllerUtils() {}
--- a/service/src/main/java/bio/terra/workspace/service/job/JobService.java
+++ b/service/src/main/java/bio/terra/workspace/service/job/JobService.java
@@ -255,7 +255,11 @@ public class JobService {
     if (resultPath == null) {
       resultPath = "";
     }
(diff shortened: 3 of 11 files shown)


service/local-dev/skaffold.yaml.template
```diff
@@ -4,7 +4,7 @@ kind: Config
 build:
   artifacts:
   - image: gcr.io/terra-kernel-k8s/terra-workspace-manager
-    context: ../
+    context: ../../
```
jaycarlton, line 7: is there a way to refer to the root dir directly so we don't have to rely on upward references?

service/src/main/java/bio/terra/workspace/common/utils/ControllerUtils.java
```diff
@@ -35,5 +38,15 @@ public static String getAsyncResultEndpoint(HttpServletRequest request, String j
     return getAsyncResultEndpoint(request, jobId, "result");
   }
 
+  /**
+   * Return the appropriate response code for an endpoint, given an async job report. For a job
+   * that's still running, this is 202. For a job that's finished (either succeeded or failed), the
```
jaycarlton, line 43: I thought we couldn't return two distinct 2xx codes from an endpoint (like I was trying to do with 200 and 204).

service/src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -206,16 +206,34 @@ public ApiJobReport mapFlightStateToApiJobReport(FlightState flightState) {
...
+                      () ->
+                          new InvalidResultStateException(
+                              "Flight failed with no exception reported"));
+          statusCode = HttpStatus.valueOf(errorCode);
+          break;
+        case SUCCEEDED:
+          FlightMap resultMap = getResultMap(flightState);
+          statusCode = resultMap.get(JobMapKeys.STATUS_CODE.getKeyName(), HttpStatus.class);
+          if (statusCode == null) {
+            statusCode = HttpStatus.OK;
```
jaycarlton, line 230: I wonder if this isn't masking a potential problem. Should we warn in this case or something? We throw for unknown status but not for unknown HttpStatus?

jaycarlton (approved):
Thanks for the fixes.

service/local-dev/skaffold.yaml.template
ddietterich: I wish there were. I tried to figure one out for the application property files when I moved `render` -> `../config`, but didn't find a way.

We could use an envvar and try to set it properly in all of the places we make these references, but that might just make things harder to figure out.

service/src/main/java/bio/terra/workspace/common/utils/ControllerUtils.java
ddietterich: Right, let me be specific. You cannot generate two distinct 2xx codes that return _different_ responses.
The code gen will pick one to be the type returned by the client.

If you check, we have a bug in the current `createCloudContext` where the 202 thinks it is returning something different from the 200. But when you generate the client code for it, the signature is:
```
public CreateCloudContextResult createCloudContext(...
```
returning the 200 result. Apparently, whatever tests we have do not look at the 202 result. Or perhaps because we finish the flight fast enough that we never have a wait (?! seems unlikely).

@zloery what do you think about fixing that as part of this PR?

ddietterich (approved):
lgtm - I did find a separate bug that you could maybe fix in this PR

service/src/main/java/bio/terra/workspace/common/utils/ControllerUtils.java
zloery: Double checking that I understand this bug correctly: our yaml has different return types for 200 vs 202, but really only returns a `CreateCloudContextResultResponse` (the 200 return type)?

From what I can tell, the code generated from this yaml only ever uses the 200 return type. The swagger-ui and client library both use the 200 return type in all cases (even as the response type for 202 in swagger-ui). Without digging too deep, this seems like a  quirk of codegen: if you give it multiple return types it just picks one (the first? lowest code? I'm not sure) and uses that everywhere.

Since the 202 return type is ignored everywhere, I've fixed it in the yaml without needing to change other places. The other async endpoints (and their result endpoints) all match 200 and 202 responses.

service/src/main/java/bio/terra/workspace/service/job/JobService.java
zloery: I don't think this is a problem - not all flights explicitly set a response status, mostly because we've been defaulting to 200 already. So this case is more "default HttpStatus" than "unknown HttpStatus".

I think throwing on the unknown jobStatus is just best practice for enums in switch statements - we've already made explicit cases for all current enum values of jobStatus, so I don't expect to ever hit this case (unless someone adds new jobStatus values incorrectly, in which case this will catch that error).

PR #383, opened 2021-06-16
QA-1504: Fix intermittent NPE in resiliency test.

ichengchang:
Bump TestRunner version to 0.0.6-SNAPSHOT to fix intermittent NPE issue in resiliency tests.

For more information, please refer to this PR as well: [https://github.com/DataBiosphere/terra-test-runner/pull/17](https://github.com/DataBiosphere/terra-test-runner/pull/17)

### DIFF
--- a/integration/src/main/resources/configs/resiliency/DeleteInitialPods.json
+++ b/integration/src/main/resources/configs/resiliency/DeleteInitialPods.json
@@ -13,10 +13,10 @@
   "testScripts": [
     {
       "name": "ServiceStatus",
-      "parameters": [120],
+      "parameters": [150],
       "numberOfUserJourneyThreadsToRun": 2,
       "userJourneyThreadPoolSize": 2,
-      "expectedTimeForEach": 120,
+      "expectedTimeForEach": 150,
       "expectedTimeForEachUnit": "SECONDS"
     }
   ],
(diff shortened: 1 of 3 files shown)


integration/src/main/resources/configs/resiliency/DeleteInitialPods.json
```diff
@@ -13,10 +13,10 @@
   "testScripts": [
     {
       "name": "ServiceStatus",
-      "parameters": [120],
+      "parameters": [150],
```
ungwudik, line 16: Nit: why did this number increase?

ungwudik (approved):

integration/src/main/resources/configs/resiliency/DeleteInitialPods.json
ichengchang: To allow more time for Kubernetes to get pods ready. A pod is ready when all application containers that support a service are in ready state. This number should be determined by app services as to what an acceptable amount of time should be for a pod to recover after some unexpected crash. I increase it arbitrarily to 150 seconds here. On average I noticed it took around 45 seconds to recover from a pod crash so 150 seconds should give enough time for 2 pods to recover from back-to-back crash.

jaycarlton (approved):

PR #384, opened 2021-06-17
DDO-1293 Update Skaffold workflow for compatibility with upcoming Helm changes

choover-broad:
This PR updates WSM's Skaffold configuration for compatibility with [upcoming Helm changes](https://docs.google.com/document/d/1B1BU_C2AXEYBbMfLAZlQzYst-4ctVOlt1EoJcaEJszI/edit?usp=sharing).

I've tested this PR by deploying to the `gmalkov` personal environment using `skaffold  run`.

Changes:
* Remove references to terra-helm.
* Convert Skaffold deployer from `helm` to [`kubectl`](https://skaffold.dev/docs/pipeline-stages/deployers/kubectl/), using terra-helmfile's `render` tool to generate manifests. (`render` is the public API for terra-helmfile, so this will continue to work even as we make changes to terra-helmfile's structure)
* Move the values specified in the local values.yaml to terra-helmfile values ([PR](https://github.com/broadinstitute/terra-helmfile/pull/1384)). If folks feel strongly about keeping these here, I can instead adapt the `render` script to accept a local values file.

### DIFF
--- a/service/local-dev/README.md
+++ b/service/local-dev/README.md
@@ -1,14 +1,8 @@
-This directory contains scripts for running continuous development builds. This 
+This directory contains scripts for running continuous development builds. This
 is not necessary for deployment, but it can be helpful for developing faster.
-The provided setup script clones the terra-helm and terra-helmfile git repos,
-and templates in the desired Terra environment/k8s namespace to target.
-If you need to pull changes to either terra-helm or terra-helmfile, rerun this script.
-
-To use this, first ensure [Helm](https://helm.sh/docs/intro/install/#from-homebrew-macos) and [Skaffold](https://skaffold.dev/) are installed on your local machine. 
-
-> Older versions of Skaffold (v1.4.0 and earlier) do not have support for Helm 3 and will fail to deploy your 
-changes. If you're seeing errors like `UPGRADE FAILED: "(Release name)" has no 
-deployed releases`, try updating Skaffold.
+The provided setup script clones the terra-helmfile git repo and renders
+Kubernetes manifests for the target Terra environment/k8s namespace.
+If you need to pull changes to terra-helmfile, rerun this script.
 
 You may need to use gcloud to provide GCR
  credentials with `gcloud auth configure-docker`. Finally, run local-run.sh with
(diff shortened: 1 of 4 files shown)


service/local-dev/values.yaml.template
```diff
@@ -1,5 +0,0 @@
-imageConfig:
-  imagePullPolicy: IfNotPresent
-global:
-  terraEnv: ENV
-initDB: true
```
wchamber, line 5: I'm not sure if `initDB: true` is a good default on personal environments (causes the WSM database to be dropped on server startup). It might actually be more appropriate only on skaffold runs for more db iteration. Or maybe never default to true. @zloery thoughts?

wchamber (commented):
Sorry, not sure I understand the migration story here. Is this all ready to merge with what's at head in terra-helmfile, or are there some other changes that need to happen?

service/local-dev/README.md
```diff
@@ -18,9 +12,9 @@ You may need to use gcloud to provide GCR
 ./setup_local_env.sh <environment>
 ```
 
-Optionally, you can provide a branch of [terra-helm](https://github.com/broadinstitute/terra-helm) and/or [terra-helmfile](https://github.com/broadinstitute/terra-helmfile) to use if you are testing PRs to either. They default to master.
+Optionally, you can provide a branch of [terra-helmfile](https://github.com/broadinstitute/terra-helmfile) to use if you are testing a PR. (The script defaults to master).
 ```
-./setup_local_env.sh <environment> <terra-helm ref> <terra-helmfile ref>
+./setup_local_env.sh <environment> <terra-helmfile ref>
```
ddietterich, line 17: Maybe add `git protocol` to the arg list docn?

service/local-dev/values.yaml.template
ddietterich: Well, this file went away, so does it matter?

In any case, I think this was the right default for "local dev", since the object of local dev to being able to build and launch into your development environment with skaffold.

wchamber: Sorry, more referencing linked PR https://github.com/broadinstitute/terra-helmfile/pull/1384/files#diff-a5a76688e62660b0d4d96da4f210efc1435b881917dde2d9af5191cb78917dd4

zloery: from personal experience I'd probably default this to false. my workflow generally involved deploying with `initDB` false, and intentionally wiping the DB from time to time instead. I generally set this to true or false explicitly instead of relying on a default, so it also isn't the most impactful to me

zloery (approved):
LGTM

service/local-dev/values.yaml.template
ddietterich: Oh yeah, I agree. I would not enable that in the helm domain

broadbot: Thanks for the comments all! Would it be best to keep a local `values.yaml` file in this repo for overriding values when deploying with Skaffold? I can make adjustments to the render script to support it.

choover-broad: Thanks for the comments all! Would it be best to keep a local `values.yaml` file in this repo for overriding values when deploying with Skaffold? I can make adjustments to the render script to support it.

ddietterich: I think it would be useful to keep a local `values.yaml`

ddietterich:
@choover-broad what is the status of this work?

choover-broad:
> @choover-broad what is the status of this work?

In-progress, just tied up with some in-progress PRs on a few other repos. If the open PR is causing clutter I'm happy to close and re-open it when I'm ready for re-review.

PR #385, opened 2021-06-17
PF-861: Bump test runner to 0.0.7

marikomedlock:
Bump the Test Runner library version. Details of the fixes that went into this version bump are in the [Test Runner PR](https://github.com/DataBiosphere/terra-test-runner/pull/18).

### DIFF
--- a/integration/build.gradle
+++ b/integration/build.gradle
@@ -10,7 +10,7 @@ plugins {
     //  fine in the service project.
     // id "com.diffplug.spotless" version "5.12.4"
     // Terra Test Runner Plugin
-    id 'bio.terra.test-runner-plugin' version '0.0.7a-SNAPSHOT'
+    id 'bio.terra.test-runner-plugin' version '0.0.7-SNAPSHOT'
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


zloery (approved):

integration/build.gradle
```diff
@@ -10,7 +10,7 @@ plugins {
     //  fine in the service project.
     // id "com.diffplug.spotless" version "5.12.4"
     // Terra Test Runner Plugin
-    id 'bio.terra.test-runner-plugin' version '0.0.6-SNAPSHOT'
+    id 'bio.terra.test-runner-plugin' version '0.0.7-SNAPSHOT'
```
jaycarlton, line 13: nit: I think there's a way to make an external property or buildscript property so we don't have to repeat the version string. It may be a rabbit hole.

jaycarlton (approved):

integration/build.gradle
marikomedlock: I think you can do it by specifying the version in a `gradle.properties` file, but then that would be inconsistent with how versions of other things are specified in the `build.gradle` file. What do you think?

jaycarlton:
It's worth keeping an eye on. I tend to have zero tolerance for duplication
and more slack for consistency, but that's just me.

PR #388, opened 2021-06-22
PF-871: Ignore case in private resource user email.

marikomedlock:
Ignore case when validating a private resource user email. Came across this when writing CLI tests for private resources.


zloery (approved):
LGTM. I dug through the code a little bit and the `assignedUser` field isn't actually passed to Sam (we infer it from the requesting user), so there shouldn't be conflicts later on in execution.

Totally out of scope for this PR but taking and storing a field we don't use feels like we're setting up future bugs. We marked `assignedUser` as deprecated, we should probably remove it from code. I'll file a ticket for that

jaycarlton:
@zloery I never understood if we could remove it completely from the API or not. It seems like it almost had a purpose but I'm not sure if that purpose is gone or if we haven't gotten to it yet.

jaycarlton (approved):

PR #389, opened 2021-06-24
[PF-506] Add omitted GCS lifecycle conditions to WSM

zloery:
This adds support for a few GCS lifecycle rules that were previously included in WSM's API but not propagated to GCP due to versioning issues with the GCS library. The newly supported lifecycle conditions are:

- noncurrentTimeBefore
- daysSinceCustomTime
- daysSinceNoncurrentTime
- customTimeBefore

This did not require updating the version of CRL that WSM was using - I think that update happened sometime between when the bug was filed and now, though we didn't add support inside WSM.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/GcsApiConversions.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/GcsApiConversions.java
@@ -212,10 +212,6 @@ public class GcsApiConversions {
     resultBuilder.setCreatedBefore(toGoogleDateTime(condition.getCreatedBefore()));
     resultBuilder.setNumberOfNewerVersions(condition.getNumNewerVersions());
     resultBuilder.setIsLive(condition.isLive());
-    resultBuilder.setMatchesStorageClass(
-        Optional.ofNullable(condition.getMatchesStorageClass())
-            .map(c -> c.stream().map(GcsApiConversions::toGcsApi).collect(Collectors.toList()))
-            .orElse(null));
     resultBuilder.setDaysSinceNoncurrentTime(condition.getDaysSinceNoncurrentTime());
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/GcsApiConversions.java
```diff
@@ -208,12 +208,18 @@ public static ApiGcpGcsBucketLifecycleRuleActionType toWsmApi(String lifecycleAc
...
-    /* TODO(PF-506): some conditions aren't in the version of the Google Storage API in the
-     *    latest version of the CRL. */
     resultBuilder.setAge(condition.getAge());
     resultBuilder.setCreatedBefore(toGoogleDateTime(condition.getCreatedBefore()));
     resultBuilder.setNumberOfNewerVersions(condition.getNumNewerVersions());
     resultBuilder.setIsLive(condition.isLive());
+    resultBuilder.setMatchesStorageClass(
+        Optional.ofNullable(condition.getMatchesStorageClass())
+            .map(c -> c.stream().map(GcsApiConversions::toGcsApi).collect(Collectors.toList()))
+            .orElse(null));
```
marikomedlock, line 218: It looks like there's a duplicate of this below. Can you remove one?

jaycarlton (approved):
Thanks for filling these in. That's great that there weren't any difficult CRL changes required.

service/src/main/java/bio/terra/workspace/service/resource/controlled/GcsApiConversions.java
zloery: good catch, thanks!

marikomedlock (approved):
LGTM Thanks!

PR #391, opened 2021-06-29
[PF-749] fix tag publish

ddietterich:
Skip the test workflow and the tag-publish workflow if the push to dev commit message is `"^bump .*"`. This will save a complete, redundant test pass and tag computation when merging code.

I added a composite action, `bump-skip`, that performs the check and sets an output. Then each following step includes an `if:` that checks against that output.

There is no "exit at this point" logic in github actions, so the `if:` approach is the recommended solution. An alternative is to use sequential jobs and put the `if:` at the job level. However, that would mean two separate checkouts, so the step level seemed like the more straightforward approach.

### DIFF
--- a/.github/actions/bump-skip/action.yml
+++ b/.github/actions/bump-skip/action.yml
@@ -1,0 +1,28 @@
+# This action must be done after the checkout action
+name: 'bump-skip'
+description: 'Set skip-out when we are doing a version bump'
+author: 'dd'
+inputs:
+  event-name:
+    description: 'github.event_name from the calling workflow'
+    required: true
+outputs:
+  is-bump:
+    description: 'yes if this is a push made by bumper; no if it is a regular push'
+    value: ${{ steps.bumptest.outputs.is-bump }}
+runs:
+  using: "composite"
+  steps:
+    - name: Bump test
+      id: bumptest
+      run: |
+        log=$(git log --pretty='%B')
+        echo "log=$log"
+        pattern="^bump .*"
+        IS_BUMP=no
+        if [[ "${{ inputs.event-name }}" == "push" && "$log" =~ $pattern ]]; then
+          IS_BUMP=yes
+        fi
+        echo "IS_BUMP=$IS_BUMP"
+        echo ::set-output name=is-bump::$IS_BUMP
+      shell: bash
--- a/.github/workflows/tag-publish.yml
+++ b/.github/workflows/tag-publish.yml
@@ -131,8 +131,10 @@ jobs:
         DOCKER_TAG=${{ steps.tag.outputs.tag }}
         echo ::set-output name=name::gcr.io/${GOOGLE_PROJECT}/${SERVICE_NAME}:${DOCKER_TAG}
     - name: Build image locally with jib
+      if: steps.skiptest.outputs.skip-out == 'no'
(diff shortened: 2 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -69,16 +77,19 @@ jobs:
...
       run: psql -h 127.0.0.1 -U postgres -f ./service/local-dev/local-postgres-init.sql
 
     - name: Set up AdoptOpenJDK 11
+      if: steps.skiptest.outputs.is-bump == 'no'
       uses: joschi/setup-jdk@v2
       with:
         java-version: 11
 
     - name: Cache Gradle packages
+      if: steps.skiptest.outputs.is-bump == 'no'
```
jaycarlton, line 92: It's amazing they don't support yaml anchors for this kind of thing https://github.community/t/support-for-yaml-anchors/16128/10

zloery (approved):

.github/actions/bump-skip/action.yml
```diff
@@ -0,0 +1,28 @@
...
+  using: "composite"
+  steps:
+    - name: Bump test
+      id: bumptest
+      run: |
+        log=$(git log --pretty='%B')
+        echo "log=$log"
+        pattern="^bump .*"
+        IS_BUMP=no
+        if [[ "${{ inputs.event-name }}" == "push" && "$log" =~ $pattern ]]; then
```
jaycarlton, line 23: do you need quotes around `$pattern`?

.github/workflows/tag-publish.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Set part of semantic version to bump
-      id: semver
+      id: controls
```
jaycarlton, line 49: I don't get this name.

.github/actions/bump-skip/action.yml
```diff
@@ -0,0 +1,28 @@
...
+          IS_BUMP=yes
+        fi
+        echo "IS_BUMP=$IS_BUMP"
```
jaycarlton, line 26: Why does it take two echo statements?

jaycarlton (approved):
This should free up a bunch of cycles. Can it be shared with other repos/the community?

.github/actions/bump-skip/action.yml
```diff
@@ -0,0 +1,28 @@
...
+    value: ${{ steps.bumptest.outputs.is-bump }}
+runs:
+  using: "composite"
+  steps:
+    - name: Bump test
+      id: bumptest
+      run: |
+        log=$(git log --pretty='%B')
+        echo "log=$log"
+        pattern="^bump .*"
```
jaycarlton, line 21: We could generalize the pattern to something like `[skip_tests]` and allow it anywhere in the log line, but I guess that's a bit ugly in the log

.github/actions/bump-skip/action.yml
ddietterich: One echo is for the log so you know whether it is a bump. The other echo is magic that sets an output variable from the action.

.github/actions/bump-skip/action.yml
ddietterich: Agreed. I made another PR. It requires a change to the common bumper action and the a compensating change to our GHAs.

PR #393, opened 2021-06-30
[AS-787] Add a snapshot list end point that matches the existing snapshot get end point

sagehen03:
App Services received feedback from the Hornet team that they would like Rawls responses to be consistent between get and list for snapshots.  https://broadworkbench.atlassian.net/browse/AS-787 We can accomplish this by making the get and list responses consistent in WSM (this PR outlines one way to do that) or we can reformat the responses in Rawls.  Initially, folks on app services leaned in favor of making changes in WSM.  More details below..

Currently the Rawls end points to get snapshots delegate to WSM:

-  List `/api/workspaces/{workspaceNamespace}/{workspaceName}/snapshots/v2` -> `/api/workspaces/v1/{workspaceId}/resources` 
- Get `/api/workspaces/{workspaceNamespace}/{workspaceName}/snapshots/v2/{snapshotId}` -> `/api/workspaces/v1/{workspaceId}/resources/referenced/datarepo/snapshots/{resourceId}`

The List mapping uses a more general end point on the WSM side to return resource (Snapshots, Buckets, BQ datasets, etc) and produces a response like this:

```
{
  "resources": [
    {
      "metadata": {
        "workspaceId": "54c76162-2337-4e65-b856-92e8202778cd",
        "resourceId": "46550c98-74f1-43c0-bf2b-a27e20d67a10",
        "name": "string",
        "description": "string",
        "resourceType": "DATA_REPO_SNAPSHOT",
        "stewardshipType": "REFERENCED",
        "cloudPlatform": "GCP",
        "cloningInstructions": "COPY_NOTHING"
      },
      "resourceAttributes": {
        "gcpDataRepoSnapshot": {
          "instanceName": "terra",
          "snapshot": "6b073ea9-caea-4a64-9cbf-745d17d5a0c8"
        }
      }
    }
  ]
} 
```

This newly proposed end point to return a list of snapshots returns a response that looks like this:
```
{
  "snapshots": [
    {
      "metadata": {
        "workspaceId": "54c76162-2337-4e65-b856-92e8202778cd",
        "resourceId": "46550c98-74f1-43c0-bf2b-a27e20d67a10",
        "name": "string",
        "description": "string",
        "resourceType": "DATA_REPO_SNAPSHOT",
        "stewardshipType": "REFERENCED",
        "cloudPlatform": "GCP",
        "cloningInstructions": "COPY_NOTHING"
      },
      "attributes": {
        "instanceName": "terra",
        "snapshot": "6b073ea9-caea-4a64-9cbf-745d17d5a0c8"
      }
    }
  ]
}
```
And is the "plural" form of the existing get a snapshot response:
```
{
  "metadata": {
    "workspaceId": "54c76162-2337-4e65-b856-92e8202778cd",
    "resourceId": "46550c98-74f1-43c0-bf2b-a27e20d67a10",
    "name": "string",
    "description": "string",
    "resourceType": "DATA_REPO_SNAPSHOT",
    "stewardshipType": "REFERENCED",
    "cloudPlatform": "GCP",
    "cloningInstructions": "COPY_NOTHING"
  },
  "attributes": {
    "instanceName": "terra",
    "snapshot": "6b073ea9-caea-4a64-9cbf-745d17d5a0c8"
  }
}
```

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
@@ -1,13 +1,7 @@
 package bio.terra.workspace.app.controller;
 
 import bio.terra.workspace.generated.controller.ReferencedGcpResourceApi;
-import bio.terra.workspace.generated.model.ApiCreateDataRepoSnapshotReferenceRequestBody;
-import bio.terra.workspace.generated.model.ApiCreateGcpBigQueryDatasetReferenceRequestBody;
-import bio.terra.workspace.generated.model.ApiCreateGcpGcsBucketReferenceRequestBody;
-import bio.terra.workspace.generated.model.ApiDataRepoSnapshotResource;
-import bio.terra.workspace.generated.model.ApiGcpBigQueryDatasetResource;
-import bio.terra.workspace.generated.model.ApiGcpGcsBucketResource;
-import bio.terra.workspace.generated.model.ApiUpdateDataReferenceRequestBody;
+import bio.terra.workspace.generated.model.*;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequestFactory;
 import bio.terra.workspace.service.resource.model.CloningInstructions;
@@ -198,14 +192,19 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
   }
 
   @Override
-  public ResponseEntity<ApiDataRepoSnapshotResource> getDataRepoSnapshotReference(
+  public ResponseEntity<ApiResourceDescription> getDataRepoSnapshotReference(
       UUID id, UUID referenceId) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     ReferencedResource referenceResource =
         referenceResourceService.getReferenceResource(id, referenceId, userReq);
-    ApiDataRepoSnapshotResource response =
-        referenceResource.castToDataRepoSnapshotResource().toApiResource();
-    return new ResponseEntity<>(response, HttpStatus.OK);
+    ApiResourceMetadata apiResourceMetadata = referenceResource.toApiMetadata();
+    ReferencedDataRepoSnapshotResource referencedDataRepoSnapshotResource =
+        referenceResource.castToDataRepoSnapshotResource();
+    ApiResourceAttributesUnion union = new ApiResourceAttributesUnion();
+    union.gcpDataRepoSnapshot(referencedDataRepoSnapshotResource.toApiAttributes());
+    return new ResponseEntity<>(
+        new ApiResourceDescription().metadata(apiResourceMetadata).resourceAttributes(union),
+        HttpStatus.OK);
   }
 
   @Override
--- a/service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
@@ -10,9 +10,12 @@ import bio.terra.workspace.service.resource.referenced.ReferencedDataRepoSnapsho
 import bio.terra.workspace.service.resource.referenced.ReferencedGcsBucketResource;
 import bio.terra.workspace.service.resource.referenced.ReferencedResource;
 import bio.terra.workspace.service.resource.referenced.ReferencedResourceService;
+import java.util.List;
 import java.util.UUID;
+import java.util.stream.Collectors;
 import javax.servlet.http.HttpServletRequest;
 import javax.validation.Valid;
+import javax.validation.constraints.Min;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -192,19 +195,14 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
   }
 
   @Override
-  public ResponseEntity<ApiResourceDescription> getDataRepoSnapshotReference(
+  public ResponseEntity<ApiDataRepoSnapshotResource> getDataRepoSnapshotReference(
       UUID id, UUID referenceId) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     ReferencedResource referenceResource =
         referenceResourceService.getReferenceResource(id, referenceId, userReq);
-    ApiResourceMetadata apiResourceMetadata = referenceResource.toApiMetadata();
-    ReferencedDataRepoSnapshotResource referencedDataRepoSnapshotResource =
-        referenceResource.castToDataRepoSnapshotResource();
-    ApiResourceAttributesUnion union = new ApiResourceAttributesUnion();
-    union.gcpDataRepoSnapshot(referencedDataRepoSnapshotResource.toApiAttributes());
-    return new ResponseEntity<>(
-        new ApiResourceDescription().metadata(apiResourceMetadata).resourceAttributes(union),
-        HttpStatus.OK);
+    ApiDataRepoSnapshotResource response =
+        referenceResource.castToDataRepoSnapshotResource().toApiResource();
+    return new ResponseEntity<>(response, HttpStatus.OK);
   }
 
   @Override
@@ -215,6 +213,22 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
         referenceResourceService.getReferenceResourceByName(id, name, userReq);
     ApiDataRepoSnapshotResource response =
         referenceResource.castToDataRepoSnapshotResource().toApiResource();
+    return new ResponseEntity<>(response, HttpStatus.OK);
+  }
+
+  @Override
+  public ResponseEntity<ApiSnapshotList> getSnapshotsForWorkspace(
+      UUID workspaceId, @Min(0) @Valid Integer offset, @Min(1) @Valid Integer limit) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    List<ReferencedResource> referencedResources =
+        referenceResourceService.enumerateSnapshots(workspaceId, offset, limit, userReq);
+    var response = new ApiSnapshotList();
+    List<ApiDataRepoSnapshotResource> result =
+        referencedResources.stream()
+            .map(ReferencedResource::castToDataRepoSnapshotResource)
+            .map(ReferencedDataRepoSnapshotResource::toApiResource)
+            .collect(Collectors.toList());
+    response.snapshots(result);
     return new ResponseEntity<>(response, HttpStatus.OK);
   }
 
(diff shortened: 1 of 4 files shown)


ddietterich (changes requested):
I would have preferred a proposal doc ahead of this PR. It would have saved the coding work.

I do not think this is the right interface approach. I would rather put a query parameter on the enumerate call that would filter the result by resource type, or set of resource types. Then we do not need to add new endpoints for every resource in order to enumerate them and different clients can look at the resources they care about.

sagehen03:
> I do not think this is the right interface approach. I would rather put a query parameter on the enumerate call that would filter the result by resource type, or set of resource types. Then we do not need to add new endpoints for every resource in order to enumerate them and different clients can look at the resources they care about.

I believe this end point already exists?  https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/Resource/enumerateResources  But, the problem as I understand it is that end point returns a more general response whose data won't match the more specific responses returned by https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/ReferencedGcpResource/getDataRepoSnapshotReference I can go the other way and add an end point that returns a single resource in the general form, which should then make things more consistent between get and list api calls.

ddietterich:
> > I do not think this is the right interface approach. I would rather put a query parameter on the enumerate call that would filter the result by resource type, or set of resource types. Then we do not need to add new endpoints for every resource in order to enumerate them and different clients can look at the resources they care about.
> 
> I believe this end point already exists? https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/Resource/enumerateResources But, the problem as I understand it is that end point returns a more general response whose data won't match the more specific responses returned by https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/ReferencedGcpResource/getDataRepoSnapshotReference I can go the other way and add an end point that returns a single resource in the general form, which should then make things more consistent between get and list api calls.

I think you are mistaken. The single object GET on the reference snapshot is `DataRepoSnapshotResource` that provides two parts:
- ResourceMetadata
- DataRepoSnapshotAttributes

The enumerate returns for each element:
- ResourceMetadata
- ResourceAttributesUnion, that for snapshots, contains DataRepoSnapshotAttributes

So all of the data you need is returned in both cases. Right?

I don't think the different in forms at the API layer is strong enough motivation to introduce another enumeration endpoint for a specific resource type.

sagehen03:
> > > I do not think this is the right interface approach. I would rather put a query parameter on the enumerate call that would filter the result by resource type, or set of resource types. Then we do not need to add new endpoints for every resource in order to enumerate them and different clients can look at the resources they care about.
> > 
> > 
> > I believe this end point already exists? https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/Resource/enumerateResources But, the problem as I understand it is that end point returns a more general response whose data won't match the more specific responses returned by https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/ReferencedGcpResource/getDataRepoSnapshotReference I can go the other way and add an end point that returns a single resource in the general form, which should then make things more consistent between get and list api calls.
> 
> I think you are mistaken. The single object GET on the reference snapshot is `DataRepoSnapshotResource` that provides two parts:
> 
>     * ResourceMetadata
> 
>     * DataRepoSnapshotAttributes
> 
> 
> The enumerate returns for each element:
> 
>     * ResourceMetadata
> 
>     * ResourceAttributesUnion, that for snapshots, contains DataRepoSnapshotAttributes
> 
> 
> So all of the data you need is returned in both cases. Right?
> 
> I don't think the minor difference in forms at the API layer is strong enough motivation to introduce another enumeration endpoint for a specific resource type.

Yes, you're right.  The data needed is present in both response types, and we could pull out the necessary pieces in Rawls and then return the data in a way that's consistent between List and Get on the Rawls.  Here's the jira ticket for context https://broadworkbench.atlassian.net/browse/AS-787 and my apologies for not providing more context sooner.  I wasn't expecting folks outside of my team to look at this straight away.  It sounds like you are strongly in favor of leaving things alone in WSM and reformatting the responses in Rawls in order to accommodate the user's request.  That's good info to have and I appreciate that feedback.  In our team discussions we leaned towards making the changes in WSM, but it's good to know where you stand.

ddietterich:
> > > > I do not think this is the right interface approach. I would rather put a query parameter on the enumerate call that would filter the result by resource type, or set of resource types. Then we do not need to add new endpoints for every resource in order to enumerate them and different clients can look at the resources they care about.
> > > 
> > > 
> > > I believe this end point already exists? https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/Resource/enumerateResources But, the problem as I understand it is that end point returns a more general response whose data won't match the more specific responses returned by https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/ReferencedGcpResource/getDataRepoSnapshotReference I can go the other way and add an end point that returns a single resource in the general form, which should then make things more consistent between get and list api calls.
> > 
> > 
> > I think you are mistaken. The single object GET on the reference snapshot is `DataRepoSnapshotResource` that provides two parts:
> > ```
> > * ResourceMetadata
> > 
> > * DataRepoSnapshotAttributes
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The enumerate returns for each element:
> > ```
> > * ResourceMetadata
> > 
> > * ResourceAttributesUnion, that for snapshots, contains DataRepoSnapshotAttributes
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So all of the data you need is returned in both cases. Right?
> > I don't think the minor difference in forms at the API layer is strong enough motivation to introduce another enumeration endpoint for a specific resource type.
> 
> Yes, you're right. The data needed is present in both response types, and we could pull out the necessary pieces in Rawls and then return the data in a way that's consistent between List and Get on the Rawls side. Here's the jira ticket for context https://broadworkbench.atlassian.net/browse/AS-787 and my apologies for not providing more context sooner. I wasn't expecting folks outside of my team to look at this straight away. It sounds like you are strongly in favor of leaving things alone in WSM and reformatting the responses in Rawls in order to accommodate the user's request. That's good info to have and I appreciate that feedback. In our team discussions we leaned towards making the changes in WSM, but it's good to know where you stand.

The PF team and the AS team coordinate on the development of WSM. But we don't always manage to sync up on everything that turns out to matter across teams. It happens...  Our next bi-weekly sync is on July 8, but I'd be happy to meet and kick things around sooner, if that would be useful.

Anyway, I am not at all opposed to making WSM changes. If you are willing, it would be great it you could make an enhancement to the existing enumerate endpoint to add filtering by resource types. In fact, we have this Jira ticket to track the work: [PF-670](https://broadworkbench.atlassian.net/browse/PF-670).

I think within Rawls, you should probably have a thin layer of insulation between the WSM API client and your internal processing of the results. (Maybe it is there already. I don't know that code at all.) In that layer, I would convert the single snapshot `ApiDataRepoSnapshotResource` return into `rawls-snapshot-resource-object`. And convert a filtered enumeration of `List<ApiResourceDescription>` into `List<rawls-snapshot-resource-object>`. If the form of ApiDataRepoSnapshotResource changes, downstream code can be insulated from that change.

davidangb:
Chiming in with some context … we have received end-user feedback from the Hornet team about the usability of these APIs. They are working directly with the list-resources API and the get-snapshot API (because those are the only ones currently exposed via Rawls), and found the difference in the response payloads to be onerous. They need different code paths/parsers to interpret the results, depending on whether they are requesting one snapshot reference or multiple snapshot references. Achieving consistency between the list-multiple and get-single APIs is one of their top 3 requests.

One consideration about modifying these APIs is whether or not to change existing APIs or create a v3 set of APIs … when we haven't even retired the v1 APIs yet. Creating new resource-specific list APIs would be purely additive and would not require a v3; of course that isn't reason on its own to go a particular route, just a consideration. If we changed the get-snapshot API to nest under a `ResourceAttributesUnion` key (so that it's consistent with the list API), we should probably make a v3.

Creating new resource-specific APIs also allows us to keep the all-resource list API, in its current state - we don't have to change that one.

I'm open to different solutions to achieve consistency, but I do think that if our first users' feedback is about API response consistency, it's worthwhile to ask if future users could have the same feedback, and address that feedback at the source (WSM) instead of massaging the responses in Rawls.

ddietterich:
> Chiming in with some context … we have received end-user feedback from the Hornet team about the usability of these APIs. They are working directly with the list-resources API and the get-snapshot API (because those are the only ones currently exposed via Rawls), and found the difference in the response payloads to be onerous. They need different code paths/parsers to interpret the results, depending on whether they are requesting one snapshot reference or multiple snapshot references. Achieving consistency between the list-multiple and get-single APIs is one of their top 3 requests.
> 
> One consideration about modifying these APIs is whether or not to change existing APIs or create a v3 set of APIs … when we haven't even retired the v1 APIs yet. Creating new resource-specific list APIs would be purely additive and would not require a v3; of course that isn't reason on its own to go a particular route, just a consideration. If we changed the get-snapshot API to nest under a `ResourceAttributesUnion` key (so that it's consistent with the list API), we should probably make a v3.
> 
> Creating new resource-specific APIs also allows us to keep the all-resource list API, in its current state - we don't have to change that one.
> 
> I'm open to different solutions to achieve consistency, but I do think that if our first users' feedback is about API response consistency, it's worthwhile to ask if future users could have the same feedback, and address that feedback at the source (WSM) instead of massaging the responses in Rawls.

I'm a confused about which APIs you are talking about here. Is the Hornet team directly using WSM APIs? WSM only has a `/v1` API.

ddietterich:
One tangential point. I have been very frustrated by the low quality APIs generated from Swagger.

I was toying with the idea of adding code to the client that would define a Java class hierarchy with a base resource class and specializations for the different resources. Then make a wrapper around the enumerate call that would convert the ugly union into a list of base-class. It would be much more natural to use from Java.

There would be a cost to that beauty: every time we changed the OpenAPI code we would have to visit the by-hand code to add in the new stuff.

davidangb:
@ddietterich sorry, I spoke poorly about the v1 vs. v2 APIs. Hornet is hitting Rawls (using Java, I think, but not sure), and Rawls delegates to the WSM APIs. Rawls has v1 and v2 APIs for snapshots, which reflect WSM's pre- and post-refactoring of WSM's v1 APIs. My previous comment mixed all this up. We could, for instance, make payload changes to WSM's v1 APIs without creating a new version, but only add a v3 to Rawls.

ddietterich:
> @ddietterich sorry, I spoke poorly about the v1 vs. v2 APIs. Hornet is hitting Rawls (using Java, I think, but not sure), and Rawls delegates to the WSM APIs. Rawls has v1 and v2 APIs for snapshots, which reflect WSM's pre- and post-refactoring of WSM's v1 APIs. My previous comment mixed all this up. We could, for instance, make payload changes to WSM's v1 APIs without creating a new version in WSM, but only add a v3 to Rawls.

Ah, I see. Yes, that is a PITA. So the current Rawls endpoint is "enumerate referenced snapshots", not "enumerate resources"?

Maybe we should make a meeting and discuss?

davidangb:
> So the current Rawls endpoint is "enumerate referenced snapshots", not "enumerate resources"?

That's correct! Sure, a meeting to discuss sounds good. @sagehen03 what do you think?

sagehen03:
> > So the current Rawls endpoint is "enumerate referenced snapshots", not "enumerate resources"?
> 
> That's correct! Sure, a meeting to discuss sounds good. @sagehen03 what do you think?

I'd be happy to join a discussion.

ddietterich:
Would it be OK if I close this PR?

PR #395, opened 2021-07-12
[no ticket] remove .git dirs and ignore repo dirs

jaycarlton:
Currently, the `setup_local_env` script performs a full branch clone on two repos under the root of the terra-workspace-manager repo. This causes some confusion in the IDE, as the directories can't be .gitignore'd, so their files look like they live in the top repo in search results.

It's also confusing on the command line if you navigate into one of the directories, as your git context switches. Since it's not a good workflow anyway to be checking out branches or looking at the log from within these nested repos, I'm deleting the `.git` directories. This also makes them `.gitignore`-able, which cleans up the UI and project structure.
Screen Shot 2021-07-12 at 4 04 28 PM

### DIFF
--- a/service/local-dev/.gitignore
+++ b/service/local-dev/.gitignore
@@ -4,6 +4,9 @@
 # file.
 # Ignore everything in this directory...
 *
+./terra-helm/**
+./terra-helmfile/**
+
 # Except...
 !.gitignore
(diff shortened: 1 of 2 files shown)


marikomedlock (approved):

ddietterich (approved):
lgtm

PR #398, opened 2021-07-15
[PF-811] Clone Referenced Resources

jaycarlton:
Create clones of all reference types that point to the original referent. This didn't involve any new flights, but reused the create reference flight.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/CloneReferencedResources.java
+++ b/integration/src/main/java/scripts/testscripts/CloneReferencedResources.java
@@ -9,6 +9,7 @@ import bio.terra.testrunner.runner.config.TestUserSpecification;
 import bio.terra.workspace.api.ReferencedGcpResourceApi;
 import bio.terra.workspace.api.WorkspaceApi;
 import bio.terra.workspace.client.ApiClient;
+import bio.terra.workspace.model.CloneReferencedGcpGcsBucketResourceResult;
 import bio.terra.workspace.model.CloneReferencedResourceRequestBody;
 import bio.terra.workspace.model.CloneReferencedResourceResult;
 import bio.terra.workspace.model.CloningInstructionsEnum;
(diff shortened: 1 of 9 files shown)


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1965,7 +2010,26 @@ components:
...
+        sourceWorkspaceId:
+          description: ID of the workspace of the source resource for this clone
+          type: string
+          format: uuid
+        sourceResourceId:
+          description: ID of the source resource
+          type: string
+          format: uuid
+        resource:
+          $ref: '#/components/schemas/ResourceDescription'
```
ddietterich, line 2032: I have mixed feelings about this. I would like to minimize the use of the union form. It is annoying to clients, as we have heard. I think we could put the clone operation at the per-resource level with a specific response and avoid it.

ddietterich (commented):
Very nice. I would like to talk about the API.

I think I would prefer to have per-resource endpoints as we do with everything except enumeration.

service/src/main/resources/api/service_openapi.yaml
jaycarlton: I'll split them out; it won't be too bad.

ddietterich (approved):
LGTM

integration/src/main/java/scripts/testscripts/CloneReferencedResources.java
```diff
@@ -0,0 +1,160 @@
...
+            .id(destinationWorkspaceId)
+            .spendProfile(getSpendProfileId())
+            .stage(getStageModel());
+    final CreatedWorkspace createdDestinationWorkspace = workspaceApi.createWorkspace(requestBody);
+    assertThat(createdDestinationWorkspace.getId(), equalTo(destinationWorkspaceId));
+
+    // create destination cloud context
+    String destinationProjectId = CloudContextMaker
+        .createGcpCloudContext(destinationWorkspaceId, workspaceApi);
+    logger.info("Created destination project {} in workspace {}", destinationProjectId, destinationWorkspaceId);
```
zloery, line 83: If we're just cloning references, I don't think we need a cloud context.

zloery (approved):
one small q in the test but LGTM otherwise!

integration/src/main/java/scripts/testscripts/CloneReferencedResources.java
jaycarlton: Awesome! Removing that takes the test time down to 30 seconds.

PR #399, opened 2021-07-15
[PF-878] Move WSM to custom project-level roles

zloery:
This changes the way WSM handles project-level permissions. Previously, it granted a collection of predefined roles for each workspace role (READER, WRITER, OWNER). Now, it grants a single custom role for each workspace role. The actual permissions are unchanged for this PR - I'll modify the list for PF-869 separately to keep changes separated.

As a reminder, we do not currently have any backfill support for old workspaces/projects using the predefined roles. This change will only affect newly created GCP contexts, though older workspaces can delete and recreate context to pick up this change. This will not affect RAWLS_WORSPACEs, which do not have cloud contexts.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
@@ -20,7 +20,7 @@ public class CloudSyncRoleMapping {
   // Note that custom roles defined at the project level cannot contain the
   // "resourcemanager.projects.list" permission, even though it was previously included here.
   // See https://cloud.google.com/iam/docs/understanding-custom-roles#known_limitations
-  private static final List<String> READER_PERMISSIONS =
+  private static final List<String> PROJECT_READER_PERMISSIONS =
       ImmutableList.of(
           "bigquery.jobs.create",
           "lifesciences.operations.get",
@@ -32,9 +32,9 @@ public class CloudSyncRoleMapping {
           "serviceusage.quotas.get",
           "serviceusage.services.get",
           "serviceusage.services.list");
-  private static final List<String> WRITER_PERMISSIONS =
+  private static final List<String> PROJECT_WRITER_PERMISSIONS =
       new ImmutableList.Builder<String>()
-          .addAll(READER_PERMISSIONS)
+          .addAll(PROJECT_READER_PERMISSIONS)
           .add(
               // TODO(wchambers): Revise service account permissions when there are controlled
               // resources for service accounts. (Also used by NextFlow)
@@ -47,9 +47,9 @@ public class CloudSyncRoleMapping {
           .build();
 
   private static final CustomGcpIamRole PROJECT_READER =
-      new CustomGcpIamRole("PROJECT_READER", READER_PERMISSIONS);
+      CustomGcpIamRole.of("PROJECT_READER", PROJECT_READER_PERMISSIONS);
   private static final CustomGcpIamRole PROJECT_WRITER =
-      new CustomGcpIamRole("PROJECT_WRITER", WRITER_PERMISSIONS);
+      CustomGcpIamRole.of("PROJECT_WRITER", PROJECT_WRITER_PERMISSIONS);
   // Currently, workspace editors, applications and owners have the same cloud permissions as
   // writers. If that changes, create a new CustomGcpIamRole and modify the map below.
   public static final ImmutableMap<WsmIamRole, CustomGcpIamRole> CUSTOM_GCP_PROJECT_IAM_ROLES =
--- a/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
@@ -1,6 +1,7 @@
 package bio.terra.workspace.service.workspace;
 
 import bio.terra.workspace.service.iam.model.WsmIamRole;
+import bio.terra.workspace.service.resource.controlled.mappings.CustomGcpIamRole;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


zloery:
connected and integration tests all (eventually) passed locally. I think timeout issues are still happening so I'll rerun the tests a couple times, but this is ready for review

ddietterich (approved):
LGTM

ddietterich:
The failures match ones I saw in the overnight test runs. It smells like IAM propagation delays.
We may want to put retries around the access checks in the tests to be more patient. I'll make a ticket for that.

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -11,34 +12,52 @@
...
+ * at the project level, and will continue to live here.
  */
 public class CloudSyncRoleMapping {
-  // Note that bigquery.jobUser is required at the project level, unlike other permissions which can
-  // be per-dataset.
+
+  // Note that custom roles defined at the project level cannot contain the
+  // "resourcemanager.projects.list" permission, even though it was previously included here.
+  // See https://cloud.google.com/iam/docs/understanding-custom-roles#known_limitations
   private static final List<String> READER_PERMISSIONS =
```
jaycarlton, line 23: Maybe name these `PROJECT_READER_PERMISSIONS`, same for writer

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -11,34 +12,52 @@
...
+  // writers. If that changes, create a new CustomGcpIamRole and modify the map below.
+  public static final ImmutableMap<WsmIamRole, CustomGcpIamRole> CUSTOM_GCP_PROJECT_IAM_ROLES =
       ImmutableMap.of(
-          // TODO: this should map to OWNER_PERMISSIONS if that's created.
-          WsmIamRole.OWNER, WRITER_PERMISSIONS,
-          // TODO: this should map to APPLICATION_PERMISSIONS if that's created.
-          WsmIamRole.APPLICATION, WRITER_PERMISSIONS,
-          WsmIamRole.WRITER, WRITER_PERMISSIONS,
-          WsmIamRole.READER, READER_PERMISSIONS);
+          // TODO: this should map to PROJECT_OWNER if that's created.
```
jaycarlton, line 57: might be simpler to make a `PROJECT_OWNER` and put this comment there. Same for `PROJECT_APPLICATION`.

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateCustomGcpRolesStep.java
```diff
@@ -9,14 +9,20 @@
...
 import java.io.IOException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
 
+/**
+ * This step creates custom role definitions in our GCP context. It does not grant these roles to
+ * any users, though other steps do.
+ */
 public class CreateCustomGcpRolesStep implements Step {
```
jaycarlton, line 26: Aside: for steps that are only meant to be used  in one flight, I wonder if it would help organization to make them static inner classes of the flights themselves. Or adopt a kludgy naming convention like `MyFlight_CreateRolesStep`.

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateCustomGcpRolesStep.java
```diff
@@ -31,37 +37,54 @@ public CreateCustomGcpRolesStep(IamCow iamCow) {
...
+    // Second, create the resource-level custom roles.
+    for (CustomGcpIamRole customResourceRole :
+        CustomGcpIamRoleMapping.CUSTOM_GCP_RESOURCE_IAM_ROLES.values()) {
+      createCustomRole(customResourceRole, projectId);
     }
     return StepResult.getStepResultSuccess();
   }
 
+  /**
+   * Utility for creating custom roles in GCP from  WSM's CustomGcpIamRole objects.
```
jaycarlton, line 58: Indicate scope (e.g. Project).

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateCustomGcpRolesStep.java
```diff
@@ -31,37 +37,54 @@ public CreateCustomGcpRolesStep(IamCow iamCow) {
...
+          customRole.getRoleName(),
+          customRole.getIncludedPermissions(),
+          projectId);
+      iamCow.projects().roles().create("projects/" + projectId, request).execute();
+    } catch (GoogleJsonResponseException googleEx) {
+      // Because this step may run multiple times, we need to handle duplicate role creation.
+      // The project was retrieved from RBS earlier in this flight, so we assume that any conflict
+      // of role names must be due to duplicate step execution.
+      if (googleEx.getStatusCode() != HttpStatus.CONFLICT.value()) {
+        throw new RetryException(googleEx);
```
jaycarlton, line 80: What's the difference in throwing this exception and returning `STEP_RESULT_FAILURE_RETRY`?

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateCustomGcpRolesStep.java
```diff
@@ -31,37 +37,54 @@ public CreateCustomGcpRolesStep(IamCow iamCow) {
...
+  private void createCustomRole(CustomGcpIamRole customRole, String projectId)
+      throws RetryException {
+    try {
+      Role gcpRole =
+          new Role()
+              .setIncludedPermissions(customRole.getIncludedPermissions())
+              .setTitle(customRole.getRoleName());
+      CreateRoleRequest request =
+          new CreateRoleRequest().setRole(gcpRole).setRoleId(customRole.getRoleName());
+      logger.info(
```
jaycarlton, line 69: maybe debug

service/src/main/java/bio/terra/workspace/service/workspace/flight/GcpCloudSyncStep.java
```diff
@@ -63,14 +63,15 @@ public StepResult doStep(FlightContext flightContext)
...
               .projects()
-              .getIamPolicy(gcpProjectName, new GetIamPolicyRequest())
+              .getIamPolicy(gcpProjectId, new GetIamPolicyRequest())
               .execute();
 
       List<Binding> newBindings = new ArrayList<>();
       // Add all existing bindings to ensure we don't accidentally clobber existing permissions.
       newBindings.addAll(currentPolicy.getBindings());
+      // Add appropriate project-level roles for each WSM IAM role.
       for (Map.Entry<WsmIamRole, String> entry : workspaceRoleGroupsMap.entrySet()) {
```
jaycarlton, line 73: `Map#forEach()` can be easier to read

service/src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
```diff
@@ -148,7 +148,8 @@ private static FlightMap createInputParameters(
    * CustomGcpIamRoleMapping}
    */
   private void assertRolesExist(Project project) throws IOException {
-    for (CustomGcpIamRole customRole : CustomGcpIamRoleMapping.CUSTOM_GCP_IAM_ROLES.values()) {
+    for (CustomGcpIamRole customRole :
+        CustomGcpIamRoleMapping.CUSTOM_GCP_RESOURCE_IAM_ROLES.values()) {
       String fullRoleName =
           "projects/" + project.getProjectId() + "/roles/" + customRole.getRoleName();
```
jaycarlton, line 154: make helper methods to build the FQ names for roles and projects

jaycarlton (approved):

service/src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
zloery: turns out there's already a helper CustomGcpIamRole. nice catch!

service/src/main/java/bio/terra/workspace/service/workspace/flight/GcpCloudSyncStep.java
zloery: it's a lot nicer having named variables instead of `entry.getKey()`. Thanks!

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
zloery: That's true, but it would also create two extra custom roles in each project which wouldn't be used. I think that would end up being more confusing in the end

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateCustomGcpRolesStep.java
zloery: As far as I know they behave the same way. I don't know if there's a subtle difference or it's just style

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateCustomGcpRolesStep.java
zloery: For one flight I think we tried making steps like that inner classes inside the flight. It was good for directory organization but got a little messy for nontrivial steps. It also ends up being messier if we're wrong about the step only being part of one flight, since we have to break it out into a separate class. Worth thinking about though

PR #402, opened 2021-07-21
[PF-869] Support polling notebook start/stop operations

zloery:
This change (re)adds support for polling notebook start/stop operations. These permissions must be granted at the project level - we previously granted them on individual notebook instances, which had no effect. This functionality was removed when we moved permissions from broad project-level roles to narrower resource-level roles (#361), though it can be supported again now that we have custom project-level roles (#399).

This change does not address the second part of PF-869 which is focused on notebook usability in the cloud console. That will be handled as part of a design change as we revisit visibility of private resources.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
@@ -25,8 +25,8 @@ public class CloudSyncRoleMapping {
           "bigquery.jobs.create",
           "lifesciences.operations.get",
           "lifesciences.operations.list",
-          "resourcemanager.projects.get",
           "monitoring.timeSeries.list",
+          "resourcemanager.projects.get",
           "serviceusage.operations.get",
           "serviceusage.operations.list",
           "serviceusage.quotas.get",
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -43,7 +43,11 @@
               "iam.serviceAccounts.list",
               "lifesciences.workflows.run",
               "lifesciences.operations.cancel",
-              "serviceusage.services.use")
+              "serviceusage.services.use",
```
jaycarlton, line 46: Could you alphabetize these please?

jaycarlton: It would be nice to have a comment on each permission or group of related ones explaining why we this user category needs it.

jaycarlton (approved):

wchamber (approved):

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
zloery: sorted!
I'm not sure how much comments would add here. All the permissions in these list (minus the already-commented service account ones) are at the project level because they can't be assigned at the resource level, and I think commenting what each permission does in GCP seems like it would be more noise than signal

PR #405, opened 2021-07-23
[no ticket] replace userReq with userRequest

jaycarlton:
Replace all occurrences of `userReq` with `userRequest`. Our style guide doesn't permit this type of abbreviation, and the name is still short enough when spelled out.


ddietterich (approved):
LGTM

zloery (approved):

PR #407, opened 2021-07-26
[PF-903] Replace 15s timeouts in tests with longer polling

zloery:
There are a few spots in our tests where we wait a static 15s for cloud IAM permissions to propagate. According to GCP these changes may take several minutes, and as a result we sometimes get flaky failures for not waiting long enough. This change replaces the static 15s waits with polling every 30s up to a 10 minute timeout, which should reduce flaky test failures.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/ControlledBigQueryDatasetLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/ControlledBigQueryDatasetLifecycle.java
@@ -118,7 +118,7 @@ public class ControlledBigQueryDatasetLifecycle extends WorkspaceAllocateTestScr
     // This is the reader's first use of cloud APIs after being added to the workspace, so we
     // retry this operation until cloud IAM has properly synced.
     var readTable = ClientTestUtils.getWithRetryOnException(() ->
-        readerBqClient.getTable(table.getTableId()), 20, Duration.ofSeconds(30));
+        readerBqClient.getTable(table.getTableId()));
     assertEquals(table, readTable);
     logger.info("Read table {} as workspace reader", tableName);
 
--- a/integration/src/main/java/scripts/testscripts/ControlledBigQueryDatasetLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/ControlledBigQueryDatasetLifecycle.java
@@ -35,6 +35,7 @@ import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.TableInfo;
 import com.google.cloud.bigquery.TableResult;
+import java.time.Duration;
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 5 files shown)


service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -158,10 +157,6 @@ public void createAiNotebookInstanceDo() throws Exception {
         controlledResourceService.getControlledResource(
             workspace.getWorkspaceId(), resource.getResourceId(), user.getAuthenticatedRequest()));
 
-    // TODO(PF-643): this should happen inside WSM.
```
zloery, line 161: Unlike the other waits, this test makes calls to GCP using the original workspace owner. I think the waiting is not necessary in that case, but if we see flakiness in this connected test we should revisit that.

integration/src/main/java/scripts/testscripts/ControlledBigQueryDatasetLifecycle.java
```diff
@@ -117,7 +115,10 @@ protected void doUserJourney(TestUserSpecification testUser, WorkspaceApi worksp
     String tableName = table.getTableId().getTable();
 
     // Workspace reader can read the table
-    var readTable = readerBqClient.getTable(table.getTableId());
+    // This is the reader's first use of cloud APIs after being added to the workspace, so we
+    // retry this operation until cloud IAM has properly synced.
+    var readTable = ClientTestUtils.getWithRetryOnException(() ->
+        readerBqClient.getTable(table.getTableId()), 20, Duration.ofSeconds(30));
```
jaycarlton, line 121: i wonder if we shouldn't use an exponential backoff, assuming the majority of cases happen within the first 15-30 seconds.

integration/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -239,7 +244,7 @@ public static boolean jobIsRunning(JobReport jobReport) {
    * @throws InterruptedException
    */
   public static @Nullable <T> T getWithRetryOnException(
-      Supplier<T> supplier,
+      SupplierWithException<T> supplier,
```
jaycarlton, line 247: does this serve to restrict the allowed suppliers? Or is it just for documentation?

jaycarlton (approved):

integration/src/main/java/scripts/utils/ClientTestUtils.java
ddietterich: The default Supplier does not have a throws clause, so cannot be used with any code that throws a checked exception.

integration/src/main/java/scripts/testscripts/ControlledBigQueryDatasetLifecycle.java
ddietterich: Two thoughts here:
 1. All of the calls to the method use the same parameters. Is it worth parameterizing? Maybe just put the values into the `getWithRetryOnException` utility. If it turns out we need different parameters sometime later, we can factor a general method.
 2. Exponential would be incrementally faster. Since this is for test, it doesn't matter a lot. My rough calculations says  if we usually succeed in under 15 seconds, then this will make a regression run take a minute longer.

ddietterich (approved):
LGTM - I do have a suggestion

integration/src/main/java/scripts/testscripts/ControlledBigQueryDatasetLifecycle.java
zloery: sgtm, inlined the parameters into the method. Exponential backoff might be a bit cleaner, but I shortened the poll delay to 15s so this shouldn't add too much additional delay

PR #408, opened 2021-07-26
[PF-746] Remove test user william.thunderlord and integration test profile

zloery:
This change removes our test user William Thunderlord from all test cases. He's a member of too many groups and used in too many places, which causes issues with Google Group syncing (and as a consequence, cloud IAM).

While removing him, I also deleted the configuration for the `integration-test` profile. This was previously used in the non-TestRunner integration tests, but it's now unused. The information included this properties file is already included in TestRunner test configuration files.


integration/src/main/resources/configs/perf/DataReferenceLifecycle.json
```diff
@@ -14,5 +14,5 @@
       "parameters": ["wm-default-spend-profile", "97b5559a-2f8f-4df3-89ae-5a249173ee0c", "terra"]
     }
   ],
-  "testUserFiles": ["william.json"]
+  "testUserFiles": ["bella.json"]
```
jaycarlton, line 17: How do you decide when to use Liam vs Bella?

jaycarlton (approved):

integration/src/main/resources/configs/perf/DataReferenceLifecycle.json
zloery: Bella has access to the dev TDR snapshot we use in tests, so I used her for tests that talk to TDR. Users are pretty interchangeable otherwise, so I picked Liam arbitrarily for the other ones

PR #410, opened 2021-07-29
[PF-909] Make listResources show private resources, support GCP cloud console

zloery:
This change modifies the behavior of WSM's resource enumeration endpoint. Previously, we only showed the controlled resources that users had permissions on, meaning that workspace readers would generally not see private resources. After this change, all members of a workspace will see all resources in that workspace, including private resources which they cannot read/write to. Filter behavior on the endpoint is unchanged, and callers will not see these controlled resources if they only request referenced resources.

Additionally, this adds a number of GCP permissions so that all workspace users can navigate GCS buckets and AI Platform Notebooks directly in the cloud console. This allows workspace users to view metadata on these resources, though they still cannot read/write to private resource if they are not the assigned user, and the cloud console generally gives informative error messages if they do. BigQuery dataset IAM is handled differently by GCP, and as a result users can still only see private datasets in the BigQuery viewer if they are the assigned user. However, they can now see metadata of those resources via the WSM resource enumeration endpoint.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/db/ResourceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/ResourceDao.java
@@ -220,12 +220,10 @@ public class ResourceDao {
     }
 
     // There are three cases for the stewardship type filter
-    // 1. If it is REFERENCED, then we ignore id list and just filter
-    //    for referenced resources.
-    // 2. If it is CONTROLLED, then we filter for
-    //    CONTROLLED resources.
-    // 3. If no filter is specified (it is null), then we want both REFERENCED
-    //    and CONTROLLED resources; that is, we want the OR of 1 and 2
+    // 1. If it is REFERENCED, then we ignore id list and just filter for referenced resources.
+    // 2. If it is CONTROLLED, then we filter for CONTROLLED resources.
+    // 3. If no filter is specified (it is null), then we want both REFERENCED and CONTROLLED
+    //    resources; that is, we want the OR of 1 and 2
     boolean includeReferenced = (stewardshipType == null || stewardshipType == REFERENCED);
     boolean includeControlled = (stewardshipType == null || stewardshipType == CONTROLLED);
 
--- a/service/src/main/java/bio/terra/workspace/db/ResourceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/ResourceDao.java
@@ -166,19 +166,14 @@ public class ResourceDao {
    * Resource enumeration
    *
    * <p>The default behavior of resource enumeration is to find all resources that are visible to
-   * the caller. If the caller has gotten this far, then they are allowed to see all referenced
-   * resources. We know which controlled resources they are allowed to see from the list provided as
-   * input.
+   * the caller. If the caller has gotten this far, then they are allowed to see all referenced and
+   * controlled resources.
    *
    * <p>The enumeration can be filtered by a resource type. If a resource type is specified, then
-   * only that type of resource is returned.
-   *
-   * <p>The enumeration can also be filtered by a stewardship type. The implementation of the
-   * stewardship type filter is more complex than simply filtering by type. That is because the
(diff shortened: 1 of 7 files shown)


service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -187,16 +182,13 @@ private boolean deleteResourceInner(UUID workspaceId, UUID resourceId) {
...
       int limit)
       throws InterruptedException {
     return DbRetryUtils.retry(
-        () ->
-            enumerateResourcesInner(
-                workspaceId, controlledResourceIds, resourceType, stewardshipType, offset, limit));
+        () -> enumerateResourcesInner(workspaceId, resourceType, stewardshipType, offset, limit));
   }
 
   @Transactional(
```
jaycarlton, line 194: We still have transactional on a private method. Is there a bug for this?

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -231,28 +222,23 @@ private boolean deleteResourceInner(UUID workspaceId, UUID resourceId) {
...
     boolean includeReferenced = (stewardshipType == null || stewardshipType == REFERENCED);
-    boolean includeControlled =
-        ((controlledResourceIds != null && !controlledResourceIds.isEmpty())
-            && (stewardshipType == null || (stewardshipType == CONTROLLED)));
+    boolean includeControlled = (stewardshipType == null || stewardshipType == CONTROLLED);
 
     final String referencedPhrase = "stewardship_type = :referenced_resource";
-    final String controlledPhrase =
-        "(stewardship_type = :controlled_resource AND resource_id IN (:id_list))";
+    final String controlledPhrase = "stewardship_type = :controlled_resource";
```
jaycarlton, line 233: What does the colon do here?

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
zloery: yep, that's tracked in PF-885, and I'd expect we probably pull that in within ~ a sprint. I didn't want to change just this method since it's an issue with all the DAO methods, and we might pull in a common-lib solution.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
zloery: It's a query parameter referring to the value `CONTROLLED.toSql()` that we put in the map earlier. I like that a bit more than string concatenation here, but it's mostly style/convenience

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
ddietterich: Yes, we should switch to the common-lib version.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -231,28 +222,23 @@ private boolean deleteResourceInner(UUID workspaceId, UUID resourceId) {
     // There are three cases for the stewardship type filter
     // 1. If it is REFERENCED, then we ignore id list and just filter
     //    for referenced resources.
-    // 2. If it is CONTROLLED, and the id list is not empty, then we filter for
-    //    CONTROLLED and require that the resources be in the id list.
+    // 2. If it is CONTROLLED, then we filter for
+    //    CONTROLLED resources.
```
ddietterich, line 226: micro-nit: line break

ddietterich (approved):
LGTM

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
jaycarlton: Oh right, it's a replacement token.

PR #411, opened 2021-08-03
[PF-814] Copy BigQuery Dataset Data

jaycarlton:
Use a copy table job to copy each table from a source to a destination dataset. This proved to be simpler and more reliable than the original plan of using the BigQuery Data Transfer Service.


service/src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -128,6 +132,24 @@ public BigQueryCow createBigQueryCow(AuthenticatedUserRequest userRequest) {
     }
   }
 
+  /**
+   * Create a vanilla Bigquery client object, for testing things that aren't in CRL yet.
```
ddietterich, line 136: Do you want to add another Jira ticket for CRL-ification of this or catch it under the current (PF-906)?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
```diff
@@ -0,0 +1,103 @@
...
+  public CompleteTableCopyJobsStep(CrlService crlService) {
+    this.crlService = crlService;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    final FlightMap workingMap = flightContext.getWorkingMap();
+    final Map<String, String> tableToJobId =
+        workingMap.get(ControlledResourceKeys.TABLE_TO_JOB_ID_MAP, new TypeReference<>() {});
```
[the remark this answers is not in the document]
ddietterich: Yup!

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
```diff
@@ -0,0 +1,103 @@
...
+    final AuthenticatedUserRequest userRequest =
+        flightContext
+            .getInputParameters()
+            .get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    final Bigquery bigQueryClient = crlService.createNakedBigQueryClient(userRequest);
```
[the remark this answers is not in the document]
ddietterich: as above, maybe track with a TODO and a Jira ticket

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
```diff
@@ -0,0 +1,100 @@
...
+                  .get(jobReference.getProjectId(), jobReference.getJobId())
+                  .setLocation(
+                      jobReference.getLocation()) // returns NOT_FOUND unless location is specified
+                  .execute();
+          final String jobState = job.getStatus().getState();
+          logger.debug("Table {} is {}", tableReference.getTableId(), jobState);
+          if ("DONE".equals(jobState)) {
+            break;
+          }
+          TimeUnit.SECONDS.sleep(5);
```
ddietterich, line 62: Maybe a static constant for the sleep time?
What led you to pick 5 seconds?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CopyBigQueryDatasetDefinitionStep.java
```diff
@@ -117,8 +117,8 @@ public StepResult doStep(FlightContext flightContext)
             .sourceWorkspaceId(sourceDataset.getWorkspaceId())
             .sourceResourceId(sourceDataset.getResourceId());
     workingMap.put(ControlledResourceKeys.CLONE_DEFINITION_RESULT, apiResult);
-    if (CloningInstructions.COPY_DEFINITION.equals(effectiveCloningInstructions)) {
-      // we're done
+    if (CloningInstructions.COPY_DEFINITION.equals(effectiveCloningInstructions)
+        || CloningInstructions.COPY_RESOURCE.equals(effectiveCloningInstructions)) {
```
ddietterich, line 121: nit: maybe a comment about why you can set the response early for the COPY_RESOURCE case?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
```diff
@@ -0,0 +1,194 @@
...
+      CrlService crlService,
+      WorkspaceService workspaceService,
+      ControlledBigQueryDatasetResource sourceDataset) {
+    this.crlService = crlService;
+    this.workspaceService = workspaceService;
+    this.sourceDataset = sourceDataset;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
```
ddietterich, line 52: If we have a pod crash in the middle of launching the jobs, on restart this step would start the same copy jobs over. IIUC, the WRITE_TRUNCATE would empty the table before the copy. Does BQ make some guarantee about only one copy job running on one table at a time?  If not, then might still get duplicate rows.

In any case, could you capture the failure-recovery design in a comment in here so the reasoning is clear to The Unknown Future Maintainer?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/DatasetCloneInputs.java
```diff
@@ -0,0 +1,50 @@
...
+        + '\''
+        + '}';
+  }
+
+  /**
+   * The dataset ID includes the project name.
+   *
+   * @return
+   */
+  public String getDatasetId() {
```
ddietterich, line 47: If you serialize this class, you will get a datasetId field. For derived fields, it would be good hygiene to add a `@JsonIgnore` to the getter.

ddietterich (commented):
This is looking good. I have one concern about restartability of the copy step. The rest of my comments are minor.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
```diff
@@ -0,0 +1,100 @@
...
+    return StepResult.getStepResultSuccess();
+  }
+
+  // Fully-qualified table IDs are of the form project-id:dataset_id.table_id
+  private TableReference tableFqIdToReference(String tableFqId) {
+    final TableReference result = new TableReference();
+    final String[] outerGroups = tableFqId.split(":");
+    result.setProjectId(outerGroups[0]);
+
+    final String[] innerGroups = outerGroups[1].split("\\.");
```
zloery, line 85: Might be worth confirming in a comment here that GCP disallows ":" and "." characters in names for all of projects, datasets, and tables.
(From what I can tell they do, though a lot more than I would expect is allowed in table names: https://cloud.google.com/bigquery/docs/tables#table_naming)

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
```diff
@@ -0,0 +1,100 @@
...
+          final Job job =
+              bigQueryClient
+                  .jobs()
+                  .get(jobReference.getProjectId(), jobReference.getJobId())
+                  .setLocation(
+                      jobReference.getLocation()) // returns NOT_FOUND unless location is specified
+                  .execute();
+          final String jobState = job.getStatus().getState();
+          logger.debug("Table {} is {}", tableReference.getTableId(), jobState);
+          if ("DONE".equals(jobState)) {
```
zloery, line 59: Does `DONE` cover both job success and failure?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
```diff
@@ -0,0 +1,194 @@
...
+
+    final DatasetCloneInputs destinationInputs = getDestinationInputs(flightContext);
+    workingMap.put(ControlledResourceKeys.DESTINATION_CLONE_INPUTS, destinationInputs);
+
+    final AuthenticatedUserRequest userRequest =
+        flightContext
+            .getInputParameters()
+            .get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    final BigQueryCow bigQueryCow = crlService.createBigQueryCow(userRequest);
+    // TODO:  remove usage of this client when it's all in CRL
```
zloery, line 67: nit: add tracking ticket or name to TODO

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
```diff
@@ -0,0 +1,194 @@
...
+          bigQueryCow
+              .tables()
+              .list(sourceInputs.getProjectId(), sourceInputs.getDatasetName())
+              .execute();
+      // Start a copy job for each source table
+      final Map<String, String> tableToJobId =
+          Optional.ofNullable(
+                  workingMap.get(
+                      ControlledResourceKeys.TABLE_TO_JOB_ID_MAP,
+                      new TypeReference<Map<String, String>>() {}))
```
zloery, line 81: I'm not sure I understand what this is doing. Is this attempting to read an existing map stored in the `workingMap`? If so, when is that populated? AFAIK working map values are only persisted to disk between steps, so this will always be the default value

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
```diff
@@ -0,0 +1,194 @@
...
+              .orElseGet(HashMap::new);
+      for (TableList.Tables table : sourceTables.getTables()) {
+        if (tableToJobId.containsKey(table.getId())) {
+          // A job already exists for this table
+          continue;
+        }
+        // If the table contains data in its streaming buffer, that data can't be copied yet
+        // https://cloud.google.com/bigquery/streaming-data-into-bigquery#dataavailability
```
zloery, line 89: TIL, that's an interesting interaction.
It seems like if the buffer is not empty, we could either copy whatever we can, or throw an error. Do we have a strong case for either one?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
```diff
@@ -0,0 +1,194 @@
...
+          logger.warn(
+              "Streaming buffer data in table {} will not be copied. Estimated rows: {}, Oldest entry time: {}",
+              table.getId(),
+              tableGetResponse.getStreamingBuffer().getEstimatedRows(),
+              Instant.ofEpochMilli(
+                      tableGetResponse.getStreamingBuffer().getOldestEntryTime().longValueExact())
+                  .toString());
+        }
+        final Job inputJob = buildTableCopyJob(sourceInputs, destinationInputs, table);
+        // bill the job to the source project
```
zloery, line 112: Should we bill the source or destination project? I'm thinking in terms of featured workspaces/resources, where (I think) it would make more sense for the destination to pay. That's probably a design-doc level discussion though

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
```diff
@@ -0,0 +1,100 @@
...
+            .getInputParameters()
+            .get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    final Bigquery bigQueryClient = crlService.createNakedBigQueryClient(userRequest);
+    try {
+      for (Map.Entry<String, String> entry : tableToJobId.entrySet()) {
+        final TableReference tableReference = tableFqIdToReference(entry.getKey());
+        final JobReference jobReference = jobFqIdToReference(entry.getValue());
+
+        // wait for job to complete
+        while (true) {
```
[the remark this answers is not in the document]
ddietterich: The problem with any timeout is that the elapsed time is dependent on the data size and location. Very difficult to predict what the timeout should be.

In the context of Stairway, I have been thinking about how to expose progress to the waiting client and how to allow cancel. Progress reporting assumes the underlying operation exposes progress. I don’t know if there is any way to determine progress of a copy job.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
ddietterich: Yes, indeed. I should have caught that! :(

One way to handle it would be to build the map in a separate step of the flight.
Then have a step that iterates through the map starting the jobs. The step would start one job, update the map, and return RERUN status to Stairway. Stairway will persist the flight state and then rerun the same step. The step would find the first, not-yet-started job and start it. We return RERUN until there are no more jobs to start. Then we return SUCCESS and move to the waiting for completion step. 

Another way would be to make a table in the WSM database to hold the state.

In all cases, I think there is a small window where a failure would cause a duplicate copy job to be submitted.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
ddietterich: I think the case where there is pending streaming input happening AND a clone happening is unlikely.
The main use case for clone is copying a featured workspace. Those won’t have streaming data pouring in.
My suggestion is that we document this behavior and move on.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
ddietterich: Yes, we should make the destination pay.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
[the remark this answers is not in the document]
jaycarlton: PF-942

integration/src/main/java/scripts/testscripts/CloneBigQueryDataset.java
```diff
@@ -94,29 +172,30 @@ protected void doUserJourney(TestUserSpecification testUser, WorkspaceApi worksp
     final String clonedDatasetDescription = "Clone of " + destinationDatasetName;
     final String jobId = UUID.randomUUID().toString();
     final CloneControlledGcpBigQueryDatasetRequest cloneRequest = new CloneControlledGcpBigQueryDatasetRequest()
-        .cloningInstructions(CloningInstructionsEnum.DEFINITION)
+        .cloningInstructions(CloningInstructionsEnum.RESOURCE)
```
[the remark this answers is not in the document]
jaycarlton: Not yet. I need to think about the best way to do that so I don't have a lot of redundancy with this test. It would be nice to have it in the CLI for that kind of test.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
jaycarlton: It was the lowest I've seen used. We could go smaller. Most in-region copies will happen almost instantly I'm hoping.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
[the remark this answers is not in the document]
jaycarlton: I'd most likely define progress as a percentage of tables copied, which is crude obviously as they can vary by 3 orders of magnitude or more in size. My original approach was going to be to sample all the jobs each time through the loop and keep a map of which ones are still running and watch that shrink. No need for something like that if we're not doing a progress indication.

Is there a master timeout for flights or something that would kill this?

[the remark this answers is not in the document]
ddietterich: There is no master timeout for flights. There is nothing that would kill this.

I have a half-baked idea for a flight cancel method in Stairway. Maybe I should up the priority of that.
That said, I think we should leave that as an open problem and not put a bound on this for now.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
jaycarlton: What values go in the map before the jobs are started? I don't have a way to create a copy job without starting it. I could make a list of tables in an earlier step if that's helpful.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
ddietterich: As @zloery points out, the working map is never persisted, so on restart we would create a new one.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
jaycarlton: I agree that this is a BigQuery limitation, which can also be seen in the Console, and we don't need to try to do anything about it but warn folks.

I still kinda want to tack this warning message onto the output somehow, but it seems like the restructuring necessary would be ugly.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
ddietterich: I think you could make a List of a POJO holding `(String table name, @Nullable String jobId)` where all the job ids would start null.

Each time you entered the copy step, you would search the list to find the first null job id. You would start that job, copy the job id into the map and return RERUN. The step boundary would persist the list. Stairway would re-run the step.

There is a window where the job is started, but the list is not persisted where we could still get duplicate jobs. There is simply no way to start the copy job and instantly persist the id. (That is why I prefer the client-provides-the-id style)

jaycarlton: If Stairway supported something like `flushWorkingMap()` so that I could force an early persistence of it during the step, I wouldn't need to restructure things to add another step and cut this one up. Would that be something worth adding?

jaycarlton: Sounds good. Implemented using rerun.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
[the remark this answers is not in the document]
jaycarlton: 10-4. leaving as is for now.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
jaycarlton: fixed by using step rerun status after each job is started.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
jaycarlton: Do we have a reusable exponential backoff loop construct?

jaycarlton: implemented

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
```diff
@@ -0,0 +1,218 @@
...
+              .orElseGet(HashMap::new);
+      final List<Tables> tables = sourceTables.getTables();
+      // Find the first table whose ID isn't a key in the map.
+      final Optional<Tables> tableMaybe =
+          tables.stream().filter(t -> !tableToJobId.containsKey(t.getId())).findFirst();
+      if (tableMaybe.isPresent()) {
+        final Tables table = tableMaybe.get();
+        // If the table contains data in its streaming buffer, that data can't be copied yet
+        // https://cloud.google.com/bigquery/streaming-data-into-bigquery#dataavailability
+        final Table tableGetResponse =
```
ddietterich, line 110: minor: it might make the algorithm more readable if this streaming check was moved to a separate `checkStreaming` method.

ddietterich (approved):
I'd like a fix for catching the InterruptedException.
Otherwise, LGTM!

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CompleteTableCopyJobsStep.java
```diff
@@ -59,7 +69,8 @@ public StepResult doStep(FlightContext flightContext)
           if ("DONE".equals(jobState)) {
             break;
           }
-          TimeUnit.SECONDS.sleep(5);
+          TimeUnit.SECONDS.sleep(sleepTimeSeconds);
+          sleepTimeSeconds = Math.min(2 * sleepTimeSeconds, 60);
```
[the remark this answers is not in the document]
jaycarlton: done

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CreateTableCopyJobsStep.java
jaycarlton: good idea

zloery (approved):
LGTM with one open question
nice use of rerun!

jaycarlton:
I think I found a good description [here](https://cloud.google.com/bigquery/docs/running-jobs).

"After calling the jobs.insert method, check the job status by calling jobs.get with the job ID and location, and check the status.state value to learn the job status. When status.state is DONE, the job has stopped running; however, a DONE status does not mean that the job completed successfully, only that it is no longer running."

"Check for job success. If the job has an errorResult property, the job has failed. The status.errorResult property holds information describing what went wrong in a failed job. If status.errorResult is absent, the job finished successfully, although there might have been some non-fatal errors, such as problems importing a few rows in a load job. Non-fatal errors are returned in the job's status.errors list."

PR #412, opened 2021-08-03
[PF-703] Consume Stairway serialization/deserialization updates.

jmczerk:
This change begins to consume the newest Stairway library, which does `FlightMap` parameter serialization/deserialization at get/put time and does not cache `Object` references in the `FlightMap`.  This exposed some places where we were dependent on this caching as ser/des was not a tested path, and in some cases does not work with current Stairway library.

Main changes forced by this change are:
- Use new Stairway `get()` overload taking a `TypeReference<T>` instead of `Class<T>`, which is required for serialization of parameterized types.  This was required for several `Map<enum, V>` serialized as Flight parameters.
- Jackson serialization in Stairway drops time zone info for `OffsetDateTime`.  Working around this by normalizing to UTC in test comparison, but this requires further discussion.

~~Making this draft for now, as it depends on [this Stairway PR](https://github.com/DataBiosphere/stairway/pull/86).~~


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateBigQueryDatasetStep.java
```diff
@@ -120,7 +121,7 @@ public StepResult doStep(FlightContext flightContext)
     // have a cleaner way of deserializing parameterized types, so we suppress warnings here.
```
jmczerk, line 121: This comment needs to be fixed up (the `TypeReference<T>` is the "cleaner way").  Not sure about the `@SuppressWarnings`, overload can still return `null`.

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResourceTest.java
```diff
@@ -58,7 +58,7 @@ public void testFlightMapSerialization() {
     serializeMap.put("resource", resource);
 
     FlightMap deserializedMap = new FlightMap();
-    deserializedMap.fromJson(serializeMap.toJson());
```
jmczerk, line 61: I think this meets the spirit of what the test was trying to accomplish...

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/UpdateGcsBucketStepTest.java
```diff
@@ -105,7 +105,9 @@ public void testDoStep() throws InterruptedException, RetryException {
 
     final LifecycleCondition rule1condition = rules.get(0).getCondition();
     assertEquals(31, rule1condition.getAge());
-    assertEquals(toGoogleDateTime(OFFSET_DATE_TIME_2), rule1condition.getCreatedBefore());
```
jmczerk, line 108: This is the issue I discussed offline: Jackson serialization serializes `OffsetDateTime` as a double-precision UTC epoch, effectively dropping the time zone info.  Jackson can be configured to keep time zone info, but it changes the representation from a double to a string, so is completely non-upward-compatible.

Note that this test change normalizes both time stamps to a UTC epoch (moral equivalent to `Instant`) and compares.  If we need to keep time zone on-disk we will need to do something special.

jmczerk:
CC: @wchamber as an FYI

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResourceTest.java
wchamber: I think just testing `.put` and `.get` is sufficient now actually. The point of this test is to verify that the `ControlledAiNotebookInstanceResource` can be serdes'd correctly by flightmap. Since flight map serialization happens at `put` and `get` now (instead of previously at `toJson -> fromJson`), we don't actually need the raw conversions.
But maybe that's not a part of the flightmap contract. Then again, I'm not sure that this use of `put -> getRaw -> putRaw -> get` is either.

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/UpdateGcsBucketStepTest.java
wchamber: FWIW, I still think it's incorrect to put OffsetDateTime in a a flightmap if it's silently dropping the timezone info, but maybe I've missed more follow up discussion.
I would rather us do explicit, custom serialization of the offending class, or use a type that can be serdes'd correctly by jackson. It's kind of surprising to me that jackson doesn't really support java8 time classes.

ddietterich: I agree that the OffsetDateTime is incorrect; the test doesn't matter so much, but the production code should be fixed.

Jackson does support Java8 time classes. John mentioned this in another communication. To get it to work, we would need to enable WRITE_DATES_AS_TIMESTAMPS. The risk of adding that is compatibility with existing serialized data. 

I added PF-933 to track the work to fix it, leaving it open whether to hand-fix the OffsetDateTime or reviewing existing Stairway users to see if there would be compatibility issues with enabling WRITE_DATES_AS_TIMESTAMPS.

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResourceTest.java
jmczerk: OK, my thinking was that this was explicitly testing that Stairway's JSON representation serialized and deserialized, but since the intent is to just make sure `FlightMap` can handle it, `put/get` make more sense (and read better).

service/src/main/java/bio/terra/workspace/service/workspace/flight/GcpCloudSyncStep.java
```diff
@@ -54,11 +55,11 @@ public StepResult doStep(FlightContext flightContext)
       throws InterruptedException, RetryException {
     String gcpProjectId = flightContext.getWorkingMap().get(GCP_PROJECT_ID, String.class);
     FlightMap workingMap = flightContext.getWorkingMap();
-    // Read Sam groups for each workspace role. Stairway does not
-    // have a cleaner way of deserializing parameterized types, so we suppress warnings here.
+    // Read Sam groups for each workspace role.
     @SuppressWarnings("unchecked")
     Map<WsmIamRole, String> workspaceRoleGroupsMap =
-        workingMap.get(WorkspaceFlightMapKeys.IAM_GROUP_EMAIL_MAP, Map.class);
+        workingMap.get(WorkspaceFlightMapKeys.IAM_GROUP_EMAIL_MAP, new TypeReference<>() {});
```
ddietterich, line 61: What happens if you remove the `@SuppressWarnings` here?

ddietterich (approved):
LGTM - one question

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResourceTest.java
```diff
@@ -54,14 +54,13 @@ public void testFlightMapSerialization() {
     ControlledAiNotebookInstanceResource resource =
         ControlledResourceFixtures.makeDefaultAiNotebookInstance().build();
 
-    FlightMap serializeMap = new FlightMap();
-    serializeMap.put("resource", resource);
```
jmczerk, line 58: @wchamber : Does this address the concern well enough?

service/src/main/java/bio/terra/workspace/service/workspace/flight/GcpCloudSyncStep.java
jmczerk: They no longer appear... have removed them.

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/UpdateGcsBucketStepTest.java
```diff
@@ -119,7 +121,9 @@ public void testDoStep() throws InterruptedException, RetryException {
 
     final LifecycleCondition rule2condition = rules.get(1).getCondition();
     assertEquals(15, rule2condition.getAge());
-    assertEquals(toGoogleDateTime(OFFSET_DATE_TIME_1), rule2condition.getCreatedBefore());
+    assertEquals(
```
jaycarlton, line 124: make a comment and reference PF-933 about the timestamps

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateBigQueryDatasetStep.java
```diff
@@ -116,20 +117,18 @@ public StepResult doStep(FlightContext flightContext)
     GcpPolicyBuilder policyBuilder =
         new GcpPolicyBuilder(resource, projectId, Policy.newBuilder().build());
 
-    // Read Sam groups for each workspace role. Stairway does not
-    // have a cleaner way of deserializing parameterized types, so we suppress warnings here.
-    @SuppressWarnings("unchecked")
+    // Read Sam groups for each workspace role.
     Map<WsmIamRole, String> workspaceRoleGroupMap =
-        workingMap.get(WorkspaceFlightMapKeys.IAM_GROUP_EMAIL_MAP, Map.class);
+        workingMap.get(WorkspaceFlightMapKeys.IAM_GROUP_EMAIL_MAP, new TypeReference<>() {});
```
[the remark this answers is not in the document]
jmczerk: I did try that, but unfortunately it doesn't work that way.

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/UpdateGcsBucketStepTest.java
jmczerk: done

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/UpdateGcsBucketStepTest.java
wchamber: If the change to use `.value` in these tests is what I think it is, consider adding `TODO(PF-933)`'s to change the test from asserting that just the value is equal back to the entire offsetdatetime.

wchamber (approved):

PR #413, opened 2021-08-03
[PF-616] Ensure assignedUser field is always stored, even if not specified

zloery:
This change ensures WSM stores the assigned user of a private resource in its database in all cases, even when `assignedUser` is not specified and we default to assigning the resource to the caller.

This ticket originally covered storing and returning the actual roles that private users held (READER, WRITER, EDITOR), but I decided against this approach after further discussion.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
@@ -5,6 +5,7 @@ import bio.terra.common.exception.ValidationException;
 import bio.terra.workspace.common.utils.ControllerUtils;
 import bio.terra.workspace.db.exception.InvalidMetadataException;
 import bio.terra.workspace.generated.controller.ControlledGcpResourceApi;
+import bio.terra.workspace.generated.model.ApiAccessScope;
 import bio.terra.workspace.generated.model.ApiCloneControlledGcpBigQueryDatasetRequest;
 import bio.terra.workspace.generated.model.ApiCloneControlledGcpBigQueryDatasetResult;
 import bio.terra.workspace.generated.model.ApiCloneControlledGcpGcsBucketRequest;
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -541,6 +532,22 @@ private ApiCloneControlledGcpBigQueryDatasetResult fetchCloneBigQueryDatasetResu
...
+   * populated for private resources, but if a resource is private then a "null" value means "assign
+   * this resource to the requesting user" and should be populated here.
+   */
+  private String assignedUserFromBodyOrToken(
+      ApiControlledResourceCommonFields commonFields, AuthenticatedUserRequest userRequest) {
+    if (commonFields.getAccessScope() != ApiAccessScope.PRIVATE_ACCESS) {
+      return null;
+    }
+    return Optional.ofNullable(commonFields.getPrivateResourceUser().getUserName())
+        .orElse(
```
jaycarlton, line 546: use `orElseGet` to avoid the Sam call in the case where it's not necessary

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -541,6 +532,22 @@ private ApiCloneControlledGcpBigQueryDatasetResult fetchCloneBigQueryDatasetResu
         result, ControllerUtils.getAsyncResponseCode(result.getJobReport()));
   }
 
+  /**
+   * Extract the assigned user from a request to create a controlled resource. This field is only
+   * populated for private resources, but if a resource is private then a "null" value means "assign
+   * this resource to the requesting user" and should be populated here.
+   */
+  private String assignedUserFromBodyOrToken(
+      ApiControlledResourceCommonFields commonFields, AuthenticatedUserRequest userRequest) {
```
jaycarlton, line 541: from a law of demeter perspective, it's a little odd to pass in the containing object here instead of just passing in the privateResourceUser.

jaycarlton (approved):

ddietterich (approved):
lgtm

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
zloery: huh, I had always assumed the argument to `orElse` was only evaluated if the optional was null. TIL, thanks!

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
zloery: This method is also evaluating the access scope from `commonFields`, since non-private resources don't have an assigned user. It either has to live inside this method or else everywhere it's called.
This could be two arguments instead. (accessScope + privateResourceUser) - I'd prefer to use the combined structure because I think it's a clearer interface and this has limited scope as a private method, but agreed it does make this a little less reusable.

PR #415, opened 2021-08-10
[PF-879] Compensating changes for renaming dev to main

ddietterich:
The changes are to documentation and the GHAs. When I do the branch rename, I will then want to immediately merge this. It cannot really be tested until I do the branch rename. So I'd like to have this reviewed and approved before I do the rename.

### DIFF
--- a/.github/workflows/tag-publish.yml
+++ b/.github/workflows/tag-publish.yml
@@ -23,6 +23,7 @@ on:
   push:
     branches:
     - main
+    - dev
     paths-ignore:
     - 'README.md'
     - '.github/**'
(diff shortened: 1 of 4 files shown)


zloery (approved):
Renaming the branch then immediately merging this sound good to me

.github/workflows/tag-publish.yml
```diff
@@ -22,7 +22,7 @@ name: Bump, Tag, Publish, and Deploy
 on:
   push:
     branches:
-    - dev
+    - main
```
jaycarlton, line 25: You could put both dev and  main in the list for these (removing dev after the smoke clears), in case you had to rename it back for some reason.

jaycarlton (approved):

.github/workflows/tag-publish.yml
ddietterich: Excellent idea. I did that. Now I can pre-merge this and then do the branch switch.

PR #416, opened 2021-08-10
[PF-885] Use common lib implementation of DB retry and transaction logic 

zloery:
Our current approach to handling DB transactions and retries is actually incorrect - the `@Transactional` annotation doesn't work with private methods where we currently use it. This change moves WSM to use the common-library code (which internally uses the Spring `@Retryable` annotation) for transactions and DB retries instead. This also has the side effect of not propagating `InterruptedExceptions` everywhere that touches DAO objects, which is nice.


service/src/test/java/bio/terra/workspace/db/DbRetryTest.java
```diff
@@ -0,0 +1,37 @@
...
+  @Autowired TransactionRetryProperties retryProperties;
+
+  private static final CannotSerializeTransactionException RETRY_EXCEPTION =
+      new CannotSerializeTransactionException("test");
+
+  @BeforeEach
+  public void setup() {
+    fakeDao.reset();
+  }
+
```
ddietterich, line 25: Maybe add test cases for InterruptedException

ddietterich (commented):
A question, a spot I think you missed, and a test suggestion.

service/src/main/java/bio/terra/workspace/service/resource/WsmResourceService.java
```diff
@@ -42,9 +41,7 @@ public WsmResourceService(
     workspaceService.validateWorkspaceAndAction(
         userRequest, workspaceId, SamConstants.SAM_WORKSPACE_READ_ACTION);
 
-    return DbRetryUtils.throwIfInterrupted(
-        () ->
-            resourceDao.enumerateResources(
-                workspaceId, resourceType, stewardshipType, offset, limit));
+    return resourceDao.enumerateResources(
```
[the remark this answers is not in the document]
ddietterich: We do not want flights to fail because we are restarting a pod. That is why I need to make a compensating fix in Stairway.
I just have to figure out how to include the exception check without hauling in piles of Spring stuff...

service/src/test/java/bio/terra/workspace/db/DbRetryTest.java
ddietterich: If the thread is killed, there will be an InterruptedException delivered at some level. Maybe it will show up inside the database client waiting for a postgres response.

I was suggesting throwing one. In theory, the retryable code will catch it and return a BackOffInterruptedException. Right?

ddietterich (approved):
LGTM - just an ignorable test suggestion

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -363,24 +332,59 @@ public boolean updateResource(UUID workspaceId, UUID resourceId, String name, St
...
+  /**
+   * This is an open ended method for constructing the SQL update statement. To use it, build the
+   * parameter list making the param name equal to the column name you want to update. The method
+   * generates the column_name = :column_name list. It is an error if the params map is empty.
+   *
+   * @param params sql parameters
+   * @param workspaceId workspace identifier - not strictly necessarily, but an extra validation
+   * @param resourceId resource identifier
+   */
+  private boolean updateResourceHelper(
```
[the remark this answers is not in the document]
jaycarlton: I'd name it `updateResourceColumns` or something maybe. Rename `params` to `columnParams`.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -363,24 +333,59 @@ public boolean updateResource(UUID workspaceId, UUID resourceId, String name, St
...
+    String[] parameterNames = params.getParameterNames();
+    if (parameterNames.length == 0) {
+      throw new InvalidDaoRequestException("Must specify some data to be updated.");
+    }
+    for (int i = 0; i < parameterNames.length; i++) {
+      String columnName = parameterNames[i];
+      if (i > 0) {
+        sb.append(", ");
+      }
+      sb.append(columnName).append(" = :").append(columnName);
```
jaycarlton, line 361: So in my training yesterday, there was a discussion on interpolating sql strings with the upshot being to make sure they don't have any syntax stuffed in them. But I guess these aren't anywhere near user input.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -363,24 +333,59 @@ public boolean updateResource(UUID workspaceId, UUID resourceId, String name, St
...
+    }
+    sb.append(" WHERE workspace_id = :workspace_id AND resource_id = :resource_id");
+
+    params
```
jaycarlton, line 365: I don't love mutating the input like this. Does this change the caller's copy of the param list? I do this so rarely I forget how it works. I'd make a new param map, copying the column params into it.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -363,24 +333,59 @@ public boolean updateResource(UUID workspaceId, UUID resourceId, String name, St
...
+    StringBuilder sb = new StringBuilder("UPDATE resource SET ");
+
+    String[] parameterNames = params.getParameterNames();
+    if (parameterNames.length == 0) {
+      throw new InvalidDaoRequestException("Must specify some data to be updated.");
+    }
+    for (int i = 0; i < parameterNames.length; i++) {
+      String columnName = parameterNames[i];
+      if (i > 0) {
+        sb.append(", ");
```
jaycarlton, line 359: do you want a comma after the last entry?

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -652,57 +634,11 @@ private WsmResource constructResource(DbResource dbResource) {
     }
   }
 
-  @Transactional(
-      propagation = Propagation.REQUIRED,
-      isolation = Isolation.SERIALIZABLE,
-      readOnly = true)
   private DbResource getDbResource(String sql, MapSqlParameterSource params) {
```
jaycarlton, line 637: Should this one be `@ReadTransaction` too?

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -180,7 +159,44 @@ public boolean updateWorkspace(UUID workspaceId, String name, String description
...
+   * @param params sql parameters
+   */
+  private boolean updateWorkspaceHelper(UUID workspaceId, MapSqlParameterSource params) {
+    StringBuilder sb = new StringBuilder("UPDATE workspace SET ");
+
+    String[] parameterNames = params.getParameterNames();
+    if (parameterNames.length == 0) {
+      throw new MissingRequiredFieldException("Must specify some data to be updated.");
+    }
+    for (int i = 0; i < parameterNames.length; i++) {
```
jaycarlton, line 180: Bust a helper method for building the substring from the column list so at least this much won't need to be duplicated.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
```diff
@@ -52,8 +51,7 @@ public void updateControlledResourceMetadata(
     if (name != null) {
       ValidationUtils.validateResourceName(name);
     }
-    DbRetryUtils.throwIfInterrupted(
-        () -> resourceDao.updateResource(workspaceId, resourceId, name, description));
+    resourceDao.updateResource(workspaceId, resourceId, name, description);
```
jaycarlton, line 54: 👏🏻

jaycarlton (commented):
Love all the red!

service/src/main/java/bio/terra/workspace/app/controller/GlobalExceptionHandler.java
```diff
@@ -53,6 +54,20 @@
...
+  @ExceptionHandler({BackOffInterruptedException.class})
+  public ResponseEntity<ApiErrorReport> retryBackoffExceptionHandler(
+      BackOffInterruptedException ex) {
+    String errorMessage =
+        "Unexpected interrupt while retrying database logic. This may succeed on a retry. "
+            + ex.getMessage();
+    ApiErrorReport errorReport =
+        new ApiErrorReport()
+            .message(errorMessage)
+            .statusCode(HttpStatus.INTERNAL_SERVER_ERROR.value());
```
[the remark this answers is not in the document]
zloery: Yeah, I'm not sure what else to use here. It definitely is a server side issue so it should be 5XX, and I don't see any common error codes that make more sense than a generic 500.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
zloery: Yep - the key here is that we're only using column names (which are hardcoded in `updateResource` above) in this concatenation. the user provided values for `name` and `description` are parameterized in normally.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
zloery: Yep, this mutates the caller's copy. changed

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
zloery: No, and I don't think this construction makes one currently

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
zloery: nope - this is a private method called by `getResource` and `getResourceByName`, both of which are annotated with `@ReadTransaction`.

jaycarlton (approved):

PR #418, opened 2021-08-11
[PF-879] Fix error reporting on start-local-server

ddietterich:
The action code has a scripting bug so it is not dumping the log file when the local WSM fails to come up.
Trying a fix for that and hoping it reveals whether anything is actually wrong with the integration tests post-rename.

I think this is still a good fix, even though it looks like the previous failures were Sam-related.

The problem was that the `timeout` statement would cause the action to terminate. By wrapping it in a subprocess and catching the output, it stops doing that.

I hope that the next time startup-local-server timesout, we will get a log of what went wrong.

### DIFF
--- a/.github/actions/start-local-server/action.yml
+++ b/.github/actions/start-local-server/action.yml
@@ -22,8 +22,7 @@ runs:
     - name: Wait for boot run to be ready
       id: wait-for-ready
       run: |
-        timeout 60 bash -c 'until echo > /dev/tcp/localhost/8080; do sleep 1; done'
-        resultStatus=$(echo > /dev/tcp/localhost/8080; echo $?)
+        resultStatus=$(timeout 60 bash -c 'until echo > /dev/tcp/localhost/8080; do sleep 1; done' && echo $?)
         if [[ $resultStatus == 0 ]]
         then
           echo "Server started successfully"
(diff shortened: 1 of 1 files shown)


.github/actions/start-local-server/action.yml
```diff
@@ -22,8 +22,7 @@ runs:
     - name: Wait for boot run to be ready
       id: wait-for-ready
       run: |
-        timeout 60 bash -c 'until echo > /dev/tcp/localhost/8080; do sleep 1; done'
-        resultStatus=$(echo > /dev/tcp/localhost/8080; echo $?)
+        resultStatus=$(timeout 60 bash -c 'until echo > /dev/tcp/localhost/8080; do sleep 1; done' && echo $?)
```
jaycarlton, line 25: I'm trying to translate into English what this does. You're saying wait up to a minute, checking every second, for localhost:8080 to spit something out. If that works, echo the exit code, which should be zero if it worked. And the thing that changed is that now we're getting the exit status of the first command and not another `echo > /dev/tcp/localhost/8080`.

I'm still confused.

ddietterich: I'll take my own medicine and add some comment. Let me try to explain...

Since `timeout` is a top-level bash command in the `run:`, when it times out it SIGTERMs which terminates the github action.

By putting the `timeout` in a subprocess, it just exits the subprocess. As a result, `resultStatus` is either empty, because the subprocess terminated, or it is 0 if we got an answer back from the echo.

jaycarlton (approved):
Thanks.

zloery (approved):

PR #419, opened 2021-08-12
[PF-945] integration test wait for grant to propagate

ddietterich:
Two changes:
1. Added a wait for the cloning user to be able to read the dataset before proceeding with the user journey
2. Made a general `assertJobSuccess` that will info-log success and error-log the ErrorReport on failure. In the current state, if we assert the job status is SUCCEEDED, then on failure we get no information about why. I applied that change to all places where we were asserting on SUCCEEDED.


jaycarlton (approved):
Thanks.

zloery (approved):

PR #424, opened 2021-08-16
AS-789 followup - do not post on bumps

ungwudik:


.github/workflows/pr-integration.yml
```diff
@@ -96,8 +96,15 @@ jobs:
...
+    - name: Skip version bump merges
+      id: bump-check
+      uses: ./.github/actions/bump-skip
+      with:
+        event-name: ${{ github.event_name }}
+
     - name: "Notify WSM Slack"
-      if: always() && github.event_name == 'push' && github.ref == 'refs/heads/main'
+      # post to WSM Slack when a regular push (i.e. non-bumper push) is made to main branch
+      if: always() && github.event_name == 'push' && github.ref == 'refs/heads/main' && steps.bump-check.outputs.is-bump == 'no'
```
ddietterich, line 107: I do not think you need the `github.ref == 'refs/heads/main'` condition. Testing for `'push'` should be sufficient.

.github/workflows/test.yml
```diff
@@ -172,7 +172,8 @@ jobs:
         fields: repo,job,workflow,commit,eventName,author,took
 
     - name: "Notify WSM Slack"
-      if: always() && github.event_name == 'push' && github.ref == 'refs/heads/main'
+      # post to WSM Slack when a regular push (i.e. non-bumper push) is made to main branch
+      if: always() && github.event_name == 'push' && github.ref == 'refs/heads/main' && steps.skiptest.outputs.is-bump == 'no'
```
ddietterich, line 176: ditto

ddietterich (approved):
Looks fine. Two comments.

.github/workflows/pr-integration.yml
ungwudik: Oh yeah, that's true, since the `on push` trigger is only targeted for main branch.

zloery (approved):

PR #425, opened 2021-08-25
[PF-953] Add listWorkspace integration test

zloery:
This adds an integration test covering WSM's `listWorkspace` endpoint. This also adds the test to the dev, alpha, and staging test suites.


integration/src/main/java/scripts/testscripts/ListWorkspaces.java
```diff
@@ -0,0 +1,93 @@
...
+    assertThat("There must be two test users to run this test",
+        testUsers.size() == 2);
+    // Primary test user (index 0) is handled in the test base.
+    TestUserSpecification nonWorkspaceUser = testUsers.get(1);
+    secondUserApi = ClientTestUtils.getWorkspaceClient(nonWorkspaceUser, server);
+    // We expect test users to never have permanent workspaces. However, they may have some
+    // workspaces due to previous test failures. Before running this test, we clean up all
+    // workspaces for both test users.
+    WorkspaceDescriptionList preTestWorkspaceList = workspaceApi.listWorkspaces(/*offset=*/0, /*limit=*/10000);
+    for (WorkspaceDescription workspace : preTestWorkspaceList.getWorkspaces()) {
```
jaycarlton, line 42: I think we need new test users for this test so there's no interference with other test runs. As written, not only is it not safe to run two instances of this test at once, but it's also possible to kill workspaces needed by someone else running integration tests in the environment.

It would be handy to have the ability to keep a test workspace around for investigation in the cloud console, but that's not possible with Janitor as I understand it.

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
```diff
@@ -0,0 +1,93 @@
...
+      workspaceApi.deleteWorkspace(workspace.getId());
+    }
+    WorkspaceDescriptionList secondUserPreTestWorkspaceList = secondUserApi.listWorkspaces(/*offset=*/0, /*limit=*/10000);
+    for (WorkspaceDescription workspace : secondUserPreTestWorkspaceList.getWorkspaces()) {
+      secondUserApi.deleteWorkspace(workspace.getId());
```
jaycarlton, line 47: I don't know if we have a use case outside of this test, but a batch endpoint would help here.

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
```diff
@@ -0,0 +1,93 @@
...
+ * run against a server at once.
+ */
+public class ListWorkspaces extends WorkspaceAllocateTestScriptBase {
+
+  private UUID workspaceId2;
+  private UUID workspaceId3;
+  private WorkspaceApi secondUserApi;
+
+  @Override
+  protected void doSetup(List<TestUserSpecification> testUsers, WorkspaceApi workspaceApi) throws Exception {
```
jaycarlton, line 32: it helps me to rename `workspaceApi` to `user1WorkspaceApi` or simlar.

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
```diff
@@ -0,0 +1,93 @@
...
+    createWorkspace(workspaceId2, getSpendProfileId(), workspaceApi);
+    workspaceId3 = UUID.randomUUID();
+    createWorkspace(workspaceId3, getSpendProfileId(), workspaceApi);
+  }
+
+  @Override
+  protected void doCleanup(List<TestUserSpecification> testUsers, WorkspaceApi workspaceApi)
+      throws Exception {
+    super.doCleanup(testUsers, workspaceApi);
+    workspaceApi.deleteWorkspace(workspaceId2);
```
jaycarlton, line 62: Why not make a helper method to delete all workspaces for the user and call that here and in the (pre-)setup?

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
```diff
@@ -0,0 +1,93 @@
...
+        Collectors.toList());
+    assertThat(callResultIdList, containsInAnyOrder(equalTo(getWorkspaceId()), equalTo(workspaceId2), equalTo(workspaceId3)));
+
+    // Assert that even with a limit > 3, we still only see the workspaces created here.
+    WorkspaceDescriptionList largeLimitResult = workspaceApi.listWorkspaces(/*offset=*/0, /*limit=*/100);
+    assertEquals(3, largeLimitResult.getWorkspaces().size());
+
+    // Validate that a different user will not see any of these workspaces.
+    WorkspaceDescriptionList secondUserResult = secondUserApi.listWorkspaces(0, 3);
+    assertEquals(0, secondUserResult.getWorkspaces().size());
```
jaycarlton, line 90: nit: assertThat & hasSize are nice

integration/src/main/resources/configs/integration/ListWorkspaces.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "ListWorkspaces",
+  "description": "List workspaces",
+  "serverSpecificationFile": "workspace-dev.json",
```
jaycarlton, line 4: I think we can make all these `workspace-local.json` because this is overridden in the GHAs anyway.

jaycarlton (commented):
How hard is it to allocate new test users so you don't step on  bella and liam?

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
```diff
@@ -0,0 +1,93 @@
...
+      workspaceApi.deleteWorkspace(workspace.getId());
+    }
+    WorkspaceDescriptionList secondUserPreTestWorkspaceList = secondUserApi.listWorkspaces(/*offset=*/0, /*limit=*/10000);
+    for (WorkspaceDescription workspace : secondUserPreTestWorkspaceList.getWorkspaces()) {
+      secondUserApi.deleteWorkspace(workspace.getId());
+    }
+
+    // Set up the test fixture workspaces. This must happen after cleaning up a user's workspaces.
+    super.doSetup(testUsers, workspaceApi);
+    workspaceId2 = UUID.randomUUID();
```
ungwudik, line 52: Nit - I like this comment you added in `deSetup`: `// Primary test user (index 0) is handled in the test base.`
Could you add a similar comment here to account for the first workspace that's created in the base class?

integration/src/main/resources/suites/FullIntegration.json
```diff
@@ -16,6 +16,7 @@
     "integration/EnumerateDataReferences.json",
     "integration/EnumerateResources.json",
     "integration/GetRoles.json",
+    "integration/ListWorkspaces.json",
```
ungwudik, line 19: Could you also add this line to [dev](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/integration/src/main/resources/suites/dev/FullIntegration.json), [alpha](https://github.com/DataBiosphere/terra-workspace-manager/tree/main/integration/src/main/resources/suites/alpha), and [staging](https://github.com/DataBiosphere/terra-workspace-manager/tree/main/integration/src/main/resources/suites/staging) suites? We should be able to avoid duplicates like this when [QA-1460](https://broadworkbench.atlassian.net/browse/QA-1460) is done.

integration/src/main/resources/configs/integration/ListWorkspaces.json
```diff
@@ -0,0 +1,18 @@
...
+    {
+      "name": "ListWorkspaces",
+      "numberOfUserJourneyThreadsToRun": 1,
+      "userJourneyThreadPoolSize": 1,
+      "expectedTimeForEach": 5,
+      "expectedTimeForEachUnit": "SECONDS",
+      "parameters": ["wm-default-spend-profile"]
+    }
+  ],
+  "testUserFiles": ["bella.json", "liam.json"]
```
ungwudik, line 17: I believe liam is registered in relevant Sam environments (particularly alpha and staging), but it's worth checking that bella is also registered.

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
zloery: Yeah, that's a good point about multiple people running tests at once. Even with separate test users I think this would be a hassle, since we generally run large test suites and bad timing could ruin both test runs.
Refactored to be resilient to multiple test runs, though with slightly less coverage

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
zloery: I'm not sure when else we would use that either. Maybe for folders, but that's a pretty underdesigned area.

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
zloery: done

integration/src/main/resources/configs/integration/ListWorkspaces.json
zloery: Good catch, I think that's a better default

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
zloery: done

integration/src/main/resources/suites/FullIntegration.json
zloery: done

integration/src/main/resources/configs/integration/ListWorkspaces.json
zloery: Just double-checked, Liam is registered in both alpha and staging but Bella was only registered in alpha.
Per your other comment, I registered Bella in staging so we can run this in any of those envs.

jaycarlton (approved):

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
```diff
@@ -64,30 +54,40 @@ protected void doCleanup(List<TestUserSpecification> testUsers, WorkspaceApi wor
...
     List<WorkspaceDescription> callResults = new ArrayList<>();
-    callResults.addAll(workspaceApi.listWorkspaces(/*offset=*/0, /*limit=*/1).getWorkspaces());
-    callResults.addAll(workspaceApi.listWorkspaces(/*offset=*/1, /*limit=*/1).getWorkspaces());
-    callResults.addAll(workspaceApi.listWorkspaces(/*offset=*/2, /*limit=*/1).getWorkspaces());
+    callResults.addAll(firstUserApi.listWorkspaces(/*offset=*/0, /*limit=*/pageSize).getWorkspaces());
+    callResults.addAll(firstUserApi.listWorkspaces(/*offset=*/pageSize, /*limit=*/pageSize).getWorkspaces());
+    // pageSize may not be divisible by 3, so cover all remaining workspaces here instead.
+    callResults.addAll(firstUserApi.listWorkspaces(/*offset=*/pageSize*2, /*limit=*/(
+        MAX_USER_WORKSPACES - 2*pageSize)).getWorkspaces());
     List<UUID> callResultIdList = workspaceList.getWorkspaces().stream().map(WorkspaceDescription::getId).collect(
```
ungwudik, line 75: Should the value of `callResultIdList` be fetched from `callResults` instead of `workspaceList.getWorkspaces()…`?

ungwudik (approved):
Approving, with one comment.

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
```diff
@@ -0,0 +1,88 @@
...
+  @Override
+  protected void doSetup(List<TestUserSpecification> testUsers, WorkspaceApi firstUserApi) throws Exception {
+    super.doSetup(testUsers, firstUserApi);
+    assertThat("There must be two test users to run this test",
+        testUsers.size() == 2);
+    // Primary test user (index 0) is handled in the test base.
+    TestUserSpecification nonWorkspaceUser = testUsers.get(1);
+    secondUserApi = ClientTestUtils.getWorkspaceClient(nonWorkspaceUser, server);
+
+    // Set up the test fixture workspaces. This must happen after cleaning up a user's workspaces.
```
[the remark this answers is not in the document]
zloery: whoops, out of date comment.

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
```diff
@@ -0,0 +1,88 @@
...
+    // For the sake of listing "all" of a user's workspaces we must have a maximum upper limit. Test
+    // users are expected to never have permanent workspaces, so this should be more than sufficient.
+    int MAX_USER_WORKSPACES = 10000;
+
+    // While test users should not have any permanent workspaces, they may have some entries from
+    // previous failures or other tests running in parallel. We therefore cannot assume they only
+    // have 3 workspaces.
+    WorkspaceDescriptionList workspaceList = firstUserApi.listWorkspaces(/*offset=*/0, /*limit=*/
+        MAX_USER_WORKSPACES);
+    List<UUID> workspaceIdList = workspaceList.getWorkspaces().stream().map(WorkspaceDescription::getId).collect(
```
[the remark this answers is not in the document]
zloery: it'll be a good day when we have spotless in the `integration` dir

integration/src/main/java/scripts/utils/WorkspaceAllocateTestScriptBase.java
```diff
@@ -71,13 +71,22 @@ public void setParameters(List<String> parameters) throws Exception {
...
       throws Exception {
     workspaceId = UUID.randomUUID();
+    createWorkspace(workspaceId, spendProfileId, workspaceApi);
+  }
+
+  /**
+   * Utility for making the WSM calls to create a workspace. Exposed as protected for test
+   * implementations which need to create additional workspaces.
+   */
+  protected CreatedWorkspace createWorkspace(UUID workspaceId, String spendProfileId, WorkspaceApi workspaceApi) throws Exception {
```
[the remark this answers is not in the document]
zloery: one problem is that we wouldn't have a way to handle workspaces created with the non-default workspaceApi object (e.g. when we use a second test user to make a second workspace)

integration/src/main/java/scripts/testscripts/ListWorkspaces.java
zloery: good catch!

PR #429, opened 2021-09-02
[IA-2861] Azure compute deploy create

jdcanas:
This implements an api to CREATE a deployment in azure. The primary use case for this exercised is to deploy a vm in azure. 

I wrote 2 tests, a unit test and it ran successfully for the internal server logic, and an integration test for the api/controller code. The integration test it doesn't do any verification, and both of these are currently disabled, as the delete logic does not exist in CRL and the necessary credentials are not in vault. Follow-up work to come. Below are screenshots of me running the app and calling the swagger endpoint, with manual verification of the azure deployment. 

image
image

This is my first time in this repo, quite possible I missed some consistency/convention issues that need addressing. This is not mergeable until the corresponding CRL PR is merged, and was developed using the package in local maven repo. 

Some call outs
- I don't handle any sort of azure credential setup at this stage. For the test, I took a managed application/resource group rob set up and plugged in the secrets/names as hardcoded strings. For this PR, I have removed the relevant secrets. Later on, we will need to set up the appropriate credentials/secret in vault, and perhaps use a different azure resource group. There's a ticket for this.
- The unit and integration tests really aren't that interesting yet. They're more or less the same as the CRL tests, just with an API layer of code wrapped around them to test. More interesting use-case specific logic will likely be layered in later. The integration test doesn't actually do verification, just calls the API. 
- I chose to take the raw template string in the API request. I considered taking a URI, but I think it makes more sense this way. It makes it bulky to call manually as a user since it involves copious amount of quote escaping, but it makes it a lot easier to use programatically since the the content of a file is irrelevant when reading as a stream. There's the option of allowing a URI specification for a template, but this required more code to read from a bucket + possible credential set-up. Something to note as an add-on later if desired.
- This is mentioned a few times, but just to be clear, the API and code has no clean-up counterpart. The values we've been passing around to test point to the same resource group to allow easy manual cleanup, since we won't have any automated methods of creation until deletion is in place.


ddietterich:
I don't think this code belongs in the mainline branch of WSM. It should not be approved or merged.

1. This interface is at the wrong level of abstraction. WSM is not a vehicle for deploying azure ARM templates. It manages the lifecycle and access control of cloud resources. Use of an azure deployment will surely be part of that, but it will not be part of the REST API of the service.
2. I believe this work is proof-of-concept. If you are going to add production code, then you need to provide design documentation and get it reviewed. I don't think the work is far enough along for that.

I am planning a WSM PoC to be done on a branch that will never be merged into main. See [WSM for Azure PoC](https://docs.google.com/document/d/1N3FlbnLw8LotLQ4bKRB8vEkJlbAF3evuz8bK9ZV-PUg/edit#). It would be worth spending a bit of time to understand if we can collaborate on that branch or if you need your own for messing around.

PR #430, opened 2021-09-07
[No ticket] Fix flakiness in PrivateControlledAiNotebookInstanceLifecycle test

zloery:
The integration test `PrivateControlledAiNotebookInstanceLifecycle` sometimes fails because GCP may respond with either a 403 or 404 when we attempt to fetch a deleted notebook, depending on timing. This change fixes the test to accept either response instead of just a 404.
I've also added messages to the assertions in this test to make debugging a little clearer.


jaycarlton (approved):
Thank you for jumping on this!

ddietterich (approved):
LGTM thanks!

PR #431, opened 2021-09-09
[PF-937] Handle removing users from private resources

zloery:
This change revokes users' access to private resources when they are removed from a workspace (i.e. when their last role in a workspace is revoked). Previously, users would maintain private resource access even when they were no longer in the workspace.

This can only handle cases where users are direct members of a workspace. WSM does not manage groups, and so cannot remove users from a workspace if they have access via a group. However, WSM can check that users have access, and will not revoke private resource access if a user still has workspace access via a group.

As part of this change, I also made two unrelated fixes:
- Controlled GCS buckets now have uniform project level access enabled. This is in line with the Terra IAM model of resources as the most granular permissions.
- Access to private Notebook service accounts now happens via Sam group instead of directly granting users cloud permissions. This is in line with how other controlled resources manage GCP permissions.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/PrivateControlledAiNotebookInstanceLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/PrivateControlledAiNotebookInstanceLifecycle.java
@@ -6,45 +6,32 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import bio.terra.testrunner.runner.config.TestUserSpecification;
 import bio.terra.workspace.api.ControlledGcpResourceApi;
 import bio.terra.workspace.api.WorkspaceApi;
 import bio.terra.workspace.client.ApiException;
-import bio.terra.workspace.model.AccessScope;
-import bio.terra.workspace.model.CloningInstructionsEnum;
-import bio.terra.workspace.model.ControlledResourceCommonFields;
-import bio.terra.workspace.model.ControlledResourceIamRole;
-import bio.terra.workspace.model.CreateControlledGcpAiNotebookInstanceRequestBody;
 import bio.terra.workspace.model.CreatedControlledGcpAiNotebookInstanceResult;
 import bio.terra.workspace.model.DeleteControlledGcpAiNotebookInstanceRequest;
 import bio.terra.workspace.model.DeleteControlledGcpAiNotebookInstanceResult;
-import bio.terra.workspace.model.GcpAiNotebookInstanceCreationParameters;
 import bio.terra.workspace.model.GcpAiNotebookInstanceResource;
-import bio.terra.workspace.model.GcpAiNotebookInstanceVmImage;
 import bio.terra.workspace.model.GrantRoleRequestBody;
 import bio.terra.workspace.model.IamRole;
 import bio.terra.workspace.model.JobControl;
-import bio.terra.workspace.model.ManagedBy;
-import bio.terra.workspace.model.PrivateResourceIamRoles;
-import bio.terra.workspace.model.PrivateResourceUser;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
-import com.google.api.services.iam.v1.model.TestIamPermissionsRequest;
 import com.google.api.services.notebooks.v1.AIPlatformNotebooks;
-import com.google.api.services.notebooks.v1.model.Instance;
 import com.google.api.services.notebooks.v1.model.StopInstanceRequest;
-import java.io.IOException;
-import java.security.GeneralSecurityException;
 import java.time.Duration;
 import java.util.List;
 import java.util.UUID;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.http.HttpStatus;
-import org.hamcrest.Matchers;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import scripts.utils.ClientTestUtils;
 import scripts.utils.CloudContextMaker;
+import scripts.utils.ResourceMaker;
 import scripts.utils.WorkspaceAllocateTestScriptBase;
 
 public class PrivateControlledAiNotebookInstanceLifecycle extends WorkspaceAllocateTestScriptBase {
@@ -89,18 +76,8 @@ public class PrivateControlledAiNotebookInstanceLifecycle extends WorkspaceAlloc
     ControlledGcpResourceApi resourceUserApi =
         ClientTestUtils.getControlledGcpResourceClient(resourceUser, server);
     CreatedControlledGcpAiNotebookInstanceResult creationResult =
-        createPrivateNotebook(resourceUser, resourceUserApi);
-    String creationJobId = creationResult.getJobReport().getId();
-    creationResult = ClientTestUtils.pollWhileRunning(
-        creationResult,
-        () -> resourceUserApi.getCreateAiNotebookInstanceResult(getWorkspaceId(), creationJobId),
-        CreatedControlledGcpAiNotebookInstanceResult::getJobReport,
-        Duration.ofSeconds(10));
-    ClientTestUtils.assertJobSuccess("create ai notebook",
-        creationResult.getJobReport(), creationResult.getErrorReport());
-    logger.info(
-        "Creation succeeded for instanceId {}",
-        creationResult.getAiNotebookInstance().getAttributes().getInstanceId());
+        ResourceMaker.makeControlledNotebookUserPrivate(
+            getWorkspaceId(), instanceId, resourceUser, resourceUserApi);
 
     UUID resourceId = creationResult.getAiNotebookInstance().getMetadata().getResourceId();
 
@@ -120,12 +97,11 @@ public class PrivateControlledAiNotebookInstanceLifecycle extends WorkspaceAlloc
             resource.getAttributes().getLocation(), resource.getAttributes().getInstanceId());
     AIPlatformNotebooks userNotebooks = ClientTestUtils.getAIPlatformNotebooksClient(resourceUser);
 
-    Instance instance = userNotebooks.projects().locations().instances().get(instanceName)
-        .execute();
-
     // TODO(PF-765): Assert that the other user does not have proxy access once we don't have
     // project level service account permissions.
-    assertUserHasProxyAccess(resourceUser, instance, resource.getAttributes().getProjectId());
+    assertTrue(ResourceMaker.userHasProxyAccess(
+        creationResult, resourceUser, resource.getAttributes().getProjectId()),
+        "Private resource user has access to their notebook");
     // The user should be able to stop their notebook.
     userNotebooks.projects().locations().instances().stop(instanceName, new StopInstanceRequest());
 
@@ -166,70 +142,5 @@ public class PrivateControlledAiNotebookInstanceLifecycle extends WorkspaceAlloc
     assertThat("Error from GCP is 403 or 404",
         notebookNotFound.getStatusCode(),
         anyOf(equalTo(HttpStatus.SC_NOT_FOUND), equalTo(HttpStatus.SC_FORBIDDEN)));
-  }
-
-  private CreatedControlledGcpAiNotebookInstanceResult createPrivateNotebook(
-      TestUserSpecification user, ControlledGcpResourceApi resourceApi)
-      throws ApiException, InterruptedException {
-    // Fill out the minimum required fields to arbitrary values.
-    var creationParameters =
-        new GcpAiNotebookInstanceCreationParameters()
-            .instanceId(instanceId)
-            .location("us-east1-b")
-            .machineType("e2-standard-2")
-            .vmImage(
-                new GcpAiNotebookInstanceVmImage()
-                    .projectId("deeplearning-platform-release")
-                    .imageFamily("r-latest-cpu-experimental"));
-
-    PrivateResourceIamRoles privateIamRoles = new PrivateResourceIamRoles();
-    privateIamRoles.add(ControlledResourceIamRole.EDITOR);
-    privateIamRoles.add(ControlledResourceIamRole.WRITER);
-    var commonParameters =
-        new ControlledResourceCommonFields()
-            .name(RandomStringUtils.randomAlphabetic(6))
-            .cloningInstructions(CloningInstructionsEnum.NOTHING)
-            .accessScope(AccessScope.PRIVATE_ACCESS)
-            .managedBy(ManagedBy.USER)
-            .privateResourceUser(
-                new PrivateResourceUser()
-                    .userName(user.userEmail)
-                    .privateResourceIamRoles(privateIamRoles));
-
-    var body =
-        new CreateControlledGcpAiNotebookInstanceRequestBody()
-            .aiNotebookInstance(creationParameters)
-            .common(commonParameters)
-            .jobControl(new JobControl().id(UUID.randomUUID().toString()));
-
-    return resourceApi.createAiNotebookInstance(body, getWorkspaceId());
-  }
-
-  /**
-   * Assert that the user has access to the Notebook through the proxy with a service account.
-   * <p>We can't directly test that we can go through the proxy to the Jupyter notebook without a
-   * real Google user auth flow, so we check the necessary ingredients instead.
-   */
-  private static void assertUserHasProxyAccess(TestUserSpecification user, Instance instance,
-      String projectId) throws GeneralSecurityException, IOException {
-    // Test that the user has access to the notebook with a service account through proxy mode.
-    // git secrets gets a false positive if 'service_account' is double quoted.
-    assertThat("Notebook has correct proxy mode access", instance.getMetadata(),
-        Matchers.hasEntry("proxy-mode", "service_" + "account"));
-
-    // The user needs to have the actAs permission on the service account.
-    String actAsPermission = "iam.serviceAccounts.actAs";
-    String serviceAccountName = String
-        .format("projects/%s/serviceAccounts/%s", projectId, instance.getServiceAccount());
-    assertThat("User may actAs notebook SA",
-        ClientTestUtils.getGcpIamClient(user)
-            .projects()
-            .serviceAccounts()
-            .testIamPermissions(
-                serviceAccountName,
-                new TestIamPermissionsRequest().setPermissions(List.of(actAsPermission)))
-            .execute()
-            .getPermissions(),
-        Matchers.contains(actAsPermission));
   }
 }
--- a/integration/src/main/java/scripts/testscripts/PrivateControlledAiNotebookInstanceLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/PrivateControlledAiNotebookInstanceLifecycle.java
@@ -32,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import scripts.utils.ClientTestUtils;
 import scripts.utils.CloudContextMaker;
 import scripts.utils.ResourceMaker;
+import scripts.utils.ResourceModifier;
 import scripts.utils.WorkspaceAllocateTestScriptBase;
 
 public class PrivateControlledAiNotebookInstanceLifecycle extends WorkspaceAllocateTestScriptBase {
@@ -99,7 +100,7 @@ public class PrivateControlledAiNotebookInstanceLifecycle extends WorkspaceAlloc
 
     // TODO(PF-765): Assert that the other user does not have proxy access once we don't have
     // project level service account permissions.
-    assertTrue(ResourceMaker.userHasProxyAccess(
+    assertTrue(ResourceModifier.userHasProxyAccess(
         creationResult, resourceUser, resource.getAttributes().getProjectId()),
         "Private resource user has access to their notebook");
     // The user should be able to stop their notebook.
--- a/integration/src/main/java/scripts/testscripts/RemoveUser.java
+++ b/integration/src/main/java/scripts/testscripts/RemoveUser.java
@@ -29,6 +29,7 @@ import org.apache.commons.lang3.RandomStringUtils;
 import scripts.utils.ClientTestUtils;
 import scripts.utils.CloudContextMaker;
 import scripts.utils.ResourceMaker;
+import scripts.utils.ResourceModifier;
 import scripts.utils.WorkspaceAllocateTestScriptBase;
 
 public class RemoveUser extends WorkspaceAllocateTestScriptBase {
@@ -80,7 +81,7 @@ public class RemoveUser extends WorkspaceAllocateTestScriptBase {
     String sharedBucketName = BUCKET_PREFIX + UUID.randomUUID();
     sharedBucket = ResourceMaker.makeControlledGcsBucketUserShared(
         ownerResourceApi, getWorkspaceId(), sharedBucketName, CloningInstructionsEnum.NOTHING);
-    ResourceMaker.addFileToBucket(sharedBucket, workspaceOwner, projectId);
+    ResourceModifier.addFileToBucket(sharedBucket, workspaceOwner, projectId);
 
     // Create a private GCS bucket for privateResourceUser with one object inside.
     String privateBucketName = BUCKET_PREFIX + UUID.randomUUID();
@@ -91,14 +92,14 @@ public class RemoveUser extends WorkspaceAllocateTestScriptBase {
     privateBucket = ResourceMaker.makeControlledGcsBucketUserPrivate(
         privateUserResourceApi, getWorkspaceId(), privateBucketName, privateResourceUser.userEmail,
         privateRoles);
-    ResourceMaker.addFileToBucket(privateBucket, privateResourceUser, projectId);
+    ResourceModifier.addFileToBucket(privateBucket, privateResourceUser, projectId);
 
     // Create a private BQ dataset for privateResourceUser and populate it.
     String datasetId = RandomStringUtils.randomAlphabetic(8).toLowerCase();
     privateDataset = ResourceMaker.makeControlledBigQueryDatasetUserPrivate(
         privateUserResourceApi, getWorkspaceId(), datasetId, privateResourceUser.userEmail,
         privateRoles, CloningInstructionsEnum.NOTHING);
-    ResourceMaker.populateBigQueryDataset(privateDataset, privateResourceUser, projectId);
+    ResourceModifier.populateBigQueryDataset(privateDataset, privateResourceUser, projectId);
     // Create a private notebook for privateResourceUser.
     String notebookInstanceId = RandomStringUtils.randomAlphabetic(8).toLowerCase();
     privateNotebook = ResourceMaker.makeControlledNotebookUserPrivate(
@@ -111,8 +112,8 @@ public class RemoveUser extends WorkspaceAllocateTestScriptBase {
     String sharedBucketName = sharedBucket.getGcpBucket().getAttributes().getBucketName();
     String privateBucketName = privateBucket.getGcpBucket().getAttributes().getBucketName();
     // Validate that setup ran correctly and users have appropriate resource access.
-    ResourceMaker.retrieveBucketFile(sharedBucketName, projectId, sharedResourceUser);
-    ResourceMaker.retrieveBucketFile(privateBucketName, projectId, privateResourceUser);
+    ResourceModifier.retrieveBucketFile(sharedBucketName, projectId, sharedResourceUser);
+    ResourceModifier.retrieveBucketFile(privateBucketName, projectId, privateResourceUser);
 
     // Remove WRITER role from sharedResourceUser. This is their only role, so they are no longer
     // a member of this workspace.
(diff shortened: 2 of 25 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
```diff
@@ -56,6 +56,10 @@ public StepResult doStep(FlightContext flightContext)
             .map(GcsApiConversions::toGcsApiRulesList)
             .orElse(Collections.emptyList()));
 
+    BucketInfo.IamConfiguration iamConfiguration =
```
ddietterich, line 59: I wonder if we will see an application that wants to use object-level IAM.
If that comes up, we can expose the parameter.

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemovePrivateResourceAccessStep.java
```diff
@@ -0,0 +1,73 @@
...
+    // to private resources.
+    if (userStillInWorkspace) {
+      return StepResult.getStepResultSuccess();
+    }
+
+    List<ControlledResource> userResources =
+        workingMap.get(
+            ControlledResourceKeys.REVOCABLE_PRIVATE_RESOURCES, new TypeReference<>() {});
+    for (ControlledResource resource : userResources) {
+      // WSM does not store which roles user have on their private resources, so here we attempt
```
ddietterich, line 53: I wonder if we should teach WSM to know which role the assigned user has on the private resource. This only applies to the assigned user, right?

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemovePrivateResourceAccessStep.java
```diff
@@ -0,0 +1,73 @@
...
+          resource, userRequest, ControlledResourceIamRole.WRITER, userToRemove);
+      samService.removeResourceRole(
+          resource, userRequest, ControlledResourceIamRole.EDITOR, userToRemove);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    // WSM does not store which roles users have on their private resources, so there is no way
```
ddietterich, line 67: Ditto

ddietterich (commented):
LGTM - I would like to decide whether to use this ticket to add the info to the WSM schema to track what role the assigned user has on the private resource. That would improve the resiliency of the flight. Alternately, we could query Sam for each possible role the user might have and keep that in the working map. That would allow undo, etc.

We could do that as a follow-on ticket or add to this one. I'm OK either way.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -416,7 +448,55 @@ public void removeWorkspaceRole(
...
+   * Wrapper around the Sam client to remove a role from the provided user on a controlled resource.
+   *
+   * <p>Similar to {@removeWorkspaceRole}, but for controlled resources. This should only be
+   * necessary for private resources, as users do not have individual roles on shared resources.
+   *
+   * <p>This call to Sam is made as the WSM SA, as users do not have permission to directly modify
+   * IAM on resources. This method still requires user credentials to validate as a safeguard, but
+   * they are not used in the role removal call.
+   */
+  @Traced
```
jaycarlton, line 460: It would help to document the params here.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -416,7 +448,55 @@ public void removeWorkspaceRole(
...
+                  role.toSamRole(),
+                  email));
+      logger.info(
+          "Removed role {} from user {} on resource {}",
+          role.toSamRole(),
+          email,
+          resource.getResourceId());
+    } catch (IOException credentialException) {
+      throw new InternalServerErrorException(
+          "Error removing resource role in Sam", credentialException);
```
jaycarlton, line 492: Seems like these two cases should have different error messages. Maybe say "internal error", though I guess the 500 gives it away.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -450,6 +530,21 @@ public void removeWorkspaceRole(
     }
   }
 
+  /** Wrapper around Sam client to fetch the list of users with a specific role in a workspace. */
+  @Traced
+  public List<String> listWorkspacePolicyMembers(
```
jaycarlton, line 535: I'd name this `listUsersWithWorkspaceRole` or similar, since policies are a Sam concept.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
```diff
@@ -56,6 +56,10 @@ public StepResult doStep(FlightContext flightContext)
...
+        BucketInfo.IamConfiguration.newBuilder().setIsUniformBucketLevelAccessEnabled(true).build();
+    bucketInfoBuilder.setIamConfiguration(iamConfiguration);
```
jaycarlton, line 61: In theory we should backfill existing controlled buckets to make this change.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/ServiceAccountPolicyStep.java
```diff
@@ -102,11 +105,15 @@ private Binding createWriterBinding(FlightMap workingMap) throws InterruptedExce
...
+    // Grant permission via private resource's Sam group so that access can be revoked via Sam.
+    Map<ControlledResourceIamRole, String> resourceRoleGroupsMap =
+        workingMap.get(
+            WorkspaceFlightMapKeys.ControlledResourceKeys.IAM_RESOURCE_GROUP_EMAIL_MAP,
+            new TypeReference<>() {});
+    String writerGroupEmail = resourceRoleGroupsMap.get(ControlledResourceIamRole.WRITER);
     return new Binding()
         .setRole(SERVICE_ACCOUNT_USER_ROLE)
-        .setMembers(ImmutableList.of("user:" + userEmail));
+        .setMembers(ImmutableList.of("group:" + writerGroupEmail));
```
jaycarlton, line 116: I think `Collections.singletonList()` might be preferred, but I don't recall why.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -390,4 +392,30 @@ public String enablePetServiceAccountImpersonation(
       throw new RuntimeException("Error enabling user's pet SA", e);
     }
   }
+
+  public void removeRole(
```
jaycarlton, line 396: javadoc here would help.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -390,4 +392,30 @@ public String enablePetServiceAccountImpersonation(
...
+      UUID workspaceId, WsmIamRole role, String userEmail, AuthenticatedUserRequest userRequest) {
+    Workspace workspace =
+        validateWorkspaceAndAction(userRequest, workspaceId, SamConstants.SAM_WORKSPACE_OWN_ACTION);
+    stageService.assertMcWorkspace(workspace, "deleteGcpCloudContext");
```
jaycarlton, line 400: this should be "removeRole".

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
jaycarlton: maybe `removeWorkspaceRoleFromUser()`?

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -390,4 +392,30 @@ public String enablePetServiceAccountImpersonation(
...
+    // specified role. Users may also be added to a workspace via managed groups, but WSM does not
+    // control membership of those groups, and so cannot remove them here.
+    List<String> roleMembers =
+        samService.listWorkspacePolicyMembers(workspaceId, role, userRequest);
+    if (!roleMembers.contains(userEmail)) {
+      return;
+    }
+    jobService
+        .newJob(
+            "Remove user from workspace " + workspaceId,
```
jaycarlton, line 411: aren't we only removing the user from the workspace if this is their last role?

jaycarlton (commented):
first batch

integration/src/main/java/scripts/testscripts/RemoveUser.java
```diff
@@ -0,0 +1,212 @@
...
+        ownerResourceApi, getWorkspaceId(), sharedBucketName, CloningInstructionsEnum.NOTHING);
+    ResourceMaker.addFileToBucket(sharedBucket, workspaceOwner, projectId);
+
+    // Create a private GCS bucket for privateResourceUser with one object inside.
+    String privateBucketName = BUCKET_PREFIX + UUID.randomUUID();
+    PrivateResourceIamRoles privateRoles = new PrivateResourceIamRoles();
+    privateRoles.addAll(PRIVATE_ROLES);
+    ControlledGcpResourceApi privateUserResourceApi =
+        ClientTestUtils.getControlledGcpResourceClient(privateResourceUser, server);
+    privateBucket = ResourceMaker.makeControlledGcsBucketUserPrivate(
```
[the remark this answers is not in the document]
ddietterich: Yes, but I have not implemented it. I just made PF-990 to track it.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
ddietterich: Since none of our resources are anywhere near production, we don't have to follow the theory :)

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemovePrivateResourceAccessStep.java
```diff
@@ -0,0 +1,73 @@
...
+    // to undo any removal that happened in the DO step. The best we can do here is log an error and
```
jaycarlton, line 68: Could you check before removing each role whether the user actually has it, then populate a map from resource ID to a set of roles, put that in the working map, and use it for undo.

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromSamStep.java
```diff
@@ -0,0 +1,44 @@
...
+    this.samService = samService;
+    this.workspaceId = workspaceId;
+    this.roleToRemove = roleToRemove;
+    this.userToRemoveEmail = userToRemoveEmail;
+    this.userRequest = userRequest;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    samService.removeWorkspaceRole(workspaceId, userRequest, roleToRemove, userToRemoveEmail);
```
jaycarlton, line 35: double-check this is idempotent

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
```diff
@@ -0,0 +1,50 @@
...
+    super(inputParameters, applicationContext);
+
+    FlightBeanBag appContext = FlightBeanBag.getFromObject(applicationContext);
+
+    UUID workspaceId =
+        UUID.fromString(inputParameters.get(WorkspaceFlightMapKeys.WORKSPACE_ID, String.class));
+    AuthenticatedUserRequest userRequest =
+        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    String userToRemove = inputParameters.get(WorkspaceFlightMapKeys.USER_TO_REMOVE, String.class);
+    WsmIamRole roleToRemove =
```
jaycarlton, line 23: Have I been sinning by serializing enums directly? It seems to work AFAICT.

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
```diff
@@ -0,0 +1,50 @@
...
+        WsmIamRole.valueOf(
+            inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, String.class));
+
+    // Flight plan:
+    // 0. (Pre-flight): Validate that the user is directly granted the specified workspace role.
+    //  WSM does not manage groups, so users with indirect group-based access cannot be removed
```
jaycarlton, line 29: How would a user have gotten such indirect access in the first place?

jaycarlton (approved):

zloery:
> LGTM - I would like to decide whether to use this ticket to add the info to the WSM schema to track what role the assigned user has on the private resource. That would improve the resiliency of the flight. Alternately, we could query Sam for each possible role the user might have and keep that in the working map. That would allow undo, etc.
> 
> We could do that as a follow-on ticket or add to this one. I'm OK either way.

I think this PR is the right place for it. My inclination is to take the second approach and query Sam for the roles so that WSM doesn't need to maintain duplicate state. I added support for that here which allows for undoing. the flight, PTAL

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
zloery: If we did that, I think we would be trusting applications to manage their own IAM and they could e.g. share data with anyone. Maybe that's actually OK for trusted/reviewed applications, but there would probably be some compliance work required.

Either way, something to think about when applications are better defined

integration/src/main/java/scripts/testscripts/RemoveUser.java
```diff
@@ -0,0 +1,212 @@
...
+    // This requires syncing google groups, so there is often a delay that we need to wait for.
+    ClientTestUtils.runWithRetryOnException(
+        () -> assertUserCannotReadBucket(sharedBucketName, sharedResourceUser));
+
+    // privateResource user can still read the shared bucket.
+    ResourceMaker.retrieveBucketFile(sharedBucketName, projectId, privateResourceUser);
+
+    // Remove READER role from privateResourceUser. They are also a writer, so they should not lose
+    // access to workspace resources because of this.
+    ownerWorkspaceApi.removeRole(getWorkspaceId(), IamRole.READER, privateResourceUser.userEmail);
```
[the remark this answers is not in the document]
zloery: WRITER is a superset of READER, but you are allowed to have both. There's not really any point to it, but it is useful for this test

integration/src/main/java/scripts/testscripts/RemoveUser.java
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * An assertion that the given user cannot read from the given bucket. This is pulled into a
+   * separate function to make retrying simpler.
+   */
+  private void assertUserCannotReadBucket(String bucketName, TestUserSpecification testUser) {
+    try {
+      ResourceMaker.retrieveBucketFile(bucketName, projectId, testUser);
+      // If nothing is thrown, that's bad! The user actually can read the bucket.
+      throw new RuntimeException(String
```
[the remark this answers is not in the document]
zloery: It does retry that exception, which is what we want since access hasn't been revoked yet. The "happy path" is getting a 403 storage exception in this case.

An alternative here is writing a `runWithRetryOnNoException` method - either way, these are looking to validate that the right exception is being thrown.

integration/src/main/java/scripts/utils/ResourceMaker.java
```diff
@@ -213,6 +229,25 @@ public static void deleteControlledGcsBucket(
...
+    return sourceOwnerStorageClient.create(blobInfo, GCS_BLOB_CONTENT.getBytes());
+  }
+
+  public static Blob retrieveBucketFile(String bucketName, String gcpProjectId,
+      TestUserSpecification bucketReader) throws IOException {
+    Storage cloningUserStorageClient = ClientTestUtils.getGcpStorageClient(bucketReader, gcpProjectId);
+    BlobId blobId = BlobId.of(bucketName, GCS_BLOB_NAME);
+
+    final Blob retrievedFile = cloningUserStorageClient.get(blobId);
+    assertNotNull(retrievedFile);
```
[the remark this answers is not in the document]
zloery: For this test the contents matching shouldn't really matter. I'm using this to validate that a user has read access, which they do if they can read anything out of the bucket

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -241,6 +241,26 @@ public boolean deleteResource(UUID workspaceId, UUID resourceId) {
...
+    MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue("workspace_id", workspaceId.toString())
+            .addValue("access_scope", AccessScopeType.ACCESS_SCOPE_PRIVATE.toSql())
+            .addValue("user_email", userEmail);
+
+    List<DbResource> dbResources = jdbcTemplate.query(sql, params, DB_RESOURCE_ROW_MAPPER);
+    return dbResources.stream()
+        .map(this::constructResource)
+        .map(WsmResource::castToControlledResource)
```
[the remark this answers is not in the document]
zloery: Good catch. I was relying on private access scope to filter for controlled resources, but it's better to be explicit in the query.
I think the name should be ok as-is, especially with the javadoc and return type specifying controlled resources.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: done

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: I think the error code should be the important bit here, but modified the error message to make it a little clearer as well.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: Good point

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
zloery: I'm operating under the principle of "we have no workspaces/resources we care about persisting" for now. It's definitely possible for us to backfill existing buckets with some manual effort, though I don't think it's worth it.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/ServiceAccountPolicyStep.java
zloery: SGTM for a one element list

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
zloery: Done, and renamed

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
zloery: good catch!

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
zloery: good point, updated description.

service/src/main/java/bio/terra/workspace/service/workspace/flight/ReadUserPrivateResourcesStep.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    FlightMap workingMap = context.getWorkingMap();
+    boolean userStillInWorkspace =
+        Optional.ofNullable(
+                workingMap.get(
+                    ControlledResourceKeys.REMOVED_USER_IS_WORKSPACE_MEMBER, Boolean.class))
+            .orElse(false);
```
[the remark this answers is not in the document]
zloery: Yep, this should be an error. I think I was following IDE suggestions too closely.
In general, is there a preference for using `Optional/orElseThrow` vs. just calling `.get()` and letting an exception surface?

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemovePrivateResourceAccessStep.java
zloery: Yep, added that in to support undo

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromSamStep.java
zloery: Double checked and added a comment for future readers

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
zloery: IIRC they would work normally, but there was an enum serialization issue when flights were recovered. I haven't kept up to date with Stairway happenings, I'll defer to @ddietterich if this has been fixed or is still an issue

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
zloery: WSM supports granting roles to Sam managed groups as well as users. So, `my-research-group@sam-group-domain.com` can be a reader of my workspace, and you would have access via group membership. In that case, WSM won't know when that group's membership changes, though it could still remove the entire group from the workspace.

service/src/main/java/bio/terra/workspace/service/workspace/flight/ReadUserPrivateResourcesStep.java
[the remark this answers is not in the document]
jaycarlton: Calling `get()` is usually a sign that you've missed an abstraction. The point of Optional is to not be subtle about such things. Plus, you  have control over the exception message this way.

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
jaycarlton: Is this one of the use cases for the Pub/Sub queue?

integration/src/main/java/scripts/testscripts/PrivateControlledAiNotebookInstanceLifecycle.java
```diff
@@ -120,12 +97,11 @@ protected void doUserJourney(TestUserSpecification testUser, WorkspaceApi worksp
             resource.getAttributes().getLocation(), resource.getAttributes().getInstanceId());
     AIPlatformNotebooks userNotebooks = ClientTestUtils.getAIPlatformNotebooksClient(resourceUser);
 
-    Instance instance = userNotebooks.projects().locations().instances().get(instanceName)
-        .execute();
-
     // TODO(PF-765): Assert that the other user does not have proxy access once we don't have
     // project level service account permissions.
-    assertUserHasProxyAccess(resourceUser, instance, resource.getAttributes().getProjectId());
+    assertTrue(ResourceMaker.userHasProxyAccess(
```
jaycarlton, line 102: nit: `ResourceMaker` is now doing more than just making resources, so maybe it needs a new name or a new class split out.

integration/src/main/java/scripts/testscripts/RemoveUser.java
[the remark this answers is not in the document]
jaycarlton: Maybe indicate in the comment that this is only designed to work with that one retry function.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/ServiceAccountPolicyStep.java
```diff
@@ -102,11 +105,15 @@ private Binding createWriterBinding(FlightMap workingMap) throws InterruptedExce
         privateIamRoles.stream().anyMatch(role -> role.equals(ControlledResourceIamRole.WRITER)),
         "A private, controlled AI Notebook instance must have the writer role or else it is not useful.");
 
-    // Assumes that the assigned user is the user requesting the resource creation.
-    String userEmail = samService.getRequestUserEmail(userRequest);
+    // Grant permission via private resource's Sam group so that access can be revoked via Sam.
+    Map<ControlledResourceIamRole, String> resourceRoleGroupsMap =
+        workingMap.get(
+            WorkspaceFlightMapKeys.ControlledResourceKeys.IAM_RESOURCE_GROUP_EMAIL_MAP,
```
jaycarlton, line 111: nit: I usually statically import `WorkspaceFlightMapKeys` for brevity

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -390,4 +392,40 @@ public String enablePetServiceAccountImpersonation(
...
   }
+
+  /**
+   * Remove a workspace role from a user. This will also remove a user from their private resources
+   * if they are no longer a member of the workspace (i.e. have no other roles) after role removal.
+   *
+   * @param workspaceId ID of the workspace user to remove user's role in
+   * @param role Role to remove
+   * @param userEmail Email identifier of user whose role is being removed
+   * @param userRequest User credentials to authenticate this removal. Must belong to a workspace
```
jaycarlton, line 403: aside: I wonder if there's a convention we could use to make this more prominent. Here we're acting as one user to remove another. Above, we are the service account acting on behalf of (but not "acting as") a user. The documentation is helpful.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -390,4 +392,40 @@ public String enablePetServiceAccountImpersonation(
...
+   *     owner, and likely do not belong to {@code userEmail}.
+   */
+  public void removeWorkspaceRoleFromUser(
+      UUID workspaceId, WsmIamRole role, String userEmail, AuthenticatedUserRequest userRequest) {
```
jaycarlton, line 407: nit: rename `userEmail` to `targetUserEmail` and `userRequest` to `executingUserRequest` or something for clarity

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -390,4 +392,40 @@ public String enablePetServiceAccountImpersonation(
...
+    // specified role. Users may also be added to a workspace via managed groups, but WSM does not
+    // control membership of those groups, and so cannot remove them here.
+    List<String> roleMembers =
+        samService.listUsersWithWorkspaceRole(workspaceId, role, userRequest);
+    if (!roleMembers.contains(userEmail)) {
+      return;
+    }
+    jobService
+        .newJob(
+            "Remove role from user in workspace " + workspaceId,
```
jaycarlton, line 421: nit: you could put the user and role in the description as well

service/src/main/java/bio/terra/workspace/service/workspace/flight/ReadUserPrivateResourcesStep.java
```diff
@@ -0,0 +1,71 @@
...
+    // Read the list of resources this user owns from WSM's DB
+    List<ControlledResource> userResources =
+        resourceDao.listPrivateResourcesByUser(workspaceId, userEmail);
+    // For each private resource, query Sam to find the roles the user has.
+    List<ResourceRolePair> resourceRolesToRemove = new ArrayList<>();
+    for (ControlledResource resource : userResources) {
+      List<ControlledResourceIamRole> userRoles =
+          samService.getUserRolesOnPrivateResource(resource, userEmail, userRequest);
+      for (ControlledResourceIamRole role : userRoles) {
+        resourceRolesToRemove.add(new ResourceRolePair(resource, role));
```
jaycarlton, line 59: nit: you can just use `addAll`, unless that doesn't exist for java.util.List.

service/src/test/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlightTest.java
```diff
@@ -0,0 +1,188 @@
...
+    ControlledBigQueryDatasetResource privateDataset = buildPrivateDataset(workspaceId, datasetId);
+    assertNotNull(privateDataset);
+
+    // Validate with Sam that secondary user can read their private resource
+    assertTrue(
+        samService.isAuthorized(
+            userAccessUtils.secondUserAccessToken().getTokenValue(),
+            privateDataset.getCategory().getSamResourceName(),
+            privateDataset.getResourceId().toString(),
+            SamControlledResourceActions.WRITE_ACTION));
```
jaycarlton, line 97: Should this be `READ_ACTION`?

jaycarlton (approved):

integration/src/main/java/scripts/testscripts/PrivateControlledAiNotebookInstanceLifecycle.java
zloery: agreed. split this into `ResourceMaker` and `ResourceModifier`

service/src/main/java/bio/terra/workspace/service/workspace/flight/ReadUserPrivateResourcesStep.java
zloery: I'm building a list of `ResourceRolePair`s here, not just adding the roles directly. I can do that via a stream and use `addAll` then, but I think this is easier to parse.

service/src/test/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlightTest.java
zloery: Either is ok. The user was added to the workspace as a writer, so they should still have both permissions at this point.

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
ddietterich: I think that serializing and deserializing standalone enums works like:
```
map.put("foo", MyEnum.FOO);
MyEnum anEnum = map.get("foo", MyEnum.class);
```
The issue I was aware of was using collections of enums. `List<MyEnum>` would not deserialize properly with `List.class`. However, now that we can use `TypeReference`, you can get that to work also.

There is a case that is untested in the Stairway regression tests: an enum embedded in another class. However, those are always tagged with the type and I believe we use them frequently.

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
ddietterich: If you mean getting notification of access changes from Sam via PubSub, yes, this is one of the use cases for that.

ddietterich (approved):
LGTM

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
zloery: sgtm, switched to directly serializing this enum

PR #432, opened 2021-09-10
[PF-765] Remove project-level SA permissions

zloery:
Followup to #427 and https://github.com/DataBiosphere/terra-cli/pull/134 which removes the project-level SA permissions WSM previously granted in workspaces. As a temporary workaround for the CLI, this continues to allow pet SA self-impersonation (tracking in PF-991).

Additionally, this lowers the role required to call the `enablePet` endpoint from WRITER to READER to align with the CLI approach of calling this endpoint as soon as users create or load a workspace. This does not change the role required to run Pipelines API jobs, which is still WRITER.

This should not merge before https://github.com/DataBiosphere/terra-cli/pull/134.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
@@ -31,10 +31,10 @@ import bio.terra.workspace.service.workspace.model.WorkspaceRequest;
 import com.google.api.services.iam.v1.model.Binding;
 import com.google.api.services.iam.v1.model.Policy;
 import com.google.api.services.iam.v1.model.SetIamPolicyRequest;
+import com.google.common.collect.ImmutableList;
 import io.opencensus.contrib.spring.aop.Traced;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
@@ -367,10 +367,12 @@ public class WorkspaceService {
     try {
       Policy saPolicy =
           crlService.getIamCow().projects().serviceAccounts().getIamPolicy(petSaName).execute();
+      // TODO(PF-991): In the future, the pet SA should not be included in this binding. This is a
+      //  workaround to support Nextflow and other applications which call Pipelines API.
       Binding saUserBinding =
           new Binding()
               .setRole(serviceAccountUserRole)
-              .setMembers(Collections.singletonList("user:" + userEmail));
+              .setMembers(ImmutableList.of("user:" + userEmail, "serviceAccount:" + petSaEmail));
       // If no bindings exist, getBindings() returns null instead of an empty list.
       List<Binding> bindingList =
           Optional.ofNullable(saPolicy.getBindings()).orElse(new ArrayList<>());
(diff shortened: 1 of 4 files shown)


jaycarlton (approved):

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -349,11 +349,10 @@ public String getRequiredGcpProject(UUID workspaceId) {
   public String enablePetServiceAccountImpersonation(
       UUID workspaceId, AuthenticatedUserRequest userRequest) {
     final String serviceAccountUserRole = "roles/iam.serviceAccountUser";
-    // Validate that the user is at least a writer in this workspace, as only writers can run
-    // workflows.
+    // Validate that the user is a member of the workspace.
     Workspace workspace =
         validateWorkspaceAndAction(
-            userRequest, workspaceId, SamConstants.SAM_WORKSPACE_WRITE_ACTION);
+            userRequest, workspaceId, SamConstants.SAM_WORKSPACE_READ_ACTION);
```
marikomedlock, line 355: What's the reason for this change? If it's just to support the CLI, then we could also handle it in the CLI (e.g. check whether a user is a reader or writer on the workspace before calling the endpoint, or just tolerating failures there, etc.). It feels weird that a reader can modify resources in the workspace project, though I suppose they can already always create a pet SA via SAM regardless of their role in the workspace.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -369,10 +368,12 @@ public String enablePetServiceAccountImpersonation(
     try {
       Policy saPolicy =
           crlService.getIamCow().projects().serviceAccounts().getIamPolicy(petSaName).execute();
+      // TODO(PF-991): In the future, the pet SA should not be included in this binding. This is a
+      //  workaround to support Nextflow and other applications which call Pipelines API.
```
marikomedlock, line 372: nit: the CLI and other applications which call the pipelines api with the pet SA's credentials.
Just noting that this restriction isn't really a problem with nextflow -- nextflow can be called with the end user's credentials. it's the CLI's quirky way of calling nextflow via a docker container that's the problem.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
zloery: This is mainly to support the CLI, but I also don't think there's a particularly strong case for this to be WRITER instead of READER. Pet impersonation is required for pipelines API jobs, but the user also needs PAPI specific permissions that only writers have for that. By itself, pet impersonation doesn't really allow users to do anything.
We also don't revoke this when a user's roles change, so even with the current system someone could do:
1. Add user X as a writer
2. Enable user X's pet SA
3. Add user X as a reader
4. Remove user X as a writer

and end up in the same place.

marikomedlock: Good points. How about keeping this as READER then, instead of tying it to WRITER/permission to run PAPI jobs? I think the CLI wouldn't need to change -- we still call enablePet each time we fetch the key file.

zloery: sounds good, this check will be READER and the pending CLI changes should work as-is.

marikomedlock (approved):
LGTM, thanks!

PR #435, opened 2021-09-12
[PF-949] use self hosted runners for nightly tests

ddietterich:
DevOps set up hosted runners for github actions. There are VMs within the Broad perimeter and therefore can access the ArgoCD server. That allows the nightly test to poke ArgoCD to update the software on the `wsmtest` cluster ahead of running the tests.

Since these are running with our own container, they do not have all of the software that comes with the github runners. So the big change in this flow is installing postgres ahead of running the tests.

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -16,11 +16,15 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      # On the self-hosted runner we need to install postgres ourselves. The process is:
-      # Create the file repository configuration.
-      # Import the repository signing key.
-      # Update the package lists.
-      # Install the latest version of PostgreSQL. If you want a specific version, use 'postgresql-12' or similar instead of 'postgresql':
+      # On the self-hosted runner we need to install additional software ourselves
+      - name: Install wget
+        run: apt-get install wget
+
+      # Install postgres:
+      # - Create the file repository configuration.
+      # - Import the repository signing key.
+      # - Update the package lists.
+      # - Install the latest version of PostgreSQL. If you want a specific version, use 'postgresql-12' or similar instead of 'postgresql':
       - name: Install the latest postgres
         run: |
             sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -16,10 +16,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


jaycarlton (approved):

.github/workflows/nightly-tests.yml
```diff
@@ -54,6 +66,13 @@ jobs:
           vault-token: ${{ steps.vault-token-step.outputs.vault-token }}
           target: wsmtest
 
+      - name: ArgoCD sync
+        run: |
+          curl --fail --silent --show-error --location --request POST \
```
marikomedlock, line 71: Is the ArgoCD sync synchronous? I'm wondering how to tell if the sync was successful. Perhaps it could be helpful to call the WSM version endpoint before/after this so we can look at logs to see if an expected sync occurred.

.github/workflows/nightly-tests.yml
```diff
@@ -66,12 +66,23 @@ jobs:
           vault-token: ${{ steps.vault-token-step.outputs.vault-token }}
           target: wsmtest
 
+      # The ArgoCD sync triggers the synchronization of the wsmtest cluster. Typically,
+      # the cluster is reset within a few seconds. However, it is possible for it to take
+      # minutes or hours. We have no way to check that the sync is complete. We sleep
+      # for a bit and just accept that in some crazy long cases, these tests will fail.
+      # The sync is bracketed by /version probes so we have the pre and post sync versions
+      # in the log.
```
jaycarlton, line 74: FWIW, I liked the idea of a separate workflow to promote the workspace manager. We could have separate reporting in Slack. But you'd have to align the cron schedules.

.github/workflows/nightly-tests.yml
```diff
@@ -66,12 +66,23 @@ jobs:
...
       - name: ArgoCD sync
         run: |
+          version=$(curl https://workspace.wsmtest.integ.envs.broadinstitute.org/version)
+          echo "pre-sync wsmtest version: $version"
           curl --fail --silent --show-error --location --request POST \
           'https://ap-argocd.dsp-devops.broadinstitute.org/api/v1/applications/workspacemanager-wsmtest/sync' \
           --header "Authorization: Bearer ${{ secrets.WSMTEST_SYNC_ARGOCD_TOKEN }}" \
           | jq .operation
+          sleep 30
```
jaycarlton, line 83: I guess this fine for now. If it's flaky you could loop and keep checking the version until it changes or matches what's in the github tag.

jaycarlton (approved):

.github/workflows/nightly-tests.yml
ddietterich: Added a sleep and version logging on either side of the argo trigger.

marikomedlock (approved):

PR #437, opened 2021-09-13
[no ticket] use SA cow

jaycarlton:
Hit an error when running in the CLI test framework that the call to get the BigQuery datasets had insufficient scopes. We should be using the WSM SA for this anyway from what I can tell.


ddietterich (approved):
LGTM
We need to test whether the cloning user has access to the resources, but I think you do that elsewhere. As long as we are checking that, we can do this operation as WSM SA.

service/src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -136,10 +136,9 @@ public BigQueryCow createBigQueryCow(AuthenticatedUserRequest userRequest) {
    * Create a vanilla Bigquery client object, for testing things that aren't in CRL yet.
    * TODO(jaycarlton): PF-942 implement needed endpoints in CRL and use them here
    *
-   * @param userRequest
    * @return
    */
-  public Bigquery createNakedBigQueryClient(AuthenticatedUserRequest userRequest) {
+  public Bigquery createNakedBigQueryClient() {
```
marikomedlock, line 141: nit: To be consistent with naming and make it clear what credentials this client uses, perhaps rename this to `createWsmSaNakedBigQueryClient` or something similar? I think this method picks up the app default creds which should be the wsm sa.

marikomedlock (approved):

PR #440, opened 2021-09-14
[PF-992] Make SamService more consistent

ddietterich:
In working on the Azure PoC, I noticed that some Sam methods were taking the auth token directly. Most were taking `AuthenticatedUserRequest`. It is a minor inconsistency, but in order to plumb in the mock Sam for Azure PoC, I need access to the `AuthenticatedUserRequest` in all cases.

This PR changes SamService so that all public methods use `AuthenticatedUserRequest`.  I think this is worth doing in the main branch.

In addition, I factored out the static `rethrowIfSamInterrupted` code into a separate module. SamService is pushing 1KLOC. There are probably other chunks that could be refactored to make it more manageable. In any case, now instead of saying
```
SamService.rethrowIfSamInterrupted(...)
```
you say
```
SamRethrow.onInterrupted(...)
```

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamRethrow.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamRethrow.java
@@ -1,0 +1,60 @@
+package bio.terra.workspace.service.iam;
+
+import bio.terra.common.sam.exception.SamExceptionFactory;
+
+/**
+ * When we use Sam in flights, we want to let InterruptedException fly and be caught by Stairway.
+ * However, when we use Sam outside of flights, we want to propagate the Interrupted exception.
+ * These static methods perform that propagation.
+ */
+public class SamRethrow {
+
+  /**
+   * For use with onInterrupted methods.
+   *
+   * @param <R>
+   */
+  @FunctionalInterface
+  public interface InterruptedSupplier<R> {
+    R apply() throws InterruptedException;
+  }
+
+  /** For use with onInterrupted. */
+  public interface VoidInterruptedSupplier {
+    void apply() throws InterruptedException;
+  }
+
+  /**
+   * Use this function outside of Flights. In that case, we do not expect SamRetry to be interrupted
+   * so the interruption is unexpected and should be surfaced all the way up as an unchecked
+   * exception, which this function does.
+   *
+   * <p>Usage: SamRethrow.onInterrupted(() -> samService.isAuthorized(...), "isAuthorized")) {
+   *
+   * @param function The SamService function to call.
+   * @param operation The name of the function to use in the error message.
+   * @param <T> The return type of the function to call.
+   * @return The return value of the function to call.
+   */
+  public static <T> T onInterrupted(InterruptedSupplier<T> function, String operation) {
+    try {
+      return function.apply();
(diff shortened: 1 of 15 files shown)


jmczerk (approved):
LGTM

service/src/main/java/bio/terra/workspace/service/iam/SamRethrow.java
```diff
@@ -0,0 +1,60 @@
...
+ * When we use Sam in flights, we want to let InterruptedException fly and be caught by Stairway.
+ * However, when we use Sam outside of flights, we want to propagate the Interrupted exception.
+ * These static methods perform that propagation.
+ */
+public class SamRethrow {
+
+  /**
+   * For use with onInterrupted methods.
+   *
+   * @param <R>
```
jaycarlton, line 15: nit: describe or delete param

service/src/main/java/bio/terra/workspace/service/iam/SamRethrow.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  @FunctionalInterface
+  public interface InterruptedSupplier<R> {
+    R apply() throws InterruptedException;
+  }
+
+  /** For use with onInterrupted. */
+  public interface VoidInterruptedSupplier {
```
jaycarlton, line 23: I'd call this `InterruptedRunner` to be consistent with the `Runnable` convention.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -116,56 +116,6 @@ public String getRequestUserEmail(AuthenticatedUserRequest userRequest)
     }
   }
```
jaycarlton, line 118: Do you want to define an interface for SamService to implement? Seems like that would help with the mocking.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -360,9 +361,8 @@ public String enablePetServiceAccountImpersonation(
     // getEmailFromToken will always call Sam, which will return a user email even if the requesting
     // access token belongs to a pet SA.
     String userEmail =
-        SamService.rethrowIfSamInterrupted(
-            () -> samService.getEmailFromToken(userRequest.getRequiredToken()),
-            "getEmailFromToken");
+        SamRethrow.onInterrupted(
+            () -> samService.getRequestUserEmail(userRequest), "getEmailFromToken");
```
jaycarlton, line 365: Should the operation description be "getRequestUserEmail"?

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
ddietterich: Not at this time. I am trying to keep existing WSM tests working as we add the PoC stuff. My plan is to have logic like the following to SamService:
```
if (azureState.isEnabled() && userRequest.getAuthType == AuthType.BASIC) {
  ...do the mock stuff...
} else {
  ...do the real stuff...
}
```
It'll do for a PoC.

PR #443, opened 2021-09-16
PF-975: Use Pet service account token for all clone operations

jaycarlton:
Clone bucket, dataset, reference, and workspace are included. Tests won't pass until Elijah is healthy.


service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -901,6 +901,36 @@ public String getOrCreatePetSaEmail(String projectId, AuthenticatedUserRequest u
...
+   * Get a pet service account token for a given project and user request
+   *
+   * @param projectId - GCP project ID
+   * @param userRequest - original user request
+   * @return - pet service account token
+   */
+  public String getOrCreatePetSaToken(String projectId, AuthenticatedUserRequest userRequest) {
+    GoogleApi googleApi = samGoogleApi(userRequest.getRequiredToken());
+    try {
+      return googleApi.getPetServiceAccountToken(projectId, new ArrayList<>(SAM_OAUTH_SCOPES));
```
jaycarlton, line 914: This is the part I'm not sure about. Do we need any more scopes for the pet than these?

ddietterich: Besides the possibility of having the token expire, I thought the reason we wanted to use the pet was to get a scope that would allow us to use the data transfer service. I would expect the tests to fail if you need more scopes.

ddietterich (approved):
LGTM

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -901,6 +901,36 @@ public String getOrCreatePetSaEmail(String projectId, AuthenticatedUserRequest u
...
+   * Get an AuthenticatedUserRequest object for the pet SA of this user.
+   * @param projectId - GCP project ID
+   * @param userRequest - User's own AuthenticatedUserRequest. May contain null email and/or subjectID
+   * @return AuthenticatedUserRequest with the token replaced by the pet token
+   */
+  public AuthenticatedUserRequest getAuthenticatedPetRequest(String projectId,
+      AuthenticatedUserRequest userRequest) {
+    final String petToken = getOrCreatePetSaToken(projectId, userRequest);
+    return new AuthenticatedUserRequest(
+        userRequest.getEmail(), userRequest.getSubjectId(), Optional.of(petToken));
```
marikomedlock, line 931: Instead of `Optional`, should this fail if `petToken` is null?

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
marikomedlock: If you're going to use the pet SA token to talk to GCP, then you'll need the `cloud-platform` OAuth scope.

marikomedlock (commented):
In stand-up we've discussed two possible reason(s) for this change.
1. So that the caller (of the WSM endpoints) does not need the `cloud-platform` OAuth scope.
  - That's fine, but I think WSM should be consistent and also do this for the create and check-access endpoints for referenced resources.
  - From the CLI perspective, this seems like a nice-to-have because the CLI can just request the pet SA token itself (with more OAuth scopes) and call WSM endpoints with it. This is what it does now for the referenced resource endpoints.
2. So that the caller's credentials don't expire during long-running WSM operations.
  - I'm not sure I'm following the WSM code flow correctly, so this may be off-base. But it looks like you're requesting the pet SA token at the controller level before launching the flight. From the expiry perspective, is this much different than just using the caller's credentials directly? e.g. The CLI requests a pet SA access token and then immediately uses it to call WSM, seems like the expiry would be very similar to if WSM requested it as soon as it got the request.
  - I was thinking you would request a pet SA token in each step that needed it, so you'd basically restart the clock each time. (Or maybe more sophisticated with caching the token and only refreshing it if needed, etc., but still requesting the token multiple times.)

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
jaycarlton: I don't know why that constructor parameter is optional and the other ones, which frequently aren't there, aren't. It's one thing I'd like to fix in that class.

jaycarlton: `Optional.of()` will throw if it's null. We could also use `.orElseThrow()`.

jaycarlton:
> * That's fine, but I think WSM should be consistent and also do this for the create and check-access endpoints for referenced resources.
Maybe. That can go in this PR.

> * I'm not sure I'm following the WSM code flow correctly, so this may be off-base. But it looks like you're requesting the pet SA token at the controller level before launching the flight. From the expiry perspective, is this much different than just using the caller's credentials directly? e.g. The CLI requests a pet SA access token and then immediately uses it to call WSM, seems like the expiry would be very similar to if WSM requested it as soon as it got the request.
I think you're right.

> * I was thinking you would request a pet SA token in each step that needed it, so you'd basically restart the clock each time. (Or maybe more sophisticated with caching the token and only refreshing it if needed, etc., but still requesting the token multiple times.)
I'd like to do this incrementally, getting this change in first and maybe studying this issue some more.

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -510,20 +523,30 @@ private ApiDeleteControlledGcpAiNotebookInstanceResult fetchNotebookInstanceDele
   public ResponseEntity<ApiCloneControlledGcpBigQueryDatasetResult> cloneBigQueryDataset(
       UUID workspaceId, UUID resourceId, @Valid ApiCloneControlledGcpBigQueryDatasetRequest body) {
     final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    final AuthenticatedUserRequest petRequest =
```
ddietterich, line 526: Maybe make this a little method in the controller?
The cut'n'paste is used many times. I don't think we ever need a userRequest except as input to the getPetCredentials.

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -552,4 +558,16 @@ private ApiCloneControlledGcpBigQueryDatasetResult fetchCloneBigQueryDatasetResu
...
     return authenticatedUserRequestFactory.from(request);
   }
+
+  private AuthenticatedUserRequest getPetRequest(UUID workspaceId) {
+    final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    return petSaService
+        .getWorkspacePetCredentials(workspaceId, userRequest)
+        .orElseThrow(
+            () ->
+                new BadRequestException(
```
marikomedlock, line 568: nit: Not sure, but would `InternalServerError` be more appropriate here? If we're trying to clone a bucket, then presumably we already have a backing GCP project and pet SA for this workspace. I usually think of 400 as an error the caller should fix, and if they've already created a GCP project then I'm not sure what action they could take to fix this.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -715,10 +718,10 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
   public ResponseEntity<ApiCloneReferencedGcpDataRepoSnapshotResourceResult>
       cloneGcpDataRepoSnapshotReference(
```
marikomedlock, line 719: Just curious, do you need the pet to check access to a TDR snapshot? We need it for the GCP resources because we're talking to the cloud directly, but I thought TDR wouldn't need additional OAuth scopes. Should be fine either way (user or pet) though.

marikomedlock (approved):
Could you please add a follow-on ticket to update the CLI to stop calling WSM endpoints with the pet SA token now that it's no longer needed?

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
zloery: You're right, we don't need to use the pet to check snapshot access. IIRC snapshots are shared with proxy groups so we either the user or pet credentials should work, but the pet credentials aren't required.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -753,4 +756,16 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+    final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    return petSaService
+        .getWorkspacePetCredentials(workspaceId, userRequest)
+        .orElseThrow(
+            () ->
+                new BadRequestException(
+                    String.format(
+                        "Pet SA credentials not found for user %s on workspace %s",
+                        userRequest.getEmail(), workspaceId)));
+  }
```
zloery: per slack discussion, this won't work for references. The underlying `getWorkspacePetCredentials` tries to create a pet in the GCP context of the provided workspace, but we may not have a GCP cloud context when we're working with references. Instead of throwing here, I think it's reasonable to fall back to end-user credentials (i.e.
```
.getWorkspacePetCredentials(workspaceId, userRequest).orElse(userRequest)
```
and renaming this method to something like `maybeGetPetRequest` or `getCloningCredentials` with a comment explaining why how it works.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -600,4 +604,16 @@ private ApiCloneWorkspaceResult fetchCloneWorkspaceResult(
     }
     return propertyMap;
   }
+
+  private AuthenticatedUserRequest getPetRequest(UUID workspaceId) {
```
zloery, line 608: Same issue here, it's valid to clone a workspace which does not have a GCP cloud context.

zloery (approved):

PR #445, opened 2021-09-20
[no ticket] gcloud command to set integration cluster

jaycarlton:
Somehow my gcloud creds were pointing to the dev cluster instead of integration, which led to an error about not connecting to the IP address. h/t @mflinn-broad

### DIFF
--- a/service/local-dev/README.md
+++ b/service/local-dev/README.md
@@ -18,6 +18,11 @@ Optionally, you can provide a branch of [terra-helm](https://github.com/broadins
 ./setup_local_env.sh <environment> [<terra-helm ref>] [<terra-helmfile ref>] [ssh|http]
 ```
 
+Ensure your `gcloud` state is pointing to the correct cluster:
+```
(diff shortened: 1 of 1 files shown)


jmczerk (approved):

mflinn-broad (approved):

ddietterich:
I think this should go into DEVELOPMENT.md and not the README.

jaycarlton:
> I think this should go into DEVELOPMENT.md and not the README.

This is a directory-specific readme for the skaffold stuff. I thought it fit here, since it goes in between those two commands. It may fit in DEVELOPMNET.md too.

ddietterich:
> > I think this should go into DEVELOPMENT.md and not the README.
> 
> This is a directory-specific readme for the skaffold stuff. I thought it fit here, since it goes in between those two commands. It may fit in DEVELOPMNET.md too.

Ah, OK. Thanks for clarifying

zloery (approved):

PR #446, opened 2021-09-20
[no ticket] Add script for cleaning up after test users

zloery:
This adds a small script to clean up a user's Sam resources leftover after WSM tests fail to clean them up.

### DIFF
--- a/scripts/cleanup-sam-resources.sh
+++ b/scripts/cleanup-sam-resources.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+#
+# A script for cleaning up a user's Sam resources, such as those not cleaned
+# up after testing. This will attempt to delete all controlled (both shared and
+# private) resources owned by a user in Sam, as well as all their workspaces.
+#
+# This script will attempt to delete all of the resources and workspaces a user
+# has any access to, including ones the user does not have permission to delete
+# (such as those owned by other test users, where the specified user is only
+# a reader). This means some delete calls will fail with a 403 error and the
+# message "You may not perform any of [DELETE] on
+# {resource type}/{resource id}". This is expected and can be ignored.
+#
+# Input:
+#   access_token: An access token used for calls to Sam. This script will delete
+#   all controlled resources and workspaces in Sam that the user owns, so be
+#   sure this belongs to the correct user.
+
+access_token=$1
+sam_url="https://sam.dsde-dev.broadinstitute.org/api"
+
+# Workspaces must be deleted last as they are parents of other resource types.
+# Sam will not allow us to delete a workspace which contains controlled resources.
+for resource_type in "controlled-user-shared-workspace-resource" "controlled-user-private-workspace-resource" "workspace"; do
+  echo "Cleaning up $resource_type"
+  # Fetch resources belonging to this user and save their IDs.
+  curl -X GET "$sam_url/resources/v2/$resource_type" \
+  -H "accept: application/json" -H "Authorization: Bearer $access_token" \
+  | jq -r ".[] | .resourceId" >$resource_type.txt
+  # For each ID, call Sam to delete the resource.
+  while read -r resource_id; do
+    echo "Deleting $resource_id"
+    curl -X DELETE "$sam_url/resources/v2/$resource_type/$resource_id" \
+    -H "accept: */*" -H "Authorization: Bearer $access_token"
+  done <$resource_type.txt
+  # Clean up the intermediate file
+  rm $resource_type.txt
+done
--- a/scripts/cleanup-sam-resources.sh
+++ b/scripts/cleanup-sam-resources.sh
@@ -27,6 +27,13 @@ for resource_type in "controlled-user-shared-workspace-resource" "controlled-use
   curl -X GET "$sam_url/resources/v2/$resource_type" \
   -H "accept: application/json" -H "Authorization: Bearer $access_token" \
   | jq -r ".[] | .resourceId" >$resource_type.txt
+  # Prompt user for confirmation
+  num_resources="$(wc -l <$resource_type.txt)"
+  read -p "This will delete $num_resources resources of type $resource_type. Continue: y/N? " yn
+  if [ "$yn" == "${yn#[Yy]}" ] ;then
+    break
+  fi
+
   # For each ID, call Sam to delete the resource.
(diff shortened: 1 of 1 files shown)


ddietterich (approved):
lgtm

scripts/cleanup-sam-resources.sh
```diff
@@ -0,0 +1,38 @@
...
+# a reader). This means some delete calls will fail with a 403 error and the
+# message "You may not perform any of [DELETE] on
+# {resource type}/{resource id}". This is expected and can be ignored.
+#
+# Input:
+#   access_token: An access token used for calls to Sam. This script will delete
+#   all controlled resources and workspaces in Sam that the user owns, so be
+#   sure this belongs to the correct user.
+
+access_token=$1
```
jaycarlton, line 19: It'd be neat to give a confirmation option, like "This will delete 2324 resources. Continue?".

scripts/cleanup-sam-resources.sh
```diff
@@ -0,0 +1,38 @@
...
+#
+# This script will attempt to delete all of the resources and workspaces a user
+# has any access to, including ones the user does not have permission to delete
+# (such as those owned by other test users, where the specified user is only
+# a reader). This means some delete calls will fail with a 403 error and the
+# message "You may not perform any of [DELETE] on
+# {resource type}/{resource id}". This is expected and can be ignored.
+#
+# Input:
+#   access_token: An access token used for calls to Sam. This script will delete
```
jaycarlton, line 15: Where/how do I get this access token? It would be safer if we could supply the email, as that's unequivocally a test user.

scripts/cleanup-sam-resources.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+#
+# A script for cleaning up a user's Sam resources, such as those not cleaned
+# up after testing. This will attempt to delete all controlled (both shared and
+# private) resources owned by a user in Sam, as well as all their workspaces.
+#
```
jaycarlton, line 6: Is it possible to put in a min_age guard, so that for example we don't delete anything less than 14 days old? That would leave any necessary resources around for test investigations, debugging, etc.

scripts/cleanup-sam-resources.sh
```diff
@@ -0,0 +1,38 @@
...
+  | jq -r ".[] | .resourceId" >$resource_type.txt
+  # For each ID, call Sam to delete the resource.
+  while read -r resource_id; do
+    echo "Deleting $resource_id"
+    curl -X DELETE "$sam_url/resources/v2/$resource_type/$resource_id" \
+    -H "accept: */*" -H "Authorization: Bearer $access_token"
+  done <$resource_type.txt
+  # Clean up the intermediate file
+  rm $resource_type.txt
+done
```
jaycarlton, line 38: Is there an endpoint or gcloud command to get the number of groups the user is in before and after the script runs? That would be nice to have.

jaycarlton (approved):
Some ideas for enhancements, but this is hugely helpful as-is.

scripts/cleanup-sam-resources.sh
```diff
@@ -0,0 +1,38 @@
...
+  -H "accept: application/json" -H "Authorization: Bearer $access_token" \
+  | jq -r ".[] | .resourceId" >$resource_type.txt
+  # For each ID, call Sam to delete the resource.
+  while read -r resource_id; do
+    echo "Deleting $resource_id"
+    curl -X DELETE "$sam_url/resources/v2/$resource_type/$resource_id" \
+    -H "accept: */*" -H "Authorization: Bearer $access_token"
+  done <$resource_type.txt
+  # Clean up the intermediate file
+  rm $resource_type.txt
```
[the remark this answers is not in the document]
zloery: I think having those files (and especially old versions of them) lying around would probably be more trouble than they're worth. This does already print the resource types and IDs while running, so it should already be manageable to track what's deleted.

scripts/cleanup-sam-resources.sh
```diff
@@ -0,0 +1,38 @@
...
+for resource_type in "controlled-user-shared-workspace-resource" "controlled-user-private-workspace-resource" "workspace"; do
+  echo "Cleaning up $resource_type"
+  # Fetch resources belonging to this user and save their IDs.
+  curl -X GET "$sam_url/resources/v2/$resource_type" \
+  -H "accept: application/json" -H "Authorization: Bearer $access_token" \
+  | jq -r ".[] | .resourceId" >$resource_type.txt
+  # For each ID, call Sam to delete the resource.
+  while read -r resource_id; do
+    echo "Deleting $resource_id"
+    curl -X DELETE "$sam_url/resources/v2/$resource_type/$resource_id" \
```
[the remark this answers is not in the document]
zloery: Nope. There's been some discussion about it, but nothing concrete.

scripts/cleanup-sam-resources.sh
```diff
@@ -0,0 +1,38 @@
...
+
+access_token=$1
+sam_url="https://sam.dsde-dev.broadinstitute.org/api"
+
+# Workspaces must be deleted last as they are parents of other resource types.
+# Sam will not allow us to delete a workspace which contains controlled resources.
+for resource_type in "controlled-user-shared-workspace-resource" "controlled-user-private-workspace-resource" "workspace"; do
+  echo "Cleaning up $resource_type"
+  # Fetch resources belonging to this user and save their IDs.
+  curl -X GET "$sam_url/resources/v2/$resource_type" \
```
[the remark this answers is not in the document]
zloery: This runs the delete requests serially and each one takes some time, so even without delays it doesn't go above ~5 QPS and is generally closer to 1-2. Looking at dev Sam monitoring from this afternoon I'm not concerned about the impact.

scripts/cleanup-sam-resources.sh
zloery: I like that idea, thanks!

scripts/cleanup-sam-resources.sh
zloery: Either through service account impersonation (what we do for tests) or using login information from the test user tracker. It's probably possible to script the first one (assuming the user also renders configs separately) but I'm not sure how, and I think automating that would probably take more time than it saves.

PR #447, opened 2021-09-21
[PF-992] Fix email extraction to call Sam when needed

zloery:
A previous change replaced a call to `getEmailFromToken` with `getRequestUserEmail`, which have subtly different (and poorly documented) differences in behavior around reading an email from a `AuthenticatedUserRequest` object directly vs. calling Sam. This is an important difference when making requests as a pet service account, as the userRequest will contain the pet's email but Sam will return the owner's email.
I've renamed the methods here to be clearer and added additional documentation. In order not to break the Azure POC changes from #440, I've also changed the interface of `getEmailFromSam` to take an `AuthenticatedUserRequest` instead of a token directly.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
@@ -549,7 +549,8 @@ public class ControlledGcpResourceApiController implements ControlledGcpResource
         .orElseGet(
             () ->
                 SamRethrow.onInterrupted(
-                    () -> samService.getEmailFromRequestOrSam(userRequest), "getRequestUserEmail"));
+                    () -> samService.getEmailFromRequestOrSam(userRequest),
+                    "getEmailFromRequestOrSam"));
   }
 
   /**
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
@@ -549,7 +549,7 @@ public class ControlledGcpResourceApiController implements ControlledGcpResource
         .orElseGet(
             () ->
                 SamRethrow.onInterrupted(
-                    () -> samService.getRequestUserEmail(userRequest), "getRequestUserEmail"));
+                    () -> samService.getEmailFromRequestOrSam(userRequest), "getRequestUserEmail"));
   }
 
   /**
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -108,22 +108,17 @@ public class SamService {
    */
   public String getEmailFromRequestOrSam(AuthenticatedUserRequest userRequest)
       throws InterruptedException {
-    Optional<String> emailMaybe = Optional.ofNullable(userRequest.getEmail());
-    if (emailMaybe.isPresent()) {
-      return emailMaybe.get();
-    } else {
-      return getEmailFromSam(userRequest);
-    }
+    return Optional.ofNullable(userRequest.getEmail()).orElse(getUserEmailFromSam(userRequest));
   }
 
   /**
-   * Fetch the email associated with user credentials directly from Sam. Unlike
-   * {@code getRequestUserEmail}, this will always call Sam to fetch an email and will never read
-   * it from the AuthenticatedUserRequest. This is important for calls made by pet service accounts,
-   * which will have a pet email in the AuthenticatedUserRequest, but Sam will return the owner's
-   * email.
-   * */
-  public String getEmailFromSam(AuthenticatedUserRequest userRequest) throws InterruptedException {
+   * Fetch the email associated with user credentials directly from Sam. Unlike {@code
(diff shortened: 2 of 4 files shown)


ddietterich (approved):
LGTM - I wish it didn't work this way, but...

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -549,7 +549,7 @@ private String assignedUserFromBodyOrToken(
         .orElseGet(
             () ->
                 SamRethrow.onInterrupted(
-                    () -> samService.getRequestUserEmail(userRequest), "getRequestUserEmail"));
+                    () -> samService.getEmailFromRequestOrSam(userRequest), "getRequestUserEmail"));
```
jaycarlton, line 552: change the description string to `getEmailFromRequestOrSam`

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -106,13 +106,29 @@ public String getWsmServiceAccountToken() throws IOException {
...
+  }
+
+  /**
+   * Fetch the email associated with user credentials directly from Sam. Unlike
+   * {@code getRequestUserEmail}, this will always call Sam to fetch an email and will never read
+   * it from the AuthenticatedUserRequest. This is important for calls made by pet service accounts,
+   * which will have a pet email in the AuthenticatedUserRequest, but Sam will return the owner's
+   * email.
+   * */
+  public String getEmailFromSam(AuthenticatedUserRequest userRequest) throws InterruptedException {
```
jaycarlton, line 126: maybe `getUserEmailFromSam()`. It kinda seems redundant on the `SamService` to have to say `fromSam` but I agree it's a necessary evil.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -185,9 +201,7 @@ public void createWorkspaceWithDefaults(AuthenticatedUserRequest userRequest, UU
     // Sam will throw an error if no owner is specified, so the caller's email is required. It can
     // be looked up using the auth token if that's all the caller provides.
     String callerEmail =
-        userRequest.getEmail() == null
-            ? getEmailFromToken(userRequest.getRequiredToken())
-            : userRequest.getEmail();
+        userRequest.getEmail() == null ? getEmailFromSam(userRequest) : userRequest.getEmail();
```
jaycarlton, line 204: I'd use `Optional.ofNullable(userRequest.getEmail()).orElseGet(() -> getEmailFromSam(userRequest))`

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -106,13 +106,29 @@ public String getWsmServiceAccountToken() throws IOException {
    * @param userRequest - request object for this user
    * @return - email address of user
    */
-  public String getRequestUserEmail(AuthenticatedUserRequest userRequest)
+  public String getEmailFromRequestOrSam(AuthenticatedUserRequest userRequest)
       throws InterruptedException {
     Optional<String> emailMaybe = Optional.ofNullable(userRequest.getEmail());
```
jaycarlton, line 111: Can we (in a separate PR) make this getter return `Optional<String>`? Same question for the subject ID.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -106,13 +106,29 @@ public String getWsmServiceAccountToken() throws IOException {
...
     if (emailMaybe.isPresent()) {
```
jaycarlton, line 112: just `return emailMaybe.orElseGet(() -> getEmailFromSam(userRequest));`. You should rarely need to call `get()` directly. IJ should have an auto-fix hint for it.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -828,7 +842,9 @@ public Boolean status() {
   private void addWsmResourceOwnerPolicy(CreateResourceRequestV2 request)
       throws InterruptedException {
     try {
-      String wsmSaEmail = getEmailFromToken(getWsmServiceAccountToken());
+      AuthenticatedUserRequest wsmRequest =
+          new AuthenticatedUserRequest().token(Optional.of(getWsmServiceAccountToken()));
```
jaycarlton, line 846: likewise (separately), we should change this signature not to take in an `Optional`, per the java guide.

jaycarlton (approved):
Thank you for these improvements. There are still many smells here, and no shortage of footguns, but at least we know which fetcher uses the getter now.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: Tracking in PF-1001. Agreed that should be a separate PR, I think it would have a lot of small changes scattered across the codebase

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: PF-1001

PR #450, opened 2021-09-21
[no ticket] Fix reference to CloneWorkspaceRequest

cbookg:


jaycarlton (approved):
Interesting. I wonder how that worked. In general, OpenAPI mistakes seem to get silently ignored, so it's good to have a fresh set of eyes and tooling on this.

ddietterich (approved):
lgtm

PR #451, opened 2021-09-22
[no ticket] Use Human email instead of pet email

jaycarlton:
When calling WSM's clone endpoint with a pet SA (to get around the lack of a `cloud_platform` scope for human users), we ran into a problem wherein the proxy was populating the `AuthenticatedUserRequest`'s email field with the pet SA's email address. When we used that email address to assign ownership of a new workspace, it turned out that Sam doesn't recognize the user's email or token when we later check access against that email address.

@zloery discovered some other endpoints in Sam that behave in unexpected ways with the pet SA email address. For example, we can't add a pet as a workspace member via their email address, and a pet can't delete a workspace it created with its email address.

The solution for the immediate problem in getting the CLI's CloneWorkspace test to run is simple: just ask Sam for the "real boy" email and use that instead of defaulting to the proxy-populated value in the `AuthenticaedUserRequest`.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -108,22 +108,17 @@ public class SamService {
    */
   public String getEmailFromRequestOrSam(AuthenticatedUserRequest userRequest)
       throws InterruptedException {
-    Optional<String> emailMaybe = Optional.ofNullable(userRequest.getEmail());
-    if (emailMaybe.isPresent()) {
-      return emailMaybe.get();
-    } else {
-      return getEmailFromSam(userRequest);
-    }
+    return Optional.ofNullable(userRequest.getEmail()).orElse(getUserEmailFromSam(userRequest));
   }
 
   /**
-   * Fetch the email associated with user credentials directly from Sam. Unlike
-   * {@code getRequestUserEmail}, this will always call Sam to fetch an email and will never read
-   * it from the AuthenticatedUserRequest. This is important for calls made by pet service accounts,
-   * which will have a pet email in the AuthenticatedUserRequest, but Sam will return the owner's
-   * email.
-   * */
-  public String getEmailFromSam(AuthenticatedUserRequest userRequest) throws InterruptedException {
+   * Fetch the email associated with user credentials directly from Sam. Unlike {@code
+   * getRequestUserEmail}, this will always call Sam to fetch an email and will never read it from
+   * the AuthenticatedUserRequest. This is important for calls made by pet service accounts, which
+   * will have a pet email in the AuthenticatedUserRequest, but Sam will return the owner's email.
+   */
+  public String getUserEmailFromSam(AuthenticatedUserRequest userRequest)
+      throws InterruptedException {
     UsersApi usersApi = samUsersApi(userRequest.getRequiredToken());
     try {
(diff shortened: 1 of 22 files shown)


service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -195,12 +195,16 @@ public void createWorkspaceWithDefaults(AuthenticatedUserRequest userRequest, UU
     ResourcesApi resourceApi = samResourcesApi(userRequest.getRequiredToken());
     // Sam will throw an error if no owner is specified, so the caller's email is required. It can
     // be looked up using the auth token if that's all the caller provides.
-    String callerEmail =
-        Optional.ofNullable(userRequest.getEmail()).orElse(getUserEmailFromSam(userRequest));
+
+    // If we called WSM as the pet SA and went through the proxy, this becomes the pet SA's email if
+    // we use the pet email. That caused an issue where the human user wasn't recognized on the
```
zloery, line 200: nit: "if we use the request email"?

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -195,12 +195,16 @@ public void createWorkspaceWithDefaults(AuthenticatedUserRequest userRequest, UU
...
+    // workspace.
+
+    String humanUserEmail = getUserEmailFromSam(userRequest);
     CreateResourceRequestV2 workspaceRequest =
         new CreateResourceRequestV2()
             .resourceId(id.toString())
-            .policies(defaultWorkspacePolicies(callerEmail));
+            .policies(defaultWorkspacePolicies(humanUserEmail));
```
marikomedlock, line 207: Are there other places WSM sets an email as a SAM resource owner? I was thinking of maybe controlled resources. But not sure if this applies to child resources also.

zloery (changes requested):
Sorry to expand the scope of this PR, but could you also change all the other call sites of `getEmailFromRequestOrSam` (which does the same thing as the `Optional` here)  to use `getUserEmailFromSam`? As-is, I think we'll hit this same bug in a few other places (e.g. assigning roles on private resources directly to pets, and then users won't have access)

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
jaycarlton: For controlled resources, the parent is the fully qualified workspace ID. In `bio.terra.workspace.service.iam.SamService#createControlledResource`, the Api is created from the user associated with the token in the user request.  For private resources, I've changed it to use the user email from Sam instead of the request email.

marikomedlock (approved):

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -679,7 +666,7 @@ public void createControlledResource(
     // Set up the master OWNER user in Sam for all controlled resources, if it's not already.
     initializeWsmServiceAccount();
     ResourcesApi resourceApi = samResourcesApi(userRequest.getRequiredToken());
-    FullyQualifiedResourceId workspaceParentFqId =
+    FullyQualifiedResourceId  workspaceParentFqId =
```
zloery, line 669: nit: remove double space

zloery (approved):

PR #452, opened 2021-09-22
[PF-1002] Refactor cloud context

ddietterich:
This refactor is in preparation for Azure and in support of the PoC.I initially made these
changes in the PoC branch, but they touch a lot of files that have nothing to do with the
PoC.

There are three main changes. First, remove cloud context from workspace object. Having
it there makes the DAO complex. Also, most use cases in flights do not need to fetch the
workspace object, but can get the cloud context with the workspace id and use it to get
the GCP project id. Added helper methods in WorkspaceService for getting the context and
the project id.

Second, add a `creating_flight` column to the cloud_context table. Use that instead of
the project id for delete checking. Azure does not have a per-context unique id to check.
The flight id provides a general solution to ensure that the undo of a cloud context
create will only delete the context that it created.

Third, moved the serdes for GcpCloudContext into the GcpCloudContext class instead of
having it in the DAO. Seems tidier to me that way: cloud context knows how to serialize
and deserialize itself.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -72,6 +72,7 @@ public class WorkspaceApiController implements WorkspaceApi {
   private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
   private final HttpServletRequest request;
   private final ReferencedResourceService referenceResourceService;
+  private final Logger logger = LoggerFactory.getLogger(WorkspaceApiController.class);
 
   @Autowired
   public WorkspaceApiController(
@@ -88,8 +89,6 @@ public class WorkspaceApiController implements WorkspaceApi {
     this.request = request;
     this.referenceResourceService = referenceResourceService;
   }
-
-  private final Logger logger = LoggerFactory.getLogger(WorkspaceApiController.class);
 
   private AuthenticatedUserRequest getAuthenticatedInfo() {
     return authenticatedUserRequestFactory.from(request);
@@ -142,9 +141,12 @@ public class WorkspaceApiController implements WorkspaceApi {
 
   private ApiWorkspaceDescription buildWorkspaceDescription(Workspace workspace) {
     ApiGcpContext gcpContext =
-        workspace.getGcpCloudContext().map(GcpCloudContext::toApi).orElse(null);
-    // Note projectId will be null here if no GCP cloud context exists.
+        workspaceService
+            .getGcpCloudContext(workspace.getWorkspaceId())
+            .map(GcpCloudContext::toApi)
+            .orElse(null);
     // When we have another cloud context, we will need to do a similar retrieval for it.
+
     return new ApiWorkspaceDescription()
         .id(workspace.getWorkspaceId())
         .spendProfile(workspace.getSpendProfileId().map(SpendProfileId::id).orElse(null))
--- a/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -72,7 +72,7 @@ public class WorkspaceApiController implements WorkspaceApi {
   private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
   private final HttpServletRequest request;
   private final ReferencedResourceService referenceResourceService;
-  private final Logger logger = LoggerFactory.getLogger(WorkspaceApiController.class);
+  private static final Logger logger = LoggerFactory.getLogger(WorkspaceApiController.class);
 
   @Autowired
   public WorkspaceApiController(
--- a/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -43,6 +43,7 @@ import bio.terra.workspace.service.resource.referenced.ReferencedResource;
 import bio.terra.workspace.service.resource.referenced.ReferencedResourceService;
 import bio.terra.workspace.service.resource.referenced.exception.InvalidReferenceException;
 import bio.terra.workspace.service.spendprofile.SpendProfileId;
+import bio.terra.workspace.service.workspace.GcpCloudContextService;
 import bio.terra.workspace.service.workspace.WorkspaceService;
 import bio.terra.workspace.service.workspace.model.GcpCloudContext;
 import bio.terra.workspace.service.workspace.model.Workspace;
@@ -72,6 +73,7 @@ public class WorkspaceApiController implements WorkspaceApi {
   private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
   private final HttpServletRequest request;
   private final ReferencedResourceService referenceResourceService;
+  private final GcpCloudContextService gcpCloudContextService;
   private static final Logger logger = LoggerFactory.getLogger(WorkspaceApiController.class);
 
   @Autowired
@@ -81,13 +83,15 @@ public class WorkspaceApiController implements WorkspaceApi {
       SamService samService,
       AuthenticatedUserRequestFactory authenticatedUserRequestFactory,
       HttpServletRequest request,
-      ReferencedResourceService referenceResourceService) {
+      ReferencedResourceService referenceResourceService,
+      GcpCloudContextService gcpCloudContextService) {
     this.workspaceService = workspaceService;
     this.jobService = jobService;
     this.samService = samService;
     this.authenticatedUserRequestFactory = authenticatedUserRequestFactory;
     this.request = request;
     this.referenceResourceService = referenceResourceService;
+    this.gcpCloudContextService = gcpCloudContextService;
   }
 
   private AuthenticatedUserRequest getAuthenticatedInfo() {
@@ -140,8 +144,12 @@ public class WorkspaceApiController implements WorkspaceApi {
   }
 
   private ApiWorkspaceDescription buildWorkspaceDescription(Workspace workspace) {
+    // This is a special case use of GcpCloudContextService. That service does not
+    // do authorization checking, so it is not normally called from controllers.
+    // However, the caller has already done the auth check, so it is safe in this
+    // context.
     ApiGcpContext gcpContext =
-        workspaceService
+        gcpCloudContextService
             .getGcpCloudContext(workspace.getWorkspaceId())
             .map(GcpCloudContext::toApi)
             .orElse(null);
--- a/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
@@ -10,6 +10,7 @@ import bio.terra.workspace.service.workspace.exceptions.DuplicateWorkspaceExcept
 import bio.terra.workspace.service.workspace.model.CloudPlatform;
 import bio.terra.workspace.service.workspace.model.Workspace;
 import bio.terra.workspace.service.workspace.model.WorkspaceStage;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
@@ -208,6 +209,11 @@ public class WorkspaceDao {
    */
   @ReadTransaction
   public List<Workspace> getWorkspacesMatchingList(List<UUID> idList, int offset, int limit) {
+    // If the incoming list is empty, the caller does not have permission to see any
+    // workspaces, so we return an empty list.
+    if (idList.isEmpty()) {
+      return Collections.emptyList();
+    }
     String sql =
         WORKSPACE_SELECT_SQL
             + " WHERE workspace_id IN (:workspace_ids) ORDER BY workspace_id OFFSET :offset LIMIT :limit";
--- a/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
@@ -7,17 +7,15 @@ import bio.terra.workspace.db.exception.WorkspaceNotFoundException;
 import bio.terra.workspace.service.spendprofile.SpendProfileId;
 import bio.terra.workspace.service.workspace.exceptions.DuplicateCloudContextException;
 import bio.terra.workspace.service.workspace.exceptions.DuplicateWorkspaceException;
-import bio.terra.workspace.service.workspace.exceptions.InvalidSerializedVersionException;
 import bio.terra.workspace.service.workspace.model.CloudPlatform;
 import bio.terra.workspace.service.workspace.model.GcpCloudContext;
 import bio.terra.workspace.service.workspace.model.Workspace;
 import bio.terra.workspace.service.workspace.model.WorkspaceStage;
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.annotations.VisibleForTesting;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -38,18 +36,30 @@ import org.springframework.stereotype.Component;
  */
 @Component
 public class WorkspaceDao {
-  /** Serializing format of the GCP cloud context */
-  @VisibleForTesting static final long GCP_CLOUD_CONTEXT_DB_VERSION = 1;
+  private final Logger logger = LoggerFactory.getLogger(WorkspaceDao.class);
+  private final NamedParameterJdbcTemplate jdbcTemplate;
 
   /** SQL query for reading a workspace, including its cloud context. */
   private static final String WORKSPACE_SELECT_SQL =
-      "SELECT W.workspace_id, W.display_name, W.description, W.spend_profile,"
-          + " W.properties, W.workspace_stage, C.context"
-          + " FROM workspace W LEFT JOIN cloud_context C"
-          + " ON W.workspace_id = C.workspace_id ";
+      "SELECT workspace_id, display_name, description, spend_profile, properties, workspace_stage"
+          + " FROM workspace W ";
 
-  private final NamedParameterJdbcTemplate jdbcTemplate;
-  private final Logger logger = LoggerFactory.getLogger(WorkspaceDao.class);
+  private static final RowMapper<Workspace> WORKSPACE_ROW_MAPPER =
+      (rs, rowNum) ->
+          Workspace.builder()
+              .workspaceId(UUID.fromString(rs.getString("workspace_id")))
+              .displayName(rs.getString("display_name"))
+              .description(rs.getString("description"))
+              .spendProfileId(
+                  Optional.ofNullable(rs.getString("spend_profile"))
+                      .map(SpendProfileId::create)
+                      .orElse(null))
+              .properties(
+                  Optional.ofNullable(rs.getString("properties"))
+                      .map(DbSerDes::jsonToProperties)
+                      .orElse(null))
+              .workspaceStage(WorkspaceStage.valueOf(rs.getString("workspace_stage")))
+              .build();
 
   @Autowired
   public WorkspaceDao(NamedParameterJdbcTemplate jdbcTemplate) {
@@ -128,9 +138,7 @@ public class WorkspaceDao {
     if (id == null) {
       throw new MissingRequiredFieldException("Valid workspace id is required");
     }
-    String sql =
-        WORKSPACE_SELECT_SQL
-            + " WHERE W.workspace_id = :id AND (C.cloud_platform = 'GCP' OR C.cloud_platform IS NULL)";
+    String sql = WORKSPACE_SELECT_SQL + " WHERE W.workspace_id = :id";
     MapSqlParameterSource params = new MapSqlParameterSource().addValue("id", id.toString());
     try {
       Workspace result =
@@ -222,49 +230,19 @@ public class WorkspaceDao {
    */
   @ReadTransaction
   public Optional<GcpCloudContext> getGcpCloudContext(UUID workspaceId) {
-    String sql =
-        "SELECT context FROM cloud_context "
-            + "WHERE workspace_id = :workspace_id AND cloud_platform = :cloud_platform";
-    MapSqlParameterSource params =
-        new MapSqlParameterSource()
-            .addValue("workspace_id", workspaceId.toString())
-            .addValue("cloud_platform", CloudPlatform.GCP.toString());
-    try {
-      return Optional.ofNullable(
-          DataAccessUtils.singleResult(
-              jdbcTemplate.query(
-                  sql,
-                  params,
-                  (rs, rowNum) -> deserializeGcpCloudContext(rs.getString("context")))));
-    } catch (EmptyResultDataAccessException e) {
-      return Optional.empty();
-    }
+    return getCloudContextWorker(workspaceId, CloudPlatform.GCP).map(GcpCloudContext::deserialize);
   }
 
   /**
-   * Create the cloud context of the workspace
+   * Create the GCP cloud context of the workspace
    *
    * @param workspaceId unique id of the workspace
    * @param cloudContext the GCP cloud context to create
    */
   @WriteTransaction
-  public void createGcpCloudContext(UUID workspaceId, GcpCloudContext cloudContext) {
-    final String sql =
-        "INSERT INTO cloud_context (workspace_id, cloud_platform, context)"
-            + " VALUES (:workspace_id, :cloud_platform, :context::json)";
-    MapSqlParameterSource params =
-        new MapSqlParameterSource()
-            .addValue("workspace_id", workspaceId.toString())
-            .addValue("cloud_platform", CloudPlatform.GCP.toString())
-            .addValue("context", serializeGcpCloudContext(cloudContext));
-    try {
-      jdbcTemplate.update(sql, params);
-      logger.info("Inserted record for GCP cloud context for workspace {}", workspaceId);
-    } catch (DuplicateKeyException e) {
-      throw new DuplicateCloudContextException(
-          String.format("Workspace with id %s already has context for GCP cloud type", workspaceId),
-          e);
-    }
+  public void createGcpCloudContext(
+      UUID workspaceId, GcpCloudContext cloudContext, String flightId) {
+    createCloudContextWorker(workspaceId, CloudPlatform.GCP, cloudContext.serialize(), flightId);
   }
 
   /**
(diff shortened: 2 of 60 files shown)


service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -222,145 +230,131 @@ private boolean updateWorkspaceColumns(UUID workspaceId, MapSqlParameterSource c
...
+      sql = sql + " AND creating_flight = :flightid";
+      params.addValue("flightid", flightId);
+    }
 
     int rowsAffected = jdbcTemplate.update(sql, params);
     boolean deleted = rowsAffected > 0;
 
     if (deleted) {
-      logger.info("Deleted GCP cloud context for workspace {}", workspaceId);
+      logger.info("Deleted {} cloud context for workspace {}", platform, workspaceId);
```
jaycarlton, line 354: should these and similar loggings be `DEBUG`?

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -312,6 +312,36 @@ public void deleteGcpCloudContext(UUID workspaceId, AuthenticatedUserRequest use
         .submitAndWait(null);
   }
 
+  /**
+   * Retrieve the optional GCP cloud context
+   *
+   * @param workspaceId workspace identifier of the cloud context
+   * @return optional GCP cloud context
+   */
+  public Optional<GcpCloudContext> getGcpCloudContext(UUID workspaceId) {
```
jaycarlton, line 321: aside: I'm wondering if we've missed an opportunity for an abstraction to keep the functions that work on a particular cloud platform together. It feels like a really fundamental categorization that's buried in the `Gcp` infix. Maybe there could be separate interfaces for the two clouds, or even separate services. In other words, representing/enforcing a major abstraction with a naming convention only seems like a code smell. Just thinking out loud.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -312,6 +312,36 @@ public void deleteGcpCloudContext(UUID workspaceId, AuthenticatedUserRequest use
...
+
+  /**
+   * Helper method for looking up the GCP project ID for a given workspace ID, if one exists. Unlike
+   * {@link #getRequiredGcpProject(UUID)}, this returns an empty Optional instead of throwing if the
+   * given workspace does not have a GCP cloud context.
+   *
+   * @param workspaceId workspace identifier of the cloud context
+   * @return optional GCP project from the cloud context
+   */
+  public Optional<String> getGcpProject(UUID workspaceId) {
```
jaycarlton, line 341: So a workspace will have 0 or 1 GCP projects and 0 or 1 Azure projects?

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -320,27 +350,14 @@ public void deleteGcpCloudContext(UUID workspaceId, AuthenticatedUserRequest use
    * @return GCP project id
    */
   public String getRequiredGcpProject(UUID workspaceId) {
-    Workspace workspace = workspaceDao.getWorkspace(workspaceId);
     GcpCloudContext gcpCloudContext =
-        workspace
-            .getGcpCloudContext()
+        workspaceDao
+            .getGcpCloudContext(workspaceId)
             .orElseThrow(
```
jaycarlton, line 356: nit: use `.map(GcpCloudContext::getGcpProjectId)` before `.orElseThrow` and just return the whole shebang.

service/src/main/java/bio/terra/workspace/service/workspace/flight/StoreGcpContextStep.java
```diff
@@ -28,16 +28,17 @@ public StepResult doStep(FlightContext flightContext) throws InterruptedExceptio
...
     return StepResult.getStepResultSuccess();
   }
 
   @Override
   public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
     String projectId = flightContext.getWorkingMap().get(GCP_PROJECT_ID, String.class);
 
-    // Delete the cloud context, but only if it is the one with our project id
-    workspaceDao.deleteGcpCloudContextWithIdCheck(workspaceId, projectId);
+    // Delete the cloud context, but only if it is the one with our flight id
```
jaycarlton, line 40: Oh, I get it now.

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
```diff
@@ -19,4 +21,36 @@ public String getGcpProjectId() {
   public ApiGcpContext toApi() {
     return new ApiGcpContext().projectId(getGcpProjectId());
   }
+
+  public static final long GCP_CLOUD_CONTEXT_DB_VERSION = 1;
```
jaycarlton, line 25: nit: define this constant  in `GcpCloudContextV1`. This class doesn't need to keep up with a mapping of versions to  classes.

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
```diff
@@ -19,4 +21,36 @@ public String getGcpProjectId() {
...
+
+  public String serialize() {
+    GcpCloudContextV1 dbContext = GcpCloudContextV1.from(getGcpProjectId());
+    return DbSerDes.toJson(dbContext);
+  }
+
+  public static GcpCloudContext deserialize(String json) {
```
jaycarlton, line 32: nit: `@Nullable` on arg and method itself

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
```diff
@@ -19,4 +21,36 @@ public String getGcpProjectId() {
...
+    GcpCloudContextV1 result = DbSerDes.fromJson(json, GcpCloudContextV1.class);
+    if (result.version != GCP_CLOUD_CONTEXT_DB_VERSION) {
+      throw new InvalidSerializedVersionException("Invalid serialized version");
+    }
+    return new GcpCloudContext(result.gcpProjectId);
+  }
+
+  public static class GcpCloudContextV1 {
+    /** Version marker to store in the db so that we can update the format later if we need to. */
+    @JsonProperty public final long version = GCP_CLOUD_CONTEXT_DB_VERSION;
```
jaycarlton, line 46: if you do it this way, how will you define it for the V2 class?

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -72,6 +72,7 @@
   private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
   private final HttpServletRequest request;
   private final ReferencedResourceService referenceResourceService;
+  private final Logger logger = LoggerFactory.getLogger(WorkspaceApiController.class);
```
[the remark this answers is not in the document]
ddietterich: Since this is an `@Component` singleton, does it make any difference to make it static?

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -222,145 +230,131 @@ private boolean updateWorkspaceColumns(UUID workspaceId, MapSqlParameterSource c
...
-      return Optional.ofNullable(
-          DataAccessUtils.singleResult(
-              jdbcTemplate.query(
-                  sql,
-                  params,
-                  (rs, rowNum) -> deserializeGcpCloudContext(rs.getString("context")))));
-    } catch (EmptyResultDataAccessException e) {
-      return Optional.empty();
-    }
+    return getCloudContextWorker(workspaceId, CloudPlatform.GCP).map(GcpCloudContext::deserialize);
```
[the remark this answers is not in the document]
ddietterich: Yes, exactly. ITFOT it will be able to have an azure cloud context also. That is what I am putting into the Azure PoC.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -222,145 +230,131 @@ private boolean updateWorkspaceColumns(UUID workspaceId, MapSqlParameterSource c
...
+  @WriteTransaction
+  public void deleteGcpCloudContext(UUID workspaceId) {
+    deleteCloudContextWorker(workspaceId, CloudPlatform.GCP, null);
+  }
+
+  /**
+   * Delete a cloud context for the workspace validating the flight id
+   *
+   * @param workspaceId workspace of the cloud context
+   * @param flightId flight id making the delete request
```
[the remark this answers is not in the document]
ddietterich: The purpose of the `creating_flight` field is to support the idempotency check in the create context flight.
The undo needs to know "did I create this context, in which case delete it; or did some other flight create it, in which case do not delete it."

So we want to compare the flight calling the delete against the original creating flight.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
ddietterich: We are consistently logging all DAO operations as info. I don't think it is too noisy. Do you?

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
ddietterich: Right

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
ddietterich: Worth thinking further about

zloery (approved):

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
ddietterich: It will have to have a `long version`. Or are you asking about the naming of the static?
I just used what was there.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -128,9 +138,7 @@ public Workspace getWorkspace(UUID id) {
     if (id == null) {
       throw new MissingRequiredFieldException("Valid workspace id is required");
     }
-    String sql =
-        WORKSPACE_SELECT_SQL
-            + " WHERE W.workspace_id = :id AND (C.cloud_platform = 'GCP' OR C.cloud_platform IS NULL)";
+    String sql = WORKSPACE_SELECT_SQL + " WHERE W.workspace_id = :id";
```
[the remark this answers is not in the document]
ddietterich: I don't understand. I think that is what I have.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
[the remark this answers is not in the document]
jaycarlton: I don't believe so. I don't see a pattern in our usage today. It's about a 60/40 split. My philosophy is that anything that could be static might as well be, since it makes refactoring (like moving constants and methods) easier.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -38,18 +36,30 @@
  */
 @Component
 public class WorkspaceDao {
-  /** Serializing format of the GCP cloud context */
-  @VisibleForTesting static final long GCP_CLOUD_CONTEXT_DB_VERSION = 1;
+  private final Logger logger = LoggerFactory.getLogger(WorkspaceDao.class);
+  private final NamedParameterJdbcTemplate jdbcTemplate;
 
   /** SQL query for reading a workspace, including its cloud context. */
   private static final String WORKSPACE_SELECT_SQL =
```
wchamber, line 43: update comment since we're no longer fetching the cloud context

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
jaycarlton: No, that's fine. It's rare for us to call lots of these methods in rapid succession.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -222,145 +230,131 @@ private boolean updateWorkspaceColumns(UUID workspaceId, MapSqlParameterSource c
...
-   * Create the cloud context of the workspace
+   * Create the GCP cloud context of the workspace
    *
    * @param workspaceId unique id of the workspace
    * @param cloudContext the GCP cloud context to create
    */
   @WriteTransaction
-  public void createGcpCloudContext(UUID workspaceId, GcpCloudContext cloudContext) {
+  public void createGcpCloudContext(
+      UUID workspaceId, GcpCloudContext cloudContext, String flightId) {
```
wchamber, line 244: probably worth a comment on what the flightId is for here.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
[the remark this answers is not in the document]
wchamber: Mind including that context in the comment? Also consider if there's a more meaningful name, originatingFlightId, owningFlightId, etc.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -222,145 +230,131 @@ private boolean updateWorkspaceColumns(UUID workspaceId, MapSqlParameterSource c
...
+    createCloudContextWorker(workspaceId, CloudPlatform.GCP, cloudContext.serialize(), flightId);
+  }
+
+  /**
+   * Delete the GCP cloud context for a workspace
+   *
+   * @param workspaceId workspace of the cloud context
+   */
+  @WriteTransaction
+  public void deleteGcpCloudContext(UUID workspaceId) {
```
wchamber, line 254: worth a comment comparing this to the other deleteGcpCloudContext method to explain when this is appropriate to use

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -118,6 +120,7 @@ public void resetFlightDebugInfo() {
   public void createAiNotebookInstanceDo() throws Exception {
     UserAccessUtils.TestUser user = userAccessUtils.defaultUser();
     Workspace workspace = reusableWorkspace(user);
+    UUID workspaceId = workspace.getWorkspaceId();
```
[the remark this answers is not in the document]
ddietterich: It is, and now it is used in even more places. Turns out I didn't need the `workspace` variable :)

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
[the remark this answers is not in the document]
ddietterich: Sounds reasonable. `static` it is!

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -222,145 +230,131 @@ private boolean updateWorkspaceColumns(UUID workspaceId, MapSqlParameterSource c
...
+            + "WHERE workspace_id = :workspace_id AND cloud_platform = :cloud_platform";
+    MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue("workspace_id", workspaceId.toString())
+            .addValue("cloud_platform", cloudPlatform.toSql());
+    try {
+      return Optional.ofNullable(
+          DataAccessUtils.singleResult(
+              jdbcTemplate.query(sql, params, (rs, rowNum) -> rs.getString("context"))));
+    } catch (EmptyResultDataAccessException e) {
```
wchamber, line 320: I think `singleResult` returns `null` instead of throwing `EmptyResultDataAccessException`. https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/dao/support/DataAccessUtils.html#singleResult-java.util.Collection-

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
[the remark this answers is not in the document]
zloery: oh, I meant using `workspace_id` instead of `W.workspace_id`, since there's no longer a `C.workspace_id`. Actually, you really don't need the `W` at all now, though it's just a style change

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -312,33 +312,63 @@ public void deleteGcpCloudContext(UUID workspaceId, AuthenticatedUserRequest use
         .submitAndWait(null);
   }
 
+  /**
+   * Retrieve the optional GCP cloud context NOTE: no user auth validation
```
wchamber, line 316: Seems risky to leave off the user validation. Can we at least make this more obvious in the name that this is by passing auth since other WorkspaceService methods do do auth?
I guess ditto all these other methods, or let's consider whether we should not be mixing auth checking with context retrieval in this class.

wchamber (approved):

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
ddietterich: Yes, you are right. Don't know why the try-catch was there. I'll remove it.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
ddietterich: OK, that is two votes for moving the GCP utility stuff elsewhere. I will pull it from here and the DAO and put it in its own service.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -312,60 +314,35 @@ public void deleteGcpCloudContext(UUID workspaceId, AuthenticatedUserRequest use
...
 
   /**
-   * Get the GcpProject with authentication for use in controllers
+   * Get the GcpProject or throw
    *
    * @param workspaceId unique workspace id
    * @param userRequest authenticated user
    * @return GCP project id
    */
   public String getRequiredGcpProject(UUID workspaceId, AuthenticatedUserRequest userRequest) {
```
jaycarlton, line 343: I'd implement this in terms of the other one: `return getGcpProject(workspaceId, userRequest).orElseThrow(() -> ...)`

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -312,60 +314,35 @@ public void deleteGcpCloudContext(UUID workspaceId, AuthenticatedUserRequest use
...
 
   /**
-   * Helper method used by other classes that require the GCP project to exist in the workspace. It
-   * throws if the project (GCP cloud context) is not set up. NOTE: no user auth validation
+   * Get an option project id from an optional GCP cloud context
    *
    * @param workspaceId unique workspace id
-   * @return GCP project id
+   * @param userRequest authenticated user
+   * @return optional project id
```
jaycarlton, line 329: Does this return `Optional.empty()` if the workspace is invalid or does it also throw? It looks like it can throw, which is unusual for `Optional`-valued methods. I guess document that in the comment (or make it a checked exception).

Same observation for `getGcpCloudContext()` which is Optional, but throws an unchecked exception. No rule against that anywhere, but I was kinda surprised.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -312,60 +314,35 @@ public void deleteGcpCloudContext(UUID workspaceId, AuthenticatedUserRequest use
...
-        .orElseThrow(
-            () -> new CloudContextRequiredException("Operation requires GCP cloud context"));
+  public Optional<String> getGcpProject(UUID workspaceId, AuthenticatedUserRequest userRequest) {
+    validateWorkspaceAndAction(userRequest, workspaceId, SamConstants.SAM_WORKSPACE_READ_ACTION);
+    return gcpCloudContextService.getGcpProject(workspaceId);
   }
 
   /**
-   * Get the GcpProject with authentication for use in controllers
+   * Get the GcpProject or throw
```
jaycarlton, line 337: indicate which exceptions are thrown for which situations

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteProjectStep.java
```diff
@@ -26,12 +26,12 @@
  */
 public class DeleteProjectStep implements Step {
```
jaycarlton, line 27: nit: maybe rename this step to `DeleteGcpProjectStep`. We may need to come up with some step naming rules.

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -212,7 +210,7 @@ private boolean updateWorkspaceColumns(UUID workspaceId, MapSqlParameterSource c
   public List<Workspace> getWorkspacesMatchingList(List<UUID> idList, int offset, int limit) {
     String sql =
```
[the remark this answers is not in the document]
ddietterich: Good catch! That is not changed code, but I'm happy to fix it.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -357,4 +316,26 @@ private void deleteCloudContextWorker(
...
+    String sql =
+        "SELECT context FROM cloud_context "
+            + "WHERE workspace_id = :workspace_id AND cloud_platform = :cloud_platform";
+    MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue("workspace_id", workspaceId.toString())
+            .addValue("cloud_platform", cloudPlatform.toSql());
+
+    return Optional.ofNullable(
+        DataAccessUtils.singleResult(
```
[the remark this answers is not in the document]
ddietterich: DataUtils is Spring code.

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.UUID;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * This service provides methods for managing GCP cloud context. These methods do not perform any
+ * access control and operate directly against the {@link bio.terra.workspace.db.WorkspaceDao}
+ */
+@Component
+public class GcpCloudContextService {
```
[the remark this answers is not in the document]
ddietterich: We want to be able to support both GCP and Azure in the same workspace, so we want both sets of code. No need for Spring injection.

I do not think there is a shared interface that can be. The methods are similar, but have different inputs and outputs.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
ddietterich: That would put the error logic in two places: in GcpCloudContextService and here. I think it is better as is.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
ddietterich: The call to validateWorkspaceAndAction will ensure that the workspace exists and that the caller has read access.
I'll add some documentation.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -215,152 +221,121 @@ private boolean updateWorkspaceColumns(UUID workspaceId, MapSqlParameterSource c
...
           e);
     }
   }
 
   /**
-   * Delete the GCP cloud context for a workspace
+   * Delete a cloud context for the workspace validating the flight id. This is for use by the cloud
+   * context create flight. We compare the incoming flight id with the creating_flight of the cloud
+   * context. We will only delete if they match. That makes sure that undoing a create does not
+   * delete a valid cloud context.
```
wchamber, line 260: good comments, thanks!

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Create the GCP cloud context of the workspace
+   *
+   * @param workspaceId unique id of the workspace
+   * @param cloudContext the GCP cloud context to create
+   */
+  public void createGcpCloudContext(
+      UUID workspaceId, GcpCloudContext cloudContext, String flightId) {
+    workspaceDao.createCloudContext(
+        workspaceId, CloudPlatform.GCP, cloudContext.serialize(), flightId);
```
wchamber, line 36: This class isn't doing a lot for me, but I don't have ideas I like much more, so I'm fine with it. I'm still going to write my comment in case it helps you think of a better idea or you like something here more, but no action requested and up to you :)

It seems like this is a very thin but incomplete wrapper around WorkspaceDao. Given that they're both available from AutoWired, if I'm a naive developer, I'm not sure what's going to direct me to this class. It's also incomplete, in that WorkspaceService still has a number of the methods needed to work with gcp cloud contexts.
I think this class exists to try to distinguish cases where we don't want to do another auth check. Couple options
- Move all cloud context methods either here or on WorkspaceService (so they're in one place), but add something about auth to the method names that might make a caller think about if it's the right thing. You might be getting conflicting reviewer comments here :)
- Reify that an auth check occurred, i.e. make an "authd workspace" object that a request receives after it's done the appropriate auth check with Sam containing the workspaceId of the authd workspace. Pass that down the stack and to flights, and then in "unauthed" methods accept that object. In the unauthed method, check that the "authd workspace" object's workspaceId matches the workspaceId of the method. This might be unwieldly, but it uses types and the stack to help keep track that auth checks have occurred.

service/src/test/java/bio/terra/workspace/db/WorkspaceDaoTest.java
```diff
@@ -216,41 +224,50 @@ void setUp() {
...
+
+      // Make sure service and dao get the same answer
+      cloudContext = gcpCloudContextService.getGcpCloudContext(workspaceId);
+      assertTrue(cloudContext.isPresent());
+      assertEquals(projectId, cloudContext.get().getGcpProjectId());
+
+      // delete with mismatched flight id - it should not delete
+      gcpCloudContextService.deleteGcpCloudContextWithCheck(workspaceId, "mismatched-flight-id");
+
+      // Make sure the contxt is still there
```
wchamber, line 244: `contxt` typo

wchamber (approved):
Update the PR description to account for the refactor?

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
ddietterich: I really like the idea of a specific object that represents an authd workspace and an authd resource! I'm going to put this into a separate ticket so I can get back to the PoC.

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Optional;
+import java.util.UUID;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * This service provides methods for managing GCP cloud context. These methods do not perform any
+ * access control and operate directly against the {@link bio.terra.workspace.db.WorkspaceDao}
+ */
+@Component
```
zloery, line 17: I'm a little concerned about this class causing more confusion in the future. From an outside caller perspective, it's not clear to me that `gcpCloudContextService.createGcpCloudContext` is a thin wrapper around the DAO with no auth, while `workspaceService.createGcpCloudContext` checks auth and launches a flight (and same for `deleteGcpCloudContext` in both places). This might just be an issue with the conflicting names though, and it might be clearer with a class name like `GcpCloudContextDbService` or something similar (and maybe method names like `persistGcpCloudContext`? not sure).

As (yet another) alternative, this class feels like it's pretty close to being `GcpCloudContextDao`, with the GCP cloud context methods currently in `WorkspaceDao` implemented directly here instead. That also tracks with the methods here not requiring authentication.

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
ddietterich: Made [PF-1007](https://broadworkbench.atlassian.net/browse/PF-1007)

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
ddietterich: Talked offline with Zach and we agreed as a way forward:
1. Rename the WorkspaceService endpoints getAuthorizedGcpContext, etc to make it the functional difference clear.
2. Merge this PR
3. Implement PF-1007 based on Will’s suggestion
4. Move the GcpContext-related endpoints from WorkspaceService to GcpCloudContextService.

PR #454, opened 2021-09-24
[PF-1011] Remove Hermione from WSM integration tests

zloery:
`hermione.owner` is a shared test user that's also used by Rawls and possibly other services' tests. For the sake of test isolation, we should use test users exclusive to WSM for the integration tests here.

Outside of this PR I've given `bella.redwalker` read access to the appropriate TDR snapshots in dev, alpha, and staging.


ddietterich (approved):
LGTM - thanks for doing this cleanup!!

ungwudik (approved):
LGTM.

PR #457, opened 2021-09-29
[IA-2861] Add azure IP controlled resource

jdcanas:
This PR adds code necessary for an Azure IP controlled resource CREATE/GET/DELETE

This was my first large contribution to WSM, and in most areas (where applicable) I have attempted to model existing controlled resource behavior/patterns. It is quite possible I missed something though, any and all feedback appreciated 😄 .    

Some notes:
- this depends on a locally published CRL module and useMavenLocal (hopefully this will be resolved before merge)
- Any SAM/Azure ACL consideration. I left an explicit todo in a function that trivial returns true for every auth question. 
- The proper WSM classification of these resources as part of an application (There is a ticket tracking this)
-  There is an integration [test](https://github.com/DataBiosphere/terra-workspace-manager/compare/azure-poc...jc-azure-crl-ip2?expand=1#diff-83025ce1db29a027a93694265cf021af86d77bc4b3b0f59efb705cbe720ef857R177) written, but I wasn't sure if the POC branch has the necessary setup yet. I saw that you added some Azure users @ddietterich. It seems that integration tests rely on an `ApiClient`, which I wasn't sure how to instantiate with those Azure users/creds in mind. 
- I included a unit test for the `CreateAzureIpStep` itself
    - this is not currently working in this branch as a recent rebase broke it. The branch `jc-azure-crl-ip2` is based off a branch where the added unit test works.
    - Are there any other unit tests worth adding? It was pretty hard to find anything to unit test that wasn't covered by existing controlled resource tests
- Note that the get/delete steps we seem to get for free as it is a controlled resource
- I included a few models needed for later iterations. They aren't used here but will be needed (in a ticket I'm working on after this), so I just left them.


service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/CreateAzureIpStepTest.java
```diff
@@ -0,0 +1,96 @@
...
+  @Mock private PublicIpAddress.DefinitionStages.WithGroup mockIpStage2;
+  @Mock private PublicIpAddress.DefinitionStages.WithCreate mockIpStage3;
+  private static final String STUB_STRING_RETURN = "stubbed-return";
+  @Mock private Region mockRegion;
+
+  @Autowired private AzureState azureState;
+
+  @BeforeEach
+  public void setup() {
+    when(mockAzureCloudContext.getAzureResourceGroupId()).thenReturn(STUB_STRING_RETURN);
```
jdcanas, line 59: this was fun.... mockito theoretically supports chain-stubbing, but it didn't seem like it cleaned up the syntax too much on cursory investigation.

service/build.gradle
```diff
@@ -102,13 +104,16 @@ dependencies {
...
+    implementation group: 'com.azure.resourcemanager', name: 'azure-resourcemanager-compute', version: '2.7.0'
+    implementation group: 'com.azure.resourcemanager', name: 'azure-resourcemanager-resources', version: '2.7.0'
 
     // Test deps
     testImplementation group: "com.google.auth", name: "google-auth-library-oauth2-http", version: "0.22.1"
     testImplementation group: "io.vavr", name: "vavr", version: "0.10.3"
     testImplementation('org.springframework.boot:spring-boot-starter-test') {
         exclude group: 'com.vaadin.external.google', module: 'android-json'
     }
+    testImplementation 'org.mockito:mockito-inline:2.13.0'
```
jdcanas, line 116: This allows us to mock final objects!!

service/src/main/java/bio/terra/workspace/app/configuration/external/AzureConfiguration.java
```diff
@@ -1,9 +1,13 @@
 package bio.terra.workspace.app.configuration.external;
 
+import com.azure.core.management.Region;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.stereotype.Component;
 
+/** Azure POC code */
+@Component
```
jdcanas, line 10: not sure if this is needed? Little fuzzy on the annotation, but the documentation was saying I need this to use it in conjunction with `@Autowired` for injection into other classes

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -44,6 +44,10 @@ public static void validateBucketName(String name) {
     }
   }
 
+  public static void validateIpName(String name) {
+    // TODO
```
jdcanas, line 48: Had trouble finding this originally https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules

I'll implement this shortly...


1-80 | Alphanumerics, underscores, periods, and hyphens.Start with alphanumeric. End alphanumeric or underscore.
-- | --




1-80 | Alphanumerics, underscores, periods, and hyphens.Start with alphanumeric. End alphanumeric or underscore.
-- | --

service/src/main/resources/application-azure.yml
```diff
@@ -2,7 +2,10 @@
 # in different environments
 workspace:
   azure:
-    tenantId: TBD
+    userTenantId: 0cb7a640-45a2-4ed6-be9f-63519f86e04b
```
ddietterich, line 5: The user part of this should not be here. That is what lives in the AzureCloudContext.
I have similar configuration defined for use in testing in `application-azure-test.yml`.

As mentioned above, I have a parallel, but different implementation for the managed app coordinates.

service/src/test/java/bio/terra/workspace/app/configuration/external/AzureStateEnabledTest.java
```diff
@@ -13,6 +14,7 @@
   @Autowired private AzureState azureState;
 
   @Test
+  @Tag("azure")
```
ddietterich, line 17: I intended this to be a unit test, not an azure test, to make sure that the explicit profile cases got the right answer.
Perhaps remove the `@Tag` and add a comment that even though this has azure written all over it, it is supposed to be a unit test.

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/CreateAzureIpStepTest.java
```diff
@@ -0,0 +1,96 @@
...
+import com.azure.resourcemanager.network.models.PublicIpAddresses;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.context.ActiveProfiles;
+
+@ActiveProfiles("azure")
+public class CreateAzureIpStepTest extends BaseUnitTest {
```
ddietterich, line 37: Derive the class from `BaseAzureTest` and you don't need to worry the profiles or use the `@Tag`. Also, you pick up the azure-test configuration that supplies the user tenant coordinates.

ddietterich (commented):
Overall, this is a very sound piece of work. Thank you!
Lots of comments and questions, of course.

service/src/main/java/bio/terra/workspace/app/configuration/external/AzureConfiguration.java
```diff
@@ -12,9 +16,23 @@
   private String managedAppClientId;
   private String managedAppClientSecret;
   private String managedAppTenantId;
+  /** tenant where WSM managed app is deployed */
```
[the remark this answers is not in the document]
jdcanas: Yeah, cleaning up now in conjunction with the AzureCloudContext you've added

service/src/main/java/bio/terra/workspace/app/configuration/external/AzureConfiguration.java
```diff
@@ -12,9 +16,23 @@
...
 
-  public String getManagedAppClientId() {
-    return managedAppClientId;
+  /** clientId for access to the MRG */
+  private String clientId;
+
+  /** clientSecret for access to the MRG */
+  private String clientSecret;
+
+  private Region defaultRegion;
```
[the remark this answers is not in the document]
jdcanas: For most of our existing resources (in leo) it's optional to provide a region, and we have a default. For a long time we didn't support arbitrary regions, as there can be conflicts if resources within a workspace are created in different regions (i.e. a user's app trying to access storage in a different region could be problematic).

I'm of the opinion that, for the POC, we can omit this parameter and have an intelligent default, since it's likely we will want all resources in the same region to avoid any unnecessary scope. I'm not attached to the idea though if you see it it differently.

service/src/main/java/bio/terra/workspace/app/controller/ResourceController.java
```diff
@@ -142,6 +135,13 @@ public ApiResourceDescription makeApiResourceDescription(
               break;
             }
 
+          case AZURE_IP:
+            {
+              ReferencedAzureIpResource resource = referencedResource.castToAzureIpResource();
```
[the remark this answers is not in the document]
jdcanas: No, this was me unsure of what exactly is needed for a controlled resource early on. I thought the 'referenced resource' was models/nomenclature regarding the DB storage of the controlled resource.  I'll remove this.

service/src/main/resources/application-azure.yml
jdcanas: Yeah, just cleanup left to do after the rebase with your `AzureCloudContext`. Great to have this to separate the 'user' supplied config vs the config WSM is responsible for.

service/src/main/java/bio/terra/workspace/app/configuration/external/AzureConfiguration.java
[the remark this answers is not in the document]
ddietterich: I'd suggest taking region as a parameter in your interface and setting the default outside of WSM.

service/build.gradle
```diff
@@ -102,13 +104,17 @@ dependencies {
     implementation group: 'com.azure', name: 'azure-identity', version: '1.3.5'
     implementation group: 'com.azure', name: 'azure-core', version: '1.19.0'
     implementation group: 'com.azure.resourcemanager', name: 'azure-resourcemanager', version: '2.8.0'
+    implementation group: 'com.azure.resourcemanager', name: 'azure-resourcemanager-compute', version: '2.7.0'
```
rtitle, line 107: Does this need to be here? I think CRL should pull in azure dependencies (can probably remove this whole block).

jdcanas: There's some initial code in WSM that use on this, written prior to CRL usage in WSM. some of which Dan's tests depend on. I can remove my additions at this point though.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureIpStep.java
```diff
@@ -0,0 +1,95 @@
...
+      logger.info("Ip {} already exists. Continuing.", resource.getIpName());
+    }
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    ComputeManagerCow computeCow = crlService.getComputeManagerCow();
+    computeCow
```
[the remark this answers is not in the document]
ddietterich: When the flight starts, there are two cases:
1. The IP address needs to be created in the workspace.
2. The IP address already exists in the workspace. This should be an error, but you treat it as a re-run of the step. And on an error in a later step, your undo would delete the already existing IP address.

I suggest that you have another step that you run ahead of CreateAzureIpStep. It would detect and error out if the IP address already existed. That way this step knows that if the IP exists, it is because this step created it. The undo can safely delete it.

There is a race condition if there are two concurrent creators of the identical thing. That is pretty much true of all of the current WSM flights. We have not implemented a solution for that problem. If it becomes a real-world problem, we can fix it by serializing in WSM.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2009,6 +2095,28 @@ components:
...
+        name:
+          description: A valid ip name per https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
+          type: string
+
+    CreatedControlledAzureIp:
+      description: Response payload for requesting a new azure IP
+      type: object
+      properties:
+        resourceId:
+          description: UUID of a newly-created resource. Null if not created yet.
```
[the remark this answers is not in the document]
jdcanas: Ahh, something left over from copy-paste

service/src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -379,4 +385,32 @@ public AzureResourceManager getResourceManager(AzureCloudContext azureCloudConte
...
+  private ComputeManagerCow buildComputeManagerCow(
+      AzureCloudContext azureCloudContext, AzureConfiguration azureConfig) {
+    TokenCredential azureCreds = getManagedAppCredentials(azureConfig);
+
+    AzureProfile azureProfile =
+        new AzureProfile(
+            azureCloudContext.getAzureTenantId(),
+            azureCloudContext.getAzureSubscriptionId(),
+            AzureEnvironment.AZURE);
+    return ComputeManagerCow.create(clientConfig, azureCreds, azureProfile);
```
rtitle, line 398: This might be slightly different in the latest CRL code, gonna pull this down and try it too..

service/src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -343,26 +353,22 @@ private ClientConfig buildClientConfig() {
...
-
-  private void assertCrlInUse() {
-    if (!crlConfig.getUseCrl()) {
-      throw new CrlNotInUseException("Attempt to use CRL when it is set not to be used");
-    }
-  }
-
-  // -- Azure Support --
+  //  Azure Support
   // TODO: this is not in CRL, but it will be, so this is a convenient place to dump the
```
rtitle, line 357: Not your comment, but CRL doesn't actually wrap the `TokenCredential` class, so clients still need to instantiate one with the appropriate clientId/secret/tenant.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureIpStep.java
```diff
@@ -0,0 +1,117 @@
...
+    // this is a redo and this step created it already.
+    PublicIpAddress existingIP;
+
+    try {
+      existingIP =
+          computeCow
+              .computeManager()
+              .networkManager()
+              .publicIpAddresses()
+              .getByResourceGroup(
```
rtitle, line 64: Instead of doing a get+create would it make sense to just `create` and recover from the conflict exception? That way we do 1 API call instead of 2.

jdcanas: The azure API does not seem to return nice exceptions like google... the status code is embedded in a generic exception message body (see recovery block below). This is the pattern other components in WSM do (see create bucket), so I think its good for consistency. Not attached to this way though, if we can find an easily-parsable status code within the `com.azure.core.management.exception.ManagementException`

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureIpStep.java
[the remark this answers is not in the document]
jdcanas: I implemented it so we intelligently don't delete it if it already existed in undo, but I did it without another step for now, with the concurrency issue of the working map that you note.

integration/src/main/java/scripts/testscripts/CreateAzureIp.java
```diff
@@ -0,0 +1,29 @@
+package scripts.testscripts;
```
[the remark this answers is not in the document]
jdcanas: I wrote a new 'unit' test that uses the real underlying services, and tests successful IP creation in azure. It is disabled by default because the credentials are not pushed, but it should be nice to be able to verify the code does indeed successfully create cloud resources as we iterate. 

Unfortunately, I did this at the level of a step/flight, and was unable to get it working through the `controlledResourceService` itself in tests due to underlying SAM complaining, at which point I punted. I left the second test there for reference with the above summarized. 

https://github.com/DataBiosphere/terra-workspace-manager/pull/457/files#diff-78d55c5069aedd76020b4b2a3638c747492d66415f806adff8a1e49aea9c3a6aR74

service/src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -379,4 +382,39 @@ public AzureResourceManager getResourceManager(AzureCloudContext azureCloudConte
...
+      AzureCloudContext azureCloudContext, AzureConfiguration azureConfig) {
+    TokenCredential azureCreds = getManagedAppCredentials(azureConfig);
+
+    AzureProfile azureProfile =
+        new AzureProfile(
+            azureCloudContext.getAzureTenantId(),
+            azureCloudContext.getAzureSubscriptionId(),
+            AzureEnvironment.AZURE);
+
+    // We must use FQDN because there are two `Defaults` symbols imported otherwise.
```
rtitle, line 396: Kinda unfortunate, maybe we can name the Azure class differently..

ddietterich: Sounds like a good idea to me.

service/src/main/resources/application-azure.yml
```diff
@@ -2,7 +2,8 @@
 # in different environments
 workspace:
   azure:
-    tenantId: TBD
     managedResourceGroupId: TBD
-    clientId: TBD
+    homeTenantId: fad90753-2022-4456-9b0a-c7e5b934e408
```
ddietterich, line 6: These names don't line up with the AzureConfiguration names.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureIpStep.java
```diff
@@ -0,0 +1,118 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    FlightMap inputMap = context.getInputParameters();
+    ApiAzureIpCreationParameters creationParameters =
+        inputMap.get(CREATION_PARAMETERS, ApiAzureIpCreationParameters.class);
+
+    ComputeManager computeManager = crlService.getComputeManager(azureCloudContext, azureConfig);
+
+    // Don't try to create it if it already exists. At this point the assumption is
```
ddietterich, line 55: I think that is a faulty assumption. You cannot tell the difference between a previously existing IP - where you should not be creating it - and a rerun of a valid create attempt.

The consequence is that the undo will not behave properly. It will leak IPs if the step is rerun and then a later step errors.

This really needs to be split into two steps. One step to determine whether the IP exists and throws if it does.
Then this step always knows for sure that if the IP exists, it is a rerun.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureIpStep.java
```diff
@@ -0,0 +1,71 @@
...
+    this.azureCloudContext = azureCloudContext;
+    this.crlService = crlService;
+    this.resource = resource;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    ComputeManager computeManager = crlService.getComputeManager(azureCloudContext, azureConfig);
+
+    computeManager
```
ddietterich, line 40: I think you still want the code here to tolerate the IP existing. On a failure after the create and before step completion, we might restart at this step and re-issue the create.

service/src/test/java/bio/terra/workspace/common/fixtures/ControlledResourceFixtures.java
```diff
@@ -1,22 +1,7 @@
...
-import bio.terra.workspace.generated.model.ApiGcpGcsBucketLifecycleRuleAction;
-import bio.terra.workspace.generated.model.ApiGcpGcsBucketLifecycleRuleActionType;
-import bio.terra.workspace.generated.model.ApiGcpGcsBucketLifecycleRuleCondition;
-import bio.terra.workspace.generated.model.ApiGcpGcsBucketUpdateParameters;
-import bio.terra.workspace.service.resource.controlled.AccessScopeType;
-import bio.terra.workspace.service.resource.controlled.ControlledAiNotebookInstanceResource;
-import bio.terra.workspace.service.resource.controlled.ControlledBigQueryDatasetResource;
-import bio.terra.workspace.service.resource.controlled.ControlledGcsBucketResource;
-import bio.terra.workspace.service.resource.controlled.ManagedByType;
+import bio.terra.workspace.generated.model.*;
```
ddietterich, line 3: We have a no `*` policy in the WSM code.
You can set IntelliJ up so it won't generate `*` imports 
In Preferences: Editor --> Code Style --> Java and set both of the "X count to use import with '*'" to 9999

ddietterich (commented):
This is looking very good. One significant comment: restoring the IP exists handling in the CreateIP step, since it can happen due to pod failure/restart.

Minor comment about your IntelliJ using `*` imports.

service/src/test/java/bio/terra/workspace/common/fixtures/ControlledResourceFixtures.java
rtitle: Oops, fixed! Thanks

PR #459, opened 2021-10-01
[PF-990] Restructure Gradle to use convention plugins

zloery:
This restructures our Gradle build structure to use [convention plugins](https://docs.gradle.org/current/samples/sample_convention_plugins.html), which allow us to pull common gradle code into separate files and apply them as plugins. This is heavily inspired by https://github.com/DataBiosphere/stairway/pull/82 and https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/95, which applied the same pattern in Stairway and CRL.

This change allows us to run Spotless in the integration directory, which previously didn't work. I'll run the command in a followup PR to reduce review noise.


ddietterich (approved):
Thank you for picking this up!

jmczerk (approved):
Looks good, thanks for doing this!

PR #460, opened 2021-10-04
[PF-990] Apply spotless to integration module

zloery:
This is the result of running `./gradlew spotlessApply` at the top level with no other changes. This is now available after the gradle refactor in #459.


ddietterich (approved):
A cursory inspection of the diffs leads me to conclude that Spotless did its thing.

jmczerk (approved):

PR #463, opened 2021-10-08
[PF-1035] Make controlled notebooks run as pet SAs

zloery:
This change makes WSM run controlled AI Platform notebooks as pet service accounts, instead of minting new SAs for each notebook. At a high level, this involves the following changes:

- The notebook creation flight no longer creates and sets up IAM policies for service accounts. Instead, it grants users permission to use their pet SA when they create a notebook.
- Remove users' permission to use their pet SA when they are removed from a workspace. This ensures they lose notebook access after leaving the workspace.

The pet service account functionality was getting a little large to be lumped in with WorkspaceService and was conceptually distinct, so I pulled it into a new service. This may change in the future as we revisit how we use pet SAs.


service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -365,14 +371,16 @@ private ApiDataReferenceDescription makeApiDataReferenceDescription(
       @PathVariable("role") ApiIamRole role,
       @RequestBody ApiGrantRoleRequestBody body) {
     ControllerValidationUtils.validateEmail(body.getMemberEmail());
+    // Sam and GCP always use lowercase email identifiers, so we do the same here for consistency.
+    String memberEmail = body.getMemberEmail().toLowerCase();
```
ddietterich, line 375: I'd suggest moving the email downcase into the SamService rather than have it in the controller.

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -0,0 +1,258 @@
...
+    // enablePetServiceAccountImpersonationWithEtag will only return an empty optional if the
+    // provided eTag does not match current policy. Because we do not use eTag checking here, this
+    // is always nonempty.
+    return enablePetServiceAccountImpersonationWithEtag(workspaceId, userAndPet, null).get();
+  }
+
+  /**
+   * Grant a user permission to impersonate their pet service account in a given workspace. Unlike
+   * other operations, this does not run as a flight because it only requires one write operation.
+   * This operation is idempotent.
```
ddietterich, line 66: Sounds fine. Do we need retrying for this? Or just let the client retry on error?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/GrantPetUsagePermissionStep.java
```diff
@@ -0,0 +1,63 @@
...
+    FlightMap workingMap = context.getWorkingMap();
+    workingMap.put(PetSaKeys.MODIFIED_PET_SA_POLICY_ETAG, modifiedPolicy.getEtag());
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    FlightMap workingMap = context.getWorkingMap();
+    String expectedEtag = workingMap.get(PetSaKeys.MODIFIED_PET_SA_POLICY_ETAG, String.class);
+    if (expectedEtag == null) {
```
ddietterich, line 51: Is there a way to close this hole?
Maybe to have a step ahead of this that reads the current etag and saves it in the map. Then query the current etag here to see if it changed or not?

It doesn't solve concurrent updating, but we have not worried concurrency at all so far.

service/src/main/java/bio/terra/workspace/service/workspace/flight/RevokePetUsagePermissionStep.java
```diff
@@ -0,0 +1,124 @@
...
+    Optional<String> validatedPetEmail = getAndValidatePet(workingMap);
+    // No need to revoke pet access if the user does not have a pet in this workspace context, or
+    // is not fully removed from the workspace.
+    if (validatedPetEmail.isEmpty()) {
+      return StepResult.getStepResultSuccess();
+    }
+    UserWithPetSa userAndPet = new UserWithPetSa(userEmailToRemove, validatedPetEmail.get());
+    String expectedEtag = workingMap.get(PetSaKeys.MODIFIED_PET_SA_POLICY_ETAG, String.class);
+    if (expectedEtag == null) {
+      // If the do step did not finish, we cannot guarantee we aren't undoing something we didn't do
```
ddietterich, line 80: Yeah, same problem here. Hmmm...

service/src/main/java/bio/terra/workspace/common/exception/GcpException.java
```diff
@@ -0,0 +1,16 @@
+package bio.terra.workspace.common.exception;
+
+import bio.terra.common.exception.ErrorReportException;
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
+import org.springframework.http.HttpStatus;
+
+/**
+ * An {@code ErrorReportException} wrapper around errors thrown by GCP. This is used to surface GCP
+ * errors directly to users.
```
ddietterich, line 9: What cases should we surface GCP errors directly to users?

I would think a more specific WSM error that includes the GCP error as a cause would be a nicer error to receive and have more context.

zloery: Looking at this again I think you're right that this is the wrong approach. If WSM unexpectedly gets a 403 on an underlying cloud resource on a request that should be correct, the WSM response really should be 500, not 403. I think there are a few spots elsewhere that we should be checking and surfacing 400s in particular, but that's really a separate issue, and I can track that separately.

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -0,0 +1,254 @@
...
+    return binding1.getRole().equals(binding2.getRole())
+        && binding1.getMembers().containsAll(binding2.getMembers())
+        && binding2.getMembers().containsAll(binding1.getMembers());
+  }
+
+  /**
+   * Returns the pet service account of a provided user in a provided project if it exists in GCP,
+   * or an empty Optional if it does not.
+   */
+  public Optional<ServiceAccountName> getArbitrarySa(
```
marikomedlock, line 226: nit: SAM has an endpoint to "gets a key for an arbitrary pet service account for the user", where "arbitrary" means any one of the pet SAs for that user. I think you mean "arbitrary" here in the sense that you want to lookup a SA without having the user's token available. Perhaps use a different word here to avoid confusion?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/GrantPetUsagePermissionStep.java
```diff
@@ -0,0 +1,63 @@
...
+    this.petSaService = petSaService;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Policy modifiedPolicy =
+        petSaService.enablePetServiceAccountImpersonation(workspaceId, userAndPet);
+    // Store the eTag value of the modified policy in case this step needs to be undone.
+    FlightMap workingMap = context.getWorkingMap();
+    workingMap.put(PetSaKeys.MODIFIED_PET_SA_POLICY_ETAG, modifiedPolicy.getEtag());
```
marikomedlock, line 43: Does the etag change even if the policy hasn't changed? e.g. if a user previously enabled their pet via the dedicated endpoint, then this should produce no iam policy changes. Asking because I was wondering if we could rollback the permissions even if they were granted outside the flight.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/GrantPetUsagePermissionStep.java
zloery: updated javadoc for this and `RevokePetUsagePermissionStep`

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/notebook/GrantPetUsagePermissionStep.java
zloery: Any call to `setIamPolicy` will update the etag value even if it doesn't actually change the policy, so this won't undo permissions granted by the `enablePet` endpoint

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -884,15 +889,38 @@ private void addPrivateResourcePolicies(
...
+   * return the email of a service account which does not exist.
+   */
+  public ServiceAccountName constructUserPetSaEmail(
+      String projectId, String userEmail, AuthenticatedUserRequest userRequest)
+      throws InterruptedException {
+    UsersApi usersApi = samUsersApi(userRequest.getRequiredToken());
+    try {
+      String subjectId = usersApi.getUserIds(userEmail).getUserSubjectId();
+      String saEmail = String.format("pet-%s@%s.iam.gserviceaccount.com", subjectId, projectId);
+      return SamRetry.retry(
```
marikomedlock, line 917: nit: Do you need `SamRetry` here? I thought this wasn't actually calling SAM, just building a string.

PR #465, opened 2021-10-12
[PF-1024] Implement and test enable, disable, get, enumerate applications in workspaces

ddietterich:
This PR provides the REST API for managing applications in workspaces.
See [WSM Application Configure - Design](https://docs.google.com/document/d/1QQ_RoN4pIDGq4bAqhaBOvnStgCU3f23bZb5xysGyLK8/edit#) for details on the semantics.

At this point, the disable operation is not fully implemented as spec'd. It performs the DAO operation, but does not delete any resources. I plan to leave that undone for a bit until I have more direction on how disabling should be done.

**Note**: I noticed that the `gradle :service:build` target was not running unit tests. I added a configuration back so that it runs the unit tests. Seems to work, but if it was removed for some reason, let me know.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApplicationApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApplicationApiController.java
@@ -1,5 +1,6 @@
 package bio.terra.workspace.app.controller;
 
+import bio.terra.workspace.common.utils.ControllerValidationUtils;
 import bio.terra.workspace.generated.controller.WorkspaceApplicationApi;
 import bio.terra.workspace.generated.model.ApiWorkspaceApplicationDescription;
 import bio.terra.workspace.generated.model.ApiWorkspaceApplicationDescriptionList;
@@ -12,7 +13,6 @@ import java.util.List;
 import java.util.UUID;
 import javax.servlet.http.HttpServletRequest;
 import javax.validation.Valid;
-import javax.validation.constraints.Min;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -20,6 +20,7 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestParam;
 
 @Controller
 public class WorkspaceApplicationApiController implements WorkspaceApplicationApi {
@@ -79,9 +80,11 @@ public class WorkspaceApplicationApiController implements WorkspaceApplicationAp
   @Override
   public ResponseEntity<ApiWorkspaceApplicationDescriptionList> listWorkspaceApplications(
       @PathVariable("workspaceId") UUID workspaceId,
-      @Min(0) @Valid Integer offset,
-      @Min(1) @Valid Integer limit) {
+      @Valid @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
+      @Valid @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit) {
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    ControllerValidationUtils.validatePaginationParams(offset, limit);
+
     List<WsmWorkspaceApplication> wsmApps =
         appService.listWorkspaceApplications(userRequest, workspaceId, offset, limit);
     var response = new ApiWorkspaceApplicationDescriptionList();
--- a/service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
@@ -8,6 +8,7 @@ import bio.terra.workspace.db.exception.InvalidApplicationStateException;
 import bio.terra.workspace.service.workspace.model.WsmApplication;
 import bio.terra.workspace.service.workspace.model.WsmApplicationState;
 import bio.terra.workspace.service.workspace.model.WsmWorkspaceApplication;
+import com.google.common.annotations.VisibleForTesting;
 import java.util.List;
 import java.util.UUID;
 import org.slf4j.Logger;
@@ -118,7 +119,7 @@ public class ApplicationDao {
   public WsmWorkspaceApplication disableWorkspaceApplication(UUID workspaceId, UUID applicationId) {
 
     // Validate that the application exists; workspace is validated in layers above this
-    getApplication(applicationId);
+    getApplicationOrThrow(applicationId);
 
     // It is an error to have application resources in the workspace if we are disabling it.
     final String countAppUsesSql =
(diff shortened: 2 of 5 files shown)


service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
```diff
@@ -76,6 +107,176 @@ public void createApplication(WsmApplication app) {
...
+   */
+  @WriteTransaction
+  public WsmWorkspaceApplication disableWorkspaceApplication(UUID workspaceId, UUID applicationId) {
+
+    // Validate that the application exists
+    getApplication(applicationId);
+
+    final String sql =
+        "DELETE FROM enabled_application"
+            + " WHERE workspace_id = workspace_id AND application_id = :application_id";
```
zloery, line 126: missing`:` in front of `workspace_id` here! I think this clause is always true otherwise

service/src/main/java/bio/terra/workspace/service/workspace/WsmApplicationService.java
```diff
@@ -28,15 +34,79 @@
...
+  public void disableWorkspaceApplication(
+      AuthenticatedUserRequest userRequest,
+      UUID workspaceId,
+      UUID applicationId,
+      String resultPath,
+      String jobId) {
+
+    workspaceService.validateWorkspaceAndAction(
+        userRequest, workspaceId, SamConstants.SAM_WORKSPACE_OWN_ACTION);
+
```
zloery, line 71: Should this (and the other application methods) also have a call to `StageService.assertMcWorkspace` for now?

service/src/main/java/bio/terra/workspace/service/workspace/flight/disable/application/DisableApplicationKeys.java
```diff
@@ -0,0 +1,6 @@
+package bio.terra.workspace.service.workspace.flight.disable.application;
+
+public class DisableApplicationKeys {
```
zloery, line 3: style note: most other flight keys are organized as static nested classes inside `WorkspaceFlightMapKeys` instead of standalone files. We should be consistent with that, though I don't have a strong opinion on which approach we take.

service/src/main/java/bio/terra/workspace/service/workspace/flight/disable/application/DisableApplicationDaoStep.java
```diff
@@ -0,0 +1,49 @@
...
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    // We want to enable the application even if the application is deprecated, so
+    // use the NoCheck variant of the DAO.
+    applicationDao.enableWorkspaceApplicationNoCheck(workspaceId, applicationId);
+
+    // Set up the flight result
+    WsmWorkspaceApplication wsmApp =
+        applicationDao.getWorkspaceApplication(workspaceId, applicationId);
+    flightContext.getWorkingMap().put(JobMapKeys.RESPONSE.getKeyName(), wsmApp);
```
zloery, line 45: Does the do step also need to store this, or is that expected to be part of a later step?

service/src/test/java/bio/terra/workspace/service/workspace/ApplicationServiceTest.java
```diff
@@ -0,0 +1,201 @@
...
+        .thenReturn(true);
+    Mockito.when(
+        mockSamService.isAuthorized(
+            Mockito.any(),
+            Mockito.eq(SamConstants.SAM_WORKSPACE_RESOURCE),
+            Mockito.any(),
+            Mockito.any()))
+        .thenReturn(true);
+
+    // Return a valid google group for cloud sync, as Google validates groups added to GCP projects.
```
zloery, line 110: This is a unit test, it shouldn't be talking to GCP. Is this in anticipation for creating application-owned resources?

service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
ddietterich: Hah! And the tests all worked anyway :)

service/src/main/java/bio/terra/workspace/service/workspace/WsmApplicationService.java
ddietterich: Yes, I will add it.

service/src/test/java/bio/terra/workspace/service/workspace/ApplicationServiceTest.java
ddietterich: No, I copied that mock from WorkspaceServiceTest which is a connected test.
I probably don't need it. CreateWorkspace doesn't call sync, right?
I just want the create workspace to succeed so I can exercise the application service.

service/src/main/java/bio/terra/workspace/db/exception/ApplicationInUseException.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.workspace.db.exception;
+
+import bio.terra.common.exception.NotFoundException;
+
+public class ApplicationInUseException extends NotFoundException {
```
zloery, line 5: 404 seems off here, I think this should be something like 400

service/src/test/java/bio/terra/workspace/service/workspace/ApplicationServiceTest.java
```diff
@@ -0,0 +1,212 @@
...
+    // enable norm - should fail - decommissioned
+    assertThrows(
+        InvalidApplicationStateException.class,
+        () -> appService.enableWorkspaceApplication(USER_REQUEST, workspaceId, NORM_UUID));
+
+    enumerateCheck(true, false, false);
+
+    // Use the DAO directly and skip the check to enable carmen so we can check that state.
+    appDao.enableWorkspaceApplicationNoCheck(workspaceId, CARMEN_UUID);
+    enumerateCheck(true, true, false);
```
marikomedlock, line 161: Not sure I'm following here -- when do you want to skip the check in normal operation?

service/src/main/java/bio/terra/workspace/db/exception/ApplicationInUseException.java
ddietterich: I intended it to be BadRequest, but it was bad cut'n'paste.

service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
```diff
@@ -76,6 +106,188 @@ public void createApplication(WsmApplication app) {
...
+   * to restore an application if the disable application flight fails.
+   *
+   * <p>It is not an error to enable an already enabled application.
+   *
+   * @param workspaceId workspace of interest
+   * @param applicationId application of interest
+   * @return workspace-application object
+   */
+  @WriteTransaction
+  public WsmWorkspaceApplication enableWorkspaceApplicationNoCheck(
```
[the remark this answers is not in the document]
ddietterich: It is useful in that test. I'll update the documentation.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApplicationApiController.java
```diff
@@ -0,0 +1,107 @@
...
+  public WorkspaceApplicationApiController(
+      AuthenticatedUserRequestFactory authenticatedUserRequestFactory,
+      HttpServletRequest request,
+      WsmApplicationService appService) {
+    this.authenticatedUserRequestFactory = authenticatedUserRequestFactory;
+    this.request = request;
+    this.appService = appService;
+  }
+
+  private AuthenticatedUserRequest getAuthenticatedInfo() {
```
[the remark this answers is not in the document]
marikomedlock: Ah gotcha, didn't see the autowiring.

PR #466, opened 2021-10-13
[PF-1043] Fix bad variable name; add test

ddietterich:
Chris Book is blocked by this bug. The fix is trivial: the path variable is `jobId` not `id`.
The rest of the code is for the integration test to verify that it works.

### DIFF
--- a/integration/src/main/resources/configs/integration/Jobs.json
+++ b/integration/src/main/resources/configs/integration/Jobs.json
@@ -1,0 +1,18 @@
+{
+  "name": "Jobs",
+  "description": "Test the JobsApi",
+  "serverSpecificationFile": "workspace-local.json",
+  "kubernetes": {},
+  "application": {},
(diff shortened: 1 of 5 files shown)


integration/src/main/resources/configs/integration/Jobs.json
```diff
@@ -0,0 +1,18 @@
...
+  "name": "Jobs",
+  "description": "Test the JobsApi",
+  "serverSpecificationFile": "workspace-local.json",
+  "kubernetes": {},
+  "application": {},
+  "testScripts": [
+    {
+      "name": "Jobs",
+      "numberOfUserJourneyThreadsToRun": 1,
+      "userJourneyThreadPoolSize": 2,
```
zloery, line 11: nit: this should also be 1, the thread pool size doesn't need to be larger than the number of threads to run

PR #467, opened 2021-10-14
[PF-947] Improve flight testing

zloery:
This change adds a number of tests using the Stairway `FlightDebugInfo` object, which allows us to intentionally fail and retry steps or entire flights.
Additionally, this adds retry rules to a number of steps which did not previously have them, and fixes idempotency of a few steps which previously could not be rerun.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceFlight.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteControlledResourceFlight.java
@@ -86,10 +86,8 @@ public class DeleteControlledResourceFlight extends Flight {
         throw new ControlledResourceNotImplementedException(
             "Delete not yet implemented for resource type " + resource.getResourceType());
     }
-
-    final RetryRule immediateRetryRule = RetryRules.shortDatabase();
     addStep(
         new DeleteMetadataStep(flightBeanBag.getResourceDao(), workspaceId, resourceId),
-        immediateRetryRule);
+        RetryRules.shortDatabase());
   }
 }


service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -277,46 +280,53 @@ void createAndDeleteWorkspace() {
...
 
-    // Test "undo" flight with lastStepFailure. Because deletion cannot be undone, this workspace
-    // will still be deleted after "undoing" the flight.
-    FlightDebugInfo debugInfo = FlightDebugInfo.newBuilder().lastStepFailure(true).build();
+    UUID createdId = workspaceService.createWorkspace(request, USER_REQUEST);
+    assertEquals(createdId, request.workspaceId());
+  }
+
+  @Test
+  void createWorkspaceUndoSteps() {
```
marikomedlock, line 315: nit: is this for MC workspaces only? if so, can you include it in the method name like you did above?

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -277,46 +280,53 @@ void createAndDeleteWorkspace() {
...
+    retrySteps.put(CreateWorkspaceAuthzStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_RETRY);
+    retrySteps.put(CreateWorkspaceStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_RETRY);
+    FlightDebugInfo debugInfo =
+        FlightDebugInfo.newBuilder().undoStepFailures(retrySteps).lastStepFailure(true).build();
     jobService.setFlightDebugInfoForTest(debugInfo);
 
-    // When a flight fails with no error message (e.g. because of debugInfo), Stairway will return
-    // an InvalidResultStateException.
+    // JobService throws a InvalidResultStateException when a synchronous flight fails without an
+    // exception, which occurs when a flight fails via debugInfo.
```
marikomedlock, line 326: nit: Not quite following the "synchronous flight" part of this comment -- I usually think of endpoints as a/sync, but not the flights themselves.

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -1086,6 +1097,275 @@ public void updateBqDatasetWithInvalidExpirationTimes() throws Exception {
...
+            .managedBy(ManagedByType.MANAGED_BY_USER)
+            .build();
+
+    ControlledGcsBucketResource createdBucket =
+        controlledResourceService.createBucket(
+            resource,
+            ControlledResourceFixtures.getGoogleBucketCreationParameters(),
+            Collections.emptyList(),
+            user.getAuthenticatedRequest());
+    assertEquals(resource, createdBucket);
```
marikomedlock, line 1209: I think these first 2 blocks are repeated in most methods, and the last block is repeated in some methods. Can you pull it out into a helper method, maybe that updates local properties or something?

service/src/main/java/bio/terra/workspace/service/workspace/flight/StoreGcpContextStep.java
```diff
@@ -27,8 +29,18 @@ public StoreGcpContextStep(GcpCloudContextService gcpCloudContextService, UUID w
   public StepResult doStep(FlightContext flightContext) throws InterruptedException {
     String projectId = flightContext.getWorkingMap().get(GCP_PROJECT_ID, String.class);
 
-    // Create the cloud context; throws if the context already exists. We let
-    // Stairway handle that.
+    // Create the cloud context; throws if another flight already created the context.
```
marikomedlock, line 32: No change requested, just an observation. I think you could also do this (check for flight id, write new context) in a single db call using a constraint on the table that includes the flight id and then modifying the create context SQL statement to do nothing, instead of throwing an exception, if that constraint is violated. That might be useful to look into if we ever run into db contention issues.

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
```diff
@@ -53,7 +54,16 @@ public StepResult doStep(FlightContext flightContext)
             .description(description)
             .build();
 
-    workspaceDao.createWorkspace(workspaceToCreate);
+    try {
+      workspaceDao.createWorkspace(workspaceToCreate);
+    } catch (DuplicateWorkspaceException ex) {
+      // This might be the result of a step re-running, or it might be an ID conflict. We can ignore
+      // this if the existing workspace matches the one we were about to create, otherwise rethrow.
+      Workspace existingWorkspace = workspaceDao.getWorkspace(workspaceId);
```
ddietterich, line 62: Since we generate the workspace ID as a random UUID, can we ever really get an ID conflict that is not the step re-running?

Not that you should remove the check; defense in depth is good. But maybe amend the comment?

service/src/main/java/bio/terra/workspace/service/workspace/flight/StoreGcpContextStep.java
ddietterich: It is a good idea. The column is nullable, but Postgres supports nulls - each considered unique from the other - in a unique constraint. So it could be done.

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -165,7 +176,7 @@ public void resetFlightDebugInfo() {
 
   @Test
   @DisabledIfEnvironmentVariable(named = "TEST_ENV", matches = BUFFER_SERVICE_DISABLED_ENVS_REG_EX)
-  public void createAiNotebookInstanceDo() throws Exception {
+  void createAiNotebookInstanceDo() throws Exception {
```
ddietterich, line 179: Does visibility matter for tests? I always make them public

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
zloery: Yep, this was really shorthand - flights aren't really sync or async, but since each flight is generally only called from one (sync or async) service endpoint, I tend to think of them as sync or async flights. 
Clarified this comment everywhere

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
zloery: As of JUnit 5, not really. The `@Test` annotation documentation just states that they can't be `private` or `static` and must not have a return type.
I have a very slight style preference for package-private since it's less typing, but that was definitely outweighed by refactoring this file.

PR #470, opened 2021-10-20
[PF-1063] Make WSM SA grant itself Role Admin on data-plane projects at handout

zloery:
This change makes the Workspace Manager service account grant itself the "Role Admin" (`roles/iam.roleAdmin`) role on data plane projects at handout time. This role cannot be given at the folder level (where most WSM SA permissions live), and must instead be applied to every project.
This will not affect Broad deployments which already grant broader roles at the folder level, but will be necessary if/when we downsize the WSM SA's data-plane permissions in these deployments in the future.

This also picked up a few unrelated spotless changes which I included so they don't interfere with other PRs.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
+++ b/service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
@@ -8,6 +8,7 @@ import bio.terra.stairway.exception.RetryException;
 import bio.terra.workspace.service.workspace.exceptions.SaCredentialsMissingException;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.services.cloudresourcemanager.v3.model.Project;
+import com.google.auth.ServiceAccountSigner;
 import com.google.auth.oauth2.ComputeEngineCredentials;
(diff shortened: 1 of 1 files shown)


service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
```diff
@@ -81,4 +85,21 @@ public static String getControlPlaneProjectId() {
...
+
+  /**
+   * Returns the email of the application default credentials, which should represent a service
+   * account in all WSM deployments.
+   */
+  public static String getWsmSaEmail(GoogleCredentials wsmCredentials) {
+    // WSM always runs as a service account, but credentials for that SA may come from different
+    // sources depending on whether it is running in GCP or locally.
+    if (wsmCredentials instanceof ComputeEngineCredentials) {
+      return ((ComputeEngineCredentials) wsmCredentials).getAccount();
```
ddietterich, line 97: NittyNit: if `getAccount()` is from a super class/interface, then you could do it in a single if.

zloery: Good point, I didn't realize the two subclasses both implemented the same interface!

PR #480, opened 2021-11-04
PF-998: When creating a new gcs-bucket with invalid bucket name, handle StorageException correctly by re-throwing an InvalidReferenceException

yuhuyoyo:
https://broadworkbench.atlassian.net/browse/PF-998

When creating a controlled resource GCS bucket with an invalid name (too long, contains goog prefix, etc), wsm ValidationUtils cannot catch these cases. Thus a StorageException with status code 400 is thrown. We should handle this case by re-throwing an InvalidReferenceException with useful messages about the error.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
@@ -5,6 +5,7 @@ import bio.terra.workspace.generated.model.ApiGcpAiNotebookInstanceCreationParam
 import bio.terra.workspace.generated.model.ApiGcpAiNotebookInstanceVmImage;
 import bio.terra.workspace.service.resource.exception.InvalidNameException;
 import bio.terra.workspace.service.resource.referenced.exception.InvalidReferenceException;
+import com.google.common.collect.ImmutableList;
 import java.util.regex.Pattern;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -41,11 +42,46 @@ public class ValidationUtils {
   public static final Pattern RESOURCE_NAME_VALIDATION_PATTERN =
       Pattern.compile("^[a-zA-Z0-9][-_a-zA-Z0-9]{0,1023}$");
 
+  private static final String GOOG_PREFIX = "goog";
+  private static final ImmutableList<String> GOOGLE_NAMES = ImmutableList.of("google", "g00gle");
+
+  /**
+   * Validates gcs-bucket name following Google documentation
+   * https://cloud.google.com/storage/docs/naming-buckets#requirements on a best-effort base.
+   *
+   * This method DOES NOT guarentee that the bucket name is valid.
+   * @param name gcs-bucket name
+   */
   public static void validateBucketName(String name) {
     if (StringUtils.isEmpty(name) || !BUCKET_NAME_VALIDATION_PATTERN.matcher(name).matches()) {
       logger.warn("Invalid bucket name {}", name);
       throw new InvalidReferenceException(
           "Invalid GCS bucket name specified. Names must be 3-222 lowercase letters, numbers, dashes, and underscores. See Google documentation for the full specification.");
+    }
+    String[] substring = name.split("\\.");
+    if (substring.length == 0 && name.length() > 63) {
+      logger.warn("Invalid bucket name {}", name);
+      throw new InvalidReferenceException(
+          "Invalid GCS bucket name specified. Bucket names must contain 3-63 characters. See Google documentation https://cloud.google.com/storage/docs/naming-buckets#requirements for the full specification.");
+    }
+    for (String s : substring) {
+      if (s.length() > 63) {
+        logger.warn("Invalid bucket name {}", name);
+        throw new InvalidReferenceException(
+            "Invalid GCS bucket name specified. Names containing dots can contain up to 222 characters, but each dot-separated component can be no longer than 63 characters. See Google documentation https://cloud.google.com/storage/docs/naming-buckets#requirements for the full specification.");
+      }
+    }
+    if (name.startsWith(GOOG_PREFIX)) {
+      logger.warn("Invalid bucket name {}", name);
+      throw new InvalidReferenceException(
+          "Invalid GCS bucket name specified. Bucket names cannot have prefix goog. See Google documentation https://cloud.google.com/storage/docs/naming-buckets#requirements for the full specification.");
+    }
+    for (String google : GOOGLE_NAMES) {
+      logger.warn("Invalid bucket name {}", name);
+      if (name.contains(google)) {
+        throw new InvalidReferenceException(
+            "Invalid GCS bucket name specified. Bucket names cannot contains google or mis-spelled google. See Google documentation https://cloud.google.com/storage/docs/naming-buckets#requirements for the full specification.");
+      }
     }
   }
 
--- a/service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
@@ -51,6 +51,8 @@ public class ValidationUtils {
    *
    * This method DOES NOT guarentee that the bucket name is valid.
    * @param name gcs-bucket name
+   * @throws InvalidNameException throws exception when the bucket name fails to conform to the
+   * Google naming convention for bucket name.
    */
   public static void validateBucketName(String name) {
     if (StringUtils.isEmpty(name) || !BUCKET_NAME_VALIDATION_PATTERN.matcher(name).matches()) {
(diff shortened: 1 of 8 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
```diff
@@ -90,6 +92,10 @@ public StepResult doStep(FlightContext flightContext)
         throw new DuplicateResourceException(
             "The provided bucket name is already in use, please choose another.", storageException);
       }
+      if (storageException.getCode() == HttpStatus.SC_BAD_REQUEST) {
+        throw new InvalidReferenceException(
+            "The provided bucket name is invalid, please follow the naming instruction on https://cloud.google.com/storage/docs/naming-buckets#requirements");
```
marikomedlock, line 97: I think a `BAD_REQUEST` may not always mean an invalid bucket name. Another property (e.g. location) may also be the problem.

Also, I think it would be better to wrap the `StorageException` in the `InvalidReferenceException` so that we have a cause and can log that original excpetion.

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -1081,6 +1082,34 @@ void createGcsBucketDo() throws Exception {
...
+            .managedBy(ManagedByType.MANAGED_BY_USER)
+            .bucketName("goog-bucket")
+            .build();
+
+    // Test idempotency of bucket-specific steps by retrying them once.
+    Map<String, StepStatus> retrySteps = new HashMap<>();
+    retrySteps.put(CreateGcsBucketStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_RETRY);
+    retrySteps.put(GcsBucketCloudSyncStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_RETRY);
+    jobService.setFlightDebugInfoForTest(
+        FlightDebugInfo.newBuilder().doStepFailures(retrySteps).build());
```
marikomedlock, line 1101: It looks like this step idempotency is already covered by the previous test `createGcsBucketDo`. Is there a reason you included again here?

marikomedlock (commented):
In a previous version of your changes that I saw, you also expanded the cases that WSM catches in its own validation. I thought that was useful -- both for a better error message and potentially fewer cloud calls -- even if it doesn't remove the need for handling 400s from GCP as you've done here.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
yuhuyoyo: Changed the error message and wrap the StorageException. PTAL

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
yuhuyoyo: oops. Removed.

yuhuyoyo:
> In a previous version of your changes that I saw, you also expanded the cases that WSM catches in its own validation. I thought that was useful -- both for a better error message and potentially fewer cloud calls -- even if it doesn't remove the need for handling 400s from GCP as you've done here.

Added back the change to the validationUtils.

yuhuyoyo:
Look like an unrelated test MdcHookTest is failing but I ran ../gradlew :service:unitTest --scan locally and all test passes. Not sure what to do about it.

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -41,12 +42,47 @@
...
+   * Validates gcs-bucket name following Google documentation
+   * https://cloud.google.com/storage/docs/naming-buckets#requirements on a best-effort base.
+   *
+   * This method DOES NOT guarentee that the bucket name is valid.
+   * @param name gcs-bucket name
+   */
   public static void validateBucketName(String name) {
     if (StringUtils.isEmpty(name) || !BUCKET_NAME_VALIDATION_PATTERN.matcher(name).matches()) {
       logger.warn("Invalid bucket name {}", name);
       throw new InvalidReferenceException(
```
zloery, line 58: I think we set a bad precedent here before, but this (and the other cases below) shouldn't be a `InvalidReferenceException` since this isn't just validating references. 
Could you make a new exception instead (something like `InvalidCloudSpecification`, status code 400)?

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/CreateGcsBucketStepTest.java
```diff
@@ -159,4 +173,31 @@ public void testCreatesBucketWithoutAllParameters() throws RetryException, Inter
...
     assertThat(info.getLifecycleRules(), empty());
   }
+
+  @Test
+  public void createBucket_invalidBucketName_throwsInvalidReferenceException() {
+    doThrow(new StorageException(400, "bad request"))
+        .when(mockStorageCow).create(bucketInfoCaptor.capture());
+
+    // A bad bucket name that fails to be caught by the WSM validation.
+    final String bucketName = uniqueName("bad-bucket-name");
```
zloery, line 183: out of curiosity, what makes this name bad? it seems valid by the guidelines

zloery (commented):
Could you include the ticket number (PF-998) in your PR title? Jira maps PRs to tickets, and we use that mapping for QA/compliance

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/CreateGcsBucketStepTest.java
yuhuyoyo: the name is fine. I'm just trying to come up with a generic name that refers to a bad bucket name that wsm's validator fails to catch. it's all mocked out in this unit test so the naming doesn't really matter.

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
yuhuyoyo: there's an existing InvalidNameException. Wdyt I just use that?

zloery: sounds good for the exceptions here, but the 400 in `CreateGcsBucketStep` can also be caused by other fields so that wouldn't be right there. 
that one already has the error from GCP included though, so it could just be a more generic `BadRequestException`

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateGcsBucketStep.java
```diff
@@ -90,6 +92,10 @@ public StepResult doStep(FlightContext flightContext)
         throw new DuplicateResourceException(
             "The provided bucket name is already in use, please choose another.", storageException);
       }
+      if (storageException.getCode() == HttpStatus.SC_BAD_REQUEST) {
+        throw new BadRequestException(
+            "Receive 400 BAD_REQUEST exception when creating a new gcs-bucket", storageException);
```
zloery, line 97: nit: Received*

zloery (approved):
LGTM. I'm not sure what's going on with the failing `processAppTest`, I don't think your PR should affect it and it passes when I check your branch out locally. I'll take a look

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -1081,6 +1081,28 @@ void createGcsBucketDo() throws Exception {
             workspace.getWorkspaceId(), resource.getResourceId(), user.getAuthenticatedRequest()));
   }
 
+  @Test
+  @DisabledIfEnvironmentVariable(named = "TEST_ENV", matches = BUFFER_SERVICE_DISABLED_ENVS_REG_EX)
+  void createGcsBucketDo_invalidBucketName_throwsInvalidReferenceException() throws Exception {
```
marikomedlock, line 1086: nit: `_throwsBadRequestException`

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -41,12 +42,47 @@
...
+   */
   public static void validateBucketName(String name) {
     if (StringUtils.isEmpty(name) || !BUCKET_NAME_VALIDATION_PATTERN.matcher(name).matches()) {
       logger.warn("Invalid bucket name {}", name);
-      throw new InvalidReferenceException(
+      throw new InvalidNameException(
           "Invalid GCS bucket name specified. Names must be 3-222 lowercase letters, numbers, dashes, and underscores. See Google documentation for the full specification.");
     }
+    String[] substring = name.split("\\.");
+    if (substring.length == 0 && name.length() > 63) {
```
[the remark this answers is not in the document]
yuhuyoyo: oh yeah you are right. removed the check

PR #482, opened 2021-11-04
[No ticket] Fix unit tests which fail on Github

zloery:
The unit test `processAppTest` was consistently failing when run on Github with `DuplicateKeyExceptions` from the DB, indicating it was not being cleaned up properly. I've switched the `DirtiesContext` annotation so the test will restart before each of the marked tests instead of after, which should handle this problem more consistently.

No ticket because this only affects unit test code.


ddietterich (approved):
Thanks for tracking this down!

yuhuyoyo (approved):

PR #484, opened 2021-11-05
PF-556: Do spend profile check as a step of CreateGoogleContextFlight

yuhuyoyo:
CheckSpendProfileStep checks for whether the user has an authorized spend profile. 

Is the spend profile is empty or unauthorized, step fails.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/exceptions/MissingSpendProfileException.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/exceptions/MissingSpendProfileException.java
@@ -8,8 +8,12 @@ import java.util.UUID;
  * profile associated with it.
  */
 public class MissingSpendProfileException extends BadRequestException {
-  public MissingSpendProfileException(UUID workspaceId) {
-    super(
+  public MissingSpendProfileException(String message) {
+    super(message);
+  }
+
+  public static MissingSpendProfileException missingSpendProfileException(UUID workspaceId) {
+    return new MissingSpendProfileException(
         String.format(
             "Spend profile id required, but none found on workspace %s", workspaceId.toString()));
   }
--- a/service/src/main/java/bio/terra/workspace/service/workspace/exceptions/MissingSpendProfileException.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/exceptions/MissingSpendProfileException.java
@@ -12,7 +12,7 @@ public class MissingSpendProfileException extends BadRequestException {
     super(message);
   }
 
-  public static MissingSpendProfileException missingSpendProfileException(UUID workspaceId) {
+  public static MissingSpendProfileException forWorkspace(UUID workspaceId) {
     return new MissingSpendProfileException(
         String.format(
             "Spend profile id required, but none found on workspace %s", workspaceId.toString()));
--- a/service/src/main/java/bio/terra/workspace/service/workspace/exceptions/NoBillingAccountException.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/exceptions/NoBillingAccountException.java
@@ -8,8 +8,12 @@ import bio.terra.workspace.service.spendprofile.SpendProfileId;
  * Profile does not have a billing account associated with it.
  */
 public class NoBillingAccountException extends BadRequestException {
-  public NoBillingAccountException(SpendProfileId spendProfileId) {
-    super(
+  public NoBillingAccountException(String message) {
+    super(message);
+  }
+
+  public static NoBillingAccountException noBillingAccountException(SpendProfileId spendProfileId) {
+    return new NoBillingAccountException(
         String.format(
             "Billing acocunt id required, but none found on spend profile %s",
             spendProfileId.id()));
--- a/service/src/main/java/bio/terra/workspace/service/workspace/exceptions/NoBillingAccountException.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/exceptions/NoBillingAccountException.java
@@ -12,7 +12,7 @@ public class NoBillingAccountException extends BadRequestException {
     super(message);
   }
 
-  public static NoBillingAccountException noBillingAccountException(SpendProfileId spendProfileId) {
+  public static NoBillingAccountException forSpendProfile(SpendProfileId spendProfileId) {
     return new NoBillingAccountException(
         String.format(
             "Billing acocunt id required, but none found on spend profile %s",
--- a/service/src/main/java/bio/terra/workspace/service/workspace/flight/CheckSpendProfileStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/flight/CheckSpendProfileStep.java
@@ -1,7 +1,6 @@
 package bio.terra.workspace.service.workspace.flight;
 
 import static bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.BILLING_ACCOUNT_ID;
-import static bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.WORKSPACE;
 
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.FlightMap;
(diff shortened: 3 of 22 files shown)


service/src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
```diff
@@ -106,29 +110,80 @@ void successCreatesProjectAndContext() throws Exception {
...
+    assertEquals(FlightStatus.ERROR, flightState.getFlightStatus());
+    assertTrue(workspaceService.getAuthorizedGcpCloudContext(workspaceId, userRequest).isEmpty());
+    String projectId =
+        flightState.getResultMap().get().get(WorkspaceFlightMapKeys.GCP_PROJECT_ID, String.class);
+    // The Project should exist, but requested to be deleted.
+    Project project = crl.getCloudResourceManagerCow().projects().get(projectId).execute();
+    assertEquals(projectId, project.getProjectId());
+    assertEquals("DELETE_REQUESTED", project.getState());
+  }
+
```
zloery, line 170: You removed a test from `WorkspaceServiceTest` which validated that an unauthorized user could not link a spend profile, can you add a test covering the same case here?

Per our [test user documentation](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/integration/src/main/resources/testusers/README.md) the user `liam.dragonmaw` does not have spend profile access, so you could use that account. You should be able to access those credentials via the `secondaryUser` methods in `UserAccessUtils`

service/src/main/java/bio/terra/workspace/service/workspace/flight/CheckSpendProfileStep.java
```diff
@@ -0,0 +1,61 @@
+package bio.terra.workspace.service.workspace.flight;
+
+import static bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.BILLING_ACCOUNT_ID;
+import static bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.WORKSPACE;
```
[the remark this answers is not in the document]
yuhuyoyo: that's very good to know. thanks!

service/src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
yuhuyoyo: thanks for the links!

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlight.java
```diff
@@ -39,6 +39,13 @@ public CreateGcpContextFlight(FlightMap inputParameters, Object applicationConte
 
     RetryRule retryRule = RetryRules.shortExponential();
 
+    addStep(
```
zloery, line 42: This step needs to be first in this flight, before `GenerateProjectIdStep` and `PullProjectFromPoolStep` up above. 
Otherwise, unauthorized users will be able to churn projects by repeatedly calling this endpoint, which could exhaust project creation quota.

service/src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
```diff
@@ -152,20 +151,16 @@ void createsProjectAndContext_emptySpendProfile_fails() throws Exception {
...
 
   @Test
   @DisabledIfEnvironmentVariable(named = "TEST_ENV", matches = BUFFER_SERVICE_DISABLED_ENVS_REG_EX)
-  void createsProjectAndContext_unauthorizedSpendProfile_fails() throws Exception {
+  void createsProjectAndContext_unauthorizedSpendProfile_flightFailsAndGcpProjectNotCreated()
+      throws Exception {
     UUID workspaceId = createWorkspace(spendUtils.defaultSpendId());
-    AuthenticatedUserRequest userRequest =
-        userAccessUtils.secondUserAuthRequest().email("liam.dragonmaw@test.firecloud.org");
+    AuthenticatedUserRequest userRequest = userAccessUtils.secondUserAuthRequest();
```
zloery, line 163: nit: to make it clearer that this is not the primary test user, could you rename this variable something like `secondaryUserRequest` or `unauthorizedUserRequest`?

zloery (approved):
LGTM. Thanks for picking this up!

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlight.java
```diff
@@ -31,14 +31,20 @@ public CreateGcpContextFlight(FlightMap inputParameters, Object applicationConte
...
         inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
 
+    RetryRule retryRule = RetryRules.shortExponential();
+    addStep(
+        new CheckSpendProfileStep(
+            appContext.getWorkspaceDao(),
+            appContext.getSpendProfileService(),
+            workspaceId,
+            userRequest),
+        retryRule);
```
marikomedlock, line 41: Stairway uses the retry rule to determine the retry pattern when the step either throws a `RetryException` or returns a `StepStatus.STEP_RESULT_FAILURE_RETRY`. Does this step do either of those?

I think the `authorizeLinking` is probably retrying calls to SAM, but I'm not sure if it will ever do that using Stairway's retry mechanism.

service/src/test/java/bio/terra/workspace/service/workspace/flight/CreateGoogleContextFlightTest.java
```diff
@@ -106,29 +111,86 @@ void successCreatesProjectAndContext() throws Exception {
...
+
+    FlightState flightState =
+        StairwayTestUtils.blockUntilFlightCompletes(
+            jobService.getStairway(),
+            CreateGcpContextFlight.class,
+            createInputParameters(workspaceId, userRequest),
+            STAIRWAY_FLIGHT_TIMEOUT,
+            FlightDebugInfo.newBuilder().build());
+
+    assertEquals(FlightStatus.ERROR, flightState.getFlightStatus());
```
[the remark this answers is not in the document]
yuhuyoyo: good point. done.

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlight.java
yuhuyoyo: Ah i see. since none of those exceptions are retyException, i just remove the retryRule.

service/src/main/java/bio/terra/workspace/service/workspace/exceptions/MissingSpendProfileException.java
```diff
@@ -8,8 +8,12 @@
  * profile associated with it.
  */
 public class MissingSpendProfileException extends BadRequestException {
-  public MissingSpendProfileException(UUID workspaceId) {
-    super(
+  public MissingSpendProfileException(String message) {
+    super(message);
+  }
+
+  public static MissingSpendProfileException missingSpendProfileException(UUID workspaceId) {
```
marikomedlock, line 15: nit: Perhaps `forWorkspace` or something else? I don't think naming a factory method the same as the class adds much information.

service/src/main/java/bio/terra/workspace/service/workspace/exceptions/NoBillingAccountException.java
```diff
@@ -8,8 +8,12 @@
  * Profile does not have a billing account associated with it.
  */
 public class NoBillingAccountException extends BadRequestException {
-  public NoBillingAccountException(SpendProfileId spendProfileId) {
-    super(
+  public NoBillingAccountException(String message) {
+    super(message);
+  }
+
+  public static NoBillingAccountException noBillingAccountException(SpendProfileId spendProfileId) {
```
marikomedlock, line 15: nit: ditto above. perhaps `forSpendProfile`?

marikomedlock (approved):

PR #485, opened 2021-11-08
PF-800: Validation of Controlled Resource Metadata Fields

yuhuyoyo:
Add char limit to description and privateUserEmail (assigned user).

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
@@ -7,6 +7,7 @@ import bio.terra.workspace.service.resource.exception.InvalidNameException;
 import bio.terra.workspace.service.resource.referenced.exception.InvalidReferenceException;
 import com.google.common.collect.ImmutableList;
 import java.util.regex.Pattern;
+import javax.annotation.Nullable;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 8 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -34,13 +34,16 @@
...
 import javax.annotation.Nullable;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 /** CRUD methods for controlled objects. */
 @Component
 public class ControlledResourceService {
 
+  private static final int MAX_ASSIGNED_USER_LENGTH = 128;
```
ddietterich, line 45: After some reading, I believe the max size of an email is 254. Suggest we use that.
If you google it you get mountains of arguments saying it is 254, 255, 256, 320...

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -38,21 +39,26 @@
...
+  /**
+   * Resource names must be 1-1024 characters, using letters, numbers, dashes, and underscores and
+   * must not start with a dash or underscore.
+   */
   public static final Pattern RESOURCE_NAME_VALIDATION_PATTERN =
       Pattern.compile("^[a-zA-Z0-9][-_a-zA-Z0-9]{0,1023}$");
 
   private static final String GOOG_PREFIX = "goog";
   private static final ImmutableList<String> GOOGLE_NAMES = ImmutableList.of("google", "g00gle");
+  private static final int MAX_RESOURCE_DESCRIPTION_NAME = 2048;
```
ddietterich, line 51: My intuition says to make this bigger, but I don't have a solid technical reason.
Maybe we leave it as is and if anyone runs ever runs in to it with a legit purpose, we can bump it up.

ddietterich (changes requested):
One change to the max email size. Otherwise, looks good.

ddietterich (approved):
lgtm

PR #486, opened 2021-11-09
[PF-1110] Have WSM pull pet credentials for validating referenced resources

zloery:
Currently, WSM validates referenced buckets and BigQuery datasets using user provided credentials. Unlike other endpoints this requires the `cloud-platform` scope. Some users cannot grant this and have been using pet SA credentials instead, so this change has WSM fetch these credentials automatically on the user's behalf if they're available for the current workspace.
For workspaces which do not have a GCP context, behavior is unchanged.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -914,8 +914,8 @@ public class SamService {
   }
 
   /**
-   * Fetch credentials of a user's pet service account in a given project. This request to Sam
-   * will create the pet SA if it doesn't already exist.
+   * Fetch credentials of a user's pet service account in a given project. This request to Sam will
+   * create the pet SA if it doesn't already exist.
    */
   public AuthenticatedUserRequest getOrCreatePetSaCredentials(
       String projectId, AuthenticatedUserRequest userRequest) throws InterruptedException {
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -919,11 +919,16 @@ public class SamService {
    */
   public AuthenticatedUserRequest getOrCreatePetSaCredentials(
       String projectId, AuthenticatedUserRequest userRequest) throws InterruptedException {
-    GoogleApi googleApi = samGoogleApi(userRequest.getRequiredToken());
+    GoogleApi samGoogleApi = samGoogleApi(userRequest.getRequiredToken());
     try {
       String petEmail = getOrCreatePetSaEmail(projectId, userRequest);
       String petToken =
-          SamRetry.retry(() -> googleApi.getPetServiceAccountToken(projectId, PET_SA_OAUTH_SCOPES));
+          SamRetry.retry(
+              () -> samGoogleApi.getPetServiceAccountToken(projectId, PET_SA_OAUTH_SCOPES));
+      // This should never happen, but it's more informative than an NPE from Optional.of
+      if (petToken == null) {
+        throw new InternalServerErrorException("Sam returned null pet service account token");
+      }
       return new AuthenticatedUserRequest().email(petEmail).token(Optional.of(petToken));
     } catch (ApiException apiException) {
       throw SamExceptionFactory.create(
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -16,6 +16,7 @@ import bio.terra.workspace.service.stage.StageService;
 import bio.terra.workspace.service.workspace.exceptions.InternalLogicException;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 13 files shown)


ddietterich (commented):
Suggest a little DRYing up

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedGcsBucketResource.java
```diff
@@ -96,7 +100,26 @@ public void validate() {
   @Override
   public boolean checkAccess(FlightBeanBag context, AuthenticatedUserRequest userRequest) {
     CrlService crlService = context.getCrlService();
-    return crlService.canReadGcsBucket(bucketName, userRequest);
+    SamService samService = context.getSamService();
+    GcpCloudContextService gcpCloudContextService = context.getGcpCloudContextService();
+    // If the resource's workspace has a GCP cloud context, use the SA from that context. Otherwise,
```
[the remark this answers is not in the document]
zloery: Good point, the pattern of "fetch workspace pet SA credentials if available" will be common for all GCP references. Pulled that out into a method in `PetSaService`

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -232,6 +233,27 @@ private static boolean bindingEquals(Binding binding1, Binding binding2) {
...
+  /**
+   * Fetches credentials for the provided user's pet service account in the current workspace's GCP
+   * context if one exists. This will create a pet SA for the user if one does not exist, but will
+   * return empty if the workspace does not have a GCP context.
+   *
+   * <p>This method does not validate that the provided credentials have appropriate workspace
+   * access.
+   */
+  public Optional<AuthenticatedUserRequest> getWorkspacePetCredentials(
+      UUID workspaceId, AuthenticatedUserRequest userRequest) {
```
ddietterich, line 245: minor: since you have the userRequest, you could make this method always return an AuthenticatedUserRequest - either the pet or the userRequest.

But maybe you need this method anyway, so could make a `getPetCredsIfTheyExist()` I suppose.
Anyway, this is mostly aesthetic, so feel free to ignore.

ddietterich (approved):
LGTM - one aesthetic comment

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -909,6 +913,24 @@ public String getOrCreatePetSaEmail(String projectId, AuthenticatedUserRequest u
     }
   }
 
+  /**
+   * Fetch credentials of a user's pet service account in a given project. This request to Sam will
+   * create the pet SA if it doesn't already exist.
+   */
+  public AuthenticatedUserRequest getOrCreatePetSaCredentials(
+      String projectId, AuthenticatedUserRequest userRequest) throws InterruptedException {
+    GoogleApi googleApi = samGoogleApi(userRequest.getRequiredToken());
```
jaycarlton, line 922: maybe name this `samGoogleApi`

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -909,6 +913,24 @@ public String getOrCreatePetSaEmail(String projectId, AuthenticatedUserRequest u
...
+    try {
+      String petEmail = getOrCreatePetSaEmail(projectId, userRequest);
+      String petToken =
+          SamRetry.retry(() -> googleApi.getPetServiceAccountToken(projectId, PET_SA_OAUTH_SCOPES));
+      return new AuthenticatedUserRequest().email(petEmail).token(Optional.of(petToken));
```
jaycarlton, line 927: If there's a possibility `petToken` can be null, use `Optional.ofNullable()`

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -232,6 +233,27 @@ private static boolean bindingEquals(Binding binding1, Binding binding2) {
...
+    Optional<String> maybeProjectId =
+        gcpCloudContextService
+            .getGcpCloudContext(workspaceId)
+            .map(GcpCloudContext::getGcpProjectId);
+    return maybeProjectId.map(
```
jaycarlton, line 250: I'd inline `maybeProjectId` and just return a longer optional chain.

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
jaycarlton: It's a  bit of an awkward contract for a public method, because the SA creds don't belong to the workspace but instead to a cloud context's GCP project. However, if you just returned an `AuthenticatedUserRequest`, you'd have no way of knowing whether it was the pet or not without inspecting the email address. Maybe there's an opportunity to make that class more expressive.

I'd keep it the way it is for now and start thinking about higher-level ways of working with different kinds of credentials.

jaycarlton (approved):
Looks good.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
zloery: I don't believe Sam will ever return a 200 with a null `petToken` here, it should throw an exception in any abnormal cases. 
I'm not a Sam expert though, so I added a check with a more informative message.

PR #487, opened 2021-11-10
PF-1165: add a lint checker test to the github action

yuhuyoyo:
i never remember to run these things so having an automatic test to check for me is useful.

Not sure if I should just merge to the main for this kind of change?


ddietterich (commented):
I'm a little puzzled. The gradle build target does `spotlessApply` and `spotbugs`. As of right now, `spotbugs` does not need to pass, but we ought to clean that up and make it a build dependency also.

Anyway, not sure why we need a separate workflow to do that. OTOH, it doesn't hurt anything.

yuhuyoyo:
> I'm a little puzzled. The gradle build target does `spotlessApply` and `spotbugs`. As of right now, `spotbugs` does not need to pass, but we ought to clean that up and make it a build dependency also.
> 
are they enforced? I'm pretty sure I've pushed PR that have unused import and format issue and didn't receive any complain. 

> Anyway, not sure why we need a separate workflow to do that. OTOH, it doesn't hurt anything.
 I can merge it to an existing workflow. i'm just erring on the side of caution and not want to break any existing important workflow.

ddietterich:
If you do not use the `build` target, then you can push things that are not "spotless".
We could enforce it in gradle at the cost of a longer debug cycle.

yuhuyoyo:
what do you mean by "not use the build target"? 

i don't think it's that big a trouble to fix the complain and it guards against some obvious flaws and dead code easily missed by human eyes. plus i often find myself adding unrelated files to the PR when i ran spotlessApply.

ddietterich (approved):
OK, let's do this

yuhuyoyo:
Thanks. i will fix the failing test and add a ticket

integration/src/main/java/scripts/testscripts/EnablePet.java
```diff
@@ -60,7 +62,7 @@ protected void doUserJourney(TestUserSpecification testUser, WorkspaceApi userWo
         ClientTestUtils.getWorkspaceClientFromToken(petSaToken, server);
     String petEnableResult = petSaWorkspaceApi.enablePet(getWorkspaceId());
     assertEquals(petSaEmail, petEnableResult);
-    Iam petIamClient = ClientTestUtils.getGcpIamClientFromToken(petSaToken);
+    Iam unused = ClientTestUtils.getGcpIamClientFromToken(petSaToken);
```
zloery, line 65: nit: this is unused because the assertion that would use it is commented out below. Instead of renaming, let's just comment this out as well.

integration/src/main/java/scripts/testscripts/PrivateControlledAiNotebookInstanceLifecycle.java
```diff
@@ -62,9 +63,10 @@ protected void doSetup(List<TestUserSpecification> testUsers, WorkspaceApi works
   }
 
   @Override
+  @SuppressFBWarnings(value = "DLS_DEAD_LOCAL_STORE")
   protected void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspaceApi)
       throws Exception {
-    String projectId = CloudContextMaker.createGcpCloudContext(getWorkspaceId(), workspaceApi);
+    String unused = CloudContextMaker.createGcpCloudContext(getWorkspaceId(), workspaceApi);
```
zloery, line 69: nit: if we aren't using this, it's probably better to just ignore the return value of this call instead

zloery (approved):

PR #489, opened 2021-11-10
[PF-1167][PF-1056] Add missing validate() calls, other validation cleanup

zloery:
This change:
- Ensures that all descendent classes of `WsmResource` call `validate()` in their constructors to ensure that all `WsmResource` objects are validated 
- Removes unnecessary explicit calls to `validate()` in a few places
- Adds name/description validation to `update` calls for both controlled and referenced resources, as these modify DB values without constructing (and thus validating) `WsmResource` objects
- Removes references to viewing logs in user-facing error messages

### DIFF
--- a/integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
@@ -219,6 +219,20 @@ public class ControlledGcsBucketLifecycle extends WorkspaceAllocateTestScriptBas
         resource.getMetadata().getDescription());
     assertEquals(UPDATED_RESOURCE_NAME, resource.getMetadata().getName());
     assertEquals(UPDATED_DESCRIPTION, resource.getMetadata().getDescription());
+    // However, invalid updates are rejected.
+    String invalidName = "!!!invalid_name!!!";
+    ApiException invalidUpdateEx =
+        assertThrows(
+            ApiException.class,
+            () ->
+                updateBucketAttempt(
+                    resourceApi,
+                    resourceId,
+                    invalidName,
+                    /*updatedDescription=*/ null,
+                    /*updateParameters=*/ null));
+    assertEquals(HttpStatusCodes.STATUS_CODE_BAD_REQUEST, invalidUpdateEx.getCode());
+
     final Bucket retrievedUpdatedBucket =
         ownerStorageClient.get(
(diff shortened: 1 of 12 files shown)


integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
```diff
@@ -219,6 +219,20 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
...
+    String invalidName = "!!!invalid_name!!!";
+    ApiException invalidUpdateEx =
+        assertThrows(
+            ApiException.class,
+            () ->
+                updateBucketAttempt(
+                    resourceApi,
+                    resourceId,
+                    invalidName,
+                    /*updatedDescription=*/ null,
```
jaycarlton, line 232: IntelliJ shows me these parameter names when I pass a literal value, so I haven't felt the need to put these comments. Is it only the pro version that does this?

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -78,7 +78,6 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
                 CloningInstructions.fromApiModel(body.getMetadata().getCloningInstructions()))
             .bucketName(body.getBucket().getBucketName())
             .build();
-    resource.validate();
```
jaycarlton, line 81: Why are these not necessary?

jaycarlton (approved):

integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
zloery: Community edition does that too, but other editors probably don't (and it's nice when I'm reading the code on Github)

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
zloery: Every subclass of `WsmResource` calls `validate` from every constructor, and since these are immutable objects this lets us enforce validity at object creation time. This call is doing duplicate validation with the `build()` call above

ddietterich (approved):
LGTM

PR #490, opened 2021-11-10
PF-1145: catch and re-sync to cloud in the cloud sync steps

yuhuyoyo:
There could be a few second delay for the cloud to get the newly created googlegroup by SAM. So a group does not exist exception is throw.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
+++ b/service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
@@ -3,12 +3,9 @@ package bio.terra.workspace.common.utils;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.FlightMap;
 import bio.terra.stairway.FlightState;
-import bio.terra.stairway.StepResult;
-import bio.terra.stairway.StepStatus;
 import bio.terra.workspace.generated.model.ApiErrorReport;
 import bio.terra.workspace.service.job.JobMapKeys;
 import bio.terra.workspace.service.workspace.exceptions.MissingRequiredFieldsException;
-import com.google.cloud.storage.StorageException;
 import java.util.Optional;
 import javax.annotation.Nullable;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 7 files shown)


jmczerk (approved):
This looks good.  Optionally, you might consider adding a static method to `bio.terra.workspace.common.utils.FlightUtils` that takes a `StorageException` as an argument and returns a `StepResult` (or rethrows if not the expected error code), to centralize the error code checking logic in one place.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
```diff
@@ -111,4 +114,20 @@ public static String getFlightErrorMessage(FlightState subflightState) {
...
+  /**
+   * Handles StorageException, when it has code 400 or 404, returns a STEP_RESULT_FAILURE_RETRY
+   * status to retry the step, otherwise, re-throw the exception.
+   *
+   * @param e storageException
+   * @return a stepResult with a retry status when the storageException has recoverable exception.
+   * @throws StorageException rethrows StorageException when the step should not be retried.
+   */
+  public static StepResult retryStorageExceptionOrRethrow(StorageException e) {
+    if (e.getCode() == HttpStatus.BAD_REQUEST.value()
```
zloery, line 127: Do we need to retry both of these cases? I'm not actually sure what error code GCP returns if a group does not exist, you might have to check the integration test logs to verify

yuhuyoyo:
For non GCS library that throws GoogleJsonResponseException, I don't need to do any changes then because the current behavior will already trigger a retry when the cloud sync step fails.

yuhuyoyo:
how do i find the failed integration test? given it's a flaky test

yuhuyoyo:
nvm i found the failed nightly test  from 7 days ago. https://github.com/DataBiosphere/terra-workspace-manager/runs/4102096317?check_suite_focus=true

zloery:
> how do i find the failed integration test? given it's a flaky test

From the ticket it looks like the failure showed up during the 11/4 nightly test. We run tests using GHA, so you can see the test history in the "Actions" tab of this repo (filtering by "Nightly Test" gives you [this page](https://github.com/DataBiosphere/terra-workspace-manager/actions/workflows/nightly-tests.yml)). The run from 11/4 [is here](https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/1420200202), which gives you the client-side logs.
For server side logs, you can look at the `wsmtest` server logs at 11/4 around 3 AM EDT (when the tests run) in the GCP log explorer. Not sure how familiar you are with the logs explorer/the env setup, happy to grab some time to walk through that if it would be helpful!

yuhuyoyo:
thanks a lot for helping with me offline with a tour of the google cloud console, Zach!

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/GcsBucketCloudSyncStep.java
```diff
@@ -45,32 +48,41 @@ public StepResult doStep(FlightContext flightContext)
...
 
-    logger.info(
-        "Syncing workspace roles to GCP permissions on bucket {}", resource.getBucketName());
-    wsmSaStorageCow.setIamPolicy(resource.getBucketName(), updatedPolicyBuilder.build());
+      logger.info(
+          "Syncing workspace roles to GCP permissions on bucket {}", resource.getBucketName());
+
+      wsmSaStorageCow.setIamPolicy(resource.getBucketName(), updatedPolicyBuilder.build());
+    } catch (StorageException e) {
+      if (e.getCode() == HttpStatus.BAD_REQUEST.value()) {
```
zloery, line 81: Could you also add a comment explaining why we retry 400s here for the future? It's a little unexpected at first glance

zloery (approved):
LGTM

PR #491, opened 2021-11-12
[PF-1169] Fix tests which fail to clean up Sam workspaces

zloery:
Recently our tests started failing because our test users were in too many groups, which was mainly due to tests not cleaning up Sam workspaces after they finish. This change adds cleanup for all connected and integration tests which weren't previously cleaning up.

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
@@ -40,6 +40,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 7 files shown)


ddietterich (commented):
LGTM - thanks for the cleanup!

service/src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
```diff
@@ -309,18 +306,16 @@ void workspaceReaderIsNotPrivateResourceReader() throws Exception {
 
   @Test
   void duplicateResourceCreateIgnored() throws Exception {
-    UUID workspaceId = createWorkspaceDefaultUser();
-
     ControlledResource bucketResource = defaultBucket(workspaceId).build();
     samService.createControlledResource(bucketResource, null, null, defaultUserRequest());
     // This duplicate call should complete without throwing.
     samService.createControlledResource(bucketResource, null, null, defaultUserRequest());
+    // Delete the bucket so we can clean up the workspace.
```
jaycarlton, line 313: Can't you just delete the workspace and let it clean up the bucket and other resources?

jaycarlton (approved):
looks good

service/src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
zloery: Normally yes, but this test is a little weird because it's only creating the controlled resource in Sam, not WSM. I think it's helpful to verify this property of `SamService` here, but we do need to explicitly clean up the resource we create.

ddietterich (approved):
LGTM

PR #495, opened 2021-11-18
[PF-1206][PF-1194] Share heavyweight Client objects across requests, remove warnings

zloery:
This change refactors behavior for WSM's calls to downstream services so that all requests now share a common HTTP client object. Previously, we created a new client object for each call to a remote service. However per documentation, these clients are heavyweight objects (e.g. each `OkHttpClient` underneath a Sam `ApiClient` creates and manages its own threadpool), and this led to a large number of idle threads (peak of ~850 while running integration tests, or ~1700 while running integration tests + CLI tests). After this change, WSM peaked at ~130 threads while running integration + CLI tests, and did not increase as significantly under additional load.

I also fixed a few warnings that kept showing up in logs.


ddietterich (approved):
LGTM - I assigned PF-1194 to you, since you fixed it in this PR

jaycarlton (approved):
LGTM

PR #498, opened 2021-11-19
Create Azure Storage operation. 

giventocode:
This PR enables the creation of azure storage accounts via WSM.
- Contains basic unit testing of the create step.
- Connected tests are pending.
- Get and delete operations are pending.


ddietterich (commented):
Some comments. It does _look_ like there is a bug in one piece of logic.

service/src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -414,6 +422,25 @@ private ComputeManager buildComputeManager(
     return manager;
   }
 
+  private StorageManager buildStorageManager(
```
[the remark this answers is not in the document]
rtitle: I think this code will remain in WSM -- we are using the Azure Java SDK directly. The CRL piece to add is the hooks for logging/cleanup (via `*RequestData` objects passed in the request context..). 

I'm guessing we'll want a new CRL `azure-resourcemanager-storage` module (although it looks like `com.azure.resourcemanager.storage.*` is already pulled in via the `azure-resourcemanager-compute` dependency).

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureStorageStep.java
```diff
@@ -0,0 +1,114 @@
...
+                          .setName(resource.getStorageAccountName())
+                          .setRegion(Region.fromName(resource.getRegion()))
+                          .setResourceGroupName(azureCloudContext.getAzureResourceGroupId())
+                          .build()))
+              .id();
+
+      setStorageAccountIdCreatedInCurrentContext(context, storageAccountId);
+
+    } catch (ManagementException e) {
+      logger.error(
```
rtitle, line 65: Do we want to handle "Conflict" and return success; and retry otherwise like we do for [other Azure resources](https://github.com/DataBiosphere/terra-workspace-manager/blob/azure-poc/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureIpStep.java#L64-L77)?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureStorageStep.java
```diff
@@ -0,0 +1,114 @@
...
+
+    String storageAccountId = getStorageAccountIdCreatedInCurrentContext(context);
+    if (storageAccountId != null) {
+      try {
+        logger.warn("Attempting to delete storage account: {}", storageAccountId);
+        storageManager.storageAccounts().deleteById(storageAccountId);
+        logger.warn("Successfully deleted storage account: {}", storageAccountId);
+      } catch (ManagementException e) {
+        logger.error("Failed to delete storage account: {}", storageAccountId);
+        return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
```
rtitle, line 100: For other resources we return success if the resource is already deleted; and retry otherwise: https://github.com/DataBiosphere/terra-workspace-manager/blob/azure-poc/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureIpStep.java#L86-L100

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/GetAzureStorageStep.java
```diff
@@ -0,0 +1,64 @@
...
+    this.azureCloudContext = azureCloudContext;
+    this.crlService = crlService;
+    this.resource = resource;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    StorageManager storageManager = crlService.getStorageManager(azureCloudContext, azureConfig);
+
+    if (!storageManager
```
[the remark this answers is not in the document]
rtitle: +1

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -6,6 +6,7 @@
 import bio.terra.workspace.common.utils.ControllerUtils;
 import bio.terra.workspace.db.exception.InvalidMetadataException;
 import bio.terra.workspace.generated.controller.ControlledAzureResourceApi;
+import bio.terra.workspace.generated.model.*;
```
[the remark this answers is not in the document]
giventocode: Fixed.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/azure/resourcemanager/storage/StorageManagerOperation.java
```diff
@@ -0,0 +1,8 @@
+package bio.terra.workspace.service.resource.controlled.flight.create.azure.resourcemanager.storage;
```
rtitle, line 1: Ah ok - these following classes (`StorageManagerOperation`, `CreateStorageAccountRequestData`, etc) will need to move to CRL I think.

rtitle (commented):
Overall looks good, a few comments above. The unit tests look good; would be nice to add a connected test as well to [CreateControlledResourceFlightTest](https://github.com/DataBiosphere/terra-workspace-manager/blob/azure-poc/service/src/test/java/bio/terra/workspace/service/workspace/flight/create/azure/CreateControlledResourceFlightTest.java).

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureStorageStep.java
```diff
@@ -0,0 +1,114 @@
...
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  /**
+   * Deletes the storage account if the current flight created it. This method is not canonically
+   * idempotent as its behavior depends on the flight the context. This approach is intended to: -
+   * Avoid unwanted deletion if two or more concurrent requests pass the get step and one or more
+   * creation requests fail. - Enforce deletion of the storage account only if created successfully
+   * but the do step failed.
```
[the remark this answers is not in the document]
giventocode: I did more testing on this. The good news is that the latest version of Create operation, in the storage resource provider, handles concurrent requests gracefully (it is an idempotent implementation). If two create requests arrive with the same parameters concurrently, both succeed - albeit the return codes are different 200, 202.  This alleviates my concern of blinding deleting in undo.

So, I changed the implementation, and it no longer depends on the working map as a mechanism to determine when to delete. 

As a follow up, I would like to validate the expected behavior in a connected test.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureStorageStep.java
ddietterich: Consistency seems good

ddietterich (approved):
LGTM - Rob has one unaddressed comment you might take a look at.

rtitle (approved):
I did another pass, this LGTM overall.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateAzureStorageStep.java
giventocode: To close this comment. I see that other flights handle the "Conflict" status code as if the resource already exists. However, the assumption is not entirely accurate. "Conflict" is returned when an operation fails because the resource's state does not allow it, and specifics of what constitutes a "conflict" varies by resource. In this PR instead, I'm using the existence check in the SDK instead of handling the specific error. 

Going forward, we should revisit this for the other resources and align the approach with either using the SDK checks or querying the resource by ID to check if it exists explicitly.  Also, if the request call is re-issued because of failure in stairway the ARM API should handle it and return a success code if the request is exactly the same.

PR #499, opened 2021-11-22
[PF-1110] Fix test flakiness via wsm/cloud retries

zloery:
This change fixes integration test flakiness I accidentally introduced in #486. That change made creating referenced resources flaky because:
- Previously, test users had access to our test referenced resources directly, so they would always have access as soon as we checked for it.
- After #486, we sometimes check referenced resource access using the test users' pet SAs instead of their credentials directly, depending on whether their workspace has a GCP context or not.
- Pet SAs have access to the referenced resources via the proxy group. That means the access check on these cloud objects now depends on Google Group propagation, which can take up to 5 minutes. Until the pet SA's membership in the proxy group propagates through GCP, the pet SA will not have access to the referenced cloud object.

This fixes the flakiness by adding retries to some WSM and GCP calls in our integration tests to handle the propagation delay.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/CloneReferencedResources.java
+++ b/integration/src/main/java/scripts/testscripts/CloneReferencedResources.java
@@ -60,7 +60,7 @@ public class CloneReferencedResources extends DataRepoTestScriptBase {
             referencedGcpResourceApi, getWorkspaceId(), bucketReferenceName);
 
     sourceBigQueryDatasetReference =
-        ResourceMaker.makeBigQueryReference(
+        ResourceMaker.makeBigQueryDatasetReference(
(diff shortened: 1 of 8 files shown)


yuhuyoyo (approved):
thanks for the fix. I was just wondering why tests are flaky :)

ddietterich (approved):
LGTM - one thing to be careful of (that I ran into): if any of those ResourceMaker calls are used in negative tests where the caller is not supposed to be able to do it, then they will always retry. Since that retry goes for a loooong time, usually the test fails with a test-level timeout.

I checked and do not see that usage for those particular calls. Just something to be mindful of in the future.

zloery:
Good point! added comments to these methods warning about using these and expecting them to fail

PR #500, opened 2021-11-29
PF-1212: Use deleteReferenceResourceForResourceType instead of deleteReferenceResource

yuhuyoyo:


ddietterich (approved):
LGTM - thanks!

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
```diff
@@ -117,8 +117,8 @@ public void deleteReferenceResource(
   }
 
   /**
-   * Delete a reference. The only state we hold for a reference is in the metadata database so we
```
zloery, line 120: Are the `deleteReferenceResource` and `ResourceDao.deleteResource` still used anywhere else? if not, you could clean them up

zloery (approved):

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
yuhuyoyo: it is still used by workspaceapicontroller#deleteDataReference. This API is [deprecated](https://github.com/DataBiosphere/terra-workspace-manager/blob/16e3dbb90e70d6c4aba606fda146c6456c6300c4/service/src/main/resources/api/service_openapi.yaml#L385) but i'm not sure if there's still any service calling this endpoint.

zloery: ah, good point. I believe Rawls still calls those endpoints in CWB, so it's fine to leave that for now.

PR #502, opened 2021-12-01
[IA-2971] Delete azure subnet when network gets deleted

gpcarr:
Tested by creating a network and validating that when it gets deleted, both the `VirtualNetwork` and `NetworkSecurityGroup` resources are deleted in Azure


ddietterich (commented):
Code looks fine. I can't speak to the ordering of operations.

Qi77Qi (approved):
thanks @gpcarr

gpcarr:
> Code looks fine. I can't speak to the ordering of operations.

if we try to delete the subnet first, we get the following error:
```
Network security group /subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/networkSecurityGroups/%scannot be deleted because it is in use by the following resources: /subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/virtualNetworks/%s/subnets/%s. In order to delete the Network security group, remove the association with the resource(s).
```
Deleting the `Network` first removes the association between the two so that the subnet can be deleted

gpcarr:
@ddietterich is there any other feedback you have? I'll merge once I get a thumb from you :)

ddietterich (approved):
LGTM

PR #503, opened 2021-12-01
[PF-1236][PF-1236Make cloneWorkspace check auth, fix and run CloneWorkspace test

zloery:


integration/src/main/resources/suites/FullIntegration.json
```diff
@@ -8,6 +8,7 @@
     "integration/CloneBigQueryDataset.json",
     "integration/CloneGcsBucket.json",
     "integration/CloneReferencedResources.json",
+    "integration/CloneWorkspace.json",
```
yuhuyoyo, line 11: does "integration/*" work? So we won't miss testing any additional tests in the future?

ddietterich: No, it doesn't. And we don't want all of the tests all of the time. For example, right now the application tests won't work on the `wsmtest` environment.

ddietterich (approved):
LGTM

yuhuyoyo (approved):
thanks for quickly fixing this.

integration/src/main/resources/suites/FullIntegration.json
zloery: I did double-check through `FullIntegration` and the only other tests not included here are the two application tests introduced in #494, which Dan is intentionally not including for now, so at least we aren't missing other tests at the moment. I'd generally prefer to explicitly enable tests by adding them to the suite, though

PR #497, opened 2021-11-19
PF-1174: Add reference type bucket file

yuhuyoyo:
This is a draft. Starting the github actions.

PR #504, opened 2021-12-03
PF-1234: Add GCS object level access check support

yuhuyoyo:
Also rename Gcs bucket file to object because that's the term cloud storage uses to refer to things in the bucket.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/ValidateReferencedResources.java
+++ b/integration/src/main/java/scripts/testscripts/ValidateReferencedResources.java
@@ -15,8 +15,8 @@ import bio.terra.workspace.client.ApiClient;
 import bio.terra.workspace.model.DataRepoSnapshotResource;
 import bio.terra.workspace.model.GcpBigQueryDataTableResource;
 import bio.terra.workspace.model.GcpBigQueryDatasetResource;
-import bio.terra.workspace.model.GcpGcsBucketFileResource;
 import bio.terra.workspace.model.GcpGcsBucketResource;
+import bio.terra.workspace.model.GcpGcsObjectResource;
 import bio.terra.workspace.model.GrantRoleRequestBody;
 import bio.terra.workspace.model.IamRole;
 import java.util.List;
@@ -39,7 +39,7 @@ public class ValidateReferencedResources extends DataRepoTestScriptBase {
   private UUID snapshotResourceId;
   // resource id of the reference to
   // gs://terra_wsm_fine_grained_test_bucket/foo/monkey_sees_monkey_dos.txt
-  private UUID bucketFileResourceId;
+  private UUID bucketTxtFileResourceId;
   // resource id of reference to gs://terra_wsm_fine_grained_test_bucket/foo/
   private UUID fooFolderResourceId;
   // resource id of reference to gs://terra_wsm_fine_grained_test_bucket/foo/**.txt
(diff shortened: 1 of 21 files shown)


service/src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -355,6 +356,25 @@ public boolean canReadGcsBucket(String bucketName, AuthenticatedUserRequest user
...
+      // If successfully get the blob, the user have at least READER access.
+      storage.get(BlobId.of(bucketName, objectName));
+      return true;
+    } catch (StorageException e) {
+      if (e.getCode() == HttpStatus.SC_FORBIDDEN) {
+        return false;
+      }
+      throw new InvalidReferenceException(
+          String.format(
+              "Error while trying to access GCS blob %s in bucket %s and status code %s",
```
zloery, line 372: You have three `%s` tokens but only pass two strings to `String.format` here

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedGcsObjectAttributes.java
```diff
@@ -3,22 +3,23 @@
...
   private final String bucketName;
-  private final String fileName;
+  private final String objectName;
 
   @JsonCreator
-  public ReferencedGcsBucketFileAttributes(
-      @JsonProperty("bucketName") String bucketName, @JsonProperty("fileName") String fileName) {
+  public ReferencedGcsObjectAttributes(
+      @JsonProperty("bucketName") String bucketName,
+      @JsonProperty("objectName") String objectName) {
```
zloery, line 13: I think changing this field from `fileName` to `objectName` will break deserialization of any GCS object references which already exist in WSM's DB. 
That might be ok since you added these pretty recently (#497), but it's worth checking existing environments.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -607,22 +607,22 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
-  /api/workspaces/v1/{workspaceId}/resources/referenced/gcp/bucket/files:
```
zloery, line 610: This is also a breaking API change, though I see the CLI PR for GCS objects is still pending and the UI doesn't use this endpoint yet, so it's probably ok.

integration/src/main/java/scripts/testscripts/ValidateReferencedResources.java
```diff
@@ -25,21 +28,34 @@
 
 public class ValidateReferencedResources extends DataRepoTestScriptBase {
 
+  // liam.dragonmaw@test.firecloud.org
   private TestUserSpecification secondUser;
+  // elijah.thunderlord@test.firecloud.org
+  private TestUserSpecification thirdUser;
```
zloery, line 34: Commenting test user names here is a little misleading since that's specified in the test configuration file and liable to change.

To make things easier to follow, can you name these test users and their API client objects by function (e.g. test user objects like `fileReader` and `noAccessUser`, client objects like `fileReaderApi`, `noAccessUserApi`)

integration/src/main/java/scripts/utils/ResourceMaker.java
```diff
@@ -152,6 +153,41 @@ public static GcpGcsBucketResource makeGcsBucketReference(
...
+      UUID workspaceId,
+      String name,
+      @Nullable CloningInstructionsEnum cloningInstructions)
+      throws InterruptedException, ApiException {
+    return makeGcsBucketReference(
+        resourceApi,
+        workspaceId,
+        name,
+        cloningInstructions,
+        /*isBucketWithFineGrainedAccess=*/ false);
```
zloery, line 175: should this be true?

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedGcsObjectAttributes.java
yuhuyoyo: there is no calling of this endpoint yet. so i think it should be OK?

service/src/main/resources/api/service_openapi.yaml
yuhuyoyo: yeah i'm planning to wait till this one goes in to update the CLI relevant PR.

ddietterich (approved):
LGTM - this would be an incompatible change, but since the previous state is not in use, go for it!

zloery (approved):

PR #506, opened 2021-12-07
PF-1177: Support updating reference target

yuhuyoyo:


service/src/main/java/bio/terra/workspace/db/DbUtils.java
```diff
@@ -22,7 +22,11 @@ public static String setColumnsClause(MapSqlParameterSource columnParams) {
       if (i > 0) {
         sb.append(", ");
       }
-      sb.append(columnName).append(" = :").append(columnName);
+      if (!columnName.equals("attributes")) {
```
ddietterich, line 25: Having the utility method have knowledge of specific column names seems fragile.
How about making a call form like
```
setColumnClause(MapSqlParameterSource columnParams, String... jsonColumns)
```
So the caller determines the column names to cast?

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -362,24 +365,70 @@ public void createReferenceResource(ReferencedResource resource)
     storeResource(resource);
   }
 
+  @WriteTransaction
+  private boolean updateResource(
```
ddietterich, line 369: You cannot put transaction control on a `private` method. The way it works is Spring inserts a layer between the DAO caller and the DAO. That layer processes the annotation and starts transactions.

The general guideline for our DAOs is to put the transaction control on public interface methods. If there are shared methods, make them internal and private. So you would not have one public transactional method call another public transaction method. If that doesn't make sense, we can talk 1:1.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -362,24 +365,70 @@ public void createReferenceResource(ReferencedResource resource)
...
+    var params = new MapSqlParameterSource();
+
+    updateParams.forEach(
+        (k, v) -> {
+          if (!StringUtils.isEmpty(k) && !StringUtils.isEmpty(v)) {
+            params.addValue(k, v);
+          }
+        });
+
+    return updateResourceColumns(workspaceId, resourceId, params);
```
ddietterich, line 381: Seems like you can collapse `updateResourceColumns` and this method into a single method that populates the params and does the update. Looks like this is the only caller.

In fact, you could also pull `getUpdateParams` into this method, making a single worker method for the two forms of `updateResource`.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -362,24 +365,70 @@ public void createReferenceResource(ReferencedResource resource)
...
-      params.addValue("description", description);
-    }
+    return updateResource(workspaceId, resourceId, updateParams);
+  }
 
-    return updateResourceColumns(workspaceId, resourceId, params);
+  /** Gets a map of the params that needs to be updated. */
+  private static Map<String, String> getUpdateParams(
+      @Nullable String name, @Nullable String description, @Nullable String attributes) {
+    Map<String, String> updateParams = new HashMap<>();
```
ddietterich, line 427: minor: given that there are only three things, it seems like a lot of effort to build a map that tells you how to build another map.

service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/update/UpdateReferenceResourceFlight.java
```diff
@@ -0,0 +1,39 @@
...
+    final ReferencedResource resource =
+        inputParameters.get(JobMapKeys.REQUEST.getKeyName(), ReferencedResource.class);
+
+    // Perform access verification
+    addStep(new ValidateReferenceStep(appContext), RetryRules.cloud());
+
+    addStep(
+        new RetrieveReferenceMetadataStep(
+            appContext.getResourceDao(), resource.getWorkspaceId(), resource.getResourceId()),
+        RetryRules.shortDatabase());
```
ddietterich, line 33: nit: the methods like `RetryRules.shortDatabase()` creates a RetryRule object. The same object can be used for more than one step in a flight (but cannot be used between flights). So you could create one shortDatabase retry rule and use it in both this step and the next one. It is a minor efficiency.

ddietterich (changes requested):
This looks good overall. I requested some changes in the DAO code. Otherwise, minor comments.

service/src/main/java/bio/terra/workspace/db/DbUtils.java
yuhuyoyo: not sure if i did what you mean. Could you take a look at the new implementation?

service/src/main/java/bio/terra/workspace/db/DbUtils.java
```diff
@@ -10,20 +13,33 @@
...
     }
+    Set<String> jsonColumnSet = new HashSet(Arrays.asList(jsonColumns));
+    for (String jsonColumn : jsonColumns) {
+      sb.append(jsonColumn).append(" = cast(:").append(jsonColumn).append(" AS jsonb)");
+    }
     for (int i = 0; i < parameterNames.length; i++) {
       String columnName = parameterNames[i];
+      // If the param needs to be cast as jsonb, it has already been appended to sb so we can
+      // skip here.
+      if (jsonColumnSet.contains(columnName)) {
```
ddietterich, line 34: I would have done
```
if (jsonColumnSet.contains(columnName) {
  sb.append(jsonColumn).append(" = cast(:").append(jsonColumn).append(" AS jsonb)");
} else {
  sb.append(columnName).append(" = :").append(columnName);
}
```
That is, the incoming columns are _possible_ json columns, but if they don't appear in the parameter source, then they have no effect. Then in the caller, you don't have to check if you are passing `attributes` and use different call forms.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -362,24 +363,75 @@ public void createReferenceResource(ReferencedResource resource)
...
+
+    if (params.hasValue("attributes")) {
+      sb.append(DbUtils.setColumnsClause(params, "attributes"));
+    } else {
+      sb.append(DbUtils.setColumnsClause(params));
+    }
+
+    sb.append(" WHERE workspace_id = :workspace_id AND resource_id = :resource_id");
+
+    MapSqlParameterSource queryParams = new MapSqlParameterSource();
```
ddietterich, line 407: I think you can just continue to append values to params here.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -362,24 +363,75 @@ public void createReferenceResource(ReferencedResource resource)
...
+  /**
+   * Update name and/or description of the resource.
+   *
+   * @param name name of the resource, may be null if it does not need to be updated
+   * @param description description of the resource, may be null if it does not need to be updated
+   */
+  @WriteTransaction
+  public boolean updateResource(
+      UUID workspaceId, UUID resourceId, @Nullable String name, @Nullable String description) {
+    return updateResource(workspaceId, resourceId, name, description, null);
```
ddietterich, line 434: This is a form I would like us to avoid: where one public method with a transaction annotation calls another public method with a transaction annotation. It is not technically wrong. It is so we always know where transactions are getting started. It mostly helps in debugging.

The fix would be to make a `updateResourceWorker` method that is a private method. It would take the same param list as the first `updateResource`. Then the two public methods would call the worker. One with attributes and the other with null. Does that make sense?

ddietterich (commented):
Thanks for this! I have two more tweaks to suggest.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
yuhuyoyo: good catch.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -362,49 +363,43 @@ public void createReferenceResource(ReferencedResource resource)
...
-   * @param workspaceId workspace identifier - not strictly necessarily, but an extra validation
-   * @param resourceId resource identifier
-   */
-  private boolean updateResourceColumns(
-      UUID workspaceId, UUID resourceId, MapSqlParameterSource columnParams) {
     StringBuilder sb = new StringBuilder("UPDATE resource SET ");
 
-    sb.append(DbUtils.setColumnsClause(columnParams));
+    if (params.hasValue("attributes")) {
+      sb.append(DbUtils.setColumnsClause(params, "attributes"));
```
ddietterich, line 391: You dont' need this conditional anymore.

ddietterich (approved):
LGTM - one minor thing

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
yuhuyoyo: Done. also added more testings for DbUtils.

integration/src/main/java/scripts/testscripts/UpdateReferenceResources.java
```diff
@@ -0,0 +1,412 @@
...
+    super.doCleanup(testUsers, workspaceApi);
+  }
+
+  @Override
+  public void setParameters(List<String> parameters) throws Exception {
+    // TODO: Refactor this function when TestRunner starts supporting parameterMap
+    super.setParameters(parameters);
+    if (parameters == null || parameters.size() < 4) {
+      throw new IllegalArgumentException(
+          "Must provide Spend Profile ID, 2 Data Repo snapshot IDs, and 2 Data Repo Instance Names in the parameters list");
```
zloery, line 115: only one TDR instance name is in the parameter list

zloery (approved):

PR #508, opened 2021-12-09
Fix wsm schema broken in last commit.

cbookg:
responses was incorrectly nested under requestBody.


yuhuyoyo (approved):

ddietterich (approved):
LGTM - how did this get by our tests?

ddietterich:
I added PF-1243 to track the question and fix for why we didn't detect this with our testing.

cbookg:
> 

swagger schema validation isn't run on presubmit.

PR #510, opened 2021-12-09
PF-1244: fix a typo to make sure that the attributes field is required as intended

yuhuyoyo:
GcpBigQueryDatasetResource should have the attributes field set as required.


marikomedlock (commented):
Change lgtm.

Also I think this object is only used in response objects and so we don't expect any new validation errors in clients as a result of this change. Is that correct?

yuhuyoyo:
> 

right. it's only used in response.

yuhuyoyo:
oops made change on the wrong branch. please do not review

ddietterich (approved):
LGTM

PR #511, opened 2021-12-09
[PF-1149] Add errorOnUnknownEnum flag to swagger

zloery:
This change means WSM will reject requests with invalid enum values with 400 status codes instead of accepting the invalid values and silently coercing them to `null`.


yonghaoy (approved):

ddietterich (approved):
LGTM - this is really a great find!

PR #513, opened 2021-12-10
[PF-1245] Add retries to direct Sam calls in integration tests

zloery:
Currently, all Sam calls from the WSM server are retried on certain failures (generally request timeouts). However, there are a few direct calls inside WSM integration tests which aren't retried and occasionally fail (mostly in `EnablePet`) due to transient issues, which leads to test flakiness.

I've intentionally pulled in SamRetry from the Terra Common Library here instead of re-using the generic `ClientTestUtils.getWithRetryOnException` so that we retry a smaller set of errors instead of also retrying legitimate failures from Sam.


melissachang (approved):

ddietterich (approved):
LGTM

PR #514, opened 2021-12-10
[PF-1197] support workspace key-value properties

ddietterich:
This turned out to be a bigger change than you might expect for a few reasons:
- Rather than add to it, I removed WorkspaceRequest. It was used for a single call. I replaced it with the Workspace object that has exactly the same fields. That mostly changed tests because of different behavior of getters. I updated the Workspace build, moving the defaulting of null display name and description in there.
- I replaced the many individual input parameters to the CreateWorkspaceFlight with the Workspace object as the request.
- That revealed that SpendProfileId (an `@AutoValue` class) would not JSON serialize properly. So I de-AutoValued it to make it work.
- I added and debugged some tests that create, clone, and update with key-value properties. It turns out that the SQL for properties column handling was broken, but since we never tried to update the properties field before we never noticed.

The semantics for the key-value properties work just like description or display name; that is:
- An update with null property array is ignored. Otherwise, the property list is a complete replacement
- A clone workspace takes a brand new set of properties

And for my trouble, I won the pulling on a thread award! YAY!!


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2271,6 +2294,9 @@ components:
         description:
           description: A description of the workspace
           type: string
+        properties:
```
zloery, line 2297: Outside scope for this PR, but I wonder if we'll eventually need "add/delete/modify properties" endpoints instead overwriting existing ones.
OTOH callers can just read-modify-write here which should be good for now, and we shouldn't add APIs until we need them

zloery (approved):

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
```diff
@@ -79,6 +82,21 @@ public static void validateRequiredEntries(FlightMap flightMap, String... keys)
     }
   }
 
+  /**
+   * Validate that all common entries are present in the flight map
+   *
+   * @param flightMap input parameters
+   */
+  public static void validateCommonEntries(FlightMap flightMap) {
+    validateRequiredEntries(
```
yuhuyoyo, line 91: can we have a static field of COMMON_ENTRIES_LIST to be shared between the two methods?

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -151,44 +154,38 @@ public Workspace getWorkspace(UUID id) {
   }
 
   @WriteTransaction
-  public boolean updateWorkspace(UUID workspaceId, String name, String description) {
-    if (name == null && description == null) {
-      throw new MissingRequiredFieldException("Must specify name or description to update.");
+  public boolean updateWorkspace(
+      UUID workspaceId, String name, String description, Map<String, String> propertyMap) {
```
yuhuyoyo, line 158: could there be a peropertyMap is not null but empty situation?

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -151,44 +154,38 @@ public Workspace getWorkspace(UUID id) {
...
-   * parameter list making the param name equal to the column name you want to update. The method
-   * generates the column_name = :column_name list. It is an error if the params map is empty.
-   *
-   * @param workspaceId workspace identifier - not strictly necessarily, but an extra validation
-   * @param columnParams sql parameters
-   */
-  private boolean updateWorkspaceColumns(UUID workspaceId, MapSqlParameterSource columnParams) {
-    StringBuilder sb = new StringBuilder("UPDATE workspace SET ");
-
-    sb.append(DbUtils.setColumnsClause(columnParams));
```
yuhuyoyo, line 183: why not use DbUtils.setColumnsClause for adding the params to the columnSet?

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
yuhuyoyo: e,g, some fields are invalid and get filtered

yuhuyoyo: also annotate with nullable

yuhuyoyo (approved):
sorry for the delay. I didn't realize I'm reviewing this.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -536,4 +563,16 @@ private ApiCloneWorkspaceResult fetchCloneWorkspaceResult(
         .errorReport(jobResult.getApiErrorReport())
         .workspace(jobResult.getResult());
   }
+
+  // Convert properties list into a map
+  private Map<String, String> propertyMapFromApi(ApiProperties properties) {
+    Map<String, String> propertyMap = new HashMap<>();
+    if (properties != null) {
+      for (ApiProperty property : properties) {
+        ControllerValidationUtils.validatePropertyKey(property.getKey());
```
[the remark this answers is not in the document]
ddietterich: I couldn't think of a useful validation.

service/src/main/resources/api/service_openapi.yaml
ddietterich: As I understand it, today these tend to be set as descriptions and not frequently changed. As you say, we'll see what the usage is.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
ddietterich: Good idea

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
ddietterich: It is not invalid to have an empty property map. If you wanted to remove all of the properties, that is what you would provide.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
ddietterich: You fix that allowed json properties wasn't in yet. I'll rebase and use that method.

yuhuyoyo: I think it is in https://github.com/DataBiosphere/terra-workspace-manager/blob/6ed5226ae3bbe21363a98b2159bc35d74b0666ad/service/src/main/java/bio/terra/workspace/db/DbUtils.java#L33

PR #515, opened 2021-12-10
PF-1201: Generate sensible resrouce id when it is not specified for Controlled resources

yuhuyoyo:
Currently, user needs to specify both resource name and the resource-type specific name (e.g. datasetId, bucket name, etc).

Instead, we can just generate a default name if it's not specified.
To create a BQ dataset, one can say
terra resource create bq-dataset --name=foo
WSM will then create a bq-dataset of resource name foo and datasetId foo given "foo" is a valid dataset name as well. If the resource name is not a valid datasetid, we will generate a unique dataset id terra_uuid_dataset.

For create a GCS bucket, we will create a bucket of name terra_uuid_bucket

For AiNotebook, we will also reuse the resource name if it is a valid ai notebook name. Otherwise, we will generate a unique ai notebook name.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
@@ -40,7 +40,6 @@ import bio.terra.workspace.service.iam.model.ControlledResourceIamRole;
 import bio.terra.workspace.service.iam.model.SamConstants;
 import bio.terra.workspace.service.job.JobService;
 import bio.terra.workspace.service.job.JobService.AsyncJobResult;
-import bio.terra.workspace.service.resource.ValidationUtils;
 import bio.terra.workspace.service.resource.WsmResourceType;
 import bio.terra.workspace.service.resource.controlled.AccessScopeType;
 import bio.terra.workspace.service.resource.controlled.ControlledAiNotebookInstanceResource;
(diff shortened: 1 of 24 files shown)


ddietterich (approved):
LGTM - a style suggestion

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
```diff
@@ -146,6 +146,10 @@ public int hashCode() {
     return result;
   }
 
+  public static String generateUniqueInstanceId() {
+    return String.format("terra-%s-notebook", UUID.randomUUID());
```
zloery, line 150: Why do notebooks need random IDs?

Agree this is necessary for buckets (which are globally namespaced), but my understanding is that you could just use the resource name here.

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -345,7 +344,10 @@ private ApiCloneControlledGcpGcsBucketResult fetchCloneGcsBucketResult(
             .accessScope(AccessScopeType.fromApi(body.getCommon().getAccessScope()))
             .managedBy(managedBy)
             .applicationId(controlledResourceService.getAssociatedApp(managedBy, userRequest))
-            .datasetName(body.getDataset().getDatasetId())
+            .datasetName(
+                body.getDataset().getDatasetId() == null
```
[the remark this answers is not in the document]
yuhuyoyo: nice!

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
yuhuyoyo: ah, you are right. i also realize that the resource name is not always a valid instance id. so I made some change.

zloery (approved):

PR #516, opened 2021-12-10
[PF-1246] Force update log4j transitive dependency

zloery:
Force transitive log4j api and slf4j library dependencies to the latest version due to security concerns. We do not use log4j core, so nothing to update there.


ddietterich (approved):
LGTM - I'm interested in the answer to @yonghaoy question

ungwudik (approved):

zarsky-broad:
See https://github.com/DataBiosphere/terra-external-credentials-manager/pull/84 and https://docs.spring.io/spring-boot/docs/2.6.1/gradle-plugin/reference/htmlsingle/#managing-dependencies.dependency-management-plugin.customizing

PR #517, opened 2021-12-10
[PF-1222] Make JobBuilder fully fluent

ddietterich:
The original WSM JobBuilder was taken from TDR where all of the constructor parameters are required. That didn't match our usage, where we often do not supply all of the inputs. Further, we were minting job ids in many different places.

This rewrites the JobBuilder in a fully fluent style, so we only set what we use. It takes charge of all validation, defaulting, and filling in of the MdcHook and TracingHook parameters. Those tasks were previously split between JobService and JobBuilder.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
+++ b/service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
@@ -1,55 +1,72 @@
 package bio.terra.workspace.service.job;
 
+import bio.terra.common.exception.MissingRequiredFieldException;
+import bio.terra.common.stairway.StairwayComponent;
+import bio.terra.common.stairway.TracingHook;
 import bio.terra.stairway.Flight;
 import bio.terra.stairway.FlightMap;
+import bio.terra.workspace.common.utils.MdcHook;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
+import bio.terra.workspace.service.job.exception.InvalidJobIdException;
 import bio.terra.workspace.service.job.exception.InvalidJobParameterException;
 import io.opencensus.contrib.spring.aop.Traced;
+import org.apache.commons.lang3.StringUtils;
 
 public class JobBuilder {
-
-  private final JobService jobServiceRef;
-  private final Class<? extends Flight> flightClass;
+  private final JobService jobService;
+  private final StairwayComponent stairwayComponent;
+  private final MdcHook mdcHook;
   private final FlightMap jobParameterMap;
-  private final String jobId;
+  private Class<? extends Flight> flightClass;
+  private String jobId;
+  private String description;
+  private Object request;
+  private AuthenticatedUserRequest userRequest;
 
-  // constructor only takes required parameters
-  public JobBuilder(
-      String description,
-      String jobId,
-      Class<? extends Flight> flightClass,
-      Object request,
-      AuthenticatedUserRequest userRequest,
-      JobService jobServiceRef) {
-    this.jobServiceRef = jobServiceRef;
-    this.flightClass = flightClass;
-    this.jobId = jobId;
-
-    // initialize with required parameters
+  public JobBuilder(JobService jobService, StairwayComponent stairwayComponent, MdcHook mdcHook) {
+    this.jobService = jobService;
+    this.stairwayComponent = stairwayComponent;
+    this.mdcHook = mdcHook;
     this.jobParameterMap = new FlightMap();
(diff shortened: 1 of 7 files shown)


zloery (approved):
LGTM, thanks for this!

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -1,55 +1,72 @@
...
+    this.mdcHook = mdcHook;
+    this.jobParameterMap = new FlightMap();
+  }
+
+  public JobBuilder flightClass(Class<? extends Flight> flightClass) {
     this.flightClass = flightClass;
+    return this;
+  }
+
+  public JobBuilder jobId(String jobId) {
```
yuhuyoyo, line 38: annotate with @Nullable.

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -1,55 +1,72 @@
...
+  }
+
+  public JobBuilder userRequest(AuthenticatedUserRequest userRequest) {
+    this.userRequest = userRequest;
+    return this;
   }
 
-  // use addParameter method for optional parameter
-  // returns the JobBuilder object to allow method chaining
   public JobBuilder addParameter(String keyName, Object val) {
```
yuhuyoyo, line 64: annotate nullable params.

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -1,55 +1,72 @@
...
     if (keyName == null) {
```
yuhuyoyo, line 65: do we need to check if val is nullable?

Also StringUtils.isBlank(keyName) instead?

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -70,6 +88,41 @@ public String submit() {
    */
   @Traced
   public <T> T submitAndWait(Class<T> resultClass) {
-    return jobServiceRef.submitAndWait(flightClass, jobParameterMap, resultClass, jobId);
+    validateAndDefault();
+    return jobService.submitAndWait(flightClass, jobParameterMap, resultClass, jobId);
+  }
+
+  // Check the inputs, supply defaults and finalize the input parameter map
+  private void validateAndDefault() {
```
yuhuyoyo, line 96: what do you think we rename this method to populateInputParams? there's not really any validation

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -70,6 +88,41 @@ public String submit() {
...
+    if (jobId == null) {
+      jobId = stairwayComponent.get().createFlightId();
+    }
+
+    // Always add the MDC logging and tracing span parameters for the mdc hook
+    addParameter(MdcHook.MDC_FLIGHT_MAP_KEY, mdcHook.getSerializedCurrentContext());
+    addParameter(
+        TracingHook.SUBMISSION_SPAN_CONTEXT_MAP_KEY, TracingHook.serializeCurrentTracingContext());
+
+    // Convert the any other members that were set into parameters. However, if they wer
```
yuhuyoyo, line 111: nit: if they were

service/src/test/java/bio/terra/workspace/service/job/JobServiceTest.java
```diff
@@ -64,8 +64,12 @@ void emptyStringJobIdTest() {
     assertThrows(
```
yuhuyoyo, line 64: can we add a test for nullJobId?

ddietterich: I modified an existing test that was making UUIDs to skip specifying the id, so that path gets exercised.

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -1,55 +1,72 @@
...
+
+  public JobBuilder flightClass(Class<? extends Flight> flightClass) {
     this.flightClass = flightClass;
+    return this;
+  }
+
+  public JobBuilder jobId(String jobId) {
+    if (jobId != null) {
+      // If clients provide a non-null job ID, it cannot be whitespace-only
+      if (StringUtils.isWhitespace(jobId)) {
```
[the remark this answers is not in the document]
ddietterich: Right, thanks

yuhuyoyo (approved):

PR #525, opened 2022-01-07
[PF-1276] Generate instance id based on user name and date

yuhuyoyo:
This logic is copied from CLI to keep the auto-gen logic consistent.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
@@ -157,8 +157,8 @@ public class ControlledAiNotebookInstanceResource extends ControlledResource {
   /**
    * Best effort mangle the user's name so that it meets the requirements for a valid instance name.
    *
-   * <p>Instance name id must match the regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', i.e. starting
-   * with a lowercase alpha character, only alphanumerics and '-' of max length 61. I don't have a
+   * <p>Instance name id must match the regex '(?:[a-z](?:[-a-z0-9]{0,63}[a-z0-9])?)', i.e. starting
+   * with a lowercase alpha character, only alphanumerics and '-' of max length 63. I don't have a
    * documentation link, but gcloud will complain otherwise.
    */
   private static String mangleUsername(String username) {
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
```diff
@@ -134,12 +140,50 @@ public void validate() {
...
+   * with a lowercase alpha character, only alphanumerics and '-' of max length 61. I don't have a
+   * documentation link, but gcloud will complain otherwise.
+   */
+  private static String mangleUsername(String username) {
+    // Strip non alpha-numeric or '-' characters.
+    String mangledName = username.replaceAll("[^a-zA-Z0-9-]", "");
+    if (mangledName.isEmpty()) {
+      mangledName = "notebook";
+    }
+    // Lower case everything, even though only the first character requires lowercase.
```
zloery, line 170: From the tooltip, I think everything has to be lowercase

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
```diff
@@ -11,15 +11,21 @@
...
+import java.time.format.DateTimeFormatter;
 import java.util.Optional;
 import java.util.UUID;
 import javax.annotation.Nullable;
+import org.apache.commons.lang3.StringUtils;
 
 /** A {@link ControlledResource} for a Google AI Platform Notebook instance. */
 public class ControlledAiNotebookInstanceResource extends ControlledResource {
+
+  protected static final int MAX_INSTANCE_NAME_LENGTH = 61;
```
zloery, line 27: per below, I think the max length is actually 63

zloery (commented):
Echoing Mariko's comments from the other PR: it's probably worth looping in someone from solutions, especially if anything changes in the actual naming scheme (e.g.  61 chars -> 63). Probably just an FYI/no-op, but better than surprising them

yuhuyoyo:
@deflaux FYI. Moving the instance id autogen logic from CLI to WSM. The format will still be <username>-<date> with the small change of allowing 63 character instead of 61.

deflaux:
> @deflaux FYI. Moving the instance id autogen logic from CLI to WSM. The format will still be - with the small change of allowing 63 character instead of 61.

This one sentence description of the change SGTM. I did not review the changes to the code. Thanks!!

zloery (approved):

ddietterich (approved):
LGTM

PR #527, opened 2022-01-12
[PF-1099] Add alpha version of list jobs in workspace

ddietterich:
I added an endpoint for listing jobs running or that have run in a workspace. The endpoint provides a set of query parameters for filtering what jobs to return.

I invented a new scheme for developing endpoints:
- We use the version `alpha1` instead of `v1` to identify the endpoint as under development. No compatibility promises.
- Provide Alpha1Controller to handle the endpoints
- Provide Alpha1Service to do the processing

We use the Stairway filtering to implement the filters we provide. Stairway can filter on its own flight attributes (like submit_time) and input parameters. So in order to get this to work, I had to pass input parameters to flights solely for the purpose of filtering. I updated the `JobBuilder` with fluent setters for those additional parameters. So many of the code changes are not about job enumeration, but are about putting the right information into the input parameters to support the enumeration.


service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
```diff
@@ -0,0 +1,238 @@
...
+            .pageToken(enumeratedJobs.getPageToken())
+            .totalResults(enumeratedJobs.getTotalResults())
+            .results(apiJobList);
+
+    return new ResponseEntity<>(result, HttpStatus.OK);
+  }
+
+  // Convert a WsmResource into the API format for enumeration
+  @VisibleForTesting
+  public ApiResourceUnion apiResourceFromWsmResource(WsmResource wsmResource, String gcpProjectId) {
```
yuhuyoyo, line 145: please annotate gcpProjectId as @Nullable.

service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
```diff
@@ -0,0 +1,238 @@
...
+                "Unimplemented controlled resource type: " + wsmResource.getResourceType());
+
+          default:
+            throw new InternalLogicException(
+                "Unknown controlled resource type: " + wsmResource.getResourceType());
+        }
+        break; // controlled
+
+      default:
+        throw new InternalLogicException("Unknown stewardship type");
```
yuhuyoyo, line 229: nit: include the stewardship type in the error message.

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -119,6 +163,28 @@ private void populateInputParams() {
...
+    if (shouldInsert(ResourceKeys.STEWARDSHIP_TYPE, stewardshipType)) {
+      addParameter(ResourceKeys.STEWARDSHIP_TYPE, stewardshipType);
+    }
+    if (shouldInsert(ResourceKeys.OPERATION_TYPE, operationType)) {
+      addParameter(ResourceKeys.OPERATION_TYPE, operationType);
+    }
+  }
+
+  private boolean shouldInsert(String mapKey, Object value) {
+    return (value != null && !jobParameterMap.containsKey(mapKey));
```
yuhuyoyo, line 187: should we handle empty/blank string cases for value?

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -119,6 +163,28 @@ private void populateInputParams() {
...
+    }
+    if (shouldInsert(ResourceKeys.STEWARDSHIP_TYPE, stewardshipType)) {
+      addParameter(ResourceKeys.STEWARDSHIP_TYPE, stewardshipType);
+    }
+    if (shouldInsert(ResourceKeys.OPERATION_TYPE, operationType)) {
+      addParameter(ResourceKeys.OPERATION_TYPE, operationType);
+    }
+  }
+
+  private boolean shouldInsert(String mapKey, Object value) {
```
yuhuyoyo, line 186: annotate value with @Nullable

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -241,6 +241,14 @@ public static void validateResourceName(String name) {
     }
   }
 
+  public static void validateOptionalResourceName(String name) {
```
yuhuyoyo, line 244: annotate name with @Nullable

yuhuyoyo: why is it called validateOptionalResourceName but still throw exception when the resource name is null?

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -203,8 +205,13 @@ public ControlledGcsBucketResource updateGcsBucket(
             .newJob()
             .description(jobDescription)
             .flightClass(UpdateControlledGcsBucketResourceFlight.class)
-            .request(resource)
+            .resource(resource)
             .userRequest(userRequest)
+            .operationType(OperationType.UPDATE)
+            .workspaceId(resource.getWorkspaceId().toString())
+            // TODO: conflict with resource name
```
yuhuyoyo, line 212: Please add ticket to TODO and specify the action for the TODO.

service/src/main/java/bio/terra/workspace/service/workspace/model/JobStateFilter.java
```diff
@@ -0,0 +1,31 @@
+package bio.terra.workspace.service.workspace.model;
+
+import bio.terra.workspace.generated.model.ApiJobStateFilter;
+
+public enum JobStateFilter {
+  ALL(ApiJobStateFilter.ALL),
```
yuhuyoyo, line 6: what does ALL mean?

service/src/main/java/bio/terra/workspace/service/workspace/model/EnumeratedJob.java
```diff
@@ -0,0 +1,48 @@
...
+
+import bio.terra.stairway.FlightState;
+import bio.terra.workspace.service.resource.WsmResource;
+import java.util.Optional;
+
+public class EnumeratedJob {
+  private FlightState flightState;
+  private String jobDescription;
+  private OperationType operationType;
+  private WsmResource resource;
```
yuhuyoyo, line 11: please annotate nullable field.

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
```diff
@@ -72,6 +70,11 @@ private ControlledResourceKeys() {}
     public static final String RESOURCE_DESCRIPTION = "resourceDescription";
     public static final String PREVIOUS_RESOURCE_DESCRIPTION = "previousResourceDescription";
     public static final String PREVIOUS_ATTRIBUTES = "previousAttributes";
+    // additional keys for job filtering
+    public static final String STEWARDSHIP_TYPE = "stewardshipType";
+    public static final String RESOURCE = "resource";
```
yuhuyoyo, line 75: would it be clearer to name it to resourceType?

service/src/main/java/bio/terra/workspace/service/workspace/model/EnumeratedJobs.java
```diff
@@ -0,0 +1,36 @@
+package bio.terra.workspace.service.workspace.model;
+
```
yuhuyoyo, line 2: this class feels like a good candidate for java autovalue. https://www.baeldung.com/introduction-to-autovalue

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
```diff
@@ -72,6 +70,11 @@ private ControlledResourceKeys() {}
...
+    // this applies outside of resources. Where should it go?
+    public static final String OPERATION_TYPE = "operationType";
```
zloery, line 77: Maybe move this to top-level `WorkspaceFlightMapKeys`? It's also used for things like "create workspace"

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -4267,6 +4333,64 @@ components:
       type: string
       enum: ['REFERENCED', 'CONTROLLED']
 
+    ResourceUnion:
```
zloery, line 4336: Mark these as `alpha1` for now?

integration/src/main/java/scripts/testscripts/EnumerateJobs.java
```diff
@@ -0,0 +1,198 @@
...
+
+    logger.info("Created {} resources", resourceList.size());
+    logger.info("Cleaning up {} resources", resourceList.size());
+
+    ResourceMaker.cleanupResources(resourceList, ownerControlledGcpResourceApi, getWorkspaceId());
+    logger.info("Cleaned up {} resources", resourceList.size());
+  }
+
+  @Override
+  public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspaceApi)
```
zloery, line 93: This test doesn't really assert much, so I think it's possible that the enumerateJobs endpoint might stop working correctly without this test failing. It seems pretty messy to validate the contents of the jobs (e.g. it's semi-random whether a job is complete or not), but you could validate things like
- the correct number of jobs are returned from each call
- for jobs which filter on resource type, validate all returned jobs match that resource type

zloery (commented):
Double-checking: will this change break currently active flights with the switch from `JobMapKeys.REQUEST` to `ResourceKeys.RESOURCE` in a couple places? 
I think that's probably fine given our low usage and that none of the modified flights run for very long, but good to confirm

integration/src/main/java/scripts/testscripts/EnumerateJobs.java
```diff
@@ -0,0 +1,198 @@
...
+  @Override
+  public void doSetup(List<TestUserSpecification> testUsers, WorkspaceApi workspaceApi)
+      throws Exception {
+
+    // initialize workspace
+    super.doSetup(testUsers, workspaceApi);
+
+    TestUserSpecification workspaceOwner = testUsers.get(0);
+
+    // TODO: if we like the alpha1 API for job enumeration, then we can maybe piggyback on
```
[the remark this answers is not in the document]
ddietterich: It is not really a TODO. I left it as a plain comment.

integration/src/main/java/scripts/testscripts/EnumerateJobs.java
```diff
@@ -0,0 +1,198 @@
...
+        ResourceMaker.makeResources(
+            ownerReferencedGcpResourceApi,
+            ownerControlledGcpResourceApi,
+            getWorkspaceId(),
+            getDataRepoSnapshotId(),
+            getDataRepoInstanceName(),
+            RESOURCE_COUNT);
+
+    logger.info("Created {} resources", resourceList.size());
+    logger.info("Cleaning up {} resources", resourceList.size());
```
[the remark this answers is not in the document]
ddietterich: This logging is on the client side and is the only way we can tell where a test is, so I have been thinking that being chatty is not a bad thing. I'm interested in your thoughts.

integration/src/main/java/scripts/testscripts/EnumerateJobs.java
```diff
@@ -0,0 +1,198 @@
...
+    logger.info("Cleaned up {} resources", resourceList.size());
+  }
+
+  @Override
+  public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspaceApi)
+      throws Exception {
+
+    // Case 1: fetch all
+    EnumerateJobsResult fetchall =
+        alpha1Api.enumerateJobs(getWorkspaceId(), null, null, null, null, null, null);
```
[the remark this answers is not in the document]
ddietterich: Our practice has been not to do that since IntelliJ auto-fills them in when you are looking at the code, but I'm happy to check in with folks about that practice.

integration/src/main/java/scripts/testscripts/EnumerateJobs.java
```diff
@@ -0,0 +1,198 @@
...
+    invalidPaginationException =
+        assertThrows(
+            ApiException.class,
+            () ->
+                alpha1Api.enumerateJobs(getWorkspaceId(), 22, "junktoken", null, null, null, null));
+    assertThat(invalidPaginationException.getMessage(), containsString("Invalid page token"));
+  }
+
+  private void logResult(String tag, EnumerateJobsResult result) {
+    logger.info(
```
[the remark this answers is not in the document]
ddietterich: I couldn't figure out a reliable way to check the exact job results, because the results depend on all of the flights that Stairway has run since it has been cleared out. That is, it would depend on the sequence of other tests. It also depends on the order of resource generation in the `ResourceMaker`.

As a separate ticket, I plan to add a `submitted_since` filter on the API. I think with that and maybe some output from `ResourceMaker` I could do more validation. Added PF-1281 and a TODO about enhancing the test for this.

service/src/main/java/bio/terra/workspace/app/configuration/external/FeatureConfiguration.java
```diff
@@ -23,19 +24,33 @@ public void setAzureEnabled(boolean azureEnabled) {
     this.azureEnabled = azureEnabled;
   }
 
+  public boolean isAlpha1Enabled() {
```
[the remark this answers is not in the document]
ddietterich: I hadn't gotten that far :)  I figured that I could use `alpha1` for going back and forth with user interface on a few endpoints before making them `v1`.

I do have a notion that we will want to revamp the entire workspace REST API and move it to v2. For that, we might want something like v2alpha1, v2alpha2, ... v2beta, v2, since we would be wanting to work out interface bugs for all of the API consumers and get some real usage before nailing it down. But I'm not really sure how we would orchestrate that process across Broad-Verily-Microsoft

service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
```diff
@@ -0,0 +1,238 @@
...
+        break; // referenced
+
+      case CONTROLLED:
+        ControlledResource controlledResource = wsmResource.castToControlledResource();
+        switch (wsmResource.getResourceType()) {
+          case AI_NOTEBOOK_INSTANCE:
+            {
+              ControlledAiNotebookInstanceResource resource =
+                  controlledResource.castToAiNotebookInstanceResource();
+              union.gcpAiNotebookInstance(resource.toApiResource(gcpProjectId));
```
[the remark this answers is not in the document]
ddietterich: Not sure. If projectId is null it would mean that someone passed an invalid resource as a parameter into a flight. I am reticent to throw, since it would make any enumeration touching that flight fail completely. OTOH, if there are errors like that with corrupt resources, it would be good to log in a way that would alert since it means we have (or had) a bug. What do you think?

@zloery do you have a perspective on this one?

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
ddietterich: I don't think so. That would mean that someone building a job had deliberately written the input parameter as empty/blank. So we wouldn't want to overwrite that.

integration/src/main/java/scripts/testscripts/EnumerateJobs.java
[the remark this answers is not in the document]
yuhuyoyo: go/java-practices/methods#call-site-literals the java style guide recommends inline comments.

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
ddietterich: If name is null, then `(name != null && !RESOURCE_NAME_VALIDATION_PATTERN.matcher(name).matches())` is false and it will not throw.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
ddietterich: Done - added PF-1282

service/src/main/java/bio/terra/workspace/service/workspace/model/JobStateFilter.java
ddietterich: I added some doc'n to the file.

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
ddietterich: There is a RESOURCE_TYPE key already. This is the RESOURCE - the whole resource object. The goal is that we can scrape Stairway to see the resources, either in the process of being created (like notebooks), or having been deleted.

service/src/main/java/bio/terra/workspace/service/workspace/model/EnumeratedJobs.java
ddietterich: The current guidance has a preference for using `immutable.io` over autovalue. I have a personal preference for being able to see the code, especially while debugging, rather than have it magicked up, so I never use them. If you want to use them, feel free, but please use immutable. It has much better and cleaner support for Jackson serdes.

integration/src/main/java/scripts/testscripts/EnumerateJobs.java
ddietterich: Good suggestions. I added assertions verifying the filtered results.

service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
```diff
@@ -0,0 +1,238 @@
...
+              ControlledGcsBucketResource resource = controlledResource.castToGcsBucketResource();
+              union.gcpGcsBucket(resource.toApiResource());
+              break;
+            }
+
+          case BIG_QUERY_DATASET:
+            {
+              ControlledBigQueryDatasetResource resource =
+                  controlledResource.castToBigQueryDatasetResource();
+              union.gcpBqDataset(resource.toApiResource(gcpProjectId));
```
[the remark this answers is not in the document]
ddietterich: > Double-checking: will this change break currently active flights with the switch from `JobMapKeys.REQUEST` to `ResourceKeys.RESOURCE` in a couple places? I think that's probably fine given our low usage and that none of the modified flights run for very long, but good to confirm

Yes, I was hoping it would be find. I started leaving both in, but that made it hard to know that I had found all of the places that needed to change.

service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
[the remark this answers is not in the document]
zloery: I'd rather not break every time we enumerate a bugged flight - I think a better way to handle this would be to log at error level and notify people with something like GCP's log-based alerting (PF-1247), which we've also looked at adding in a few other places.

zloery (approved):
LGTM!

integration/src/main/java/scripts/testscripts/EnumerateJobs.java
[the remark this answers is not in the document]
yuhuyoyo: Ah I see. Cleaning vs cleaned. That is reasonable to me.

service/src/main/java/bio/terra/workspace/service/workspace/model/JobStateFilter.java
yuhuyoyo: where is the doc?

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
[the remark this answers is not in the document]
ddietterich: I coded a check with logging on null in a subroutine, so we can easily change our minds about this.

service/src/main/java/bio/terra/workspace/service/workspace/model/JobStateFilter.java
ddietterich: Huh. It is there when I look at the code in the branch and it is there on my machine. I wonder why you are not seeing it. Here is the link:
https://github.com/DataBiosphere/terra-workspace-manager/blob/8fc595e458f6f1d265b7b0ddac617da5d2c0942c/service/src/main/java/bio/terra/workspace/service/workspace/model/JobStateFilter.java#L7

PR #528, opened 2022-01-13
[no ticket] Fix Azure VM creation failure due to invalid image

rtitle:
This should fix the failing VM test.

At first I thought I'd use a stable Linux image, but discovered VM creation as implemented only accepts a custom Linux image. It might be nice in the future to make the custom image optional, and support more generic OS types.


Qi77Qi (approved):

rtitle:
VM test fixed, now there's a failure in this PR with `CreateAzureStorageStep`. Will investigate and fix..

PR #530, opened 2022-01-14
[PF-1264] Split up WSM application testing profiles

zloery:
This change separates the real application configuration used for the `PrivateResourceCleanupServiceTest` connected test from the fake application configuration used by the `ApplicationConfigurationTest` unit test. Future tests which need an application should use the `app-test` profile to include this valid application configuration.

This change allows me to override the real application used in non-Broad deployment tests without also having to ignore/break the hardcoded values being checked in the unit test.

### DIFF
--- a/service/src/test/java/bio/terra/workspace/connected/ApplicationAccessUtils.java
+++ b/service/src/test/java/bio/terra/workspace/connected/ApplicationAccessUtils.java
@@ -9,7 +9,7 @@ import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Component;
 
 @Component
-@Profile("connected-test & configuration-test")
+@Profile("connected-test & app-test")
 public class ApplicationAccessUtils {
 
   @Autowired UserAccessUtils userAccessUtils;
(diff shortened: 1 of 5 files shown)


ddietterich:
This doesn't make sense to me. The test data is coming from a test resource file `application-configuration-test.yml` and does not need to be real or valid. Some of it is fabricated! This test is just making sure that we can read the configuration properly.

zloery:
Is `application-configuration-test.yml` only supposed to be for this unit test? 
Right now, the same file is also used for the `PrivateResourceCleanupServiceTest` connected test which creates an application-controlled resource, where it does need to be real and valid. Using it in both tests is what causes the problem.

If this configuration is only supposed to be for this configuration test, I can create a different application configuration file for that test and other connected tests which need real credentials.

ddietterich:
OK, got it. I had meant it to be a one-off only used in that test. Should've documented in the property file better!

One idea would be for the PrivateResourceCleanupServiceTest to call the `ApplicationDao` directly to create and enable the application, instead of using the configuration-test profile. We would have to mark the test as polluting the environment so the DB gets reset for the next test.

zloery:
Refactored this so that the `configuration-test` profile is now explicitly only used for this unit test and a new `app-test` profile is available for connected tests which need a real application, ptal.

ddietterich (approved):
LGTM

yuhuyoyo (approved):

PR #531, opened 2022-01-14
IA-3097 delete OS disk and nic during VM deletion

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3097

See [this doc](https://docs.google.com/document/d/1yT3Io_FbOj_RybYl-rtvRAEcEuCgdmhCEMkQCta3XmY/edit) for a bit more context

Tested locally via swagger

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/CreateAndDeleteAzureControlledResourceFlightTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/CreateAndDeleteAzureControlledResourceFlightTest.java
@@ -35,12 +35,11 @@ import bio.terra.workspace.service.resource.controlled.flight.delete.DeleteContr
 import bio.terra.workspace.service.resource.model.CloningInstructions;
 import bio.terra.workspace.service.workspace.WorkspaceService;
 import bio.terra.workspace.service.workspace.flight.create.azure.CreateAzureContextFlight;
+import com.azure.resourcemanager.compute.ComputeManager;
+import com.azure.resourcemanager.compute.models.VirtualMachine;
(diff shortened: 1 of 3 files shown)


ddietterich (commented):
Huh. So the NIC and disk are implicitly created when the VM is created, but need to be explicitly deleted? Interesting.

Is it possible to make a test that would verify this is functioning as expected?

Qi77Qi:
Yea. It's weird. Will add a test

Qi77Qi:
@ddietterich I expanded the existing create Vm test a bit to include deletion....lmk what you think

ddietterich (approved):
LGTM - thank you!

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/CreateAndDeleteAzureControlledResourceFlightTest.java
```diff
@@ -252,6 +256,47 @@ public void createAzureVmControlledResource() throws InterruptedException {
...
+    while (vmTemp == null) {
+      try {
+        vmTemp =
+            computeManager
+                .virtualMachines()
+                .getByResourceGroup(
+                    azureTestUtils.getAzureCloudContext().getAzureResourceGroupId(),
+                    creationParameters.getName());
+      } catch (com.azure.core.exception.HttpResponseException ex) {
+        if (ex.getResponse().getStatusCode() == 404) Thread.sleep(10000);
```
jdcanas, line 272: Seems a bit risky to not have an exit after a certain amount of tries here

jdcanas (approved):
small nit, otherwise looks good

PR #533, opened 2022-01-19
PF-1085: move default location logic from CLI to WSM

yuhuyoyo:
Currently in CLI, when location is not specified, a default location is set.

Another way is to allow user to specify a location when creating a workspace. If that location is set, all the resources will use that location if location is not specified. It adds more complexity to the change and I don't feel strongly one way or another so I opted for the simpler change.


ddietterich (commented):
The target state is for the location to be settable as a default or as a requirement on a workspace. I could see us picking up the workspace default location from the user's default location. I am not sure if we should provide a default for the system.

cbookg:
> The target state is for the location to be settable as a default or as a requirement on a workspace. I could see us picking up the workspace default location from the user's default location. I am not sure if we should provide a default for the system.

Right now the UI is hardcoding a specific region.  It would be nice if WSM either selected a default, or some way to read the default.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
```diff
@@ -24,6 +24,7 @@
 /** A {@link ControlledResource} for a Google AI Platform Notebook instance. */
 public class ControlledAiNotebookInstanceResource extends ControlledResource {
 
+  private static final String LOCATION = "us-central1-a";
```
ddietterich, line 27: Do we need three location constants or should we share one?
Since this is an intermediate, it isn't too important.

ddietterich (approved):
LGTM - one suggestion

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledAiNotebookInstanceResource.java
yuhuyoyo: I created a resourceConstant file under controlled/

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedGitRepoResource.java
```diff
@@ -0,0 +1,39 @@
+package bio.terra.workspace.service.resource.referenced;
```
zloery, line 1: I don't think this file should be in this PR - did it get pulled in from a different branch you're working on?

zloery (commented):
LGTM, minus one file which looks like it got pulled in from another ticket

zloery (approved):

PR #534, opened 2022-01-20
[PF-1287] Reroute and pretty up slack messages

ddietterich:
This change moves test notifications to the new `#terra-wsm-alerts` channel.
I also changed how the emoji works, so it reflects the status result.
I added `@channel` to the message on failure, but I have not been able to confirm that works.
As usual, it is hard to test push tests without doing a push...

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -137,8 +137,15 @@ jobs:
           L4="Res: ${{ steps.resiliency-test.outcome }}"
           text="Link to <https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/$GITHUB_RUN_ID|test run>"
           bold="$L2 | $L3 | $L4"
+          if [ "${{ job.status }}" == "success" ]; then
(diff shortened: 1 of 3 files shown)


ungwudik (approved):

zloery (approved):
LGTM

.github/workflows/nightly-tests.yml
```diff
@@ -137,8 +137,15 @@ jobs:
           L4="Res: ${{ steps.resiliency-test.outcome }}"
           text="Link to <https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/$GITHUB_RUN_ID|test run>"
           bold="$L2 | $L3 | $L4"
+          if [ "${{ job.status }}" == "success" ]; then
+            emoji=":white_check_mark:"
+          else
+            emoji=":no_entry:"
+            text="@channel $text"
```
jaycarlton, line 144: idea: Consider defining an alias for our team/first responders like `@wsm-devteam` and using that instead of `@channel`.

PR #535, opened 2022-01-21
[PF-1285] Wait for database row before async return

ddietterich:
For async resource create, wait until the row is present in the database (or the flight fails) to reply to the client. If that doesn't happen within a set time, return service unavailable.

This change is made so that an enumerateResources call performed while a resource is being created will show the resource. The related change for querying jobs allows the UI to know the state of the async resource create.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/db/ResourceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/ResourceDao.java
@@ -585,6 +585,28 @@ public class ResourceDao {
     jdbcTemplate.update(sql, params);
   }
 
+  /**
+   * Wait for a resource row to appear in the database. This is used so async resource creation
+   * calls return after the database row has been inserted. That allows clients to see the resource
+   * with a resource enumeration.
+   *
+   * @param workspaceId workspace where the resource is being created
+   * @param resourceId resource being created
+   */
+  @ReadTransaction
+  public boolean checkForResourceRow(UUID workspaceId, UUID resourceId) {
+    final String sql =
+        "SELECT COUNT(1) FROM resource"
+            + " WHERE workspace_id = :workspace_id AND resource_id = :resource_id";
+    MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue("workspace_id", workspaceId.toString())
+            .addValue("resource_id", resourceId.toString());
+
+    Integer count = jdbcTemplate.queryForObject(sql, params, Integer.class);
+    return (count != null && count > 0);
+  }
+
   private void validateUniqueGcsBucket(ControlledGcsBucketResource bucketResource) {
     String bucketSql =
         "SELECT COUNT(1)"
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -585,6 +585,28 @@ public void setPrivateResourcesStateForWorkspaceUser(
...
+  /**
+   * Wait for a resource row to appear in the database. This is used so async resource creation
+   * calls return after the database row has been inserted. That allows clients to see the resource
+   * with a resource enumeration.
+   *
+   * @param workspaceId workspace where the resource is being created
+   * @param resourceId resource being created
+   */
+  @ReadTransaction
+  public boolean checkForResourceRow(UUID workspaceId, UUID resourceId) {
```
jaycarlton, line 597: Since it's a boolean function, I'd name it like `resourceExists()`. It's implied that we're talking about a DB row since we're in the Dao.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -630,4 +641,35 @@ private JobBuilder commonDeletionJobBuilder(
...
+   * @param jobId id of the create flight.
+   */
+  private void waitForResourceOrJob(UUID workspaceId, UUID resourceId, String jobId) {
+    try {
+      for (int i = 0; i < RESOURCE_ROW_MAX_RETRIES; i++) {
+        if (resourceDao.checkForResourceRow(workspaceId, resourceId)) {
+          return;
+        }
+        FlightState flightState = jobService.getStairway().getFlightState(jobId);
+        if (flightState.getCompleted().isPresent()) {
```
jaycarlton, line 664: Are you sure there's no race condition where we get here and the job completed successfully? Should we predicate the first branch on the  job being successfully started but not finished?

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -55,7 +58,10 @@
 @Component
 public class ControlledResourceService {
 
-  private static final int MAX_ASSIGNED_USER_LENGTH = 254;
+  // These are chosen to retry a maximum length so we return under a 30 second
```
yuhuyoyo, line 61: I'm lost how the math works out for the 30 seconds?

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
ddietterich: It is possible that we get here and the job completed successfully. That is OK. We return the job status to the client. The client calls the `/result` endpoint and gets the successful result.
It is unexpected, because we would expect to see the initial step complete before the whole flight completes, but it could happen.
Does that make sense? Perhaps I can resolve this by adding more commentary above?

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
jaycarlton: That makes sense. It smelled like there might be coupling or an assumption about timing, but it sounds like you've worked that out. A blurb about this possibility in a comment couldn't hurt I suppose. Thanks.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
ddietterich: I switched to using the timer, FYI.

PR #536, opened 2022-01-21
PF-1184: Create a git repo referenced resource and its relevant API

yuhuyoyo:
https://docs.google.com/document/d/13pEGiu9OOk_zhdpi-LPJhwjPEbUVdeQQT43PqLhc2uE/edit?usp=sharing&resourcekey=0-qrxEly_ybyUr4d9t68fT6w


ddietterich (commented):
Your code ran into a broader question about whether referenced resources should have a cloud platform or require a cloud context. I've asked Jay and Zach to chime in. So we need to decide about the approach to "MULTI" cloud platform before this goes in.

There are some typos in the openapi, but otherwise looks good.

service/src/main/java/bio/terra/workspace/service/workspace/model/CloudPlatform.java
```diff
@@ -6,7 +6,8 @@
 
 public enum CloudPlatform {
   GCP("GCP", ApiCloudPlatform.GCP),
-  AZURE("AZURE", null);
+  AZURE("AZURE", null),
+  MULTI("MULTI", null);
```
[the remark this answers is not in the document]
jaycarlton: I agree that this enum is only applicable in reference to a particular cloud. I don't know why references might need to be cloud-bound, but I never quite got a handle on all the use cases a reference supports. If we wanted a general URI reference type, it would be cloud-independent, though I don't know what you'd do with it.

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -65,6 +65,9 @@
   public static final Pattern RESOURCE_NAME_VALIDATION_PATTERN =
       Pattern.compile("^[a-zA-Z0-9][-_a-zA-Z0-9]{0,1023}$");
 
+  public static final Pattern GIT_URL_PATTERN =
```
jaycarlton, line 68: Is this a URI more than a URL?

service/src/main/java/bio/terra/workspace/service/resource/WsmResourceType.java
```diff
@@ -88,7 +89,13 @@
       "AZURE_STORAGE_ACCOUNT",
       ApiResourceType.AZURE_STORAGE_ACCOUNT,
       null,
-      ControlledAzureStorageResource.class);
+      ControlledAzureStorageResource.class),
+  GIT_REPO(
+      CloudPlatform.MULTI,
```
jaycarlton, line 94: We might think of backing this property out of this enum, since there are non-cloud-specific resources now. Or we can make it optional.

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedGitRepoResource.java
```diff
@@ -0,0 +1,162 @@
...
+    return WsmResourceType.GIT_REPO;
+  }
+
+  @Override
+  public String attributesToJson() {
+    return DbSerDes.toJson(new ReferencedGitRepoAttributes(gitUrl));
+  }
+
+  @Override
+  public boolean checkAccess(FlightBeanBag context, AuthenticatedUserRequest userRequest) {
```
jaycarlton, line 62: document why this is always true

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2417,6 +2531,17 @@ components:
...
           type: string
 
+    UpdateGitRepoReferenceRequestBody:
+      type: object
+      properties:
+        name:
+          $ref: "#/components/schemas/Name"
+        description:
+          type: string
+        gitUrl:
```
jaycarlton, line 2541: This name could become ambiguous if we later add support for other URLs associated with a repo, like tags, SHAs, individual files, etc.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -4662,7 +4845,12 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/CloneReferencedGcpDataRepoSnapshotResourceResult'
-
+    CloneReferencedGitRepoResourceResponse:
+      description: Response for successful Data Repo Snapshot reference clone
```
jaycarlton, line 4849: copypasta

service/src/test/java/bio/terra/workspace/service/resource/ValidationUtilsTest.java
```diff
@@ -239,4 +239,54 @@ public void validateBqDataTableName_invalidName_throwsException() {
...
+    ValidationUtils.validateGitRepoUrl(
+        "https://username:password@github.com/username/repository.git");
+  }
+
+  @Test
+  public void validateGitHubRepoHttpsUrl_httpUrl_throwsException() {
+    assertThrows(
+        InvalidReferenceException.class,
+        () ->
+            ValidationUtils.validateGitRepoUrl(
```
jaycarlton, line 259: It's not obvious why some of these are invalid. Is it because they're missing a .git?

service/src/main/java/bio/terra/workspace/service/workspace/model/CloudPlatform.java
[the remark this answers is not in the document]
zloery: I think this is a spot our object model doesn't fit quite right. E.g. we already support references to TDR snapshots - those have the "GCP" cloud platform, but they're not really GCP resources. In retrospect I think `CloudPlatform` should probably have been a ControlledResource property, not a top-level WsmResource property, though changing that now would be a big mess for the API. The only value I can think of for having this at the WsmResource level would be filtering the results of an EnumerateResources call, though we don't have that as a parameter now and I'm not sure any users have asked for it.

References don't (and definitely shouldn't!) require a cloud context.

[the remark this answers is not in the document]
ddietterich: Thanks @jaycarlton and @zloery! I think we can live with this structure more or less how Yu has it.
Yu: I'll leave it to your aesthetic sense to pick a name.

ITFOT, we can move the cloud platform to the controlled resources. I think the API is OK. Saying `/gcp/bucket` and (just for example)`/aws/bucket` seems fine to me to identify the cloud resource.

ddietterich (approved):
LGTM given the typo cleanup

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
yuhuyoyo: does check for git specific elements, https (https://<host>/<organization>/<repo>.git) or ssh url (git@github.com:<organization>/<repo>.git)

service/src/test/java/bio/terra/workspace/service/resource/ValidationUtilsTest.java
yuhuyoyo: this one is because it's a http url instead of https. the one below is missing .git in the end. I will update the test name to make it more obvious.

service/src/main/java/bio/terra/workspace/service/workspace/model/CloudPlatform.java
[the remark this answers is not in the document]
yuhuyoyo: used ANY. Added comment and a TODO to move it to controlled resources.

service/src/main/java/bio/terra/workspace/service/resource/WsmResourceType.java
yuhuyoyo: added a TODO. I tried making it optional and it requires a lot more change in the ResourceDao so I think it should be in a separate PR.

service/src/main/resources/api/service_openapi.yaml
yuhuyoyo: true. This is only for the cloning url. I updated the name

jaycarlton (approved):

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2565,7 +2690,7 @@ components:
 
     SystemVersion:
       type: object
-      required: [gitTag, gitHash, github, build]
+      required: [gitTag, gitHash, gitHub, build]
```
zloery, line 2693: why this change? this entry in the `required` field no longer matches the actual `SystemVersion.properties.github` field

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -65,6 +65,9 @@
   public static final Pattern RESOURCE_NAME_VALIDATION_PATTERN =
       Pattern.compile("^[a-zA-Z0-9][-_a-zA-Z0-9]{0,1023}$");
 
+  public static final Pattern GIT_CLONE_URL_PATTERN =
```
zloery, line 68: This pattern of allowing users to pass arbitrary URLs and having WSM call them (which we would need to do for access validation) scares me a little from a security perspective.

What we did for TDR snapshots was to configure a list of allowed instances (e.g. for dev WSM, allow dev TDR with a known URL) and have requests indicate which instance they were referring to. Could we do something similar here?

E.g., we would explicitly only allow git references from [Github, Gitlab, Azure, Github enterprise, GCP cloud source repository, others added as needed]. The "createGitReference" request would have fields like "instance" with enum values like "GITHUB", "GITLAB", ..., plus fields like "repoName", and WSM could construct the URL from the combination of the two. That way, users couldn't force WSM to make calls to arbitrary URLs.

zloery (changes requested):

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
jaycarlton: I don't think we could enumerate and then keep up with all the git origins out there. For example, what about internal company servers that we wouldn't know about? I like the idea of passing URL parts separately. Plus it expands to tags/releases and commits naturally.

If we did have a fixed list, we'd want an API to return it so the UI could have a dropdown for the entries.

ddietterich: Is WSM directly making requests on the URL? We are aiming for an enterprise github, so github.com might not be one the allowed list.

A whitelist probably makes sense in the fullness of time. I wonder if there is a wider issue of controlling external access from a workspace that is more in the realm of policy and security perimeters rather than point solutions for single URLs.

yuhuyoyo: I was going to just have a github referenced resource but then changed to a generic git url because i was trying to handle as many cases as possible (e.g. servers that we don't know about) but security concern is a good point. 

Maybe we can start with constructing the url for github enterprise case first and add to the api as needed then? 

an enterprise github url would still look like https://github.com/organization/project.git or git@github.com:organization/project.git so I would need user input for organization and project name and whether they want ssh url or https url.

zloery: Agreed, the number + variation of sources makes it messier to maintain a list like this than it is in the TDR case. Whitelisting URLs also sgtm if we'd rather the API still take in whole URLs instead of instance identifiers.

IIUC yes, the current design would have WSM directly making requests to the provided URL using the user's credentials to validate a user's repo access before creating the reference (though we don't right now since validation isn't implemented). I think there's probably a broader policy question of limiting the things a workspace can reach, but I think this is a much more local security issue of "don't let users make your service call arbitrary URLs".

yuhuyoyo: right now the user case I'm imaging is that the jupyter notebook uses the git repo references to clone some workspaces in the notebook instance. I'm not sure if there's value to directly make GIT operation in WSM.

yuhuyoyo: Thanks for all the suggestions. I listed the three options here https://docs.google.com/document/d/13pEGiu9OOk_zhdpi-LPJhwjPEbUVdeQQT43PqLhc2uE/edit?resourcekey=0-qrxEly_ybyUr4d9t68fT6w#bookmark=id.xuuyb98gkeza. Would we like to settle on option 3: having a set of allowlisted hostname?

ddietterich: I think doing the allow list is best, even though it means more work to maintain it.
Is the hostname sufficient for the allow list? I think so, but I'm not a security expert.

Implementing allow list will be a separate piece of work and need some design around how we maintain and propagate that list. Is it deployment wide or a policy that propagates down folders? Or something else? Maybe need some product requirements.

yuhuyoyo: added an allowlist of hostname as an external/gitreporeferenceconfiguration. not sure if that's the right directory to put it. please take a look.

so for https:// and ssh:// url, we parse it to a uri and then check hostname. If it's a git@ ssh uri, we use regex to get the host name and then check whether it is in the allowlist.

yuhuyoyo:
@ddietterich re-adding you because after merging with your PR, I put the git repo referenced resource and attributes in cloud/any/ to match with the CloudPlatform type.

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -116,6 +131,48 @@ public static void validateBucketName(String name) {
     }
   }
 
+  /** Validate whether the input URI is a valid GitHub Repo https uri. */
```
ddietterich, line 134: nit: I think the comment is not quite right - it doesn't have to be https, right?

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2417,6 +2531,17 @@ components:
...
           type: string
 
+    UpdateGitRepoReferenceRequestBody:
+      type: object
+      properties:
+        name:
+          $ref: "#/components/schemas/Name"
+        description:
+          type: string
+        gitCloneUrl:
```
ddietterich, line 2541: What makes this a "clone" URL and not just the git URL? Won't users do other things with it besides clone?

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
```diff
@@ -116,6 +131,48 @@ public static void validateBucketName(String name) {
...
+  public void validateGitRepoCloneUri(String gitUri) {
```
ddietterich, line 135: The term "clone URI" is confusing in our code. We use the term clone to refer to our own workspace and resource cloning operations. (ref later comment)

ddietterich (commented):
I have a question about using the name "gitCloneUrl". Otherwise, looks good!

service/src/main/resources/api/service_openapi.yaml
yuhuyoyo: this was changed from gitUrl based on Jay's previous comment because we can later support passing in git url for a specific branch or file. 

What can user do besides cloning though? I'm lack of a better word.. gitUrlMain to imply that it's from the main branch? gitUrlDefault?

ddietterich: Maybe gitRepoUrl? Because it is pointing at the whole repo?

jaycarlton: I like gitRepoUrl.

service/src/main/java/bio/terra/workspace/service/resource/ValidationUtils.java
yuhuyoyo: renamed.

service/src/main/resources/api/service_openapi.yaml
yuhuyoyo: renamed.

service/src/test/java/bio/terra/workspace/service/resource/ValidationUtilsTest.java
```diff
@@ -239,4 +255,61 @@ public void validateBqDataTableName_invalidName_throwsException() {
...
+            validationUtils.validateGitRepoUri(
+                "http://github.com/DataBiosphere/terra-workspace-manager.git"));
+  }
+
+  @Test
+  public void test() throws URISyntaxException {
+    URI uri = new URI("ssh://github.com/project.git");
+
+    uri.getHost();
+  }
```
zloery, line 293: does this test anything, or is it left over from development?

zloery (approved):
LGTM!

service/src/test/java/bio/terra/workspace/service/resource/ValidationUtilsTest.java
yuhuyoyo: oops nice catch

PR #537, opened 2022-01-25
[PF-1241] Grant proxy group `actAs` permission, instead of user and pet SA

melissachang:
Before this PR, `WorkspaceService.enablePet` gave the following `actAs` permission for the user's pet SA:

- User (needed for using notebook in GCP console)
- Pet SA (needed for running Nextflow from a notebook)

This needs to change for Verily deployment. (`enablePet` was added for Verily deployment.) In Verily deployment, workspaces are created in google.com org. google.com org has a policy restriction that users cannot have `actAs` permission. (For Verily deployment, we have a temporary exception.) To get around this policy restriction, in this PR, instead of granting above two `actAs` permission, we grant for:

- User proxy group

Tested: Ran `enablePet` locally. In GCP console, instead of user/pet having `actAs` permission, proxy group has `actAs` permission.

Note that `SamService.getProxyGroupEmail()` takes in userEmail instead of AuthenticatedUserRequest because of `WorkspaceService.removeWorkspaceRoleFromUser()`. When User `A` removes User `B` from workspace, we want to get `B`'s proxy group, not `A`'s.


service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -74,12 +77,13 @@ public Policy enablePetServiceAccountImpersonation(UUID workspaceId, UserWithPet
...
    * @param workspaceId ID of the workspace to enable pet SA in
    * @param userAndPet The user and the pet SA the user is being granted permission to
+   * @param proxyGroupEmail User's proxy group
    * @param eTag GCP eTag which must match the pet SA's current policy. If null, this is ignored.
    * @return The new IAM policy on the user's pet service account, or empty if the eTag value
    *     provided is non-null and does not match current IAM policy on the pet SA.
    */
   public Optional<Policy> enablePetServiceAccountImpersonationWithEtag(
-      UUID workspaceId, UserWithPetSa userAndPet, @Nullable String eTag) {
+      UUID workspaceId, UserWithPetSa userAndPet, String proxyGroupEmail, @Nullable String eTag) {
```
jaycarlton, line 86: Do you still need a `UserWithPetSa` for this method if we're only getting the user email from it now? You could just take in the user email directly.

melissachang: On line 89, we get the pet email (to look up pet SA).

jaycarlton: my bad.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -525,7 +525,9 @@ private ApiCreateCloudContextResult fetchCreateCloudContextResult(
                     gcpCloudContextService.getRequiredGcpProject(workspaceId), userRequest),
             "enablePet");
     UserWithPetSa userAndPet = new UserWithPetSa(userEmail, petSaEmail);
-    petSaService.enablePetServiceAccountImpersonation(workspaceId, userAndPet);
+    String proxyGroupEmail =
```
ddietterich, line 528: Should we put these two lines into PetSaService?  If it is used elsewhere, it might be worth it.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -525,7 +525,9 @@ private ApiCreateCloudContextResult fetchCreateCloudContextResult(
                     gcpCloudContextService.getRequiredGcpProject(workspaceId), userRequest),
             "enablePet");
     UserWithPetSa userAndPet = new UserWithPetSa(userEmail, petSaEmail);
```
ddietterich, line 527: This is getting complex enough that I think we should move the logic out of the controller and into a service method. I'd suggest either the workspace service or the PetSaService.

In general, we don't want implementation in the controller; just dispatching.

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -179,10 +183,7 @@ public Policy enablePetServiceAccountImpersonation(UUID workspaceId, UserWithPet
       Binding bindingToRemove =
```
ddietterich, line 183: Is the warning above correct now? Aren't we revoking proxy group access to the pet SA?
(Or am I totally confused?? which is totally possible!)

Same comment/question for the enable call.

service/src/main/java/bio/terra/workspace/service/workspace/flight/RevokePetUsagePermissionStep.java
```diff
@@ -90,7 +93,7 @@ public StepResult undoStep(FlightContext context) throws InterruptedException {
       return StepResult.getStepResultSuccess();
     }
     petSaService.enablePetServiceAccountImpersonationWithEtag(
-        workspaceId, userAndPet, expectedEtag);
+        workspaceId, userAndPet, proxyGroupEmail, expectedEtag);
```
ddietterich, line 96: Would it make sense to extend the UserWithPetSa class to carry the proxy group instead of modifying the interfaces all the way through?

ddietterich (commented):
Looks fine. A few questions I'd like to understand before approving.

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
zloery: I think we should remove all the TODOs pointing to PF-991, which is a "Closed, won't do" ticket.

Originally we thought allowing pets to `actAs` themselves would be a temporary fix, so I marked all the places we'd have to remove. That changed and we expect to keep that functionality, so the TODOs are now out of date.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -130,6 +130,15 @@ public String getUserEmailFromSam(AuthenticatedUserRequest userRequest)
     }
   }
 
+  public String getProxyGroupEmail(String userEmail) throws InterruptedException {
```
zloery, line 133: nit: add javadoc for public methods
Your note about this method in the PR description is helpful, could you also include it in javadoc here?

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -130,6 +130,15 @@ public String getUserEmailFromSam(AuthenticatedUserRequest userRequest)
...
+    GoogleApi googleApi = samGoogleApi(getWsmServiceAccountToken());
```
zloery, line 134: I think this method should also take in an `AuthenticatedUserRequest` and use those credentials here, instead of pulling WSM's SA credentials. 
WSM's SA has broad permissions and it makes determining the real caller from logs harder, so we generally avoid using it unless we need to.

melissachang:
Friendly ping :)

integration/src/main/java/scripts/testscripts/EnablePet.java
```diff
@@ -48,8 +47,9 @@ protected void doUserJourney(TestUserSpecification testUser, WorkspaceApi userWo
     Iam userIamClient = ClientTestUtils.getGcpIamClient(testUser);
     assertFalse(canImpersonateSa(userIamClient, petSaEmail));
 
-    String returnedPetSaEmail = userWorkspaceApi.enablePet(getWorkspaceId());
-    assertEquals(petSaEmail, returnedPetSaEmail);
+    // FIX BEFORE MERGING
```
jaycarlton, line 50: It says to fix before merging.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -130,6 +130,22 @@ public String getUserEmailFromSam(AuthenticatedUserRequest userRequest)
...
   }
 
+  /**
+   * Gets proxy group email.
+   *
+   * <p>This takes in userEmail instead of AuthenticatedUserRequest because of
+   * WorkspaceService.removeWorkspaceRoleFromUser(). When User A removes User B from workspace, we
+   * want to get B's proxy group, not A's.
+   */
+  public String getProxyGroupEmail(String userEmail, String token) throws InterruptedException {
```
jaycarlton, line 140: Is this user A or B?

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -54,35 +53,49 @@ public PetSaService(
...
    * particular GCP eTag value.
    */
-  public Policy enablePetServiceAccountImpersonation(UUID workspaceId, UserWithPetSa userAndPet) {
+  public Policy enablePetServiceAccountImpersonation(
+      UUID workspaceId, AuthenticatedUserRequest userRequest) {
     // enablePetServiceAccountImpersonationWithEtag will only return an empty optional if the
     // provided eTag does not match current policy. Because we do not use eTag checking here, this
     // is always nonempty.
-    return enablePetServiceAccountImpersonationWithEtag(workspaceId, userAndPet, null).get();
+    return enablePetServiceAccountImpersonationWithEtag(workspaceId, userRequest, null).get();
```
jaycarlton, line 61: I'd use `orElseThrow()` instead of `get()` so you get a more meaningful error message when it's not found.

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -147,42 +155,49 @@ public Policy enablePetServiceAccountImpersonation(UUID workspaceId, UserWithPet
...
           new Binding()
               .setRole(SERVICE_ACCOUNT_USER_ROLE)
-              .setMembers(
-                  ImmutableList.of(
-                      "user:" + userAndPet.getUserEmail(),
-                      "serviceAccount:" + petServiceAccount.email()));
+              .setMembers(ImmutableList.of("group:" + proxyGroupEmail));
       // If no bindings exist, getBindings() returns null instead of an empty list. If there are
       // no policies, there is nothing to revoke, so this method is finished.
       List<Binding> oldBindingList = saPolicy.getBindings();
```
jaycarlton, line 203: Aside: it might be nice to have a wrapper method like `getBindingsList()` that returned an empty list instead of null.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateControlledResourceFlight.java
```diff
@@ -243,6 +250,7 @@ public CreateControlledResourceFlight(FlightMap inputParameters, Object beanBag)
 
   private void addNotebookSteps(
       UserWithPetSa userAndPet,
```
jaycarlton, line 252: It feels like we're passing in user informatino in three pieces, and it's not obvious which one is authoritative (i.e. is the user email supposed to be in the `userAndPet` or the `userRequest`. I guess just maybe comment on that for now, and think about a higher-level `UserBundle` object or something that knows the user, pet, and proxy.

jaycarlton (commented):
"FI X  BEFORE MERGING" in several places.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateControlledResourceFlight.java
ddietterich: I think that is another piece of work. It would be useful to have a consistent collection of the parts. Perhaps we could extend the AuthenticatedUserRequest rather than make another structure to pass around. Just an idea.

ddietterich (approved):
LGTM - as I wrote, we should think about some consolidation of the user information, but that is beyond the scope of this PR.

ddietterich:
I added PF-1311 to track reviewing how we are passing around user information.

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -371,7 +371,10 @@ void createAiNotebookInstanceUndo() throws Exception {
     String serviceAccountEmail =
         samService.getOrCreatePetSaEmail(projectId, user.getAuthenticatedRequest());
     UserWithPetSa userAndPet = new UserWithPetSa(user.getEmail(), serviceAccountEmail);
-    petSaService.disablePetServiceAccountImpersonation(workspace.getWorkspaceId(), userAndPet);
+    String proxyGroupEmail =
```
zloery, line 374: This fetches the proxy group email, but never uses or checks it

service/src/main/java/bio/terra/workspace/service/workspace/flight/RevokePetUsagePermissionStep.java
```diff
@@ -34,18 +34,21 @@
...
   private final String userEmailToRemove;
+  private final String proxyGroupEmail;
   private final PetSaService petSaService;
   private final GcpCloudContextService gcpCloudContextService;
   private final AuthenticatedUserRequest userRequest;
 
   public RevokePetUsagePermissionStep(
       UUID workspaceId,
       String userEmailToRemove,
+      String proxyGroupEmail,
```
zloery, line 45: This is passed in, but never used in this step

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateControlledResourceFlight.java
```diff
@@ -243,6 +250,7 @@ public CreateControlledResourceFlight(FlightMap inputParameters, Object beanBag)
...
+      String proxyGroupEmail,
```
zloery, line 253: Same issue with this flight: `proxyGroupEmail` is never passed to/used by steps

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
melissachang: Done

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
melissachang: A. Though I don't think we need to clarify in a comment. It's just a token the caller passed in, for calling SAM.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
melissachang: Done

zloery (commented):
Mostly LGTM, one issue with the undo step on RevokePetUsagePermission

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -54,35 +53,51 @@ public PetSaService(
...
+   * @param userRequest The user whose proxy group will be granted permission.
    * @param eTag GCP eTag which must match the pet SA's current policy. If null, this is ignored.
    * @return The new IAM policy on the user's pet service account, or empty if the eTag value
    *     provided is non-null and does not match current IAM policy on the pet SA.
    */
   public Optional<Policy> enablePetServiceAccountImpersonationWithEtag(
-      UUID workspaceId, UserWithPetSa userAndPet, @Nullable String eTag) {
+      UUID workspaceId, AuthenticatedUserRequest userRequest, @Nullable String eTag) {
+    // userRequest might not have user email, so fetch user email explicitly.
+    String userEmail =
```
jaycarlton, line 85: I guess you could try to get the email from the request first, before calling Sam.

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -509,24 +508,16 @@ private ApiCreateCloudContextResult fetchCreateCloudContextResult(
...
-        SamRethrow.onInterrupted(
-            () ->
-                samService.getOrCreatePetSaEmail(
-                    gcpCloudContextService.getRequiredGcpProject(workspaceId), userRequest),
-            "enablePet");
-    UserWithPetSa userAndPet = new UserWithPetSa(userEmail, petSaEmail);
-    petSaService.enablePetServiceAccountImpersonation(workspaceId, userAndPet);
-    return new ResponseEntity<>(petSaEmail, HttpStatus.OK);
+    petSaService.enablePetServiceAccountImpersonation(
+        workspaceId, userRequest.getEmail(), userRequest.getRequiredToken());
```
zloery, line 519: I'm not positive, but I think this email will still need to come from Sam, i.e. keeping the previous
```
 String userEmail =
        SamRethrow.onInterrupted(() -> samService.getUserEmailFromSam(userRequest), "enablePet");
```
The email field of the userRequest gets populated by the proxy in front of WSM, so in places where we run WSM without the proxy (i.e. tests and locally), this might cause issues.

zloery (approved):
LGTM, thank you for this!

one note that might cause test failures, but I'm not totally sure about that

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
melissachang: Done. I realized this as well, you beat me to it. :) Changed another file as well.

PR #538, opened 2022-01-26
[PF-1295] Reorganize into per-cloud, per-resource directories in main and test

ddietterich:
First step in the "improving resource dev" epic. No code change at all. Just renames.


jaycarlton (approved):
I checked out the branch and played with it in the project view in IJ. The only thing that's kinda funny is seeing azure and gcp on the same level as exception and flight. It almost felt like the different cloud types should be alone together in one package.

jaycarlton (approved):

zloery (approved):

yuhuyoyo (approved):

PR #539, opened 2022-01-27
[PF-1291] Update testrunner to 0.1.0, named parameters for all test scripts

zloery:
The latest version of TestRunner lets test scripts specify parameters as key-value pairs instead of relying on the order they were specified. This change pulls in the new TestRunner version and migrates all existing test scripts, but does not change any of the parameters or values passed to tests.


zloery:
Adding @ichengchang as an FYI as this touches QA's results upload scripts, though this doesn't change any values so it should be a no-op.

integration/src/main/java/scripts/testscripts/UpdateReferenceResources.java
```diff
@@ -102,18 +104,15 @@ protected void doCleanup(List<TestUserSpecification> testUsers, WorkspaceApi wor
   }
 
   @Override
-  public void setParameters(List<String> parameters) throws Exception {
-    // TODO: Refactor this function when TestRunner starts supporting parameterMap
+  public void setParameters(Map<String, String> parameters) throws Exception {
     super.setParameters(parameters);
-    if (parameters == null || parameters.size() < 4) {
+    if (parameters == null
```
ddietterich, line 109: I wonder if we could have some util methods to handle these and DRY it up a bit. This is similar to the FlightMap as an anonymous map, so two things that might be useful:
1. Something like the FlightUtil Jay wrote to test that all expected keys are present in the map
2. Something that does "get key XYZ and throw if you don't find it"
I think that would make the code more readable

ddietterich (commented):
I have some suggestions for DRYing up the code.
Otherwise, it looks excellent.

yuhuyoyo (approved):

integration/src/main/java/scripts/testscripts/UpdateReferenceResources.java
zloery: Added suggestion 2, that cleans up the configs nicely. 

I don't think we need 1 since calling "get key X or throw" for each param we need already validates keys before the tests run. Including the full map check just meant each base test script had a "check for all keys or throw" line, then the "get key or throw" lines,  which seemed redundant.

ddietterich (approved):
I like it. Thanks!

zloery:
Integration test is failing because I messed up publishing version `0.1.0-SNAPSHOT` of TestRunner (I somehow published the existing code to that version first), and our GHA now has the wrong jar cached. It's working locally, I'll try to fix the GHA cache

PR #540, opened 2022-01-28
[PF-1308] Make projectId an attribute of notebooks and controlled buckets.

ddietterich:
In working on the larger resource handler implementation I ran into a problem. Notebooks and controlled buckets returned their GCP projectId in the resource return, but did not save the projectId in their attributes. That meant that we had to do a cloud context lookup in the database in those specific cases.

This change stores projectId as part of the attributes for those objects. The projectId is immutable, so it is safe to store with the resource. This is upward compatible, but not backward compatible. The new version of attributes can deserialize the old version of attributes, but not vice versa. This is only an issue during a rolling upgrade, so not something I think we need to worry about right now.

To implement the upward compatibility, I needed to call the GcpCloudContextService and get the projectId. I cannot do that from the resource object, so I put in the initial resource handler hierarchy. I just implemented the minimum of the two handlers for these objects.

I put you three on the reviewer list to keep you in the loop on these structural changes. Let me know if you'd prefer I spread it around.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
@@ -200,7 +200,7 @@ public class Alpha1ApiController implements Alpha1Api {
             {
               ControlledAiNotebookInstanceResource resource =
                   controlledResource.castToAiNotebookInstanceResource();
-              union.gcpAiNotebookInstance(resource.toApiResource(gcpProjectId));
+              union.gcpAiNotebookInstance(resource.toApiResource());
               break;
             }
           case GCS_BUCKET:
--- a/service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
@@ -14,6 +14,7 @@ import bio.terra.workspace.generated.model.ApiResourceUnion;
 import bio.terra.workspace.generated.model.ApiStewardshipType;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequestFactory;
+import bio.terra.workspace.service.iam.model.SamConstants.SamWorkspaceAction;
 import bio.terra.workspace.service.job.JobService;
 import bio.terra.workspace.service.resource.ValidationUtils;
 import bio.terra.workspace.service.resource.controlled.cloud.gcp.ainotebook.ControlledAiNotebookInstanceResource;
@@ -39,7 +40,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
-import javax.annotation.Nullable;
 import javax.servlet.http.HttpServletRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -87,8 +87,9 @@ public class Alpha1ApiController implements Alpha1Api {
     // Make sure Alpha1 is enabled
     features.alpha1EnabledCheck();
 
-    // Prepare the inputs
     final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    workspaceService.validateWorkspaceAndAction(userRequest, workspaceId, SamWorkspaceAction.READ);
+
     ControllerValidationUtils.validatePaginationParams(0, limit);
     ValidationUtils.validateOptionalResourceName(name);
 
@@ -103,10 +104,6 @@ public class Alpha1ApiController implements Alpha1Api {
             StewardshipType.fromApiOptional(stewardship),
             name,
             JobStateFilter.fromApi(jobState));
-
-    // projectId
-    String gcpProjectId =
-        workspaceService.getAuthorizedGcpProject(workspaceId, userRequest).orElse(null);
 
     // Convert the result to API-speak
     List<ApiEnumeratedJob> apiJobList = new ArrayList<>();
@@ -127,8 +124,7 @@ public class Alpha1ApiController implements Alpha1Api {
               .jobDescription(enumeratedJob.getJobDescription())
               .operationType(enumeratedJob.getOperationType().toApiModel())
               .resourceType(optResource.map(r -> r.getResourceType().toApiModel()).orElse(null))
-              .resource(
-                  optResource.map(r -> apiResourceFromWsmResource(r, gcpProjectId)).orElse(null));
+              .resource(optResource.map(this::apiResourceFromWsmResource).orElse(null));
       apiJobList.add(apiJob);
     }
 
@@ -143,8 +139,7 @@ public class Alpha1ApiController implements Alpha1Api {
 
   // Convert a WsmResource into the API format for enumeration
   @VisibleForTesting
-  public ApiResourceUnion apiResourceFromWsmResource(
-      WsmResource wsmResource, @Nullable String gcpProjectId) {
+  public ApiResourceUnion apiResourceFromWsmResource(WsmResource wsmResource) {
 
     var union = new ApiResourceUnion();
     switch (wsmResource.getStewardshipType()) {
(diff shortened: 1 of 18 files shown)


service/src/main/java/bio/terra/workspace/app/controller/Alpha1ApiController.java
```diff
@@ -200,7 +195,7 @@ public ApiResourceUnion apiResourceFromWsmResource(
             {
               ControlledAiNotebookInstanceResource resource =
                   controlledResource.castToAiNotebookInstanceResource();
-              union.gcpAiNotebookInstance(resource.toApiResource(gcpProjectId));
+              union.gcpAiNotebookInstance(resource.toApiResource());
```
jaycarlton, line 198: This is definitely an improvement. It was awkward having to supply that project ID separately.

service/src/main/java/bio/terra/workspace/app/controller/ResourceController.java
```diff
@@ -102,9 +101,7 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
         workspaceService.getAuthorizedGcpProject(workspaceId, userRequest).orElse(null);
 
     List<ApiResourceDescription> apiResourceDescriptionList =
-        wsmResources.stream()
-            .map(r -> makeApiResourceDescription(r, gcpProjectId))
-            .collect(Collectors.toList());
+        wsmResources.stream().map(r -> makeApiResourceDescription(r)).collect(Collectors.toList());
```
jaycarlton, line 104: nit: you can use a method referenece

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import bio.terra.workspace.db.model.DbResource;
+import bio.terra.workspace.service.resource.model.WsmResource;
+import bio.terra.workspace.service.resource.model.WsmResourceHandler;
+import bio.terra.workspace.service.workspace.GcpCloudContextService;
+import java.util.Optional;
+import javax.annotation.PostConstruct;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
```
jaycarlton, line 13: Do you want/need `@Scope("singleton")`?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public WsmResource makeResourceFromDb(DbResource dbResource) {
+    // Old version of attributes do not have project id, so in that case we look it up
+    ControlledAiNotebookInstanceAttributes attributes =
+        DbSerDes.fromJson(dbResource.getAttributes(), ControlledAiNotebookInstanceAttributes.class);
+    String projectId =
+        Optional.ofNullable(attributes.getProjectId())
+            .orElse(gcpCloudContextService.getRequiredGcpProject(dbResource.getWorkspaceId()));
```
jaycarlton, line 39: use `.orElseGet(() -> gcpCloudContextService...)` so you don't do the lookup unless it's needed.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookInstanceResource.java
```diff
@@ -112,15 +111,15 @@ public InstanceName toInstanceName(String workspaceProjectId) {
...
         .metadata(toApiMetadata())
-        .attributes(toApiAttributes(workspaceProjectId));
+        .attributes(toApiAttributes());
   }
 
-  public ApiGcpAiNotebookInstanceAttributes toApiAttributes(String workspaceProjectId) {
+  public ApiGcpAiNotebookInstanceAttributes toApiAttributes() {
     return new ApiGcpAiNotebookInstanceAttributes()
-        .projectId(workspaceProjectId)
+        .projectId(projectId)
```
jaycarlton, line 122: maybe call `getProjectId()` to make it parallel

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/bqdataset/ControlledBigQueryDatasetHandler.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Override
+  public WsmResource makeResourceFromDb(DbResource dbResource) {
+    // Old version of attributes do not have project id, so in that case we look it up
+    ControlledBigQueryDatasetAttributes attributes =
+        DbSerDes.fromJson(dbResource.getAttributes(), ControlledBigQueryDatasetAttributes.class);
+    String projectId =
+        Optional.ofNullable(attributes.getProjectId())
+            .orElse(gcpCloudContextService.getRequiredGcpProject(dbResource.getWorkspaceId()));
```
jaycarlton, line 39: `orElseGet()`

jaycarlton (approved):
Looking good.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
```diff
@@ -0,0 +1,60 @@
...
+  @Autowired
+  public ControlledAiNotebookHandler(GcpCloudContextService gcpCloudContextService) {
+    this.gcpCloudContextService = gcpCloudContextService;
+  }
+
+  public static ControlledAiNotebookHandler getHandler() {
+    return theHandler;
+  }
+
+  @PostConstruct
```
yuhuyoyo, line 27: i've never seen postConstruct before and it seems to be deprecated? https://levelup.gitconnected.com/stop-using-postconstruct-in-your-java-applications-2a66fb202cb8

Could you autoWire ControlledAiNoteBookHandler into ResourceDao instead of using getHandler?

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -498,10 +498,9 @@ private ApiDeleteControlledGcpAiNotebookInstanceResult fetchNotebookInstanceDele
         controlledResourceService.getControlledResource(workspaceId, resourceId, userRequest);
     try {
       ApiGcpAiNotebookInstanceResource response =
-          controlledResource
-              .castToAiNotebookInstanceResource()
-              .toApiResource(
-                  workspaceService.getAuthorizedRequiredGcpProject(workspaceId, userRequest));
+          controlledResource.castToAiNotebookInstanceResource().toApiResource();
+      // TODO: security check for: workspaceService.getAuthorizedRequiredGcpProject(workspaceId,
```
yuhuyoyo, line 502: resolve TODO or add ticket?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
ddietterich: I could `@Autowire`, but that defeats my goal for new resources _not_ to have to touch `ResourceDao`.

AFAIK, `@PostConstruct` is not deprecated, but it must be used with care. In our system, we do not start taking requests or running any flights until all of the Spring objects have been built. The problems in using `@PostConstruct` are caused by a system in operation, making a race between usage and the execution of the `@PostConstruct`.

All `@PostConstruct` methods are done executing before Spring does injection. All of that is before our `StartupInitializer` is called.

zloery (approved):
LGTM.

I'm wondering if this might break CLI tests in the same way as `DeleteGcpContextWithControlledResource` (i.e. a 400 response becoming a 404), but we can keep an eye out for that in tonight's tests

yuhuyoyo (approved):

PR #542, opened 2022-01-28
Jaycarlton fix pet check

jaycarlton:
Change the user check from using subjectId to validating read access on the workspace. This required making workspaceId required when creating a new Job.


service/src/main/java/bio/terra/workspace/service/workspace/WsmApplicationService.java
```diff
@@ -106,6 +106,7 @@ private WsmWorkspaceApplication commonAbleJob(
             .description(description)
             .flightClass(ApplicationAbleFlight.class)
             .userRequest(userRequest)
+            .workspaceId(workspaceId.toString())
```
ddietterich, line 109: workspace id is set immediately below. Probably should delete line 110.

service/src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -394,15 +397,27 @@ private FlightMap getResultMap(FlightState flightState) {
...
-      if (!StringUtils.equals(flightSubjectId, userRequest.getSubjectId())) {
-        throw new JobUnauthorizedException("Unauthorized");
-      }
+      UUID workspaceId = inputParameters.get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
+
+      // If the workspace doesn't exist, don't try to validate it
+      final WorkspaceService workspaceService = flightBeanBag.getWorkspaceService();
+      final WorkspaceDao workspaceDao = flightBeanBag.getWorkspaceDao();
+
+      workspaceDao.getWorkspaceIfExists(workspaceId).ifPresent(w ->
```
ddietterich, line 418: This is not ideal, since someone can probe completed flights of a deleted workspace.
Is the only case we need to allow the delete workspace case? Unfortunately, there is no information in the FlightState to tell us what the flight class is.

Anyway, I do not have a  better idea for fix right now.

ddietterich (approved):
LGTM - one nit

service/src/main/java/bio/terra/workspace/service/job/JobService.java
jaycarlton: Delete workspace is the primary offender, and it caused tons of test failures. I'm not sure if it's the only one.

jaycarlton: You're saying this change allows probing that would otherwise not be permitted?

ddietterich: Yes. It is a bit obscure. Imagine that I know the UUID of a workspace and the flight ids of some flights that interest me. I am not a member of the workspace. The workspace gets deleted. I can now get the parameters of those flights when previously, I couldn't see them.

Since we often serialize the auth token, there is a window between workspace deletion and token timeout where I could swipe the token and act as the job submitter.

The best defense would be to either stop or delay deleting the Sam permission information. If it existed until all traces of the workspace were gone from the Stairway database, then it would be safe to delete.

How about I make a ticket for this and we go ahead with the change as a short term fix.

ddietterich: I made https://broadworkbench.atlassian.net/browse/PF-1313

service/src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -394,15 +397,30 @@ private FlightMap getResultMap(FlightState flightState) {
...
       FlightState flightState = stairwayComponent.get().getFlightState(jobId);
       FlightMap inputParameters = flightState.getInputParameters();
-      String flightSubjectId =
-          inputParameters.get(JobMapKeys.SUBJECT_ID.getKeyName(), String.class);
-      if (!StringUtils.equals(flightSubjectId, userRequest.getSubjectId())) {
-        throw new JobUnauthorizedException("Unauthorized");
-      }
+      UUID workspaceId = inputParameters.get(WorkspaceFlightMapKeys.WORKSPACE_ID, UUID.class);
+
+      // If the workspace doesn't exist, don't try to validate it
```
zloery, line 414: I'm not totally following why we need this exception. Is the problem that we currently have tests which delete a workspace and then read jobs from that workspace afterwards, which are now failing? 

403 seems like a fine behavior for queries reading jobs from a deleted workspace. If that's the real behavior we want, could we update tests to reflect that instead?

service/src/main/java/bio/terra/workspace/service/job/JobService.java
ddietterich: You are right. WSM does not return the parameter information. So this is not so bad.

service/src/main/java/bio/terra/workspace/service/job/JobService.java
ddietterich: The semantics of all of our deletes is to return success if we find the object already deleted. The thought behind that is to make it easier for callers to simply retry rather than interpret the error.

jaycarlton: The main problem I saw was that deleteWorkspace went from being a no-op when the workspace wasn't found to throwing `MissingWorkspaceException`. This broke all kinds of tests.

zloery: I don't think that's currently our semantic for deleting workspaces, right now we throw a 404 for workspaces which are already deleted.

I think the problem is that without this check `WorkspaceService. deleteWorkspace` will always fail, not just when the workspace wasn't found (which is current behavior). The problem is that `deleteWorkspace` can't read the result of the job it submits because the workspace is gone.

Instead of allowing anyone to read jobs from deleted workspaces, I think it would be cleaner to fix `WorkspaceService. deleteWorkspace`. Either:

- This method calls a new JobService endpoint which gets the job result without doing an auth check. `deleteWorkspace` already verifies the user's permission before doing the delete
- This method catches a `WorkspaceNotFoundException` from the `submitAndWait` call and treats it as success
- Something else?

service/src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -394,15 +415,23 @@ private FlightMap getResultMap(FlightState flightState) {
     return resultMap;
   }
 
+  /**
+   * Ensure the user in the user request has permission to read the workspace associated with the
+   * Job ID. Throws ForbiddenException if not. If the workspace does not exist, treat this as the
+   * user having access (i.e. skip validateWorkspaceAndAction check).
```
zloery, line 421: nit: update comment

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -107,7 +107,7 @@ public UUID createWorkspace(Workspace workspace, AuthenticatedUserRequest userRe
       createJob.addParameter(
           WorkspaceFlightMapKeys.SPEND_PROFILE_ID, workspace.getSpendProfileId().get().getId());
     }
-    return createJob.submitAndWait(UUID.class);
+    return createJob.submitAndWait(UUID.class, false);
```
zloery, line 110: nit: add comment explaining why this skips access check

zloery (approved):
LGTM, thank you for this!

PR #543, opened 2022-02-01
[IA-3185] Add leo app config

jdcanas:
The service-account field will be populated by a terra-helmfile PR soon to come, the ID will be static across envs.

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -101,6 +101,14 @@ workspace:
     - # A special spend profile with no billing account associated with it.
       id: no-billing-account
 
+  application:
+    configurations:
+      identifier: 6397C5FF-F83C-4837-95DB-059198FD89BF
+      name: Leonardo
+      description: Leonardo application SA
+      service-account: leonardo-dev@broad-dsde-dev.iam.gserviceaccount.com
+      state: operating
+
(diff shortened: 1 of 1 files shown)


service/src/main/resources/application.yml
```diff
@@ -101,6 +101,14 @@ workspace:
     - # A special spend profile with no billing account associated with it.
       id: no-billing-account
 
+  application:
+    configurations:
+      identifier: 6397C5FF-F83C-4837-95DB-059198FD89BF
+      name: Leonardo
+      description: Leonardo application SA
+      service-account: leonardo-dev@broad-dsde-dev.iam.gserviceaccount.com
```
ddietterich, line 109: Typically, we make the application.yml the production state and override for other environments. It helps prevents accidental exposure of test stuff.

Note that you can inject this into the configuration via `scripts/write-config.sh`. It makes different settings/gets different secrets for different environments. It generates a file that is optionally included application.yml.

Qi77Qi, line 109: shall we leave this as blank here?

ddietterich (approved):
One concern re: using the dev SA. Otherwise, looks fine to me.

Qi77Qi (approved):
:shipit:

PR #544, opened 2022-02-01
[PF-1264] Make referenced resources integration test parameters

zloery:
This change pulls all referenced resources in the WSM integration tests out to be test parameters instead of hardcoded values. This makes it much easier to write tests which reference different resources, as not all deployments have access to the hardcoded Broad resources.

I had originally planned to refactor some of the larger tests here (specifically, breaking up the multi-resource functionality tests like `UpdateReferenceResources` into resource-type specific tests like `ReferencedGcsBucketLifecycle`), to reduce the number of tests which require many parameters, but that's a big refactor and I think it's ultimately out-of-scope here. PF-1318 to track that.


ddietterich (approved):
LGTM That is a big lift.

yuhuyoyo (approved):
sorry for the delay. i thought i stamped it

PR #546, opened 2022-02-01
PF-1319: set a default post-startup script when none is set.

yuhuyoyo:


yuhuyoyo:
> Is the intent to remove the default startup script from the CLI?
> 
> I also wonder if we should always run our own startup script, even if the user wants to specify something that is launched?

the intent is to move the default logic to WSM so all the clients can have the same default. I moved the script to WSM because I want to move it closest to the call site.

What do you mean by specifying something that is launched? I'd imagine not a lot of user will write their own post-startup-script but if they do, I don't think the ai notebook api allows setting up sequences of post-startup-script (if that's what you mean)

cbookg:
> > Is the intent to remove the default startup script from the CLI?
> > I also wonder if we should always run our own startup script, even if the user wants to specify something that is launched?
> 
> the intent is to move the default logic to WSM so all the clients can have the same default. I moved the script to WSM because I want to move it closest to the call site.
> 
> What do you mean by specifying something that is launched? I'd imagine not a lot of user will write their own post-startup-script but if they do, I don't think the ai notebook api allows setting up sequences of post-startup-script (if that's what you mean)

Basically if the user ever specifies their own script, it will likely break everything (unless they embed all of our logic).  For example if we add support to clone their referenced git repos, it may be confusing that this does't work.

But yeah, there's no easy solution.  So maybe it should just be documented that if you do this, you should build from the default.

yuhuyoyo:
> But yeah, there's no easy solution. So maybe it should just be documented that if you do this, you should build from the default.

added documentation to the poststartupscript attribute in the openapi yaml file.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
```diff
@@ -45,6 +45,10 @@
  * <p>Undo deletes the created notebook instance.
  */
 public class CreateAiNotebookInstanceStep implements Step {
+
+  /** Default post-startup-script when starting a notebook instance. */
+  protected static final String DEFAULT_POST_STARTUP_SCRIPT =
+      "https://raw.githubusercontent.com/DataBiosphere/terra-workspace-manager/main/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh";
```
jaycarlton, line 51: Could/should this be in a configuration file? Do we think we'll ever need to customize that script per-environment?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
jaycarlton, line 1: This is copied from somewhere?

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
yuhuyoyo: this is copied from https://github.com/DataBiosphere/terra-cli/blob/main/notebooks/post-startup.sh. I just want the script to be close to the call site so it's easy to find.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
yuhuyoyo: It does not have environment specific content *yet*. I'd say we keep it simple for now and move it when we need to?

ddietterich (approved):
LGTM

PR #550, opened 2022-02-06
[PF-1297] Refactor DAO uniqueness check

ddietterich:
Add the method `getUniquenessCheckParameters` to all controlled resource classes. The method is called from ResourceDao to get the parameters to check. The DAO then generates a SQL statement and executes the check.

This moves per-resource code out of ReesourceDao. Making the POJO to describe the check seemed better than letting the resources pass in SQL to execute. See what you think.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/db/model/UniquenessCheckParameters.java
+++ b/service/src/main/java/bio/terra/workspace/db/model/UniquenessCheckParameters.java
@@ -1,0 +1,50 @@
+package bio.terra.workspace.db.model;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.lang3.tuple.Pair;
+
+/**
+ * This class is returned from resource handlers to the ResourceDao. It describes
+ * the uniqueness check that the ResourceDao should do. At this time, only
+ * string compares are supported, since those are all that are in use. We can
+ * make this more complex if we need other datatypes.
+ */
+public class UniquenessCheckParameters {
+  /** The scope of the uniqueness check */
+  public enum UniquenessScope {
+    /** search all resources of this type in the WSM database */
+    WSM,
+    /** search resources of this type within the workspace */
+    WORKSPACE;
+  }
+
+  private final UniquenessScope uniquenessScope;
+  private final List<Pair<String, String>> parameters;
+
+  public UniquenessCheckParameters(UniquenessScope uniquenessScope) {
+    this.uniquenessScope = uniquenessScope;
+    this.parameters = new ArrayList<>();
+  }
+
+  public UniquenessScope getUniquenessScope() {
+    return uniquenessScope;
+  }
+
+  public List<Pair<String, String>> getParameters() {
+    return parameters;
+  }
+
+  /**
+   * Add a parameter to be filtered. The filters are JSONB references of the
+   * form:  attributes->>'name' = value
+   *
+   * @param name name of the attribute to check
+   * @param value value of the attribute to check
+   * @return this - for fluent style
+   */
+  public UniquenessCheckParameters addParameter(String name, String value) {
+    parameters.add(Pair.of(name, value));
+    return this;
+  }
+}
--- a/service/src/main/java/bio/terra/workspace/db/model/UniquenessCheckParameters.java
+++ b/service/src/main/java/bio/terra/workspace/db/model/UniquenessCheckParameters.java
@@ -5,10 +5,9 @@ import java.util.List;
 import org.apache.commons.lang3.tuple.Pair;
 
 /**
- * This class is returned from resource handlers to the ResourceDao. It describes
- * the uniqueness check that the ResourceDao should do. At this time, only
- * string compares are supported, since those are all that are in use. We can
- * make this more complex if we need other datatypes.
+ * This class is returned from resource handlers to the ResourceDao. It describes the uniqueness
+ * check that the ResourceDao should do. At this time, only string compares are supported, since
+ * those are all that are in use. We can make this more complex if we need other datatypes.
  */
 public class UniquenessCheckParameters {
   /** The scope of the uniqueness check */
@@ -36,8 +35,8 @@ public class UniquenessCheckParameters {
   }
 
   /**
-   * Add a parameter to be filtered. The filters are JSONB references of the
-   * form:  attributes->>'name' = value
+   * Add a parameter to be filtered. The filters are JSONB references of the form:
+   * attributes->>'name' = value
    *
    * @param name name of the attribute to check
    * @param value value of the attribute to check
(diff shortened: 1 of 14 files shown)


service/src/main/java/bio/terra/workspace/db/model/UniquenessCheckParameters.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Add a parameter to be filtered. The filters are JSONB references of the form:
+   * attributes->>'name' = value
+   *
+   * @param name name of the attribute to check
+   * @param value value of the attribute to check
+   * @return this - for fluent style
+   */
+  public UniquenessCheckParameters addParameter(String name, String value) {
```
jaycarlton, line 45: Yeah, I could see the whole  query getting built here, but that would make this class depend on the details of the database we're using. If we needed to port/migrate from our postgres version, we'd potentially have to pull in and test these classes. So what you have is the best way to split it I think.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/ControlledAzureDiskResource.java
```diff
@@ -71,6 +73,14 @@ public ControlledAzureDiskResource(DbResource dbResource) {
     validate();
```
jaycarlton, line 73: Aside: I wonder if we'd benefit from an ugly separation of the parts of the controlled resource names, e.g. `Controlled_Azure_DiskResource`. It'd be easier to parse at a glance, as we'd know which parts went to which concept.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/ControlledAzureVmResource.java
```diff
@@ -88,6 +90,14 @@ public ControlledAzureVmResource(DbResource dbResource) {
     validate();
   }
 
+  /** {@inheritDoc} */
+  @Override
+  public Optional<UniquenessCheckParameters> getUniquenessCheckParameters() {
+    return Optional.of(
+        new UniquenessCheckParameters(UniquenessScope.WORKSPACE)
+            .addParameter("vmName", getVmName()));
```
jaycarlton, line 98: vm name is globally unique, like bucket names?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookInstanceResource.java
```diff
@@ -92,6 +94,15 @@ public Builder toBuilder() {
         .projectId(getProjectId());
   }
 
+  /** {@inheritDoc} */
+  @Override
+  public Optional<UniquenessCheckParameters> getUniquenessCheckParameters() {
+    return Optional.of(
+        new UniquenessCheckParameters(UniquenessScope.WORKSPACE)
+            .addParameter("instanceId", getInstanceId())
```
jaycarlton, line 102: ISTR this was only unique within the project ID.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/bqdataset/ControlledBigQueryDatasetResource.java
```diff
@@ -80,6 +82,14 @@ public Builder toBuilder() {
         .projectId(projectId);
   }
 
+  /** {@inheritDoc} */
+  @Override
+  public Optional<UniquenessCheckParameters> getUniquenessCheckParameters() {
+    return Optional.of(
+        new UniquenessCheckParameters(UniquenessScope.WORKSPACE)
+            .addParameter("datasetName", getDatasetName()));
```
jaycarlton, line 90: same question about project ID.

service/src/main/java/bio/terra/workspace/db/model/UniquenessCheckParameters.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.lang3.tuple.Pair;
+
+/**
+ * This class is returned from resource handlers to the ResourceDao. It describes the uniqueness
+ * check that the ResourceDao should do. At this time, only string compares are supported, since
+ * those are all that are in use. We can make this more complex if we need other datatypes.
+ */
+public class UniquenessCheckParameters {
```
jaycarlton, line 12: I'd call these attributes rather than parameters, to match usage elsewhere. They  only become parameters when you sql-ify them.

jaycarlton (approved):
I think this makes sense. It'll keep the dao from growing too fast.

service/src/main/java/bio/terra/workspace/db/model/UniquenessCheckParameters.java
```diff
@@ -0,0 +1,49 @@
...
+  /** The scope of the uniqueness check */
+  public enum UniquenessScope {
+    /** search all resources of this type in the WSM database */
+    WSM,
+    /** search resources of this type within the workspace */
+    WORKSPACE;
+  }
+
+  private final UniquenessScope uniquenessScope;
+  private final List<Pair<String, String>> parameters;
```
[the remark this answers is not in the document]
ddietterich: I considered using a `Set`.  It is probably bit-bumming on my part, but using Set or Map means an index structure gets allocated when I just need a list, typically of one thing. So I used Pair.

And yes, good point: I'll change the constructor and do fluent setting for scope as well..

service/src/main/java/bio/terra/workspace/db/model/UniquenessCheckParameters.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * This class is returned from resource handlers to the ResourceDao. It describes the uniqueness
+ * check that the ResourceDao should do. At this time, only string compares are supported, since
+ * those are all that are in use. We can make this more complex if we need other datatypes.
+ */
+public class UniquenessCheckParameters {
+  /** The scope of the uniqueness check */
+  public enum UniquenessScope {
+    /** search all resources of this type in the WSM database */
+    WSM,
```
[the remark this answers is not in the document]
ddietterich: I'm OK with that. But for something like bucket, we do not test globally. We only test what is in the WSM database.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/ControlledAzureVmResource.java
ddietterich: I just transcribed the existing code. It could be it is wrong.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookInstanceResource.java
yuhuyoyo: can there be multiple projects from different cloud in one workspace?

jaycarlton: But never more than one context per cloud platform, right?

ddietterich:
Right!

PR #551, opened 2022-02-07
[PF-1326] move casts to resource classes

ddietterich:
This PR removes the cast methods from the ControlledResource and ReferenceResource classes and puts them in each resource class.

I went looking for a way to make an abstract method that would be required to be implemented by subclasses, but still cast the return to the subclass. I did this in WsmResource:
```
public abstract <T> T castByEnum(WsmResourceType expectedType);
```
And then each subclass has exactly the same code snippet:
```
  @SuppressWarnings("unchecked")
  public <T> T castByEnum(WsmResourceType expectedType) {
    if (getResourceType() != expectedType) {
      throw new BadRequestException(String.format("Resource is not a %s", expectedType));
    }
    return (T) this;
  }
```

I _really_ like that it can be abstract. There is a penalty: you cannot cascade directly off the result; that is,
it seems you cannot do this:
```
  WsmResource resource = ...
  String name = resource.castByEnum(CONTROLLED_GCP_GCS_BUCKET).getBucketName();
```
You have to do this:
```
  ControlledGcpGcsBucket bucketResource = resource.castByEnum(CONTROLLED_GCP_GCS_BUCKET);
  String name =  bucketResource.getBucketName();
```

I started by using static members, but that is annoying because they cannot easily be mocked in the tests, and we do that a lot. Anyway, I'll be interested in your thoughts on this approach.


jaycarlton:
You should be able to put the implementation in the base class (or in an interface with a `default` method) if it's the same snippet in each child class. Unless there's something tricksy about the template that makes it necessary to paste everywhere.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/ControlledAzureDiskResource.java
```diff
@@ -73,9 +74,23 @@ public ControlledAzureDiskResource(DbResource dbResource) {
...
 
+  public static ControlledAzureDiskResource.Builder builder() {
+    return new ControlledAzureDiskResource.Builder();
+  }
+
   /** {@inheritDoc} */
   @Override
-  public Optional<UniquenessCheckAttributes> getUniquenessCheckParameters() {
+  @SuppressWarnings("unchecked")
+  public <T> T castByEnum(WsmResourceType expectedType) {
```
jaycarlton, line 84: I'd name this `toResourceType()` instead. The fact that it's a cast is an implementation detail.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
```diff
@@ -147,4 +147,21 @@ public static String getFlightErrorMessage(FlightState subflightState) {
...
+   * Get a value from one of the flight maps and check that it is not null. If it is null, throw.
+   *
+   * @param flightMap input or working map
+   * @param key string key to lookup in the map
+   * @param tClass class to return
+   * @param <T> generic
+   * @return T
+   */
+  public static <T> T getRequired(FlightMap flightMap, String key, Class<T> tClass) {
+    var value = flightMap.get(key, tClass);
```
jaycarlton, line 161: I'm very impressed that this line works.

service/src/main/java/bio/terra/workspace/service/resource/controlled/model/ControlledResource.java
```diff
@@ -153,54 +144,6 @@ public void validate() {
     }
   }
 
-  // Double-checked down casts when we need to re-specialize from a ControlledResource
```
jaycarlton, line 156: 👯‍♀️

ddietterich:
> You should be able to put the implementation in the base class (or in an interface with a default method) if it's the same snippet in each child class. Unless there's something tricksy about the template that makes it necessary to paste everywhere.

There is small magic to get the correct result class. When you dispatch (`resource.castByEnum(...)`), Java traverses the class hierarchy and runs the method of the subclass that returns its type. If you put the snippet in the base class, all you do is cast to the base class.

jaycarlton (approved):

jaycarlton:
> > You should be able to put the implementation in the base class (or in an interface with a default method) if it's the same snippet in each child class. Unless there's something tricksy about the template that makes it necessary to paste everywhere.
> 
> There is small magic to get the correct result class. When you dispatch (`resource.castByEnum(...)`), Java traverses the class hierarchy and runs the method of the subclass that returns its type. If you put the snippet in the base class, all you do is cast to the base class.

I wonder if you could minimize the copied code by doing the type check in the base class and having a call to get the runtime class be abstract. It's more of an academic exercise than anything though.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
ddietterich: Wow! Didn't notice that. PFM

ddietterich:
> I wonder if you could minimize the copied code by doing the type check in the base class and having a call to get the runtime class be abstract. It's more of an academic exercise than anything though.

Hmm... I don't really need to check the resource type, I suppose. I could check `instanceof`, but I like having the enum target in the caller so that when you read the code you understand the intention.

yuhuyoyo (approved):

PR #552, opened 2022-02-07
[PF-1298] Retrieve all cloud contexts in a step

ddietterich:
For GCP, we have a step that retrieves the cloud context and puts it in the working map. For Azure, we were retrieving the context in the Flight constructor and passing it as a parameter. This consolidates the cloud context retrieval in a new step calls `GetCloudContextStep`.

In general, avoid doing any interesting code in the flight constructor. The error handling from exceptions thrown in the constructor are not handled as well as from a step. Also, there is efficiency gained if the flight is paused and restarted.

This is part 1 of 2 that will extract resource steps into the resource object and remove the big switch statements. All part of making it simpler and more localized to add resources.


Qi77Qi (approved):
lgtm...once tests pass

zloery (approved):
LGTM!

PR #553, opened 2022-02-08
[PF-1298] Make resource supply steps for create and delete

ddietterich:
Continuing with refactoring, this removes the big switch statements from create and delete resource and moves the code into the resource objects.

I made one change on the delete path beyond this refactor. I re-used the GetCloudContextStep I made for the create flight. That lets the steps get the context from the working map instead of passing around the DAO and making repeated calls to it.


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/ControlledAzureDiskResource.java
```diff
@@ -82,6 +88,32 @@ public ControlledAzureDiskResource(DbResource dbResource) {
...
   }
 
+  /** {@inheritDoc} */
+  @Override
+  public void addCreateSteps(
+      CreateControlledResourceFlight flight,
+      String petSaEmail,
+      AuthenticatedUserRequest userRequest,
+      FlightBeanBag flightBeanBag) {
+    RetryRule cloudRetry = RetryRules.cloud();
```
jaycarlton, line 98: Interesting. I had assumed these would return a `Step` or list of steps, rather than be void and take in the target flight. Might be more mockable/unit testable that way.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/ControlledAzureDiskResource.java
```diff
@@ -82,6 +88,32 @@ public ControlledAzureDiskResource(DbResource dbResource) {
...
+        cloudRetry);
+    flight.addStep(
+        new CreateAzureDiskStep(
+            flightBeanBag.getAzureConfig(), flightBeanBag.getCrlService(), this),
+        cloudRetry);
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public void addDeleteSteps(DeleteControlledResourceFlight flight, FlightBeanBag flightBeanBag) {
```
jaycarlton, line 110: might want to pass in the services you need explicitly instead of adding usages of the bean bag if you're planning to pull it out soon. Also from a Law of Demeter standpoint it makes sense to hand the cashier the money instead of the pocket with the wallet in it.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DeleteAzureDiskStep.java
```diff
@@ -20,40 +19,30 @@
...
-    var wsmResource = resourceDao.getResource(workspaceId, resourceId);
-    var disk = wsmResource.castToControlledResource().castToAzureDiskResource();
+    final AzureCloudContext azureCloudContext =
+        context
+            .getWorkingMap()
+            .get(ControlledResourceKeys.AZURE_CLOUD_CONTEXT, AzureCloudContext.class);
 
     ComputeManager computeManager = crlService.getComputeManager(azureCloudContext, azureConfig);
     var azureResourceId =
         String.format(
```
jaycarlton, line 41: Seems like an odd place for this knowledge to live. I'd put a method on the resource itself to return the fully qualified ID.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/ip/DeleteAzureIpStep.java
```diff
@@ -68,7 +57,9 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
   @Override
   public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
```
jaycarlton, line 58: I keep meaning to make a `OneWayStep` class that handles the erroring on undo, and/or a variant of that with a no-op undo, since we have to write this function unnecessarily in a lot of places.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/DeleteAzureVmStep.java
```diff
@@ -5,59 +5,49 @@
...
-    var wsmResource = resourceDao.getResource(workspaceId, resourceId);
-    var vm = wsmResource.castToControlledResource().castToAzureVmResource();
+    final AzureCloudContext azureCloudContext =
+        context
+            .getWorkingMap()
+            .get(ControlledResourceKeys.AZURE_CLOUD_CONTEXT, AzureCloudContext.class);
 
     ComputeManager computeManager = crlService.getComputeManager(azureCloudContext, azureConfig);
     var azureResourceId =
         String.format(
```
jaycarlton, line 38: same question: is there a better place for this logic to live?

jaycarlton (approved):
Thanks for tackling this!

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DeleteAzureDiskStep.java
ddietterich: Retrieving the cloud context is a database transaction. I don't want to hit the database on every step of a flight, so factoring the fetch into a step and passing the data through seemed reasonable.

We could directly attach the cloud context to the resources in ControlledResource class and fetch that (if applicable) whenever we fetch the resource. That might be a worthy (but good sized) refactor:
- we would not have to go looking for the project id for GCP resources
- we would not need the separate transaction to lookup the cloud context
- it would make clear the requirement to have the context in place for resource creation, because it would be in the resource object.

I'll make a ticket for this work.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/ip/DeleteAzureIpStep.java
ddietterich: We should put that into Stairway. There are two undo cases that we code all the time:
- undo that returns success
- undo that triggers dismal failure
I'd like both of those as Step derivatives in Stairway

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/bqdataset/ControlledBigQueryDatasetResource.java
```diff
@@ -91,6 +97,33 @@ public Builder toBuilder() {
...
+  public void addCreateSteps(
+      CreateControlledResourceFlight flight,
+      String petSaEmail,
+      AuthenticatedUserRequest userRequest,
+      FlightBeanBag flightBeanBag) {
+    RetryRule cloudRetry = RetryRules.cloud();
+    // Unlike other resources, BigQuery datasets set IAM permissions at creation time to avoid
+    // unwanted defaults from GCP.
+    flight.addStep(
+        new CreateBigQueryDatasetStep(
```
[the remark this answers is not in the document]
ddietterich: Sure, we could do that. The only drawback is that the resource classes are getting pretty big and that would make them a good bit bigger.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/CreateControlledResourceFlight.java
```diff
@@ -40,6 +23,12 @@
   private final RetryRule gcpRetryRule = RetryRules.cloud();
   private final RetryRule dbRetryRule = RetryRules.shortDatabase();
 
+  // addStep is protected in Flight, so make an override that is public
+  @Override
```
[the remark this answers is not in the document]
ddietterich: I want to change Stairway to make addStep public and remove this crutch. There is no reason it should be protected.

zloery (approved):
LGTM!

PR #554, opened 2022-02-08
[PF-1318] Refactor all integration tests

zloery:
At a high level, this refactors a number of WSM's integration tests from being function-based to resource type-based. In particular, we had a number of tests which executed the same functionality across all resource types, and these tests were getting larger and more difficult to maintain as we added more resource types. To fix that, I've done the following:

- Break up `CloneReferencedResources`, `ValidateReferencedResources`, and `UpdateReferencedResources`. The functionality covered in these tests is now inside resource type-specific tests.
- Created `ReferencedBigQueryResourceLifecycle`, `ReferencedGcsResourceLifecycle`, `ReferencedGitRepoLifecycle`, and `ReferencedTdrSnapshotLifecycle` tests to cover all operations on specific resource types. As we add new resource types in the future, contributors can add similar tests exercising resource behavior rather than adding to large functionality-based tests.
- Deleted the `CloneBigqueryDataset` and `CloneGcsBucket` tests and moved their coverage into `ControlledBigQueryDatasetLifecycle` and `ControlledGcsBucketLifecycle`, respectively.
- Reduced the number and kinds of resources being created for `CloneWorkspace`, `EnumerateResources`, and `EnumerateJobs`. Previously, we were creating one of every GCP resource type in each of those tests, despite their implementations being mostly generic. This approach should be simpler to maintain as we add more resource types.
- I didn't modify the tests using the deprecated snapshot endpoints, my understanding is we'll be removing those soon.

Although it's not included in the full integration suite, I've also locally verified that the modified `EnumerateJobs` runs as expected.


integration/src/main/java/scripts/testscripts/ReferencedTdrSnapshotLifecycle.java
```diff
@@ -0,0 +1,185 @@
+package scripts.testscripts;
```
ddietterich, line 1: Elsewhere we call this DataRepoSnapshot.

ddietterich (commented):
Thanks for this! Since we don't have all the resource in the enumerate test, I think we should have each resource test enumerating to make sure it is hooked up properly.

integration/src/main/java/scripts/utils/MultiResourcesUtils.java
```diff
@@ -33,131 +26,72 @@ private static String makeName() {
...
+        GcsBucketUtils.makeGcsBucketReference(
+            privateBucket.getAttributes(),
+            referencedGcpResourceApi,
+            workspaceId,
+            makeName(),
+            CloningInstructionsEnum.NOTHING);
+    GcpBigQueryDatasetResource datasetReference =
+        BqDatasetUtils.makeBigQueryDatasetReference(
+            sharedDataset.getAttributes(), referencedGcpResourceApi, workspaceId, makeName());
+    return List.of(
```
jaycarlton, line 87: You could make a map from resoruce type to metadata instead of a list.

integration/src/main/resources/configs/integration/ReferencedBigQueryResourceLifecycle.json
```diff
@@ -0,0 +1,23 @@
...
+    {
+      "name": "ReferencedBigQueryResourceLifecycle",
+      "numberOfUserJourneyThreadsToRun": 1,
+      "userJourneyThreadPoolSize": 1,
+      "expectedTimeForEach": 5,
+      "expectedTimeForEachUnit": "MINUTES",
+      "parameters": {
+        "spend-profile-id": "wm-default-spend-profile",
+        "bq-dataset": "projects/terra-kernel-k8s/datasets/terra_wsm_test_dataset",
+        "bq-table": "projects/terra-kernel-k8s/datasets/terra_wsm_test_dataset/tables/terra wsm test data table",
```
jaycarlton, line 17: I'm confused how the spaces work in the table name. Like this doesn't look like a valid FQID.

integration/src/main/resources/configs/integration/ReferencedGitRepoLifecycle.json
```diff
@@ -0,0 +1,21 @@
...
+  "testScripts": [
+    {
+      "name": "ReferencedGitRepoLifecycle",
+      "numberOfUserJourneyThreadsToRun": 1,
+      "userJourneyThreadPoolSize": 1,
+      "expectedTimeForEach": 5,
+      "expectedTimeForEachUnit": "MINUTES",
+      "parameters": {
+        "spend-profile-id": "wm-default-spend-profile",
+        "git-repo": "git@github.com:DataBiosphere/terra-workspace-manager.git"
```
jaycarlton, line 16: If you teach the code about its own repo location, it'll activate SKYNET.

integration/src/main/resources/configs/integration/ReferencedTdrSnapshotLifecycle.json
```diff
@@ -0,0 +1,23 @@
...
+    {
+      "name": "ReferencedTdrSnapshotLifecycle",
+      "numberOfUserJourneyThreadsToRun": 1,
+      "userJourneyThreadPoolSize": 1,
+      "expectedTimeForEach": 5,
+      "expectedTimeForEachUnit": "MINUTES",
+      "parameters": {
+        "spend-profile-id": "wm-default-spend-profile",
+        "data-repo-instance": "terra",
+        "snapshot-id": "97b5559a-2f8f-4df3-89ae-5a249173ee0c",
```
jaycarlton, line 17: why not make the snapshot ids an array-valued field?

jaycarlton (approved):
I went through most of it and assumed the rest was good. Thanks for sorting this out!

integration/src/main/java/scripts/testscripts/ReferencedBigQueryResourceLifecycle.java
```diff
@@ -0,0 +1,354 @@
...
+    GcpBigQueryDataTableResource fetchedDataTable =
+        referencedGcpResourceApi.getBigQueryDataTableReference(
+            getWorkspaceId(), bqDataTableResourceId);
+    assertEquals(referencedDataTable, fetchedDataTable);
+
+    // Create a second workspace to clone references into, owned by the same user
+    destinationWorkspaceId = UUID.randomUUID();
+    createWorkspace(destinationWorkspaceId, getSpendProfileId(), workspaceApi);
+    // Clone references
+    CloneReferencedGcpBigQueryDatasetResourceResult datasetCloneResult =
```
yuhuyoyo, line 109: similar to update, can we extract a cloneReferenceAndVerify

integration/src/main/java/scripts/testscripts/ReferencedBigQueryResourceLifecycle.java
```diff
@@ -0,0 +1,354 @@
...
+        BqDatasetUtils.makeBigQueryDataTableReference(
+            referencedBqTableAttributes,
+            referencedGcpResourceApi,
+            getWorkspaceId(),
+            MultiResourcesUtils.makeName(),
+            CloningInstructionsEnum.REFERENCE);
+    UUID bqDataTableResourceId = referencedDataTable.getMetadata().getResourceId();
+
+    // Get the references
+    GcpBigQueryDatasetResource fetchedDataset =
```
yuhuyoyo, line 97: extract getReferencesAndVerify

integration/src/main/java/scripts/testscripts/ReferencedGcsResourceLifecycle.java
```diff
@@ -0,0 +1,364 @@
...
+    GcpGcsObjectResource referencedGcsFolder =
+        GcsBucketObjectUtils.makeGcsObjectReference(
+            gcsFolderAttributes,
+            referencedGcpResourceApi,
+            getWorkspaceId(),
+            MultiResourcesUtils.makeName(),
+            CloningInstructionsEnum.REFERENCE);
+    folderResourceId = referencedGcsFolder.getMetadata().getResourceId();
+
+    // Get the references
```
yuhuyoyo, line 117: extract separate method, same for other resourceType test

integration/src/main/java/scripts/testscripts/ReferencedGcsResourceLifecycle.java
```diff
@@ -0,0 +1,364 @@
...
+    assertEquals(fineGrainedBucket, fetchedFineGrainedBucket);
+    GcpGcsObjectResource fetchedGcsFile =
+        referencedGcpResourceApi.getGcsObjectReference(getWorkspaceId(), fileResourceId);
+    assertEquals(referencedGcsFile, fetchedGcsFile);
+    GcpGcsObjectResource fetchedGcsFolder =
+        referencedGcpResourceApi.getGcsObjectReference(getWorkspaceId(), folderResourceId);
+    assertEquals(referencedGcsFolder, fetchedGcsFolder);
+
+    // Create a second workspace to clone references into, owned by the same user
+    destinationWorkspaceId = UUID.randomUUID();
```
yuhuyoyo, line 132: extract separate method for cloning

yuhuyoyo (approved):

integration/src/main/java/scripts/testscripts/ReferencedTdrSnapshotLifecycle.java
zloery: sgtm, renamed the test to `ReferencedDataRepoSnapshotLifecycle`

integration/src/main/java/scripts/utils/MultiResourcesUtils.java
zloery: We could, but at least right now callers don't make any assumptions about the types resources in the returned list, which I prefer

integration/src/main/resources/configs/integration/ReferencedBigQueryResourceLifecycle.json
zloery: It looks a bit weird, but spaces are explicitly allowed in BQ table names: https://cloud.google.com/bigquery/docs/tables#table_naming

integration/src/main/resources/configs/integration/ReferencedTdrSnapshotLifecycle.json
zloery: The current test config schema has `parameters` as a `Map<String,String>`. We could pass an array as a String and parse it ourselves, but that seemed messier than just using a second key. Additionally this is the only test that uses > 1 snapshot ID, and there's one other test which uses a single snapshot ID.

integration/src/main/java/scripts/testscripts/CloneWorkspace.java
```diff
@@ -204,52 +182,49 @@ protected void doSetup(
             null,
             CloningInstructionsEnum.RESOURCE);
 
-    // Create reference to GCS bucket with COPY_REFERENCE
-    final ApiClient apiClient = ClientTestUtils.getClientForTestUser(sourceOwnerUser, server);
-    ReferencedGcpResourceApi referencedGcpResourceApi = new ReferencedGcpResourceApi(apiClient);
+    // Create reference to the shared GCS bucket in this workspace with COPY_REFERENCE
+    ReferencedGcpResourceApi referencedGcpResourceApi =
+        ClientTestUtils.getReferencedGpcResourceClient(sourceOwnerUser, server);
```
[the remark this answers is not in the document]
zloery: A ~10 month old typo, nice find!

integration/src/main/java/scripts/testscripts/ReferencedBigQueryResourceLifecycle.java
zloery: Done - I've pulled out separate methods for get/enumerate, validate, clone, and update for each of the referenced resource lifecycle tests.

PR #555, opened 2022-02-09
[WOR-122] Fix azure cloud context creation

aherbst-broad:
## Context
Azure cloud context creation fails with a 500, since it tries to marshal the result AzureCloudContext to a GcpCloudContext (this also affects the `getCloudContextResult` endpoint).

## This PR
* Creates a `CloudContextHolder` class that holds references to both a `GcpCloudContext` and an `AzureCloudContext`. This class is used by both flavors of cloud context creation flights as the result format.
* Adjusts `WorkspaceApiController#getCreateCloudContextResult` to conditionally return an azure cloud context as well as a GcpCloudContext if present (the previous implementation would 500 if called on an azure cloud context creation flight, due to the aforementioned marshalling issues).
* Adds `createGetDeleteAzureContext` test to the `WorkspaceServiceTest`
   * A corresponding delete azure cloud context flight was not present. As part of this PR I have implemented a basic implementation of such a flight for consumption by the test that just deletes the cloud context DB record. I expect we'll need to hook into this and augment it for deleting azure cloud contexts via the API


ddietterich:
Quick note on the openapi change on the draft. I don't think you should bother with the cloud platform parameter. This is just a database read, so retrieving both contexts in one go is not a problem. (We do want to support multiple cloud contexts in one workspace.)

aherbst-broad:
@ddietterich Not sure I understand -- this is retrieving the results of the creation of a specific cloud context, either azure or gcp. Retrieving the workspace details should definitely retrieve all cloud contexts, but in this instance it seems like we should be returning just the context that was being created.

ddietterich:
I should have started with a question: why do you need a cloud platform parameter on the async completion call?
The cloud platform was specified on the request. The client got an id back for the request. So the client can't really change their mind about what platform they asked to be made.

aherbst-broad:
@ddietterich a fair point, agreed that it's awkward to ask the client to restate the cloud platform to fetch the results during a subsequent API call to `getCloudContextResult`. In the `getCloudContextResult` codepath it seems we need to know the correct class to serialize the job result to so we can call `jobService.retrieveAsyncJobResult`  with the correct generic type (https://github.com/DataBiosphere/terra-workspace-manager/blob/43accccbdce5c253f93e13659fd8ab32b559a1fb/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java#L488 )

Alternatively, we may want to adjust the type on the fetch of the `AsyncJobResult` to `Void` and just return the high-level job status without the context details. This would require callers to make a second call to get the workspace to fetch the context details should they be needed. This is fine for my use-case, not sure about elsewhere.

ddietterich:
I'd suggest making a little POJO with both cloud context pointers in it and return that from the flights.
Then we easily copy the output of that into the new API response.

aherbst-broad:
@ddietterich 👍 excellent idea, I've take a cut at that in this PR.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -54,7 +56,9 @@
 import org.mockito.Mockito;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ActiveProfiles;
 
+@ActiveProfiles({"azure", "azure-test"})
```
aherbst-broad, line 61: I am unsure about adding these profiles to this test, would love thoughts on this one (factor into another azure specific workspace service test?)

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2913,6 +2913,8 @@ components:
           $ref: '#/components/schemas/JobReport'
         gcpContext:
           $ref: '#/components/schemas/GcpContext'
+        azureContext:
```
aherbst-broad, line 2916: This is an additive API modification to allow for the inclusion of an azure cloud context in a cloud context result response body.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -343,6 +339,26 @@ public void deleteGcpCloudContext(UUID workspaceId, AuthenticatedUserRequest use
         .submitAndWait(null);
   }
 
+  public void deleteAzureCloudContext(UUID workspaceId, AuthenticatedUserRequest userRequest) {
```
ddietterich, line 342: Thank you for implementing this!

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
ddietterich: Good question. Since azure is not enabled in all environments, I would suggest continuing to segregate the azure tests. You can make derive your test class from `BaseAzureTest` and not worry about setting the profiles.

ddietterich (commented):
Thanks for doing this work! It is looking good. I have a number of comments and suggestions.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -477,26 +478,43 @@ private ApiDataReferenceDescription makeApiDataReferenceDescription(
...
-    } else {
-      gcpContext = null;
+      gcpContext =
+          Optional.ofNullable(jobResult.getResult().getGcpCloudContext())
+              .map(c -> new ApiGcpContext().projectId(c.getGcpProjectId()))
+              .orElse(null);
+    }
+
+    ApiAzureContext azureContext = null;
+    if (jobResult.getJobReport().getStatus().equals(StatusEnum.SUCCEEDED)) {
```
cahrens, line 501: Why not put this in the same conditional block as above (line 493)?

service/src/test/java/bio/terra/workspace/service/workspace/AzureWorkspaceTest.java
```diff
@@ -0,0 +1,64 @@
...
+    workspaceService.createAzureCloudContext(
+        request.getWorkspaceId(), jobId, userRequest, "/fake/value", azureCloudContext);
+    jobService.waitForJob(jobId);
+
+    assertNull(jobService.retrieveJobResult(jobId, Object.class, userRequest).getException());
+    assertTrue(
+        testUtils
+            .getAuthorizedAzureCloudContext(request.getWorkspaceId(), userRequest)
+            .isPresent());
+    workspaceService.deleteAzureCloudContext(request.getWorkspaceId(), userRequest);
```
cahrens, line 60: Is it important to ensure that the cloud context is deleted? If so, this should be put into a finally block or test cleanup method.

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteAzureContextFlight.java
```diff
@@ -0,0 +1,26 @@
...
+  public DeleteAzureContextFlight(FlightMap inputParameters, Object applicationContext) {
+    super(inputParameters, applicationContext);
+
+    FlightBeanBag appContext = FlightBeanBag.getFromObject(applicationContext);
+    UUID workspaceId =
+        UUID.fromString(inputParameters.get(WorkspaceFlightMapKeys.WORKSPACE_ID, String.class));
+
+    RetryRule retryRule = RetryRules.cloudLongRunning();
+
+    // TODO cleanup any azure resources here
```
marctalbott, line 20: Can we create a ticket for this or will it be done as part of a different ticket?

marctalbott (approved):

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteAzureContextFlight.java
aherbst-broad: @marctalbott good idea, [done](https://broadworkbench.atlassian.net/browse/WOR-139)

service/src/test/java/bio/terra/workspace/service/workspace/AzureWorkspaceTest.java
aherbst-broad: I don't think so. We're using a mock sam here so no external resources are leaked and we're not allocating any other resources as part of this test. In the event that we create real resources, we can implement an `@AfterAll` to cleanup anything remaining.

cahrens (approved):

ddietterich (approved):
LGTM

ddietterich:
I went ahead and merged this, since it fixes a separate bug that is in @Qi77Qi's way.
I will follow it up with a PR that will include a test of the problem they ran into.

Qi77Qi:
thanks @ddietterich !

PR #557, opened 2022-02-15
[WOR-137] Point at new snapshots for ITs

aherbst-broad:
## Context
The [`ReferencedDataRepoSnapshotLifecycle`](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/integration/src/main/java/scripts/testscripts/ReferencedDataRepoSnapshotLifecycle.java) IT is failing because the snapshots it was consuming were deleted. 

## This PR
* Points at 2 newly created snapshots that were created and explicitly named for this test
   * I need to look into creating a dev group that can own these snapshots, for now they are under my dev acct and readable by the IT SAs.

### DIFF
--- a/integration/src/main/resources/configs/integration/DataReferenceLifecycle.json
+++ b/integration/src/main/resources/configs/integration/DataReferenceLifecycle.json
@@ -14,7 +14,7 @@
       "parameters": {
         "spend-profile-id": "wm-default-spend-profile",
         "data-repo-instance": "terra",
-        "snapshot-id": "97b5559a-2f8f-4df3-89ae-5a249173ee0c"
+        "snapshot-id": "220a2d18-56f7-492a-b7b3-2bc7d79e54b2"
       }
     }
   ],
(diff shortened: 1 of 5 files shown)


ddietterich:
Thank you for picking this up! I had started working in parallel on it, but had not gotten far.

zloery:
One thing that confused me a bit: there are a bunch of other tests which reference snapshot `97b5559a-2f8f-4df3-89ae-5a249173ee0c`, but only this one failed. I did a bit of digging and it looks like we actually didn't lose that snapshot (or it's been restored), but we did lose our old secondary snapshot (`de189c9b-ef96-4ade-9970-5829056083e3`) which was only used in this one test.

For the sake of consistency, could you either revert the change to `97b5559a-2f8f-4df3-89ae-5a249173ee0c` here, or if you'd prefer to use the new snapshots for easier bookkeeping on the TDR side, update the other test configs where we also reference that snapshot? In total I see it used in:

- `integration/DataReferenceLifecycle.json`
- `integration/EnumerateDataReferences.json`
- `integration/ReferencedDataRepoSnapshotLifecycle.json`
- `perf/DataReferenceLifecycle.json`
- `perf/EnumerateDataReferences.json`

zloery (approved):
Thank you for the fix!

ddietterich (approved):
LGTM

aherbst-broad:
@zloery @ddietterich no problem. I am a bit mystified what happened in this case with the secondary snapshot, but these should work.

PR #559, opened 2022-02-16
[PF-1350] Modify test to exercise azure context create service call

ddietterich:
The missing parameter to azure context create was fixed by WOR-122.
There was no test that caught that problem. I modified one of the service tests to exercise that path so this won't regress.


aherbst-broad (approved):
lgtm

PR #560, opened 2022-02-16
[PF-1090] Mark bucket and bigquery response properties as required

cbookg:
They are listed as optional, but will be present on success.


ddietterich (approved):
LGTM

zloery (approved):

PR #562, opened 2022-02-18
Refactor Gradle File

jaycarlton:
Split `build.gradle` into include files based on purpose. With this change, the build file is only 36 lines long, and the longest (but simplest) included file is 87 lines.

Also tried to add helpful/accurate comments in places.

Based on original PR https://github.com/DataBiosphere/terra-workspace-manager/pull/363/.

### DIFF
--- a/service/gradle/deploy.gradle
+++ b/service/gradle/deploy.gradle
@@ -1,0 +1,24 @@
+import java.time.ZonedDateTime
+
+// Deploy config
+jib {
+  from {
+    // see https://github.com/broadinstitute/dsp-appsec-blessed-images/tree/main/jre
+    // Google's distroless images are openjdk, this is the simplest with adoptopenjdk
+    image = "us.gcr.io/broad-dsp-gcr-public/base/jre:11-debian"
+  }
+  extraDirectories {
+    paths = [file(jibExtraDirectory)]
+  }
+  container {
+    filesModificationTime = ZonedDateTime.now().toString() // to prevent ui caching
+    jvmFlags = [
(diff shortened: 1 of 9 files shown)


service/gradle/deploy.gradle
```diff
@@ -0,0 +1,24 @@
...
+    // Google's distroless images are openjdk, this is the simplest with adoptopenjdk
+    image = "us.gcr.io/broad-dsp-gcr-public/base/jre:11-debian"
+  }
+  extraDirectories {
+    paths = [file(jibExtraDirectory)]
+  }
+  container {
+    filesModificationTime = ZonedDateTime.now().toString() // to prevent ui caching
+    jvmFlags = [
+            "-agentpath:" + cloudProfilerLocation + "/profiler_java_agent.so=" +
```
ddietterich, line 16: That variable is defined in the profiler part. I think it would be better if we avoided references across the leaf parts.

So maybe combine or set in the top-level file?

service/gradle/profiler.gradle
```diff
@@ -0,0 +1,19 @@
+// Download and extract the Cloud Profiler Java Agent
+ext {
+  // where to place the Cloud Profiler agent in the container
+  cloudProfilerLocation = "/opt/cprof"
```
ddietterich, line 4: This is used outside of this leaf file.

ddietterich (commented):
Looks good. One concern

ddietterich (approved):
LGTM

zloery (approved):

PR #563, opened 2022-02-22
[PF-1333]: add service account scope when creating notebook instance

yuhuyoyo:
update the notebook dependency because the old one doesn't have the service account scope yet.


ddietterich (approved):
LGTM

PR #566, opened 2022-02-23
[PF-1347] Use BQ result tables instead of temporary tables in tests

zloery:
This change removes the places in integration tests where we currently implicitly use BigQuery temporary tables to hold query results. Instead, all queries which return results now have the destination table explicitly specified. This allows us to run these tests in cloud environments where Domain Restricted Sharing is enforced - otherwise, BigQuery will attempt to grant IAM policies on temporary tables directly to end-users, which causes errors.


ddietterich (approved):
LGTM

yuhuyoyo (approved):

PR #567, opened 2022-02-24
[PF-1363] Create caller's pet SA on GCP context create

zloery:
This change adds a new step to the "create GCP context" flight to create a pet service account in the newly-created GCP project for the caller. Previously, pet SAs would only be created the first time they were explicitly requested from Sam. This is a small optimization to prevent the following failure case which has appeared a few times:

1. Sam immediately adds new pet SAs to the appropriate proxy group on creation. However, there can be a small delay (generally a few seconds) before GCP permissions granted to the proxy group apply to the new group member.
2. A common use of pet SAs in Workspace Manager is checking access to referenced cloud resources, which happens on referenced resource creation. If this is the first time a user's pet is needed in the workspace, WSM will request Sam create the pet SA and then immediately try to use the pet credentials.
3. WSM attempts to check permissions on a cloud resource using the pet SA. However, the SA was just created and GCP permissions have not propagated, leading to a false negative IAM check.

This is especially common while cloning workspaces, where WSM may attempt to read referenced resources from the source workspace immediately after creating the destination workspace.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlightV2.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlightV2.java
@@ -72,6 +72,7 @@ public class CreateGcpContextFlightV2 extends Flight {
     addStep(
         new SyncSamGroupsStep(appContext.getSamService(), workspaceId, userRequest), shortRetry);
     addStep(new GcpCloudSyncStep(crl.getCloudResourceManagerCow()), cloudRetry);
+    addStep(new CreatePetSaStep(appContext.getSamService(), userRequest), shortRetry);
 
     // Store the cloud context data and unlock the database row
     // This must be the last step, since it clears the lock. So this step also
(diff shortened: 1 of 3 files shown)


yuhuyoyo (approved):

ddietterich (approved):
LGTM - nice analysis of the problem.
This does not remove the code to create the pet on demand. Is there a reason to leave it? If not, a separate PR for that is fine with me.

zloery:
We can't really remove the on-demand code since the underlying endpoint Sam exposes is "get or create pet SA" - as long as we need to fetch the pet SA credentials, we'll still be calling the same endpoint.

This also only solves the problem for whichever user calls the "create GCP context endpoint", so we still need the on-demand creation. This fixes cloning and single-user workspaces, but additional users might still see false negatives the first time they try to add a referenced resource in workspace, and they might have to retry the call. I looked for a better solution here by trying to verify permissions before returning from `PetSaService` but didn't find anything great yet

PR #570, opened 2022-02-26
[PF-1302] split openapi

ddietterich:
This PR is a reorganization of our OpenAPI code. There are several significant changes:

### New Project: openapi

I created a new project called `openapi` that is where all of the split up parts of our REST API now live.
The YAML pieces are in `src`. The shell scripts and Python script that merges all of the parts are in `tools`.
There is a `build.gradle` file that provides tasks to configure a python virtual environment and install the libraries required for the build. The `README` file explains the layout of the YAML.

### Removed Deprecated Interfaces

I removed the old `DataReference` interfaces from the API, server code, and integration test code. In the process, I found that those integration tests were some of the only tests run in our perf suite, alpha, and staging environments. I added PF-1372 to track development of additional perf tests. I added WOR-164 to track the need to add more integration testing for alpha and staging. (@ungwudik - FYI)

### Documentation

I added `ADDING_RESOURCES` documentation in the top-level. There are probably missing pieces that folks should feel free to add as they add new resources.


cbookg:
So as a user of this API I think this will be nicer to read through.  But I'm not sure how the UI is going to work with this structure.

Currently the UI is generating the typescript client based on the github service_openapi.yaml URL.  It seems like there will be no equivalent way to do that (I am assuming you will remove this file in a later step?).

Perhaps we need to move typescript client generation into this repo, or some kind of distinct repo that publishes npm package?

ddietterich:
> Perhaps we need to move typescript client generation into this repo, or some kind of distinct repo that publishes npm package?

Ah yes, this change moves and renames the file. It would be great to build the typescript client from this repo and publish it as part of our normal build process. The same way we do with the Java client. I'm happy (well, willing :)) to help with the gradle if you can tell me what it should do.

cbookg:
> Ah yes, this change moves and renames the file. It would be great to build the typescript client from this repo and publish it as part of our normal build process. The same way we do with the Java client. I'm happy (well, willing :)) to help with the gradle if you can tell me what it should do.

Unfortunately I have no idea how to do this either.  Would require some investigation.

openapi/tools/merge_openapi.py
```diff
@@ -0,0 +1,144 @@
...
+import argparse
+import yaml
+from glob import glob
+from openapi_spec_validator import validate_spec
+from openapi_spec_validator.readers import read_from_filename
+from pathlib import Path
+
+global_had_load_error = False
+
+def validate_openapi(filepath):
```
jaycarlton, line 11: most of these top-level methods can be private methods of your class.

openapi/tools/merge_openapi.py
```diff
@@ -0,0 +1,144 @@
...
+    components = sortDictionary(indict['components'])
+    result = dict()
+    result['openapi'] = indict['openapi']
+    result['info'] = indict['info']
+    result['paths'] = paths
+    result['components'] = components
+    result['security'] = indict['security']
+    return result
+
+def main():
```
jaycarlton, line 99: Ideally your `main` would be 2 lines: instantiate the class and dump the output to file.

jaycarlton (commented):
first pass review of the python file.

ddietterich:
I added PF-174 to track adding a typescript client to the WSM build

openapi/README.md
```diff
@@ -0,0 +1,40 @@
...
+Many resources share their attribute and resource definitions between a controlled and a
+referenced version. Those definitions are also referenced by the enumeration
+endpoints. Therefore, we place all resource attribute and definition files here.
+
+In the future, we may drive auto-generation of enumeration unions by scanning this
+directory. Files should be named:
+```
+resource_{resource-type}
+```
+For example, the file describing a GCP GCS bucket would be `resource_gcp_gcs_bucket.yaml`.
```
jaycarlton, line 34: Why not have a `resource ` subdirectory so you don't have confusion of which underscore is a delimieter. If you don't want to do that, consider `resource__gcp_gcs_bucket.yaml`.

openapi/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+
+    doLast {
+        def script = "${projectDir}/tools/configure_venv.sh"
+        exec {
+            commandLine '/bin/bash', '-c', "${script} ${toolsDir} ${outDir}"
+        }
+    }
+}
+
+tasks.register('buildOpenApi') {
```
jaycarlton, line 33: by convention, this should just be named `build`

openapi/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    dependsOn('configureVenv')
+    ConfigurableFileTree yamlFiles = fileTree(dir: 'src')
+    yamlFiles.include '**/*.yaml'
+
+    inputs.files(yamlFiles)
+            .withPropertyName('yamlFiles')
+            .withPathSensitivity(PathSensitivity.RELATIVE)
+    outputs.dir(file("${outDir}"))
```
jaycarlton, line 45: does `file(outDir)` not work here?

openapi/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+// the inputs and outputs so it can skip tasks if there have been no changes.
+//
+// There may be better ways to do this. Gradle is vast and complex.
+// All I know is that this is a way that works.
+
+def outDir = "${buildDir}/openapi"
+def toolsDir = "${projectDir}/tools"
+def srcDir = "${projectDir}/src"
+
+tasks.register('configureVenv') {
```
jaycarlton, line 20: Please add descriptions to custom tasks: https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:adding_a_description_to_a_task

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -128,15 +127,6 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     return new ResponseEntity<>(referenceResource.toApiResource(), HttpStatus.OK);
   }
```
jaycarlton, line 129: Why are there Java changes? Is this a clean merge?

ADDING_RESOURCES.md
```diff
@@ -0,0 +1,148 @@
+# Adding Resources to Workspace Manager
+
+*ddietterich@verily.com*
+
+*created: 2022-02-25*
```
jaycarlton, line 5: we don't need this, since it's in source control.

jaycarlton:
I built this locally. Here's what the rebuilt [OpenAPI file](https://gist.github.com/jaycarlton/f52061a33be8c5f49ccca6041392a1c2) looks like.

openapi/tools/merge_openapi.py
```diff
@@ -0,0 +1,144 @@
...
+from pathlib import Path
+
+global_had_load_error = False
+
+def validate_openapi(filepath):
+    spec_dict = read_from_filename(filepath)[0]
+    validate_spec(spec_dict)
+
+
+def yaml_loader(filepath):
```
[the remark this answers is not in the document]
jaycarlton: It would be nice to `trim()` all strings on load, so we don't get ugly trailing newlines. I don't know if it's possible to honor all the text wrapping directives though.

```
      summary: 'Enumerate resources in a workspace. The visible resources depend on
        the permissions of the caller.
        These are presented sorted by ascending resource name. The offset and limit
        parameters allow paging
        through the results. You can also filter by a resource type and by a stewardship
        type.
        '
```

openapi/src/common/schemas.yaml
```diff
@@ -0,0 +1,232 @@
+# Common schemas
```
jaycarlton, line 1: We can connect the comments to the nodes they care about by using a node that's ignored by swagger like `x-comment`. i'm looking for the exact documentation of this...

openapi/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+    inputs.files(files("${toolsDir}/requirements.txt", "${toolsDir}/configure_venv.sh"))
```
jaycarlton, line 21: Does this task know not to repeat itself if the venv stuff is already there?

openapi/tools/merge_openapi.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
jaycarlton, line 1: I bet you could do this in the Gradle task and lose this file.

jaycarlton (commented):
I've got to get back to my on-call duties. Here's what I've seen so far.

I played with the tasks a little bit. For some reason the `openapi/build` directory got hidden in IJ, though it was there the first time I built it. Do you not need a `.gitignore` entry there?

yuhuyoyo (commented):
documentation lgtm

openapi/tools/merge_openapi.sh
ddietterich: I did not see how I could activate the virtual environment and run the script from within Gradle. The standard Exec task has a single command.

openapi/tools/merge_openapi.py
```diff
@@ -0,0 +1,144 @@
...
+        make_dict(self.data['components'], 'responses')
+
+def lookup_dict(yfile, ypath):
+    tpath = yfile.data
+    for p in enumerate(ypath):
+        tpath = tpath[p[1]]
+    return tpath
+
+def check_duplicates(target, source, ypath):
+    tdict = lookup_dict(target, ypath)
```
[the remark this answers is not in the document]
ddietterich: Better, since if there is more than one duplicate key it will can report them all.

openapi/tools/merge_openapi.py
```diff
@@ -0,0 +1,144 @@
...
+    for k in indict['components'].keys():
+        indict['components'][k] = sortDictionary(indict['components'][k])
+    components = sortDictionary(indict['components'])
+    result = dict()
+    result['openapi'] = indict['openapi']
+    result['info'] = indict['info']
+    result['paths'] = paths
+    result['components'] = components
+    result['security'] = indict['security']
+    return result
```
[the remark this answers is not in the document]
ddietterich: How would I know?

I kinda like it, since I kinda don't like knowing I am at the end of a block by indentation. But that is a whole different discussion...

openapi/src/common/schemas.yaml
ddietterich: It would also have to be ignored by the validator

openapi/build.gradle
ddietterich: Yes. By virtue of specifying the input files and the output, if the inputs have not changed, it does not remake the output.

openapi/tools/merge_openapi.py
```diff
@@ -0,0 +1,144 @@
...
+    with open(filepath,'r')as file_descriptor:
+        try:
+            data = yaml.safe_load(file_descriptor)
+        except yaml.YAMLError as err:
+            global_had_load_error = True
+            print("Parse error: ", err)
+            return None
+    return data
+
+def make_dict(data, ypath):
```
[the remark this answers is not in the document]
ddietterich: I named it `ensure_dict_exists` which is what it does

openapi/src/common/schemas.yaml
jaycarlton: I think it was when I tried it.

openapi/tools/merge_openapi.py
```diff
@@ -0,0 +1,144 @@
...
+    party = list(map(YFile, yaml_list))
+
+    # Bail if we did not get a clean load
+    if global_had_load_error:
+        print("Exiting due to load error")
+        return 1
+
+    # Merge each section
+    for p in party:
+        merge(mainy, p, ['paths'])
```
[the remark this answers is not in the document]
ddietterich: I updated the comment to explain that the order doesn't matter.

openapi/tools/merge_openapi.sh
jaycarlton: You can use `doLast` to get another task in there, or do a `"my command".execute()`. https://stackoverflow.com/a/37520139/611672

openapi/tools/merge_openapi.py
```diff
@@ -0,0 +1,144 @@
...
+    outfile = open(outfilepath, 'w')
+    outfile.write(yaml.safe_dump(result, sort_keys=False))
+    outfile.close()
+
+    # Validate the resulting openapi file
+    validate_openapi(outfilepath)
+    return 0
+
+if __name__ == '__main__':
+    sys.exit(main())
```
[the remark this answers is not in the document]
ddietterich: Oof! I suppose, but yet another set of tooling and gradle :(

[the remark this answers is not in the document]
jaycarlton: that can come later. we're exercising the output well enough I guess.

openapi/tools/merge_openapi.py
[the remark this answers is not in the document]
ddietterich: My thought was to standardize on one text wrapping directive that makes it through this transition the best. There are some controls on the generation side that might help, too. It will require some investigation. I made PF1378 to track this.

openapi/tools/merge_openapi.py
```diff
@@ -0,0 +1,144 @@
...
+        try:
+            data = yaml.safe_load(file_descriptor)
+        except yaml.YAMLError as err:
+            global_had_load_error = True
+            print("Parse error: ", err)
+            return None
+    return data
+
+def make_dict(data, ypath):
+    if ypath not in data:
```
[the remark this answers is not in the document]
ddietterich: I could not get this expression to parse:
```
       if yaml_path not in input_dict or input_dict[yaml_path] is None:
```
I tried with parens and that didn't. So right now it is two `if` statements.

openapi/tools/merge_openapi.py
ddietterich: I moved the class-related ones into the class.

openapi/tools/merge_openapi.py
ddietterich: I am not seeing the benefit here. I could add multi-tree methods to the tree class, but that isn't how I think classes should be constructed. Can you elaborate on what you'd like to see?

ADDING_RESOURCES.md
```diff
@@ -0,0 +1,148 @@
...
+   cloning, then add your resource to the `CloneWorkspace` test as well.
+
+## More Information
+
+### Asynchronous Interfaces
+
+We have a strong bias for synchronous APIs. If you have an operation that runs over the
+network time more that 10% of the time, it is appropriate to make it an asynchronous
+interface. See
+[MC Terra - Async REST API Specification](https://docs.google.com/document/d/1PTd4xvmV9xnEkWaIgFc6d3VUyJwwymQTFdhpKrdsUKw/edit#heading=h.ol9mx3vfhjjj)
```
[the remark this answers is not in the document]
ddietterich: Yes, we could convert it to markdown and put it in the `terra` repo.

integration/src/main/resources/suites/FullIntegration.json
```diff
@@ -8,11 +8,9 @@
     "integration/CloneWorkspace.json",
     "integration/ControlledBigQueryDatasetLifecycle.json",
     "integration/ControlledGcsBucketLifecycle.json",
-    "integration/DataReferenceLifecycle.json",
```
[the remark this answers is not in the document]
ddietterich: It saved me from converting over the deprecated APIs.

openapi/README.md
ddietterich: It is redundant, since they are in a `resources` directory. I'll rename them.

openapi/build.gradle
ddietterich: I was shy of doing that since the tasks are driven from other projects rather than an independent build.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
ddietterich: This is part of removing the deprecated APIs

ADDING_RESOURCES.md
```diff
@@ -0,0 +1,148 @@
...
+
+This section summarizes the steps to add a new resource. I recommend this order to be able
+to compile your code as you go.
+
+1. Create your resource type and attributes in the `openapi` sub-project.
+   - Add your resource file to `src/resources/` directory, defining the resource and the
+     attributes objects.
+   - Edit the `resource_type.yaml` file in that directory, creating an enumeration for the
+     resource and adding the appropriate resource element to the union objects.
+   - Run `gradle :openapi:buildOpenApi` to make sure your syntax and references are correct.
```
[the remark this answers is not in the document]
jaycarlton: Usually `build` is the first thing I try when I'm in a new project. You'll have to change `buildEnvironment` to something like `setupEnvironment` if you change it though.

openapi/src/common/schemas.yaml
ddietterich: OK, that reference is from V2. They are now called [Specification Extensions](https://swagger.io/specification/)
I'm going to attach that to the PF-1378 ticket

openapi/tools/merge_openapi.sh
ddietterich: Those commands will not guaranteed run in the same shell, so the sourcing to set the venv will not be there in the second command.

ADDING_RESOURCES.md
```diff
@@ -0,0 +1,148 @@
...
+1. Create your Java resource package in `service/src/main/java/bio.terra/workspace/service/resources/`
+      in the appropriate package
+    - `controlled/{cloud}/{your resource}`
+    - `referenced/{cloud}/{your resource}`
+1. Create classes for your resource
+   - Attributes class describing the fields beyond the common metadata
+   - Resource class. This is the most complex class to implement and supplies the bulk of
+     your resource code
+1. If you are making a controlled resource you will also need to add:
+   - Step classes as required for object create and update. These will be needed to
```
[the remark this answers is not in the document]
ddietterich: I added a reference to the Flight Developer Guide markdown.

ADDING_RESOURCES.md
```diff
@@ -87,7 +84,7 @@ Most of the code you make is in files specific to the new resource. These are:
 | integration | src/main/java/scripts/testscripts | `{resource}Lifecycle.java` |  |
 | integration | src/main/resources/configs/integration | `{resource}Lifecycle.java` |  |
 | openapi | src/parts | `{stewardship}_{cloud}_{resource}.yaml` | |
-| openapi | src/resources | `resource_{cloud}_{resource}.yam`l |  |
+| openapi | src/resources | `{cloud}_{resource}.yam` |  |
```
jaycarlton, line 87: you lost an `l`.

DEVELOPMENT.md
```diff
@@ -42,17 +44,29 @@ integration environments), update the
...
-  brew install postgresql@13
-  ```
-  - [The app](https://postgresapp.com/downloads.html) may be easier, just make sure to download the right version. It'll manage things for you and has a useful menulet where the server can be turned on and off. Don't forget to create a server if you go this route.
-- Install AdoptOpenJDK Java 11 (Hotspot). Here's an easy way on Mac, using [jEnv](https://www.jenv.be/) to manage the active version:
+#### Postgres
+We are currently using Postgres 13.1.
+
+You have choices for how to install Postgres:
+- Directly from the Postgresql website: https://www.postgresql.org/download/
+- Via homebre:
```
jaycarlton, line 56: homebrew

DEVELOPMENT.md
```diff
@@ -42,17 +44,29 @@ integration environments), update the
...
+```sh
+brew install postgresql@13
+```
+- By installing the Mac application: [Postgresql app](https://postgresapp.com/downloads.html). This
+is the easiest. Just make sure to download the right version. It'll manage things for you and has a
+useful menulet where the server can be turned on and off. Don't forget to create a server if you
+go this route.
+
```
jaycarlton, line 64: There's also the Docker option, which I use. It's documented someplace...

DEVELOPMENT.md
```diff
@@ -70,26 +84,33 @@ integration environments), update the
...
+```
+
+In the automation, we use the `venv` to make virtual environments. The advantage of using `venv`
+is that it is built in so does not require extra installation steps.
+However, it is not so useful as a developer tools because it has no way to de-activate a virtual
+environment. Consider installing `virtualenv` for your developer use. It can be installed using
+`pip` like so:
+```
+python3 -m pip install virtualenv
+```
```
jaycarlton, line 113: You've just recommended three different virtual environment solutions for Python: `pyenv`, `venv`, and `virtualenv`. Can we not settle on one (even if it means adding cruft to the dockerfile)?

openapi/tools/merge_openapi.py
jaycarlton: Largely preference. I like my `main` methods dumb, like turning the key in the ignition and letting the classes do the work. I just figured your class would be the point of knowledge about all this, but you could have more than one class if that makes sense. It's totally fine not to do that and have top-level methods if you want, since this isn't intended to be loaded as a module (and even then it might not matter too much).

openapi/tools/merge_openapi.py
```diff
@@ -6,126 +7,139 @@
...
+    # the files we loaded. It is ordered below.
+    for p in part_tree_list:
+        merge(main_tree, p, ['paths'])
+        merge(main_tree, p, ['components', 'schemas'])
+        merge(main_tree, p, ['components', 'parameters'])
+        merge(main_tree, p, ['components', 'responses'])
 
     # Put the merged parts into a pretty order
-    result = order_merge(mainy.data)
+    result = order_merge(main_tree.data)
```
jaycarlton, line 142: This looks like you're reaching into the guts of the class and rearranging them. Why not have the class be responsible for maintaining its own invariants and keep `data` private?

service/gradle/openapi.gradle
```diff
@@ -20,5 +20,16 @@ swaggerSources {
...
 }
 
-generateSwaggerCodeServer.dependsOn(tasks.getByPath(':openapi:buildOpenApi'))
+// Copy the result of the openapi build into the built resources
+// so it is available for the swagger_ui.html endpoint
+task copyOpenApiToResources(type: Copy) {
+  dependsOn(tasks.getByPath(':openapi:buildOpenApi'))
+  from tasks.getByPath(':openapi:buildOpenApi').openApiSource
+  into 'build/resources/main/api'
+  rename { String filename -> "service_openapi.yaml" }
```
jaycarlton, line 29: Document for external consumption where this file lives and how it's updated.

jaycarlton (approved):

DEVELOPMENT.md
ddietterich: I reworded. It is discussed later in the document.

DEVELOPMENT.md
ddietterich: `pyenv` is not a virtual environment solution.

The docker Broad has running on their hosted servers is deliberately minimal. My instructions from devops are to install what I need in the GHA. So I prefer using venv rather than adding another installation step.

openapi/src/openapi_main.yaml
```diff
@@ -0,0 +1,57 @@
+# DEVELOPER NOTES
+#
+# There are three sections of the file bracketed with ## ## that are deprecated. They are the
+# data references interface that we abandoned in favor of the unified resources interface.
```
zloery, line 4: outdated comment?

openapi/tools/merge_openapi.py
ddietterich: Good call. I moved this and the sort function into the class.

zloery (approved):
Documentation and removing the deprecated snapshot resource code LGTM, but I haven't reviewed the yaml or python files thoroughly.

openapi/src/openapi_main.yaml
ddietterich: Removed

PR #571, opened 2022-03-02
[PF-1227] Increase retries for GcpCloudSyncStep

melissachang:
There is an IAM performance issue for `GcpCloudSyncStep`, that only affects Verily deployment. When performance issue is fixed, retries will go back to normal.


service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlightV2.java
```diff
@@ -43,6 +43,7 @@ public CreateGcpContextFlightV2(FlightMap inputParameters, Object applicationCon
 
     RetryRule shortRetry = RetryRules.shortExponential();
     RetryRule cloudRetry = RetryRules.cloud();
```
jmczerk, line 45: Nit, but maybe `bufferRetry` for consistency?

jmczerk (approved):

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlightV2.java
```diff
@@ -80,4 +82,4 @@ public CreateGcpContextFlightV2(FlightMap inputParameters, Object applicationCon
         new UpdateDbGcpCloudContextStep(workspaceId, appContext.getGcpCloudContextService()),
         shortRetry);
   }
-}
+}
```
ddietterich, line 85: nit: Not sure it matters, but no final new line

ddietterich (approved):
LGTM

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlightV2.java
melissachang: Done. Changed Intellij preference so this doesn't happen again.

PR #572, opened 2022-03-03
make dependencies explicit

jaycarlton:
Fix a bunch of warnings of the form 
```
- Gradle detected a problem with the following location: '/Users/jaycarlton/repos/terra-workspace-manager/service/build/openapi/src/main/java/bio/terra/workspace/generated/controller/Alpha1Api.java'. Reason: Task ':service:spotlessJava' uses this output of task ':service:generateSwaggerCodeServer' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.0/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

Updated task trees are attached.
[service_build_tasks.txt](https://github.com/DataBiosphere/terra-workspace-manager/files/8179535/service_build_tasks.txt)
[client_build_tasks.txt](https://github.com/DataBiosphere/terra-workspace-manager/files/8179565/client_build_tasks.txt)

### DIFF
--- a/client/build.gradle
+++ b/client/build.gradle
@@ -6,3 +6,4 @@ plugins {
 def includeDir = "$projectDir/gradle"
 apply(from: "$includeDir/dependencies.gradle")
 apply(from: "$includeDir/openapi.gradle")
+tasks.spotlessJava.dependsOn(generateSwaggerCodeClient)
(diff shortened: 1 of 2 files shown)


jaycarlton:
Somehow now Spotless is [running on generated files](https://github.com/DataBiosphere/terra-workspace-manager/runs/5410676706?check_suite_focus=true) in the `client` project. That's not an improvement.

jaycarlton:
This is the wrong approach. We don't want the dependency, implicitly or explicitly.

PR #575, opened 2022-03-04
[no ticket] check return value of getWithRetryOnException()

jaycarlton:
In investigating test failures, I discovered that the `CloneWorkspace` test was failing to setup resources as test fixtures, but not flunking the test until much later. This change asserts that we didn't run out of retries in the helper method, failing faster.


yuhuyoyo (approved):

zloery (commented):
Instead of changing the callsites everywhere, would it make sense to just throw the last exception (or a new exception) from `get/runWithRetryOnException` when we run out of retries? I don't think it's ever the expected behavior to hit the max number of retries, that should always be a failure

jaycarlton:
> Instead of changing the callsites everywhere, would it make sense to just throw the last exception (or a new exception) from `get/runWithRetryOnException` when we run out of retries? I don't think it's ever the expected behavior to hit the max number of retries, that should always be a failure

Yeah, I was thinking along those lines myself.

zloery (approved):
Thanks for this!

yuhuyoyo (approved):

PR #576, opened 2022-03-05
TOAZ-71 WIP add hybrid connection APIs

Qi77Qi:
https://broadworkbench.atlassian.net/browse/TOAZ-71


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/relayNamespace/ControlledAzureRelayNamespaceResource.java
```diff
@@ -122,6 +123,11 @@ public void addDeleteSteps(DeleteControlledResourceFlight flight, FlightBeanBag
         new DeleteAzureRelayNamespaceStep(
             flightBeanBag.getAzureConfig(), flightBeanBag.getCrlService(), this),
         RetryRules.cloud());
+//    //TODO: get all the hybrid connections this has
```
Qi77Qi, line 126: @ddietterich think I'll need some suggestion here...Basically relay namespace is kinda parent resource in Azure for hybrid connections. So when we delete relay namespace, we delete all hybrid connections as well....What's the best way to model that in WSM? currently namespace and hybrid connections are treated as separate resources, which I think is mostly fine(but lmk if you have better suggestion). In delete Azure namespace, we need to somehow find all the hybrid connection resources and delete all related sam resources etc, what's the best way to retrieve that info?

ddietterich: We took the direction awhile back that WSM would not maintain relationships between resources. Leo would be in charge of that. If we continue in that direction, then Leo should track that parent-child relationship.

I think Leo would need to call WSM to delete all of the resources, even if some of them were implicitly deleted. That cleans up WSM resource state. Since the semantics of delete are to return success if the object does not exist, Leo would not see errors from WSM.

PR #580, opened 2022-03-08
[PF-1330] Revise how applications are configured

ddietterich:
This PR changes how applications are configured. Implementation details are in [Fix WSM Application Config](https://docs.google.com/document/d/1fQRdVzkjfhu-lL6ABMjEUSmoQeKSKt93Ql5BsFS54cA/edit#heading=h.6rt3cgdzsac)

The bulk of the file changes are trivial and are for switching from UUID to String.

@zloery - tagged you because this will break Verily-side tests


integration/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -51,7 +51,7 @@
 
   public static final String RESOURCE_NAME_PREFIX = "terratest";
   // We may want this to be a test parameter. It has to match what is in the config or in the helm
-  public static final UUID TEST_WSM_APP = UUID.fromString("E4C0924A-3D7D-4D3D-8DE4-3D2CF50C3818");
+  public static final String TEST_WSM_APP = "TestWsmApp";
```
jaycarlton, line 54: You're going to run out of GUIDs if you keep wasting them like this 😞

openapi/src/parts/workspace_application.yaml
```diff
@@ -117,7 +126,7 @@ components:
         applicationId:
           description: The ID of the application
           type: string
-          format: uuid
+          format: string
```
jaycarlton, line 129: I think this is the default if you wanna leave it out

service/src/main/java/bio/terra/workspace/app/configuration/external/WsmApplicationConfiguration.java
```diff
@@ -19,20 +17,11 @@
 @ConfigurationProperties(prefix = "workspace.application")
 public class WsmApplicationConfiguration {
   public static class App {
-    private String identifier;
     private String name;
```
jaycarlton, line 20: Do these need to match the config file's order? Document if so. Also, how does `state` get populated for the test app?

service/src/test/java/bio/terra/workspace/service/privateresource/PrivateResourceCleanupServiceTest.java
```diff
@@ -54,22 +54,25 @@
   // The name of the "member" policy created by default for groups. If this is ever used for
   // implementation code, it should live in SamConstants instead.
   private static final String SAM_GROUP_MEMBER_POLICY = "member";
```
jaycarlton, line 56: can this go with the other Sam constants/enums?

service/src/test/java/bio/terra/workspace/service/privateresource/PrivateResourceCleanupServiceTest.java
```diff
@@ -204,20 +207,23 @@ void cleanupResourcesSuppressExceptions_cleansApplicationPrivateResource_succeed
         "grantWorkspaceRole");
     // Enable the WSM test app in this workspace. This has a test user as the "service account" so
     // we can delegate credentials normally.
-    App appConfig = getAppBySa(applicationAccessUtils.getApplicationSaEmail());
-    UUID appId = UUID.fromString(appConfig.getIdentifier());
+    WsmApplication app = applicationDao.getApplication(TEST_WSM_APP);
+    AccessToken saAccessToken = userAccessUtils.generateAccessToken(app.getServiceAccount());
```
jaycarlton, line 211: nit: usually I expect classes ending in "Utils" to be all static methods.

service/src/test/java/bio/terra/workspace/service/workspace/ApplicationUnitTest.java
```diff
@@ -31,8 +30,8 @@
   // These strings have to match the text in WsmApplicationService or the test will fail
   private static final String ERROR_MISSING_REQUIRED =
       "Invalid application configuration: missing some required fields (identifier, service-account, state)";
-  private static final String ERROR_BAD_UUID =
-      "Invalid application configuration: invalid UUID format";
+  private static final String ERROR_BAD_ID =
```
jaycarlton, line 33: BTW, is it safe to have an application ID start with a digit? If so, that limits the contexts where it can be used (in scripts, etc).

service/src/test/resources/application-app-test.yml
```diff
@@ -2,9 +2,9 @@ workspace:
   application:
     test-app-sa: Elizabeth.Shadowmoon@test.firecloud.org
     configurations:
```
jaycarlton, line 4: So there's no yaml schema from what I've seen. How does it know to build a map from this structure?

jaycarlton (approved):
Looks good.

service/src/main/java/bio/terra/workspace/app/configuration/external/WsmApplicationConfiguration.java
ddietterich: They do not need to match. Spring is doing the mapping work here.

The application.yml file has an optional include for a write-config.sh output file. That lets us inject the test application and the Azure config in the local configuration. So when we run `local` configuration tests the application is there.

service/src/test/java/bio/terra/workspace/service/privateresource/PrivateResourceCleanupServiceTest.java
ddietterich: Not my code. As I understand it, the `userAccessUtils` is an `@Component` so it can get some configuration.

service/src/test/java/bio/terra/workspace/service/workspace/ApplicationUnitTest.java
ddietterich: Yes, it can start with a number. The pattern is `"^[a-zA-Z0-9][-_a-zA-Z0-9]{0,1023}$"`

I don't think we need the application id to be useful as a shell variable name, so I think that is fine. Many WSM, GCP and Azure identifiers allow leading numbers.

service/src/test/resources/application-app-test.yml
ddietterich: See [Inject a Map from a YAML File with Spring](https://www.baeldung.com/spring-yaml-inject-map)

service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
```diff
@@ -342,12 +344,30 @@ public void updateApplication(WsmApplication app) {
...
             .addValue("description", app.getDescription())
             .addValue(SERVICE_ACCOUNT, app.getServiceAccount())
             .addValue("state", app.getState().toDb());
 
     jdbcTemplate.update(sql, params);
   }
+
+  /**
+   * Temporary method to delete the old Leo application. This is based on the understanding that
+   * there is no use of the application anywhere. No enabling in any workspace and no dependent
```
jdcanas, line 358: Yeah, I just generated this UUID to make the config. Shouldn't be used anywhere.

service/src/main/resources/application.yml
```diff
@@ -103,11 +103,11 @@ workspace:
 
   application:
     configurations:
-      identifier: 6397C5FF-F83C-4837-95DB-059198FD89BF
-      name: Leonardo
-      description: Leonardo application SA
-      service-account: leonardo-prod@broad-dsde-prod.iam.gserviceaccount.com
-      state: operating
+      leo:
```
jdcanas, line 106: this looks good to me!

jdcanas (approved):
Leo config changes look good to me! Thanks for doing this, will be a lot better than needing to use an index for sure.

PR #581, opened 2022-03-10
[PF-1330] Return empty application map from config instead of null

zloery:
This changes the application configuration reader to return an empty `Map` instead of null when no applications are configured. This is similar to the previous `List` behavior prior to #580.

Without this change, WSM will always [throw a NPE](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/workspace/WsmApplicationService.java#L261) on startup if no applications are configured. This was hidden in this repo because one application is present in the default configuration values, but the Verily default configuration did not have this.


ddietterich (approved):
Good catch. Thanks for the fix.

jaycarlton (approved):
Good find.

PR #585, opened 2022-03-17
Fix flaky test PrivateControlledAiNotebookInstanceLifecycle

melissachang:
Before PR:
1. Line 87: Create first notebook
2. In `createAControlledAiNotebookInstanceWithoutSpecifiedInstanceId_validInstanceIdIsGenerated()`
    - Create second notebook
    - Delete second notebook
3. In `createAControlledAiNotebookInstanceWithoutSpecifiedInstanceId_specifyLocation()`
    - Create third notebook
    - Delete third notebook
4. Line 157: List notebooks and assert there are 3 notebooks

4 was sometimes failing because second notebook was deleted, so there were 2 notebooks instead of 3.

I moved 4 to before 2. Instead of 3 notebooks, I assert there is 1 notebook.


jaycarlton (approved):
LGTM. Thanks!

yuhuyoyo (approved):

PR #588, opened 2022-03-21
[PF-1256] Unicode-escape certain characters in WSM JSON responses

zloery:
Informational finding from a pentest, recommendation was unicode-escaping certain characters in all JSON responses to prevent potential XSS attacks if the responses is rendered improperly in a UI.
This is difficult to test directly as the Java client our integration tests use automatically un-escapes the characters in responses, but I tested manually as follows:

1. Create a workspace with description `test string with characters " \n \t & < > \ /`
2. Prior to the change, calls to GET this workspace return description `test string with characters \" \\n \\t & < > \\ /` in the response. The UI displays a string matching the original description.
3. After this change, calls to GET this workspace return description `test string with characters \" \\n \\t \u0026 \u003C \u003E \\ /` in the response. The UI displays a string matching the original description.

Both the CLI and ET UI already display these characters properly, so this change is invisible to end-users.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java
+++ b/service/src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java
@@ -3,6 +3,8 @@ package bio.terra.workspace.app.configuration.spring;
 import bio.terra.workspace.app.StartupInitializer;
 import bio.terra.workspace.app.configuration.external.WorkspaceDatabaseConfiguration;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.core.SerializableString;
+import com.fasterxml.jackson.core.io.CharacterEscapes;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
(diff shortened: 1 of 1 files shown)


service/src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java
```diff
@@ -22,13 +24,49 @@ public NamedParameterJdbcTemplate getNamedParameterJdbcTemplate(
     return new NamedParameterJdbcTemplate(config.getDataSource());
   }
 
+  public static class HTMLCharacterEscapes extends CharacterEscapes {
```
jaycarlton, line 27: Presumably all of our services will want to do this. Did you consider putting it in the common lib?

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java
```diff
@@ -22,13 +24,49 @@ public NamedParameterJdbcTemplate getNamedParameterJdbcTemplate(
...
+    /**
+     * Return the escape codes used for non-ASCII characters, or ASCII characters with escape code
+     * set to ESCAPE_CUSTOM.
+     *
+     * <p>This is required because CharacterEscapes is abstract. We don't need additional escaping
+     * for any characters, so per documentation this can return null for all inputs.
+     */
+    @Override
+    public SerializableString getEscapeSequence(int ch) {
+      return null;
```
yuhuyoyo, line 56: should this actually return something?

zloery: We don't need any additional custom escape behavior (either escaping non-ASCII characters or ASCII characters set to `CharacterEscapes.ESCAPE_CUSTOM` above), so this should return `null` for all inputs. Per superclass documentation:
`Returns:
Escape sequence to use for the character, if any; null if not`

(but agreed, it looks kind of odd)

service/src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java
zloery: Part of the problem is that this is one setting for the Jackson `ObjectMapper`, and right now each service creates its own `ObjectMapper` object. I think we could reasonably provide a pre-configured one as a component in TCL including this change, but that's a bit out of scope here (filed PF-1424 to track it, though)

yuhuyoyo (approved):

PR #591, opened 2022-03-24
[PF-1428] don't check access on referenced resource create or update

jaycarlton:
Remove the access step check from the `CreateReferencedResourceFlight`, but don't delete it, as it will be needed again soon for the on-demand access check functionality.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/create/CreateReferenceResourceFlight.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/create/CreateReferenceResourceFlight.java
@@ -21,10 +21,7 @@ public class CreateReferenceResourceFlight extends Flight {
 
     FlightBeanBag appContext = FlightBeanBag.getFromObject(beanBag);
 
-    // Perform access verification
-    addStep(new ValidateReferenceStep(appContext), RetryRules.cloud());
-
-    // If all is well, then store the reference metadata
+    // Store the reference metadata
     addStep(
         new CreateReferenceMetadataStep(appContext.getResourceDao()), RetryRules.shortDatabase());
   }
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/create/CreateReferenceResourceFlight.java
```diff
@@ -21,10 +21,7 @@ public CreateReferenceResourceFlight(FlightMap inputParameters, Object beanBag)
 
     FlightBeanBag appContext = FlightBeanBag.getFromObject(beanBag);
 
-    // Perform access verification
-    addStep(new ValidateReferenceStep(appContext), RetryRules.cloud());
-
-    // If all is well, then store the reference metadata
+    // Store the reference metadata
```
jaycarlton, line 24: Apparently we didn't have any integration tests asserting that creation failed when there was no access. 🙃

service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/create/CreateReferenceResourceFlight.java
```diff
@@ -21,10 +21,7 @@ public CreateReferenceResourceFlight(FlightMap inputParameters, Object beanBag)
 
     FlightBeanBag appContext = FlightBeanBag.getFromObject(beanBag);
 
-    // Perform access verification
-    addStep(new ValidateReferenceStep(appContext), RetryRules.cloud());
```
zloery, line 25: Can you add a comment/documentation to this step indicating that it's dead code (for now), but might be needed again soon?

zloery, line 25: We also have Stairway idempotency tests for this step (https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/test/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceServiceTest.java#L250 and the `undo` test below that), can you either remove or comment those lines out as well?

zloery (approved):

service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/create/CreateReferenceResourceFlight.java
jaycarlton: I don't understand what to comment out. The tests still pass, and are targeting a flight that is still idempotent, right?

yuhuyoyo:
we might need to do CLI test change. I vaguely remember there are test that try to add references that the user doesn't have access to.

yuhuyoyo (approved):

PR #593, opened 2022-03-24
ouptut test resutls dir and use it to upload to github

jaycarlton:


.github/workflows/pr-integration.yml
```diff
@@ -133,3 +133,4 @@ jobs:
         name: wsm-logs
         path: |
           wsm.log
+          ${{ steps.integration-test.outputs.results_dir }}
```
yuhuyoyo, line 136: that seems nice :) but how do i get the log?

jaycarlton: I hope it will got to the Artifacts section on the run summary page, along with wsm.log (which is there now).

yuhuyoyo: i see! thanks :)

yuhuyoyo (approved):

cbookg (approved):

.github/workflows/pr-integration.yml
cbookg: If you want to test this pre-submit, you can just temporarily include a test fail in this PR

PR #597, opened 2022-03-29
[PF-1449] Suppress unnecessary spotbugs warning

zloery:
Spotbugs warns that we're exposing a `private final` array via public method as required by the `CharacterEscapes` class. This is expected, and the class documentation notes "Caller is not to modify contents of this array, since this is expected to be a shared copy" on the return value of `getEscapeCodesForAscii`.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java
+++ b/service/src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java
@@ -26,24 +26,23 @@ public class BeanConfig {
   }
 
   public static class HTMLCharacterEscapes extends CharacterEscapes {
-    private final int[] asciiEscapes;
+    private static final int[] asciiEscapes;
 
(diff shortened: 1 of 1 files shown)


service/src/main/java/bio/terra/workspace/app/configuration/spring/BeanConfig.java
```diff
@@ -40,6 +41,9 @@ public HTMLCharacterEscapes() {
 
     /** Return the escape codes used for ASCII characters. */
     @Override
+    @SuppressFBWarnings(
+        value = "EI",
+        justification = "per interface doc, callers should not modify returned value")
     public int[] getEscapeCodesForAscii() {
       return asciiEscapes;
```
jaycarlton, line 48: why can't this be a static membmer?

jaycarlton (approved):

yuhuyoyo (approved):

PR #600, opened 2022-03-31
plug spring vulnerability

dvoet:


ddietterich:
Do we need this change or can we upgrade? I see this: [[CVE report published for Spring Cloud Function](https://spring.io/blog/2022/03/29/cve-report-published-for-spring-cloud-function)](https://spring.io/blog/2022/03/29/cve-report-published-for-spring-cloud-function)

dvoet:
I recommend upgrading to latest spring boot but that may be a bigger change. In the mean time I think this is required

ddietterich:
Since WSM Spring boot usage does not match the vulnerability, I would rather take the time to upgrade rather than make this patch.

ddietterich:
Doesn't make any sense to me. Not clear this is a comprehensive fix. Trevyn seems to be making the more general fix.
But whatever. Easy enough to remove later.

ddietterich (approved):

PR #603, opened 2022-03-31
WOR-164 - setup snapshots tests on alpha and staging

ungwudik:
https://broadworkbench.atlassian.net/browse/WOR-164

- Create TDR datasets/snapshots alpha and staging (done outside of this PR)
- Setup the snapshots for use in WSM alpha and staging tests.
- Add test for `getDataRepoSnapshotReferenceByName`


marctalbott (approved):
All looks good to me

cahrens (approved):
This looks OK, but I'm not familiar with how these tests work, so I've been somewhat reluctant to be the second thumb.

ungwudik:
> This looks OK, but I'm not familiar with how these tests work, so I've been somewhat reluctant to be the second thumb.

That's good feedback for me. I'll add a reviewer from PF team since they also have context around this test. Obvsiously still good for Workspace team's awareness as we pick up more WSM stuff.

ddietterich (approved):
lgtm

PR #604, opened 2022-03-31
[PF-1486] Break down workflows into jobs

jaycarlton:
Jobs are the unit of concurrency and workflow reuse. Additionally, having more than one job will make the UI more fine-grained.


jaycarlton:
We're not going to keep these changes. It was cool to see the linter and static analysis run in parallel, but it wasn't a time savings.

PR #608, opened 2022-04-04
[PF-1446] Return destination workspace ID immediately

jaycarlton:
Simply populate the `workspace` object in the result with what we already know.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -370,10 +370,11 @@ public class WorkspaceApiController extends ControllerBase implements WorkspaceA
     Optional<SpendProfileId> spendProfileId =
         Optional.ofNullable(body.getSpendProfile()).map(SpendProfileId::new);
 
+    UUID destinationWorkspaceId = UUID.randomUUID();
     // Construct the target workspace object from the inputs
     Workspace destinationWorkspace =
         Workspace.builder()
-            .workspaceId(UUID.randomUUID())
+            .workspaceId(destinationWorkspaceId)
             .spendProfileId(spendProfileId.orElse(null))
             .workspaceStage(WorkspaceStage.MC_WORKSPACE)
             .displayName(body.getDisplayName())
@@ -385,7 +386,13 @@ public class WorkspaceApiController extends ControllerBase implements WorkspaceA
         workspaceService.cloneWorkspace(
             workspaceId, petRequest, body.getLocation(), destinationWorkspace);
 
+    // Set destination workspace ID ASAP so UI can build a landing page for it
     final ApiCloneWorkspaceResult result = fetchCloneWorkspaceResult(jobId, getAuthenticatedInfo());
+    final ApiClonedWorkspace clonedWorkspaceStub = new ApiClonedWorkspace()
+        .destinationWorkspaceId(destinationWorkspaceId)
+            .sourceWorkspaceId(workspaceId);
+    result.setWorkspace(clonedWorkspaceStub);
+
     return new ResponseEntity<>(result, getAsyncResponseCode(result.getJobReport()));
   }
 
--- a/integration/src/main/java/scripts/testscripts/CloneWorkspace.java
+++ b/integration/src/main/java/scripts/testscripts/CloneWorkspace.java
@@ -261,7 +261,8 @@ public class CloneWorkspace extends WorkspaceAllocateTestScriptBase {
 
     final String jobId = cloneResult.getJobReport().getId();
     logger.info("Clone Job ID {}", jobId);
-
+    assertNotNull(cloneResult.getWorkspace().getDestinationWorkspaceId(),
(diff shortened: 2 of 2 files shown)


service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -385,7 +386,13 @@ private ApiCreateCloudContextResult fetchCreateCloudContextResult(
         workspaceService.cloneWorkspace(
             workspaceId, petRequest, body.getLocation(), destinationWorkspace);
 
+    // Set destination workspace ID ASAP so UI can build a landing page for it
     final ApiCloneWorkspaceResult result = fetchCloneWorkspaceResult(jobId, getAuthenticatedInfo());
+    final ApiClonedWorkspace clonedWorkspaceStub = new ApiClonedWorkspace()
+        .destinationWorkspaceId(destinationWorkspaceId)
```
cbookg, line 392: Question (for a potential follow-up PR): When this API returns, will we be able to call GetWorkspace using this ID immediately?  ie does the API wait for the flight to add the db entry?  This would be inline with how CreateAiNotebookInstance was modified.

cbookg (approved):

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
jaycarlton: It's not guaranteed to be there yet, and if it is, it might not have its cloud context included, but it is the first step in the flight.

cbookg: Ok.  Then just so its clear, returning the ID is part of what the UI needs.  But it will also needs the GetWorkspace API to not return 404.

jaycarlton: So I lied; it's not the first step. I'm looking at it, and that's going to be tough. The workspace create happens in a nested flight, so you can't really get there from here. We'd need some notion of a reserved ID in the DB, but all that would do for you would be to give you a non-404; it wouldn't have things that are normally expected to be there. 

How nasty is it to poll until you get a 200? It should just be a few seconds if you don't need the GCP project (cloud context) with it. On the UX side, you can either have it be transactional, where you don't show anything but a spinner while it's cloning, or try to show incrementally cloned things by polling the resource enumerate point during the clone. The resources aren't going to have DB entries right away either; as part of the resource clone it runs the create flight, and you get a DB entry sometime in there.

@ddietterich any ideas here?

ddietterich: I think we should discuss what we want the whole clone process to look like to the user before we code it up. This seems like the first worm in the can.

The next thing the UI will need is what? Will users expect to/be able to start using the partially cloned workspace before the clone is complete? Is that what we want? I think we have nothing to prevent that, and it could cause other parts of the clone operation to fail.

Do we need to create all workspace and resource metadata first - synchronously - and then do the actual work, so that all list operations work from the moment the initial clone call returns?

If users cannot use the workspace or resources within it, should we show it? Do we need additional state to note that the workspace is in-process?

jaycarlton: v0 could do what the CLI does and treat it as one big transaction. Just put a spinner up and show the workspace and all its goodies N minutes later.

cbookg: For sure we should discuss.
Based on the initial mocks, my rough plan was for clone (and create) to:
- Redirect to the new page URL based on the created ID.
- Have that page detect that creation is still in progress and show some kind of 'being constructed' content (you can see this in the mocks too).

This PR addresses part 1 above, but I would like any clean workspace page load to be able to reflect the state that the creation is in progress.  The new jobs API already tells us this, which is great.  Ideally the GetWorkspace API would also return as well.

The workaround listed above (such as keeping track of this in-memory, or polling) are possible, but really not ideal from the point of view of the UI code.

We should probably chat about this.

ddietterich: Yes, I think we should make a plan on paper before making this code change. @jaycarlton can you drive?

jaycarlton: Are we good with merging this now? We had the meeting but didn't commit it to paper yet.

jaycarlton: Filed [PF-1552](https://broadworkbench.atlassian.net/browse/PF-1552) for the 404 issue.

ddietterich: I would prefer not to merge this. I think we should implement the API changes we have discussed and return the target workspace id only when the caller can successfully enumerate it.

PR #609, opened 2022-04-05
[PF-900] Address Clone unit test debt

jaycarlton:
A couple of unit tests. Had to do a fair bit of mocking, but I think these are still worthwhile.

A regression in the integration test led  me down a path of simplifying the handling of cloning instructions. Now, instead of each step checking what the instructions are, we handle that in the flight constructor and document the assumptions in the steps.


zloery (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/StoragetransferConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class StoragetransferConfiguration {
+
+  @Bean
+  public Storagetransfer getStoragetransfer() {
+    try {
+      return StorageTransferServiceUtils.createStorageTransferService();
```
zloery, line 15: Looking at the test failures: `StorageTransferServiceUtils.createStorageTransferService()` tries to read Application Default Credentials, but those aren't (and shouldn't be) provided for unit tests. Putting this in a Bean this way means that startup is now dependent on having ADCs available, which breaks unit tests.

In other places we need ADCs, we've either gated the call behind a config option that we disable for unit tests ([like this for Cloud Resource Library](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/crl/CrlService.java#L78)), or caught the exception -> logged a warning saying "this is fine for unit tests but not anything else" -> mock out the object that needs ADC whenever it's actually used ([like this for SamService](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/iam/SamService.java#L158))

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/StoragetransferConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+import java.io.IOException;
+import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class StoragetransferConfiguration {
+
+  @Bean
+  public Storagetransfer getStoragetransfer() {
```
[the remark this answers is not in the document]
jaycarlton: Maybe, but that's a larger blast radius than I'm hoping for with this PR.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/StoragetransferConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class StoragetransferConfiguration {
+  private static final Logger logger = LoggerFactory.getLogger(StoragetransferConfiguration.class);
+
+  @Bean
+  public GoogleCredentials getGoogleCredentials() {
```
ddietterich, line 19: What is the advantage of having this be a Bean?
What was broken about the previous implementation?
Could we use a "normal" singleton that makes the object on first reference? It would properly barf if called during unit tests.

ddietterich (commented):
I'm not so sure about this. Seems like the bean construct is more trouble than it is worth.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/StoragetransferConfiguration.java
jaycarlton: I made `Storagetransfer` into a bean so I could mock it out in the unit test. It depends on this `GoogleCredentials` object in a non-springy way, so I made it a bean too. You're right that it just kicks the can down the road a bit in terms of failing to load the ADCs, but at least now it fails to create a bean with the name of the thing it can't find. ✌🏻 

I'm not sure what a normal singleton is in this context. A spring Component is the usual way to do that, right?

ddietterich: You don't need something to be a bean to be able to mock. You can mock any class or interface.
One way to do this is to make a @Component that has a (non-static) method that does a getOrCreate of the Storagetransfer object. Then in your test, mock that method.

Alternately, you could make a non-Spring static class that did the getOrCreate. I think you could use `MockedStatic` do mock it, but I have never used that myself so don't know if it is weird.

jaycarlton: Haven't tried `MockedStatic`. Your first method sounds like a way to make a bean with more steps, since Components are beans too.

ddietterich: The difference is that the bean gets created all the time. The indirection through a method only happens when we need to real transfer.

jaycarlton: Split the bean via profiles.

jaycarlton:
The integration test failure indicates a pretty serious bug, which I'm working on now. Somehow even though the bucket is marked `COPY_NOTHING`, we're exercising storage transfer code and failing. One problem is that we check the copy instructions pretty late in the game, in `CopyGcsBucketDefinitionStep`. I think we should actually do this in the flight constructor and omit all the steps that don't apply when it's a no-op. Should speed things up too.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/StoragetransferConfiguration.java
```diff
@@ -0,0 +1,20 @@
+package bio.terra.workspace.service.resource.controlled.flight.clone.bucket;
+
+import com.google.api.services.storagetransfer.v1.Storagetransfer;
+import java.io.IOException;
+import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class StoragetransferConfiguration {
```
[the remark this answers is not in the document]
jaycarlton: I was calling it `StorageTransferService`, but that's misleading. It made it seem like I'd written a service to wrap this one. Not that that's necessarily a bad idea. So for now, I'm calling it what it is.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CopyGcsBucketDefinitionStep.java
```diff
@@ -23,45 +23,35 @@
...
-              .effectiveCloningInstructions(cloningInstructions.toApiModel())
-              .bucket(null)
-              .sourceWorkspaceId(sourceBucket.getWorkspaceId())
-              .sourceResourceId(sourceBucket.getResourceId());
-      FlightUtils.setResponse(flightContext, noOpResult, HttpStatus.OK);
-      return StepResult.getStepResultSuccess();
-    } // todo: handle COPY_REFERENCE PF-811, PF-812
+    FlightUtils.validateRequiredEntries(inputParameters,
+        ControlledResourceKeys.DESTINATION_WORKSPACE_ID);
+     // todo: handle COPY_REFERENCE PF-811, PF-812
```
ddietterich, line 54: I think the comment is obsolete. PF-811 is DONE and PF-812 is CLOSED

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CreateStorageTransferServiceJobStep.java
```diff
@@ -63,45 +71,45 @@ public StepResult doStep(FlightContext flightContext)
...
-      //   referenced bucket.
-      final String transferServiceSAEmail =
-          workingMap.get(ControlledResourceKeys.STORAGE_TRANSFER_SERVICE_SA_EMAIL, String.class);
-      logger.debug("Storage Transfer Service SA: {}", transferServiceSAEmail);
+    } catch (GoogleJsonResponseException e) {
+      logger.info("No pre-existing transfer job named {} found.", transferJobName);
+    } catch (IOException e) {
+      return new StepResult(
+          StepStatus.STEP_RESULT_FAILURE_FATAL,
+          new IllegalStateException("Failed to check for existing storage transfer job.", e));
```
ddietterich, line 92: Any reason to retry the step here? Or for the other transfer service calls below?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CreateStorageTransferServiceJobStep.java
```diff
@@ -114,29 +122,28 @@ private void createTransferJob(
...
+        storagetransfer.transferJobs().create(transferJobInput).execute();
     logger.debug("Created transfer job {}", transferJobOutput);
   }
 
   // Remove the Storage Transfer Job. Any data in the destination bucket will be deleted in the
   // previous step's undo method.
   @Override
   public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
-    return StorageTransferServiceUtils.deleteTransferJobStepImpl(flightContext);
+    return StorageTransferServiceUtils.deleteTransferJobStepImpl(flightContext, storagetransfer);
```
ddietterich, line 146: A failure here will be a DISMAL FAILURE from Stairway. Maybe that is the right thing; perhaps add a comment to that effect?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/SkipBucketCloneStep.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    final ApiClonedControlledGcpGcsBucket noOpResult =
+        new ApiClonedControlledGcpGcsBucket()
+            .effectiveCloningInstructions(ApiCloningInstructionsEnum.NOTHING)
+            .bucket(null)
+            .sourceWorkspaceId(sourceBucket.getWorkspaceId())
+            .sourceResourceId(sourceBucket.getResourceId());
+    FlightUtils.setResponse(context, noOpResult, HttpStatus.OK);
```
ddietterich, line 33: So this step doesn't skip anything. Maybe call it SetNullBucketCloneResponseStep or some such?
It would be nice if there was one place in the flight that populated the response.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/StoragetransferConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+    } catch (IOException e) {
+      throw new BeanCreationException(e.getMessage());
+    }
+  }
+
+  /**
+   * We don't need ADC in unit tests, but we do need to provide something that looks like them
+   * in unit tests.
+   */
+  @Bean
```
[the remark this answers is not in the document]
ddietterich: I think it is best to keep the two sides of the condition in the same place as they are here.

ddietterich (commented):
This is looking good. I appreciate the value of the bean for transfer service.
I have scattered questions.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/BucketCloneRolesService.java
```diff
@@ -8,16 +8,18 @@
...
 import com.google.cloud.Role;
+import javax.annotation.Nullable;
+import javax.validation.constraints.Null;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Component;
+import org.springframework.stereotype.Service;
 
-@Component
-public class BucketCloneRolesComponent {
+@Service
```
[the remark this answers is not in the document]
jaycarlton: It was confusing me to have both, so I made this one a service. I read a little article on it last week, and from what I can tell, `Component` is a slightly higher level of abstraction.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CopyGcsBucketDefinitionStep.java
jaycarlton: Yeah, nobody has asked for clone-resource-to-reference; it was just implied by the structure of the cloning instructions as I read it.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CreateStorageTransferServiceJobStep.java
jaycarlton: You mean reason not to retry it? I guess we can retry.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
```diff
@@ -91,7 +91,7 @@ public static void validateRequiredEntries(FlightMap flightMap, String... keys)
     for (String key : keys) {
       if (null == flightMap.getRaw(key)) {
         throw new MissingRequiredFieldsException(
-            String.format("Required entry with key '%s' missing from flight map.", key));
+            String.format("Required entry with key %s missing from flight map.", key));
```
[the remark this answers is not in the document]
jaycarlton: They are getting encoded as `&amp;#39` or some such. I believe this was for security reasons, but I didn't think single quotes would get escaped.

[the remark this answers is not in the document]
ddietterich: OK, that is ugly. I wonder if there is a safe character to use.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/BucketCloneRolesService.java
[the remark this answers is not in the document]
ddietterich: It is purely for documentation. It has no additional function. I don't object to using it, but we should probably be consistent across the code base.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CopyGcsBucketDefinitionStep.java
ddietterich: Why leave the comment, though?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CreateStorageTransferServiceJobStep.java
ddietterich: Our general strategy has been to go fatal and then discover through integration testing what cases we should retry. It is not necessarily a bad strategy. What I want to know is, do we know of any error responses from the transfer service that we should retry? Or do we have to find and weed them later?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CreateStorageTransferServiceJobStep.java
ddietterich: I think I was wrong, though. The utility method does not error, it logs and continues. Thought I deleted this comment.

ddietterich (approved):
LGTM

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/DeleteStorageTransferServiceJobStep.java
```diff
@@ -5,27 +5,32 @@
...
+ *
+ * <p>Preconditions: Cloning instructions are COPY_RESOURCE. Transfer Service Job exists in the
+ * control plane project.
+ *
+ * <p>Post conditions; STS Job no longer exists.
  */
 public class DeleteStorageTransferServiceJobStep implements Step {
 
+  private final Storagetransfer storagetransfer;
+  private final CloningInstructions cloningInstructions;
```
zloery, line 22: looks like this is now unused

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/StoragetransferConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.IOException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+
+@Configuration
+public class StoragetransferConfiguration {
```
zloery, line 17: Given that this is making globally available Beans, I thnk this file should be next to other `@Configuration`s in `bio/terra/workspace/app/configuration` instead of deeply nested inside  `workspace/service`

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/StoragetransferConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(StoragetransferConfiguration.class);
+
+  @Bean
+  @Profile("!unit-test")
+  public GoogleCredentials getGoogleCredentials() {
```
zloery, line 22: Should this (and fakeGoogleCredntials) be in a separate file, something like `GoogleCredentialsConfiguration`? If I'm trying to track down where an autowired `GoogleCredentials` object came from, I wouldn't expect it to be in `StoragetransferConfiguration`

zloery (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/StoragetransferConfiguration.java
jaycarlton: In IJ ultimate, you can just click on the symbol next to the bean injection site...but good point.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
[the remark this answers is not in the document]
jaycarlton: I see it in other places too: ```Error getting proxy group from Sam: Missing the required parameter &amp;#39;email&amp;#39; when calling getProxyGroup(Async)```

PR #610, opened 2022-04-07
[PF-1500] determine if job exists before trying to delete it

jaycarlton:
Fix a bug where a failed UndoStep causes a dismal failure.


yuhuyoyo (approved):

ddietterich (approved):
lgtm - thanks!!

PR #612, opened 2022-04-07
[PF-1499] install java in the notebook instance if it's not present

yuhuyoyo:


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -44,6 +44,14 @@ sudo -u "${JUPYTER_USER}" sh -c "/opt/conda/bin/nbstripout --install --global"
 # Install Nextflow. Use an edge release that allows overriding the default compute engine SA and VPC network
 export NXF_VER=21.05.0-edge
 export NXF_MODE=google
+
+if [[ -n "$(which java)" ]];
+then
+  echo "java is installed"
+else
+  sudo apt-get -y install default-jre
```
jaycarlton, line 52: are you sure this is the same JRE we test the CLI with? Is it Java 11? Maybe we need to peg it to a version.

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
yuhuyoyo: it is default to 11. I specified it.

zloery (approved):

jaycarlton (commented):
Make sure to test this on images that already have java, so that it doesn't error out. I think you're fine.

PR #613, opened 2022-04-07
[DDO-1993] Have WSM's nightly tests provide IAP auth header

jack-r-warren:
DevOps is preparing to move ArgoCD out from behind the VPN-only firewall, and instead use Identity Aware Proxy.

Right now, this action uses a self-hosted GitHub runner that's allowed in the firewall, and that's enough to "see" Argo. With IAP, actions will need to pass a Proxy-Authorization header ([docs](https://docs.google.com/document/d/1FCVPfCjJMF_ljBTeG6bJwbMUCe52kSsbKWTXCqdO7Nw/edit#heading=h.8gl88su5pqu2)) if they want to talk directly to Argo.

While I was at it, I updated the action here to use ArgoCD's CLI instead of the API directly. It'll be a bit more resilient to updates we do to Argo, but more importantly it will wait for the sync itself to complete before continuing. It's not perfect, because it doesn't wait for _WSM_ to come back online, but it waits for all the Kubernetes work to finish, which is more than before.

I've tested the action [here](https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/2116665541) and confirmed the syncing behavior looking at Argo itself.

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   nightly-tests:
 
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
 
     permissions:
       contents: 'read'
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -36,9 +36,10 @@ jobs:
           workload_identity_provider: 'projects/1038484894585/locations/global/workloadIdentityPools/github-wi-pool/providers/github-wi-provider'
           service_account: 'dsp-tools-iap-access@dsp-tools-k8s.iam.gserviceaccount.com'
           # Create `steps.auth.outputs.id_token` that can access Identity Aware Proxy
-          token_format: 'id_token'
-          id_token_audience: '${{ secrets.IAP_ID_TOKEN_AUDIENCE }}'
-          id_token_include_email: true
+          #token_format: 'id_token'
+          #id_token_audience: '${{ secrets.IAP_ID_TOKEN_AUDIENCE }}'
+          #id_token_include_email: true
+          token_format: 'access_token'
 
       # Install postgres:
       # - Create the file repository configuration.
@@ -96,6 +97,7 @@ jobs:
       # for a bit and just accept that in some crazy long cases, these tests will fail.
       # The sync is bracketed by /version probes so we have the pre and post sync versions
       # in the log.
+      # --header "Proxy-Authorization: Bearer ${{ steps.auth.outputs.id_token }}" \
       - name: ArgoCD sync
         run: |
           version=$(curl https://workspace.wsmtest.integ.envs.broadinstitute.org/version)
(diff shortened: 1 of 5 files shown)


ddietterich:
I tagged myself and @zloery as reviewers. Zach specifically to understand any Verily impact.
@jaycarlton FYI - if you are interested, by all means take a look

.github/workflows/nightly-tests.yml
```diff
@@ -74,23 +90,31 @@ jobs:
...
 
-      # The ArgoCD sync triggers the synchronization of the wsmtest cluster. Typically,
-      # the cluster is reset within a few seconds. However, it is possible for it to take
-      # minutes or hours. We have no way to check that the sync is complete. We sleep
-      # for a bit and just accept that in some crazy long cases, these tests will fail.
+      # The ArgoCD sync triggers the synchronization of the wsmtest cluster. The ArgoCD CLI
+      # will wait for the initial sync to complete, but it won't necessarily wait for WSM's
+      # new version to actually come fully online--which could take minutes or hours in rare
+      # cases. We sleep for two minutes to try to allow WSM to come online.
       # The sync is bracketed by /version probes so we have the pre and post sync versions
```
jaycarlton, line 97: Could you put a step here specifically to ping WSM somehow until it's online? For example, it could curl the unauthenticated `/status` endpoint.

.github/workflows/nightly-tests.yml
```diff
@@ -74,23 +90,31 @@ jobs:
...
+            --auth-token '${{ secrets.WSMTEST_SYNC_ARGOCD_TOKEN }}' \
+            --header 'Proxy-Authorization: Bearer ${{ steps.auth.outputs.id_token }}' \
+            app sync workspacemanager-wsmtest
           sleep 120
           version=$(curl https://workspace.wsmtest.integ.envs.broadinstitute.org/version)
           echo "$(date "+%Y-%m-%dT%H:%M:%S") post-sync wsmtest version: $version"
+      - name: Remove ArgoCD CLI
+        if: always()
+        run: |
+          rm -f ./argocd
```
jaycarlton, line 117: Is this necessary? It'll all get cleaned up when the container goes down, right?

jack-r-warren: In theory, but these are self-hosted runners right now which aren't as ephemeral as GitHub's own (self-hosted runners actually aren't ephemeral by default). It _should_ be unnecessary, but it seemed zero-cost to make sure.

.github/workflows/nightly-tests.yml
jack-r-warren: WSM doesn't go down, it's that the rolling zero-downtime eventually-consistent update across an effectively arbitrary number of replicas is hard to introspect without additional work.

We're actually working on that problem, among others, for our "on-demand Terra environments with specific app versions" project. My ticket here is just for the proxy-authorization header so I can get that work unblocked, and I figured a switch to Argo's CLI would be an easy win while I was at it. I'm on a bit of a time crunch so if you think it makes sense to do a narrower change and pass the header through the old curl command, I'm fine with that too. This flow for interacting with Argo isn't one we actively support--we just don't want to break it if we can help it.

zloery (approved):
LGTM! The test failure looked unrelated to this change so I kicked off a rerun.

@ddietterich no concerns from me about the Verily side since we don't use Argo and maintain separate actions

.github/workflows/nightly-tests.yml
ddietterich: Good idea, but I think it is out of scope for this change.

jaycarlton: No, you're good. I just didn't approve b/c I wasn't a designated reviewer.

ddietterich (approved):
LGTM

jack-r-warren:
Thanks all!

PR #614, opened 2022-04-08
[PF-1530] Make connected test dataset names more unique

zloery:
We recently had a failure due to a BQ dataset name conflict in `ControlledResourceServiceTest` resources. This should make those conflicts much less likely.

I also fixed a `Logger` bug in `ControlledBigQueryDatasetLifecycle`, I think the wrong class got copied when this test was written, which gave us slightly misleading logs.


jaycarlton (approved):

ddietterich (approved):
LGTM

PR #616, opened 2022-04-12
[PF-900] connected tests for workspace and bucket clone

jaycarlton:


service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -481,6 +515,86 @@ void createGoogleContextRawlsStageThrows() throws Exception {
                 request.getWorkspaceId(), jobId, USER_REQUEST, "/fake/value"));
   }
 
+  @Test
+  public void cloneGcpWorkspace() {
```
ddietterich, line 519: Is this really a connected test? It looks like you have mocked out everything that touches the cloud platform.

ddietterich (approved):
LGTM - one question

service/src/test/java/bio/terra/workspace/connected/WorkspaceConnectedTestUtils.java
```diff
@@ -1,11 +1,16 @@
...
 
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doReturn;
 
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.iam.model.SamConstants;
 import bio.terra.workspace.service.job.JobService;
 import bio.terra.workspace.service.spendprofile.SpendConnectedTestUtils;
+import bio.terra.workspace.service.spendprofile.SpendProfile;
```
zloery, line 11: nit: it looks like you're not using any of the new imports, remove?
I'm surprised spotless didn't clean these up

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -481,6 +515,86 @@ void createGoogleContextRawlsStageThrows() throws Exception {
...
+    // destination WS should exist
+    final Workspace retrievedDestinationWorkspace = workspaceService.getWorkspace(
+        destinationWorkspace.getWorkspaceId(), USER_REQUEST);
+    assertEquals("Destination Workspace", retrievedDestinationWorkspace.getDisplayName().orElseThrow());
+    assertEquals("Copied from source", retrievedDestinationWorkspace.getDescription().orElseThrow());
+    assertEquals(WorkspaceStage.MC_WORKSPACE, retrievedDestinationWorkspace.getWorkspaceStage());
+
+    // Destination Workspace should have a GCP context
+    assertNotNull(gcpCloudContextService.getGcpCloudContext(destinationWorkspace.getWorkspaceId()).orElseThrow());
+
```
zloery, line 592: It looks like you're also cloning a bucket from the source workspace, should this have any assertions about the cloned bucket?

zloery (approved):

service/src/test/java/bio/terra/workspace/connected/WorkspaceConnectedTestUtils.java
jaycarlton: Spotless appears to be not working on the test source set anymore. I might have broken that.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -63,33 +87,43 @@ class WorkspaceServiceTest extends BaseConnectedTest {
...
-                Mockito.eq(SamSpendProfileAction.LINK)))
+                eq(SamSpendProfileAction.LINK)))
         .thenReturn(true);
+    final String policyGroup = "terra-workspace-manager-test-group@googlegroups.com";
     // Return a valid google group for cloud sync, as Google validates groups added to GCP projects.
     Mockito.when(mockSamService.syncWorkspacePolicy(any(), any(), any()))
-        .thenReturn("terra-workspace-manager-test-group@googlegroups.com");
+        .thenReturn(policyGroup);
+
+    doReturn(policyGroup).when(mockSamService).syncResourcePolicy(
```
ddietterich, line 124: This seems redundant to the previous mock, just with different syntax. It takes `any()` which I understand to be a superset of the `any(ControlledResource.class)` you have here.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
ddietterich: I think the problem is more general. This test started just testing workspace. Then it grew some cloud context tests. Then it grew a delete workspace test with a reference to make sure we deleted those. Now it is growing to actually create a controlled object. I am thinking this test shouldn't be in this module.

We would have a module that has only basic workspace tests, like duplicate name and bad name, etc. Keep that separate from tests that create resources.

jaycarlton: That's a good question in general about the utility of connected vs integration tests. I don't really know the strategy (and certainly didn't when I initially filed the ticket). Cloning is slow, so this adds considerably to the test burden.

We could split out more complex tests from this test class (if that's what you mean by module), but that would introduce some duplication in the form of the application context setup. Maybe we can make a shareable app context for this kind of test, but we're getting out of scope for this PR. I did add a bucket clone only test to the `ControlledResourceServiceTest`, but I thought cloning a whole workspace wouldn't fit there (as it might bring in other dependencies). So it's a square peg either way.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
jaycarlton: The previous mock is for the function `syncWorkspacePolicy()` and this one is for `syncResourcePolicy()`. Using a naked `any()` makes me feel icky, like I'm writing JavaScript.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
ddietterich: Agreed that this is a square peg and agree that we do not have a clean separation of connected and integration tests. Let’s go ahead with this one as is. I think it would be worth sorting out the tests and maybe running different sets as PR vs nightly tests.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
ddietterich: Ah, I missed that. Thanks for explaining!

PR #617, opened 2022-04-12
[PF-1485] Update dependencies, Spring Boot version

zloery:
This bumps the versions for a number of dependencies, including Spring Boot. Many of the transitive dependencies here had identified vulnerabilities in scans of other services, so I figured I'd update them at the same time as Spring. I've also updated the Artifactory gradle plugin to the most recent version (matching RBS), as older versions had some conflicts with newer gradle versions.

I've also removed two previous changes specific to older spring versions:
- Removed the `log4j` version pinning, the current version of Spring Boot already pulls in a newer version of log4j-api and log4j-to-slf4j dependencies
- Removed `SetDisallowedBinderFields` added in #600, this was a patch until we could upgrade to the latest version of Spring Boot.


service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/CreateAzureStorageStepTest.java
```diff
@@ -43,7 +43,7 @@ public class CreateAzureStorageStepTest extends BaseAzureTest {
   @Mock private AzureConfiguration mockAzureConfig;
   @Mock private AzureCloudContext mockAzureCloudContext;
   @Mock private StorageManager mockStorageManager;
-  @Mock private ManagementException mockException;
+  private ManagementException resourceNotFoundException;
```
zloery, line 46: It seems like the new version of Logback doesn't like the way we previously logged this mock's fields, which led to `createStorage_failsToCreate` throwing an NPE inside the step code.

I'm also not sure why we needed to mock out an exception in the first place, though, so I replaced it with a real object here.

service/gradle/dependencies.gradle
```diff
@@ -14,23 +14,23 @@ dependencies {
...
 
   // Get stairway via TCL
-  implementation group: "bio.terra", name: "terra-common-lib", version: "0.0.54-SNAPSHOT"
+  implementation group: "bio.terra", name: "terra-common-lib", version: "0.0.56-SNAPSHOT"
   implementation group: "org.broadinstitute.dsde.workbench", name: "sam-client_2.12", version: "0.1-61135c7"
   implementation group: "bio.terra", name: "terra-resource-buffer-client", version: "0.4.3-SNAPSHOT"
 
   // Cloud Resource Library
-  implementation group: 'bio.terra', name: 'terra-cloud-resource-lib', version: "1.0.4-SNAPSHOT"
+  implementation group: 'bio.terra', name: 'terra-cloud-resource-lib', version: "1.1.1-SNAPSHOT"
```
jaycarlton, line 22: Yay, bumptagbot is back in business. Only took 9 PRs.

jaycarlton (approved):
Thanks for volunteering to do all these!

ddietterich (approved):
LGTM - thanks

PR #620, opened 2022-04-18
[PF-1575] validate cloning instructions

jaycarlton:
Make sure cloning instructions are allowed per stewardship type. In retrospect, we probably should've had two separate enums for cloning instructions.


service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -75,6 +77,8 @@ public ResponseEntity<ApiCreatedControlledAzureDisk> createAzureDisk(
       UUID workspaceId, ApiCreateControlledAzureDiskRequestBody body) {
     features.azureEnabledCheck();
 
+    ResourceValidationUtils.validateCloningInstructions(
```
zloery, line 80: Since this method should be called every time you create a resource, could this call be in `WsmResource.validate()` instead? We already call `validate()` on subclasses (which always call the base class validation method) as part of every resource type's `build()` method, and this matches other static validation we do (like name/description checks) that live in that method.

jaycarlton: Capital idea!

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gcsbucket/GcsApiConversions.java
```diff
@@ -258,16 +257,19 @@ public static DateTime toGoogleDateTime(@Nullable OffsetDateTime offsetDateTime)
...
             Duration.ofSeconds(offsetDateTime.getOffset().getTotalSeconds()).toMinutes()));
   }
+
   @Nullable
   public static DateTime toGoogleDateTimeDateOnly(@Nullable OffsetDateTime offsetDateTime) {
     if (offsetDateTime == null) {
       return null;
     }
-    return new DateTime(true,
+    return new DateTime(
```
jaycarlton, line 266: I got spotless working by doing `./gradlew build -x test`.

jmczerk (approved):

PR #622, opened 2022-04-19
[PF-1548] update cloning instructions for buckets

jaycarlton:
Add `cloningInstructions` to `updateParameters` and treat it as nullable.


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gcsbucket/UpdateGcsBucketStep.java
```diff
@@ -108,6 +113,13 @@ private StepResult updateBucket(@Nullable ApiGcpGcsBucketUpdateParameters update
       bucketCowBuilder.setStorageClass(gcsStorageClass);
     }
 
+    final boolean replaceCloningInstructions = null != updateParameters.getCloningInstructions();
```
ddietterich, line 116: I think this is in the wrong place. I think updating the cloning instructions should follow the update metadata path and not be part of this step. Also, I think it can be done in the same update transaction as the other DAO update. We do not need to run a separate transaction.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -490,6 +490,20 @@ public boolean updateResource(
     return updateResourceWorker(workspaceId, resourceId, name, description, /*attributes=*/ null);
   }
 
+  @WriteTransaction
+  public boolean updateResourceCloningInstructions(
```
ddietterich, line 494: Why do we need this method? I think we should generalize the `updateResource()` method. We have been making incremental changes. Perhaps we need a more general way to say: here is a resource object, please update these fields. Maybe?

In any case, we should update the metadata in one transaction.

ddietterich (changes requested):
I don't think we should run a DAO transaction while updating the Cow. Also think we should generalize the `ResourceDao.updateResource()` to update the cloning instruction rather than make a new method.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
jaycarlton: What you're describing is starting to sound like a data transfer object, and you can use Spring to build repositories around such objects that give you many kinds of queries for "free". I could do this a couple of ways. One way is to pass in a map of key-value pairs which gets added to the params map. Another would be to walk all the fields of the resource structure and update all the non-null values, though in that case we'd frequently be updating values that didn't change, which is theoretically harmless, if unapetizing.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gcsbucket/UpdateGcsBucketStep.java
jaycarlton: Good point.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -474,8 +477,10 @@ public boolean updateResource(
       UUID resourceId,
       @Nullable String name,
       @Nullable String description,
-      @Nullable String attributes) {
-    return updateResourceWorker(workspaceId, resourceId, name, description, attributes);
+      @Nullable String attributes,
+      @Nullable CloningInstructions cloningInstructions) {
```
zloery, line 481: It's a little concerning to me that changing this signature doesn't require updating any unit tests. Out of scope for this PR, but I think our unit test coverage needs some work

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledResourceMetadataStep.java
```diff
@@ -50,7 +52,8 @@ public StepResult undoStep(FlightContext flightContext) throws InterruptedExcept
     final String previousName = workingMap.get(ResourceKeys.PREVIOUS_RESOURCE_NAME, String.class);
     final String previousDescription =
         workingMap.get(ResourceKeys.PREVIOUS_RESOURCE_DESCRIPTION, String.class);
-    resourceDao.updateResource(workspaceId, resourceId, previousName, previousDescription);
+    resourceDao.updateResource(
+        resource.getWorkspaceId(), resource.getResourceId(), previousName, previousDescription);
```
ddietterich, line 56: This undo will not restore the cloning instructions. Shouldn't it?

ddietterich (commented):
Looking good. There is one spot where I _think_ you need to restore previous cloning instructions.

zloery (commented):
This LGTM, but I don't think this functionality is usable unless you also update the generated API + endpoint in the controller - is the plan to do that in a separate PR?

jaycarlton:
> This LGTM, but I don't think this functionality is usable unless you also update the generated API + endpoint in the controller - is the plan to do that in a separate PR?

Whoa, good catch. I just spaced on this. I'll fix it in the present PR. I need to adjust an integration test to exercise the controller.

ddietterich (approved):
LGTM

openapi/src/parts/controlled_gcp_gcs_bucket.yaml
```diff
@@ -192,6 +192,8 @@ components:
           $ref: '#/components/schemas/GcpGcsBucketDefaultStorageClass'
         lifecycle:
           $ref: '#/components/schemas/GcpGcsBucketLifecycle'
+        cloningInstructions:
```
zloery, line 195: style q: the other metadata fields `name` and `description` are properties of the request object (`UpdateControlledGcpGcsBucketRequestBody`), so one level up from `GcpGcsBucketUpdateParameters`, while `defaultStorageClass` and `lifecycle` are GCS-specfic things we change in GCP. Should cloningInstructions be a property alongside the other resource metadata, or are these split for a different reason?

zloery (approved):
LGTM - one question about API structure

openapi/src/parts/controlled_gcp_gcs_bucket.yaml
jaycarlton: Good question. I figured it was a less intrusive change to add it to the update parameters, where other new updatable things would likely go. Didn't consider the GCP angle.

PR #624, opened 2022-04-20
PF-1442: start ssh agent and clone all the git repos references

yuhuyoyo:


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -76,9 +76,12 @@ fi
 # Log in with app-default-credentials
 sudo -u "${JUPYTER_USER}" sh -c "terra auth login --mode=APP_DEFAULT_CREDENTIALS"
 
-# Get all the git repvoio clone urls from the workspace resource list with type GIT_REPO and try to clone all of them.
+sudo -u "${JUPYTER_USER}" sh -c "mkdir -p /home/${JUPYTER_USER}/.ssh"
+cd /home/${JUPYTER_USER}/.ssh
+sudo -u "${JUPYTER_USER}" sh -c "terra user ssh-key get --format=JSON | jq -r '.privateSshKey'> id_rsa"
```
zloery, line 81: What happens if the user hasn't already registered an SSH key? I don't know how notebooks handle failure in startup scripts. 

also, the previous approach allowed users to clone public repos without setting up an ssh key with Github, but it seems like that wouldn't happen now. Is that behavior we care about, or was that just useful before we had ssh keys?

cbookg: +1, I think we still need to try the repos below even if the key fails.  and likely ensure we try them all such that we can clone a public one even if a private one fails?

yuhuyoyo: good point. done

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -76,9 +76,15 @@ fi
...
+sudo -u "${JUPYTER_USER}" sh -c "mkdir -p /home/${JUPYTER_USER}/.ssh"
 cd /home/${JUPYTER_USER}
-sudo -u "${JUPYTER_USER}" sh -c "terra resource list --type=GIT_REPO --format=JSON" | jq -r 'unique_by(.gitRepoUrl) | .[] | {gitRepoUrl} | join (" ") ' | while read -r repoUrl; do
-  sudo -u "${JUPYTER_USER}" sh -c "git clone ${repoUrl}" && echo "cloned ${repoUrl}" || echo "clone ${repoUrl} failed"
-done
+readonly TERRA_SSH_KEY=$(sudo -u "${JUPYTER_USER}" sh -c "terra user ssh-key get --format=JSON")
+if [[ -n "${TERRA_SSH_KEY}" ]]; then
+  printf '%s' "${TERRA_SSH_KEY}" | sudo -u "${JUPYTER_USER}" sh -c "jq -r '.privateSshKey'> .ssh/id_rsa"
+  sudo -u "${JUPYTER_USER}" sh -c 'chmod go-rwx .ssh/id_rsa'
+  echo eval '"$(ssh-agent -s)"' >> .bash_profile
```
cbookg, line 85: is this starting the agent?  i think we should start the agent even if we don't have a key

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -76,9 +76,15 @@ fi
...
 
-# Get all the git repvoio clone urls from the workspace resource list with type GIT_REPO and try to clone all of them.
+sudo -u "${JUPYTER_USER}" sh -c "mkdir -p /home/${JUPYTER_USER}/.ssh"
 cd /home/${JUPYTER_USER}
-sudo -u "${JUPYTER_USER}" sh -c "terra resource list --type=GIT_REPO --format=JSON" | jq -r 'unique_by(.gitRepoUrl) | .[] | {gitRepoUrl} | join (" ") ' | while read -r repoUrl; do
-  sudo -u "${JUPYTER_USER}" sh -c "git clone ${repoUrl}" && echo "cloned ${repoUrl}" || echo "clone ${repoUrl} failed"
-done
+readonly TERRA_SSH_KEY=$(sudo -u "${JUPYTER_USER}" sh -c "terra user ssh-key get --format=JSON")
+if [[ -n "${TERRA_SSH_KEY}" ]]; then
+  printf '%s' "${TERRA_SSH_KEY}" | sudo -u "${JUPYTER_USER}" sh -c "jq -r '.privateSshKey'> .ssh/id_rsa"
```
jaycarlton, line 83: nit: space before `>`.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -76,9 +76,15 @@ fi
...
+  sudo -u "${JUPYTER_USER}" sh -c 'chmod go-rwx .ssh/id_rsa'
+  echo eval '"$(ssh-agent -s)"' >> .bash_profile
+  sudo -u "${JUPYTER_USER}" sh -c 'eval `ssh-agent -s`; ssh-add .ssh/id_rsa; ssh-keyscan -H github.com >> ~/.ssh/known_hosts'
+fi
+
+sudo -u "${JUPYTER_USER}" sh -c 'terra git clone --all'
```
jaycarlton, line 89: So right now we're limited to one key for all repos (except public ones). I'm trying to think if that impacts the script in any way or if we should just document that.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -76,9 +76,15 @@ fi
 # Log in with app-default-credentials
 sudo -u "${JUPYTER_USER}" sh -c "terra auth login --mode=APP_DEFAULT_CREDENTIALS"
```
jaycarlton, line 77: I don't _think_ we need the curly braces here. Does `"$JUPYPTER_USER"` amount to the same thing?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -76,9 +76,15 @@ fi
...
 
-# Get all the git repvoio clone urls from the workspace resource list with type GIT_REPO and try to clone all of them.
+sudo -u "${JUPYTER_USER}" sh -c "mkdir -p /home/${JUPYTER_USER}/.ssh"
 cd /home/${JUPYTER_USER}
-sudo -u "${JUPYTER_USER}" sh -c "terra resource list --type=GIT_REPO --format=JSON" | jq -r 'unique_by(.gitRepoUrl) | .[] | {gitRepoUrl} | join (" ") ' | while read -r repoUrl; do
-  sudo -u "${JUPYTER_USER}" sh -c "git clone ${repoUrl}" && echo "cloned ${repoUrl}" || echo "clone ${repoUrl} failed"
-done
+readonly TERRA_SSH_KEY=$(sudo -u "${JUPYTER_USER}" sh -c "terra user ssh-key get --format=JSON")
```
jaycarlton, line 81: Does this error out if there's no SSH key, or just return an empty structure?

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
yuhuyoyo: i believe it does. i just wanted to keep the style consistent. i can remove it here tho..

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
yuhuyoyo: it returns empty

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
yuhuyoyo: done.

zloery (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -76,9 +76,19 @@ fi
...
+
+# Start the ssh-agent. Set this command in bash_profile so everytime user starts a shell, we start the ssh-agent.
+echo eval '"$(ssh-agent -s)"' >> .bash_profile
+if [[ -n "$TERRA_SSH_KEY" ]]; then
+  printf '%s' "$TERRA_SSH_KEY" | sudo -u "${JUPYTER_USER}" sh -c "jq -r '.privateSshKey' > .ssh/id_rsa"
+  sudo -u "$JUPYTER_USER" sh -c 'chmod go-rwx .ssh/id_rsa'
+  sudo -u "$JUPYTER_USER" sh -c 'eval `ssh-agent -s`; ssh-add .ssh/id_rsa; ssh-keyscan -H github.com >> ~/.ssh/known_hosts'
```
zloery, line 88: Line 84 has `echo eval '"$(ssh-agent -s)"' >> .bash_profile` is already a line above, do you also need `eval `ssh-agent -s` here?

yuhuyoyo: good catch. this was left over from before i added line 84.

PR #625, opened 2022-04-22
[PF-1576, PF-1596] Add Workspace.userFacingId and plumb through to API

melissachang:
Tested with swagger: create succeeds, create fails, update succeeds, update fails.

If someone tries to create a workspace with an existing userFacingId, we show friendly error:

image

Same with update workspace:

image

----

[In Workspace.build()](https://github.com/DataBiosphere/terra-workspace-manager/blob/19c461a3d0e94c5be0f4cf3f674a0c7f3ffb566d/service/src/main/java/bio/terra/workspace/service/workspace/model/Workspace.java#L170), I can't do this:

```
      if (userFacingId == null) {
        userFacingId = "";
      }
```
Because for tests that don't set `userFacingId`, it will be set to `""` in DB. Creating the first workspace works. Creating the second workspace fails; the uniqueness constraint for `userFacingId` fails because there is already a workspace with `userFacingId = ""`.

In PF-1596, I will auto-populate `userFacingId` if it isn't set. In pseudocode:

```
      if (userFacingId == null) {
        generated userFacingId based on displayName or workspaceId
      }

```
---

Instead of creating `DuplicateUserFacingIdException`, my first attempt was to add an enum field to `DuplicateWorkspaceException` indicating which field was being duplicated. I ran into problems with `StairwayExceptionSerializer`; it wasn't worth adding that enum field to `StairwayExceptionFields`.

### DIFF
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -333,7 +333,7 @@ components:
         userFacingId:
           description: |
             Human-settable, mutable id. Optional. If this isn't set, one will
-            be generated based on displayName or id.
+            be generated based id.
           type: string
         displayName:
           description: The human readable name of the workspace
@@ -425,7 +425,7 @@ components:
 
     WorkspaceDescription:
       type: object
-      required: [id]
+      required: [id, userFacingId]
       properties:
         id:
           description: The ID of the workspace
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -333,7 +333,7 @@ components:
         userFacingId:
           description: |
             Human-settable, mutable id. Optional. If this isn't set, one will
-            be generated based id.
+            be generated based on id.
           type: string
         displayName:
           description: The human readable name of the workspace
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -330,6 +330,11 @@ components:
           description: The ID of the workspace
           type: string
           format: uuid
+        userFacingId:
+          description: |
+            Human-settable, mutable id. Optional. If this isn't set, one will
+            be generated based on displayName or id.
+          type: string
         displayName:
           description: The human readable name of the workspace
           type: string
@@ -404,6 +409,11 @@ components:
     UpdateWorkspaceRequestBody:
       type: object
       properties:
+        userFacingId:
+          description: |
+            Human-settable, mutable id. Optional. If this isn't set, one will
+            be generated based on displayName or id.
+          type: string
         displayName:
           description: The human readable name of the workspace
           type: string
@@ -422,6 +432,9 @@ components:
           description: The ID of the workspace
           type: string
           format: uuid
+        userFacingId:
+          description: Human-settable, mutable id
+          type: string
         displayName:
           description: The human readable name of the workspace
           type: string
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -411,8 +411,7 @@ components:
       properties:
         userFacingId:
           description: |
-            Human-settable, mutable id. Optional. If this isn't set, one will
-            be generated based on displayName or id.
+            Human-settable, mutable id
           type: string
         displayName:
           description: The human readable name of the workspace
--- a/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -112,10 +112,14 @@ public class WorkspaceApiController extends ControllerBase implements WorkspaceA
     Optional<SpendProfileId> spendProfileId =
         Optional.ofNullable(body.getSpendProfile()).map(SpendProfileId::new);
 
+    // ET uses userFacingId; CWB doesn't. Schema enforces that userFacingId must be set. CWB doesn't pass
+    // userFacingId in request, so use id. Prefix with "a" because userFacingId must start with letter.
+    String userFacingId = body.getUserFacingId() != null ? body.getUserFacingId() : "a" + body.getId();
+
     Workspace workspace =
         Workspace.builder()
             .workspaceId(body.getId())
-            .userFacingId(body.getUserFacingId())
+            .userFacingId(userFacingId)
             .displayName(body.getDisplayName())
             .description(body.getDescription())
             .spendProfileId(spendProfileId.orElse(null))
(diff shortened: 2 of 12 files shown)


openapi/src/parts/workspace.yaml
```diff
@@ -422,6 +431,9 @@ components:
           description: The ID of the workspace
           type: string
           format: uuid
+        userFacingId:
```
ddietterich, line 434: nit: just to be tidy, I think this should be in the required list.

ddietterich (changes requested):
I did not see the code for create workspace that generates a user facing id if the create request does not supply one. Did I just miss it? I think that needs to be here, or perhaps in PF-1581, before this goes in.

openapi/src/parts/workspace.yaml
melissachang: Done

melissachang (commented):
> I did not see the code for create workspace that generates a user facing id if the create request does not supply one. Did I just miss it? I think that needs to be here, or perhaps in PF-1581, before this goes in.

Done

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -97,13 +100,25 @@ public UUID createWorkspace(Workspace workspace) {
...
+      if (e.getMessage()
+          .contains("duplicate key value violates unique constraint \"workspace_pkey\"")) {
+        throw new DuplicateWorkspaceException(
+            String.format(
+                "Workspace with id %s already exists - display name %s stage %s",
+                workspaceId,
+                workspace.getDisplayName().toString(),
+                workspace.getWorkspaceStage().toString()),
+            e);
+      } else {
```
cbookg, line 113: suggestion: don't assume its the ufid key that fails here as 'else', and have the 'else' be a fallback.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -165,17 +181,23 @@ public Workspace getWorkspace(UUID id) {
...
+      @Nullable String userFacingId,
       @Nullable String name,
       @Nullable String description,
       @Nullable Map<String, String> propertyMap) {
-    if (name == null && description == null && propertyMap == null) {
+    if (userFacingId == null && name == null && description == null && propertyMap == null) {
       throw new MissingRequiredFieldException(
-          "Must specify name, description, or properties to update.");
+              // "ID" instead of "userFacingId" because end user sees this.
+              "Must specify ID, name, description, or properties to update.");
```
cbookg, line 191: suggestion: just "No update provided."

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -193,14 +215,28 @@ public boolean updateWorkspace(
...
+      throw e;
+    }
 
+    boolean updated = rowsAffected > 0;
     logger.info(
-        "{} record for workspace {}",
-        (updated ? "Updated" : "No Update - did not find"),
-        workspaceId);
-
+            "{} record for workspace {}",
```
cbookg, line 237: adding indent here looks wrong?

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
```diff
@@ -39,6 +40,10 @@ public StepResult doStep(FlightContext flightContext)
       if (!workspace.equals(existingWorkspace)) {
         throw ex;
       }
+    } catch (DuplicateUserFacingIdException ex) {
```
cbookg, line 43: why catch this just to throw it?

cbookg (approved):

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
melissachang: Done, changed to `"Must specify field to update."`

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
melissachang: Done

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
melissachang: Done. (This was leftover from when I used the same exception for both cases.)

melissachang (commented):
Ping for @ddietterich

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -112,13 +112,18 @@ public ResponseEntity<ApiCreatedWorkspace> createWorkspace(
     Optional<SpendProfileId> spendProfileId =
         Optional.ofNullable(body.getSpendProfile()).map(SpendProfileId::new);
 
+    // ET uses userFacingId; CWB doesn't. Schema enforces that userFacingId must be set. CWB doesn't pass
+    // userFacingId in request, so use id. Prefix with "a" because userFacingId must start with letter.
+    String userFacingId = body.getUserFacingId() != null ? body.getUserFacingId() : "a" + body.getId();
```
ddietterich, line 117: You can also do this using Optional like
```
String userFacingId = Optional.ofNullable(body.getUserFacingId()).orElse("a" + body.getId());
```
Not that you have to. I am not afraid of nulls, but there seems to be a growing preference for using Optional for this case.

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
```diff
@@ -6,6 +6,7 @@
 import bio.terra.stairway.exception.RetryException;
 import bio.terra.workspace.common.utils.FlightUtils;
 import bio.terra.workspace.db.WorkspaceDao;
+import bio.terra.workspace.service.workspace.exceptions.DuplicateUserFacingIdException;
```
ddietterich, line 9: nit: Is this import used?

ddietterich (approved):
LGTM - two nits to maybe check out

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java
melissachang: Done

PR #626, opened 2022-04-22
[PF-1576] Validate Workspace.userFacingId format

melissachang:


service/src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -84,4 +92,13 @@ public static void validateCloudPlatform(ApiCloudPlatform platform) {
             "Invalid cloud platform. Currently, only AZURE and GCP are supported.");
     }
   }
+
+  public static void validateUserFacingId(String userFacingId) {
+    if (!USER_FACING_ID_VALIDATION_PATTERN.matcher(userFacingId).matches()) {
+      logger.warn("User provided invalid userFacingId: " + userFacingId);
+      // "ID" instead of "userFacingId" because user sees this.
+      throw new ValidationException(
+          "ID must have 3-63 characters, contain lowercase letters, numbers, dash, or underscore, and start with lowercase letter");
```
cbookg, line 101: dash -> dashes

cbookg: underscore -> underscores

service/src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -84,4 +92,13 @@ public static void validateCloudPlatform(ApiCloudPlatform platform) {
             "Invalid cloud platform. Currently, only AZURE and GCP are supported.");
     }
   }
+
+  public static void validateUserFacingId(String userFacingId) {
```
cbookg, line 96: do we have tests for these, or that the validation runs?

melissachang: I tested with Swagger. Unit tests wouldn't exercise this code (since unit test calls service directly). Didn't seem worth making an integration test for this. (And there isn't a ControllerValidationUtils.java for this file. I could make one, but the other parts of this file aren't being unit tested.)

ddietterich: I'd suggest adding tests in `WorkspaceLifecycle.java` integration test. Right now it has no negative cases in it at all which seems like a problem. In general, our integration tests are not doing a good job at testing invalid inputs.

melissachang: Done.

Just curious, is it possible to write a unit test that calls controller (workspaceApi?) instead of service?

ddietterich: I don't know of any way, but that does not mean there isn't one.

cbookg (approved):

PR #627, opened 2022-04-26
[PF-1552] Create Workspace Synchronously

jaycarlton:
Create the destination workspace first synchronously and return its ID directly so that the UI can navigate to a landing page for the workspace while cloning is taking place.


ddietterich (approved):
LGTM

PR #629, opened 2022-04-27
rename workspaceId to workspaceUuid

jaycarlton:
With the new user-facing ID being called "workspace ID" casually in the mocks and specs, I've renamed `workspaceId` to `workspaceUuid` for clarity in all the places it's safe to do so. Notably, this does not include DB column names, API files, or JSON properties. Rename was executed with "Replace in Files" in IntelliJ plus some manual fixup.


melissachang (approved):
Nice!

ddietterich (approved):
I'm not going to look! :)

PR #633, opened 2022-04-28
Fix WorkspaceLifecycle perf test

melissachang:
Fix this failure:

https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/2237678331

### DIFF
--- a/integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
@@ -20,9 +20,12 @@ import scripts.utils.WorkspaceApiTestScriptBase;
 
 public class WorkspaceLifecycle extends WorkspaceApiTestScriptBase {
   private static final Logger logger = LoggerFactory.getLogger(WorkspaceLifecycle.class);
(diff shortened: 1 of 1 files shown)


integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
```diff
@@ -21,7 +21,7 @@
 public class WorkspaceLifecycle extends WorkspaceApiTestScriptBase {
   private static final Logger logger = LoggerFactory.getLogger(WorkspaceLifecycle.class);
   // Perf tests run same test in different threads. Use uuid to make userFacingIds unique.
-  String uuid = UUID.randomUUID().toString();
+  private static final String uuid = UUID.randomUUID().toString();
```
jaycarlton, line 24: nit: these should all be UPPER_SNAKE_CASE.

jaycarlton (approved):

integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
melissachang: Done. Left `uuid` since `UUID` is taken.

PR #634, opened 2022-04-30
Really fix WorkspaceLifecycle perf test

melissachang:
Passing run:

https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/2248615536

Note: `ThreadLocal` doesn't fix the problem:

```
  private static final ThreadLocal<String> uuid =
      ThreadLocal.withInitial(() -> UUID.randomUUID().toString());
```

Because the same thread runs the test multiple times. For example, for WorkspaceLifecycle, there are 2 threads and test needs to run 10 times. So each thread runs the test around 5 times.


ddietterich (approved):
LGTM

cbookg (approved):
LGTM, but I see this is passing in autopush + devel, and only failing in staging + preprod.

If the test was broken why isn't autopush failing?

zloery:
> LGTM, but I see this is passing in autopush + devel, and only failing in staging + preprod.
> 
> If the test was broken why isn't autopush failing?

This bug only shows up for perf tests where we run the same test case multiple times in parallel, we don't run those against verily autopush/devel

PR #636, opened 2022-05-02
[PF-1585] Add workspaceByUserFacingId endpoint

melissachang:
Manually tested with swagger (normal case, workspace missing, workspace no permission)


openapi/src/common/parameters.yaml
```diff
@@ -96,3 +96,11 @@ components:
...
         type: string
         format: uuid
+
+    WorkspaceUserFacingId:
+      name: workspaceUserFacingId
+      in: path
+      description: Human-settable, mutable id.
+      required: true
+      schema:
+        type: string
```
jaycarlton, line 106: nit: missing newline at EOF. Really only matters for older POSIX tools.

openapi/src/parts/workspace.yaml
```diff
@@ -100,6 +100,27 @@ paths:
...
 
+  /api/workspaces/v1/workspaceByUserFacingId/{workspaceUserFacingId}:
+    parameters:
+      - $ref: '#/components/parameters/WorkspaceUserFacingId'
+    get:
+      summary: Get an existing Workspace by userFacingId.
+      operationId: getWorkspaceByUserFacingId
+      tags: [Workspace]
+      responses:
+        '200':
```
jaycarlton, line 111: Aside, and maybe a question for @ddietterich: should we draw a distinction between listing workspaces and reading them, or is it our policy that you need read access to know it exists? ISTR it being the latter, which is consistent with my understanding of this new endpoint.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -170,14 +170,14 @@ private ApiWorkspaceDescription buildWorkspaceDescription(Workspace workspace) {
     // When we have another cloud context, we will need to do a similar retrieval for it.
     return new ApiWorkspaceDescription()
         .id(workspace.getWorkspaceId())
+        .userFacingId(workspace.getUserFacingId())
```
jaycarlton, line 173: nit: maybe alphabetize these? There's no natural ordering.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -176,10 +176,10 @@ public Optional<Workspace> getWorkspaceIfExists(UUID id) {
   @ReadTransaction
   public Workspace getWorkspace(UUID id) {
     return getWorkspaceIfExists(id)
```
jaycarlton, line 178: nit: change `id` to `workspaceUuid` for consistency.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -269,6 +269,25 @@ public List<Workspace> getWorkspacesMatchingList(List<UUID> idList, int offset,
...
+    MapSqlParameterSource params = new MapSqlParameterSource().addValue("user_facing_id", userFacingId);
+    try {
+      UUID result =
+              DataAccessUtils.requiredSingleResult(
+                      jdbcTemplate.query(sql, params, UUID_ROW_MAPPER));
+      logger.info("Retrieved workspace uuid {}", result);
+      return result;
+    } catch (EmptyResultDataAccessException e) {
+      throw new WorkspaceNotFoundException(
+              String.format("Workspace %s not found.", userFacingId));
```
jaycarlton, line 287: nit: add `userFacingId` to the log messages, since some UFIDs can look like GUIDs.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -169,6 +169,13 @@ public Workspace getWorkspace(UUID id, AuthenticatedUserRequest userRequest) {
     return validateWorkspaceAndAction(userRequest, id, SamConstants.SamWorkspaceAction.READ);
   }
 
+  /** Retrieves an existing workspace by userFacingId */
+  @Traced
+  public Workspace getWorkspaceByUserFacingId(String userFacingId, AuthenticatedUserRequest userRequest) {
+    UUID uuid = workspaceDao.getUuidByUserFacingId(userFacingId);
```
jaycarlton, line 175: nit: name `workspaceUuid` for consistency.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -217,7 +262,7 @@ void duplicateWorkspaceUserFacingIdRequestsRejected() {
             DuplicateUserFacingIdException.class,
             () -> workspaceService.createWorkspace(duplicateUserFacingId, USER_REQUEST));
     assertEquals(
-        String.format("Workspace with user facing ID %s already exists", userFacingId),
+        String.format("Workspace with ID %s already exists", userFacingId),
```
jaycarlton, line 265: In the logs, I think it's important to distinguish these, since they won't always look different.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -307,19 +353,19 @@ void testUpdateWorkspace() {
...
     assertEquals(propertyMap2, updatedWorkspace.getProperties());
 
     // Sending through empty strings and an empty map clears the values.
     Map<String, String> propertyMap3 = new HashMap<>();
     Workspace thirdUpdatedWorkspace =
         workspaceService.updateWorkspace(
             USER_REQUEST, workspaceUuid, userFacingId, "", "", propertyMap3);
-    assertEquals("", thirdUpdatedWorkspace.getDisplayName().orElse(null));
-    assertEquals("", thirdUpdatedWorkspace.getDescription().orElse(null));
+    assertEquals("", thirdUpdatedWorkspace.getDisplayName().get());
```
jaycarlton, line 365: nit: never call `get()` without checking/asserting `isPresent()`. Even in test code, it makes me look up to see if you've checked it.

jaycarlton (approved):
Looks great!

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -269,6 +269,25 @@ public List<Workspace> getWorkspacesMatchingList(List<UUID> idList, int offset,
     return jdbcTemplate.query(sql, params, WORKSPACE_ROW_MAPPER);
   }
 
+  @ReadTransaction
+  public UUID getUuidByUserFacingId(String userFacingId) {
```
ddietterich, line 273: This could retrieve the workspace directly in one transaction instead of looping out and back with the UUID to look up the same row again.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -169,6 +169,13 @@ public Workspace getWorkspace(UUID id, AuthenticatedUserRequest userRequest) {
     return validateWorkspaceAndAction(userRequest, id, SamConstants.SamWorkspaceAction.READ);
   }
 
+  /** Retrieves an existing workspace by userFacingId */
+  @Traced
+  public Workspace getWorkspaceByUserFacingId(String userFacingId, AuthenticatedUserRequest userRequest) {
```
ddietterich, line 174: I would prefer to do this operation in one DB transaction instead of two.
You would put the Sam authz call here instead of the `validateWorkspaceAndAction`

If we get a second userFacingId-based API, we will probably want a `validateWorkspaceAndActionByUFID` method.

ddietterich (changes requested):
I would prefer a different structure. What do you think?

openapi/src/parts/workspace.yaml
cbookg: Regardless of this API, you will be able to determine this by trying to create a workspace with a given UFID to see if it is rejected.

cbookg: In fact, I was planning to use this API in the "Create Workspace" flow to determine if a provided UFID is taken.

So I was expecting this API to return the UUID, not the full object.  I think before it returns the full object, you would have to have access to use it?

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -176,10 +176,10 @@ public Optional<Workspace> getWorkspaceIfExists(UUID id) {
   @ReadTransaction
   public Workspace getWorkspace(UUID id) {
     return getWorkspaceIfExists(id)
-        .orElseThrow(
-            () ->
-                new WorkspaceNotFoundException(
-                    String.format("Workspace %s not found.", id.toString())));
+            .orElseThrow(
```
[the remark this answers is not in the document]
melissachang: If you click and drag, you'll see these are spaces. Intellij has a `Continuation indent` preference that's 8 by default:

image

4 does look better to me, than 8. I changed mine to 4, and I added a step to Terra Onboarding.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
melissachang: Done, makes sense

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
melissachang: Done

openapi/src/parts/workspace.yaml
melissachang: `getWorkspaceByUserFacingId` will be used for:

- UI: User loads workspace page. Url has userFacingId
- CLI ` terra workspace set --id=userFacingId`

`getWorkspaceByUserFacingId` will only return the workspace if Sam says you have READ access.

For UI determining "is UFID taken", I was thinking that would be a separate endpoint: PF-1605. Something like `/api/workspaces/v1/isUserFacingIdTaken/{workspaceUserFacingId}`. @cbookg do you want separate endpoint, or do you want to call `getWorkspaceByUserFacingId` and 403/404 means ufId is taken?

openapi/src/common/parameters.yaml
```diff
@@ -96,3 +96,11 @@ components:
       schema:
         type: string
         format: uuid
+
+    WorkspaceUserFacingId:
+      name: workspaceUserFacingId
+      in: path
+      description: Human-settable, mutable id.
```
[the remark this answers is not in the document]
ddietterich: I think this description gets shown to people using the swagger UI, so it is nice to have here for them. Not that we have great descriptions for other things! Just trying to get better incrementally.

openapi/src/parts/workspace.yaml
ddietterich: For Jay's question: we only list workspaces that the caller has some access to. We do this by first doing a Sam call to get the list of WORKSPACE objects the caller has access to and then filtering our DAO query with that list.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -559,6 +578,9 @@ public String getUserEmail() {
     }
   }
 
+  private static final RowMapper<UUID> UUID_ROW_MAPPER =
```
ddietterich, line 581: I don't think you need this anymore

ddietterich (approved):
LGTM - a couple nits

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
melissachang: Done

openapi/src/common/parameters.yaml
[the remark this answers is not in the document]
melissachang: Ah, I see. Added to all the relevant places, thanks.

melissachang (commented):
Will wait for Chris's review before merging

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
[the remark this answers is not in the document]
jaycarlton: If you use the Google Style theme instead of Default, you should get that and other settings for free. I didn't have to change my Continuation indent 
Screen Shot 2022-05-03 at 10 11 36 AM
to 4, for example.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
jaycarlton: Yes it will fail, but not with your assertion failure, so if you have a description in your assertion, you won't see it. One of these libraries has `assertContains(T value, Optional<T> o)` that lets you do it without the get.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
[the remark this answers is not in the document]
ddietterich: Here are the instructions from our DEVELOPER.md describing the standard formatting we are using for MC Terra.

### IntelliJ Setup

1. Open the repo normally (File -> Open)
2. In project structure (the folder icon with a little tetromino over it in the upper
   right corner), make sure the project SDK is set to Java 11. If not, IntelliJ should
   detect it on your system in the dropdown, otherwise click "Add JDK..." and navigate to
   the folder from the last step.
3. Set up [google-java-format](https://github.com/google/google-java-format). We use the
   spotless checker to force code to a standard format. Installing the IntelliJ plug-in
   and library makes it easier to get it in the right format from the start.
4. See some optional tips below in the ["Tips"](#tips) section.

openapi/src/parts/workspace.yaml
cbookg: > For UI determining "is UFID taken", I was thinking that would be a separate endpoint: PF-1605. Something like /api/workspaces/v1/isUserFacingIdTaken/{workspaceUserFacingId}. @cbookg do you want separate endpoint, or do you want to call getWorkspaceByUserFacingId and 403/404 means ufId is taken?

Either option is fine with me.  If it is more efficient for the backend to check existing of an ID than to populate the full get result, a distinct ID might be better.

ddietterich: I vote for fewer endpoints. The lookup is a single row in the workspace table and the Sam access check, so returning the full result is not a big cost.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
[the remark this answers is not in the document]
melissachang: I have `google-java-format` plugin enabled. Continuation indent was still 8. I'm guessing [default changed to 8 recently](https://intellij-support.jetbrains.com/hc/en-us/community/posts/4403453685778-Why-is-HTML-continuation-indent-default-8-spaces-), maybe that's why you didn't have to change it @ddietterich .

The [Google scheme](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml) does set continuation indent to 4. Should I add Google Schema to the `IntelliJ Setup` instructions? It's unfortunate that `google-java-format` doesn't take care of everything.

(Side note - There was one instance where even though I used `google-java-format` plugin, `spotlessCheck` still failed. That's why I changed Intellij to run `./gradlew spotlessApply`. I'll let you know next time that happens.)

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
melissachang: Ah, it makes sense to assert `isPresent()` to get a friendlier error message. Done.

openapi/src/parts/workspace.yaml
melissachang: SG. Closed PF-1605 and updated PF-1606.

PR #638, opened 2022-05-04
[PF-1616] Have CloneWorkspace work with userFacingId

melissachang:
Before this PR:
- CloneWorkspaceRequest doesn't have `userFacingId` field
- In db, destination workspace dosen't have userFacingId set

After:
- CloneWorkspaceRequest has `userFacingId` field
- In db, destination workspace has userFacingId set

Before we validated (`ControllerValidationUtils.validateUserFacingId`) `userFacingId` in:

- controller `createWorkspace` `updateWorkspace`

In this PR, I added validation to:

- controller `cloneWorkspace`
- dao `createWorkspace` `updateWorkspace`

I added validation to dao to be extra-safe -- for example, what if there's some new endpoint in the future that calls `dao.createWorkspace`.

Dao validation meant I had to change a bunch of unit/connected tests to set `userFacingId`. (In normal application, this is set in controller, which unit/connected tests don't call.)

FYI @cbookg

### DIFF
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -256,7 +256,7 @@ components:
         destinationUserFacingId:
           description: |
             Human-settable, mutable id. Must have 3-63 characters, contain lowercase letters, numbers, dashes, or
-            underscores, and start with lowercase letter. Required.
+            underscores, and start with lowercase letter.
           type: string
         resources:
           type: array
@@ -280,8 +280,9 @@ components:
       properties:
         userFacingId:
           description: |
-            Human-settable, mutable id. ID must have 3-63 characters, contain lowercase letters, numbers, dashes, or
-            underscores, and start with lowercase letter. Optional. If this isn't set, one will be generated.
+            Human-settable, mutable id for cloned workspace. ID must have 3-63 characters, contain lowercase letters,
+            numbers, dashes, or underscores, and start with lowercase letter. Optional. If this isn't set, one will be
+            generated.
           type: string
         displayName:
           description: The human readable name of the workspace
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -273,6 +273,12 @@ components:
       required:
         - spendProfile
       properties:
+        userFacingId:
+          description: |
+            Human-settable, mutable id. ID must have 3-63 characters, contain lowercase letters, numbers, dashes, or
+            underscores, and start with lowercase letter. Optional. If this isn't set, one will be generated based on id
+            (guid).
+          type: string
         displayName:
           description: The human readable name of the workspace
           type: string
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -253,6 +253,11 @@ components:
         destinationWorkspaceId:
           type: string
           format: uuid
+        destinationUserFacingId:
+          description: |
+            Human-settable, mutable id. Must have 3-63 characters, contain lowercase letters, numbers, dashes, or
+            underscores, and start with lowercase letter. Required.
+          type: string
         resources:
           type: array
           items:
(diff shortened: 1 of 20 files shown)


openapi/src/parts/workspace.yaml
```diff
@@ -253,6 +253,11 @@ components:
         destinationWorkspaceId:
           type: string
           format: uuid
+        destinationUserFacingId:
+          description: |
+            Human-settable, mutable id. Must have 3-63 characters, contain lowercase letters, numbers, dashes, or
+            underscores, and start with lowercase letter. Required.
```
jaycarlton, line 259: This is an output property so you don't need to mark it required.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -396,10 +396,18 @@ public ResponseEntity<ApiCloneWorkspaceResult> cloneWorkspace(
     Optional<SpendProfileId> spendProfileId =
         Optional.ofNullable(body.getSpendProfile()).map(SpendProfileId::new);
     final UUID destinationWorkspaceId = UUID.randomUUID();
+
+    // ET uses userFacingId; CWB doesn't. Schema enforces that userFacingId must be set. CWB doesn't
+    // pass userFacingId in request, so use id. Prefix with "a" because userFacingId must start with
+    // letter.
+    String destinationUserFacingId = Optional.ofNullable(body.getUserFacingId()).orElse("a" + destinationWorkspaceId);
```
jaycarlton, line 403: nit: a hyphen after the 'a' would make it more obvious that it's a name and not a guid in log messages and other contexts where it's maybe confusable.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -84,6 +85,7 @@ public UUID createWorkspace(Workspace workspace) {
             + " cast(:properties AS jsonb), :workspace_stage)";
 
     final String workspaceUuid = workspace.getWorkspaceId().toString();
+    ControllerValidationUtils.validateUserFacingId(workspace.getUserFacingId());
```
jaycarlton, line 88: I think this makes sense, but I don't know if we intended this method to be called from below the controller layer. @zloery do you think it matters?

ddietterich (approved):
LGTM

openapi/src/parts/workspace.yaml
melissachang: Done

openapi/src/parts/workspace.yaml
```diff
@@ -273,6 +278,11 @@ components:
       required:
         - spendProfile
       properties:
+        userFacingId:
+          description: |
```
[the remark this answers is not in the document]
melissachang: Done

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
melissachang: Done. Updated clone and create.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
melissachang: I checked with @ddietterich, he's ok with this. I'm calling this in relevant controller places. I also call in Dao in case there's some controller place I missed, or some new controller in the future. If my first PR called in dao, this bug (PF-1616) wouldn't have happened.

Added comment since it's strange to see `ControllerValidationUtils` here, thanks.

zloery: strictly speaking if we checked every endpoint where a user-facing ID could be generated/passed in we shouldn't need this, but it's a static validation method so it doesn't really cost much. 

+1 we should be calling it from all controller endpoints where user facing IDs could come in/be generated/be modified, but also having it in the DAO with a comment sgtm

PR #639, opened 2022-05-04
IA-3120 fix createAzureVm undo step

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3120

When we delete VM, we also need to delete OS disk and nic automatically

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
@@ -40,6 +40,7 @@ public class CreateAzureVmStep implements Step {
   private final CrlService crlService;
   private final ControlledAzureVmResource resource;
   private final ResourceDao resourceDao;
+  private final String NIC_NAME;
 
   public CreateAzureVmStep(
       AzureConfiguration azureConfig,
@@ -50,6 +51,7 @@ public class CreateAzureVmStep implements Step {
     this.crlService = crlService;
     this.resource = resource;
     this.resourceDao = resourceDao;
+    this.NIC_NAME = String.format("nic-%s", resource.getVmName());
   }
 
   @Override
@@ -106,7 +108,7 @@ public class CreateAzureVmStep implements Step {
           computeManager
               .networkManager()
               .networkInterfaces()
-              .define(String.format("nic-%s", resource.getVmName()))
+              .define(NIC_NAME)
               .withRegion(resource.getRegion())
               .withExistingResourceGroup(azureCloudContext.getAzureResourceGroupId())
               .withExistingPrimaryNetwork(existingNetwork)
@@ -187,9 +189,7 @@ public class CreateAzureVmStep implements Step {
               .virtualMachines()
               .deleteByResourceGroup(azureCloudContext.getAzureResourceGroupId(), resource.getVmName());
 
-      resolvedVm
-              .networkInterfaceIds()
-              .forEach(nic -> computeManager.networkManager().networkInterfaces().deleteById(nic));
+      computeManager.networkManager().networkInterfaces().deleteByResourceGroup(azureCloudContext.getAzureResourceGroupId(), NIC_NAME);
 
       // Delete the OS disk
       computeManager.disks().deleteById(resolvedVm.osDiskId());
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
@@ -177,9 +177,22 @@ public class CreateAzureVmStep implements Step {
     ComputeManager computeManager = crlService.getComputeManager(azureCloudContext, azureConfig);
 
     try {
+      VirtualMachine resolvedVm =
+              computeManager
+                      .virtualMachines()
+                      .getByResourceGroup(
+                              azureCloudContext.getAzureResourceGroupId(), resource.getVmName());
+
       computeManager
-          .virtualMachines()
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
```diff
@@ -177,9 +177,22 @@ public StepResult undoStep(FlightContext context) throws InterruptedException {
     ComputeManager computeManager = crlService.getComputeManager(azureCloudContext, azureConfig);
 
     try {
+      VirtualMachine resolvedVm =
+              computeManager
+                      .virtualMachines()
+                      .getByResourceGroup(
+                              azureCloudContext.getAzureResourceGroupId(), resource.getVmName());
+
       computeManager
```
jaycarlton, line 186: Is this delete idempotent? We don't want any exceptions coming from undo if it's already deleted.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
```diff
@@ -177,9 +177,22 @@ public StepResult undoStep(FlightContext context) throws InterruptedException {
...
-          .deleteByResourceGroup(azureCloudContext.getAzureResourceGroupId(), resource.getVmName());
+              .virtualMachines()
+              .deleteByResourceGroup(azureCloudContext.getAzureResourceGroupId(), resource.getVmName());
+
+      resolvedVm
+              .networkInterfaceIds()
+              .forEach(nic -> computeManager.networkManager().networkInterfaces().deleteById(nic));
+
+      // Delete the OS disk
+      computeManager.disks().deleteById(resolvedVm.osDiskId());
```
jaycarlton, line 195: I wonder if you'll need to keep coming back here to add things to be deleted. That would be easy to miss, so I wonder if you wouldn't be better served by a central `deleteVmAndEverythingInIt()` method, especially if you need to do this in another place, like a delete endpoint.

jaycarlton (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
```diff
@@ -177,9 +177,22 @@ public StepResult undoStep(FlightContext context) throws InterruptedException {
...
+                      .virtualMachines()
+                      .getByResourceGroup(
+                              azureCloudContext.getAzureResourceGroupId(), resource.getVmName());
+
       computeManager
-          .virtualMachines()
-          .deleteByResourceGroup(azureCloudContext.getAzureResourceGroupId(), resource.getVmName());
+              .virtualMachines()
+              .deleteByResourceGroup(azureCloudContext.getAzureResourceGroupId(), resource.getVmName());
+
```
ddietterich, line 189: I am not sure I understand the dependencies, but I think this undo code is not idempotent. If the pod crashed (or was stopped by K8s) at this point, then a rerun of this undo would not be able to locate the nics or the disks. The VM would be deleted so the retrieval at line 180 would not find it.

The fix would be to find the objects by name, the same way you find them in the `do()` code.

ddietterich (changes requested):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
Qi77Qi: where would you but this central method

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
Qi77Qi: good point..I can fix it for nic, but OS disk doesn't have a known name, so Idk if we can fix this completely

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
jaycarlton: Wherever your delete endpoint is, or a utility class. If you only ever need it here, then there's no need to break it out.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
ddietterich: Probably my misunderstanding about the disk. I see at line 85, a disk is retrieved by name. Is that separate from the OS disk? Or is that the disk you are supposed to delete?

If the OS disk is automatically created by Azure, but not automatically deleted by Azure, I don't see a way to fix the idempotency.

Qi77Qi: >Probably my misunderstanding about the disk. I see at line 85, a disk is retrieved by name. Is that separate from the OS disk? Or is that the disk you are supposed to delete?

yea OS disk and the user disk are 2 disks. We define user disk but not OS disk.

>If the OS disk is automatically created by Azure, but not automatically deleted by Azure, I don't see a way to fix the idempotency.
yes.

ddietterich: OK, then would you document the problem case (orphan NIC and disk in a small failure window) in the undo code for our future selves? I'll approve the PR now.

ddietterich (approved):

Qi77Qi:
@ddietterich  nic should be good now...OS disk is the only thing, I have a TODO for now...will look at this a bit more once I get the main ticket I'm working on in a good state

PR #640, opened 2022-05-05
[PF-1622] In startup script, authenticate before setting workspace

melissachang:
`--defer-login` is going away, so authenticate before calling `terra workspace set`. See PF-1622 for why `--defer-login` is going away.


yuhuyoyo (approved):

ddietterich (approved):
lgtm

PR #641, opened 2022-05-06
TOAZ-128 Create VM endpoint should handle version parameter for the VM image

sergiygetlin:
Extend VM image with version parameter.


Qi77Qi (approved):
lgtm..will test this out once it's in dev

ddietterich (approved):
LGTM

PR #643, opened 2022-05-06
IA-3120 IA-3390 Take an optional resource ID

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3120

### DIFF
--- a/openapi/src/common/schemas.yaml
+++ b/openapi/src/common/schemas.yaml
@@ -70,6 +70,9 @@ components:
           $ref: '#/components/schemas/ManagedBy'
         privateResourceUser:
           $ref: '#/components/schemas/PrivateResourceUser'
+        resourceId:
+          type: string
+          format: uuid
   
     DeleteControlledAzureResourceRequest:
       type: object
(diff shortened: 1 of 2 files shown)


ddietterich (approved):
That was easier than I expected. Nice!

Qi77Qi:
@ddietterich do we need to do any duplicate check? Or that'll be handled properly already?

ddietterich:
> do we need to do any duplicate check? Or that'll be handled properly already?

Good point. The create will fail because the UUID is a primary key, but the error may not be pretty - probably a 500 as opposed to a 400 - so maybe check the DAO to see what it throws for a PK collision on the resource table.

You should also add a test to cause duplicate UUIDs and verify they are being properly handled.

rtitle (approved):
Looks good to me

Qi77Qi:
@ddietterich  gotcha..I'm gonna add a TODO for now since it will fail anyway. Really trying to get sth else working

ddietterich:
Please add a ticket to the TODO and we are good.

jaycarlton (approved):
Looks like @ddietterich's refactoring crusades are really paying off. This should have been harder.

PR #644, opened 2022-05-07
[PF-1538] jdk17

ddietterich:
This work moves WSM to JDK17. The client generates Java 11 code, so it will work with components still using JDK11.

With this merge, we will be able to use Java 17 features in the **service** component. Integration tests and the client will remain at JDK11 until all of the services have moved to JDK17.

It looks like this PR caused some reformatting. Somehow code is going in that is not following the spotless rules.


gradle.properties
```diff
@@ -1,2 +1,9 @@
 org.gradle.caching=true
 org.gradle.parallel=true
+# needed for Spotless to be able to run google-java-format
```
Qi77Qi, line 3: good to know...maybe this is what I need in listener repo

Qi77Qi (approved):

jaycarlton:
Maybe spotless rules are different for Java 17? Make sure you run spotlessApply by doing a full `./gradlew build -x test`. I think I broke standalone `spotlessApply`. There's a ticket PF-1559 that is similar.

jaycarlton:
see https://github.com/DataBiosphere/terra-workspace-manager/pull/647 for a stand-alone format fix changeset

client/gradle/openapi.gradle
```diff
@@ -6,6 +6,7 @@ swaggerSources {
   client {
     inputFile = file(tasks.getByPath(':openapi:buildOpenApi').openApiSource)
     code {
+      jvmArgs = ['--add-opens=java.base/java.util=ALL-UNNAMED'] // for Java 16+
```
jaycarlton, line 9: Is this for module system compatibility?

integration/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -226,6 +229,10 @@ private static ApiClient buildClient(
       apiClient.setAccessToken(accessToken.getTokenValue());
     }
 
+    ClientConfig clientConfig = new ClientConfig();
```
jaycarlton, line 232: Can you comment what all this is for? Why do we only need to do it in the test config? I'd have thought you'd want the same config for test and production.

jaycarlton (approved):
This is exciting. Someone should do a chalk talk on the new features in Java 12-17?

.github/workflows/nightly-tests.yml
```diff
@@ -52,10 +52,11 @@ jobs:
             sudo apt-get update
             sudo apt-get -y install postgresql
 
-      - name: Set up AdoptOpenJDK 11
-        uses: joschi/setup-jdk@v2
+      - name: Set up JDK
+        uses: actions/setup-java@v2
```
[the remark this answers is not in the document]
ddietterich: There wasn't a built-in java set up before.

DEVELOPMENT.md
```diff
@@ -64,29 +64,49 @@ useful menulet where the server can be turned on and off. Don't forget to create
...
+```sh
+    # In getting set up initially, I made some mess, but I think this will work.
+    brew install temurin17
+    jenv add /Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home
+ ```
 
-    # follow instructions from https://github.com/AdoptOpenJDK/homebrew-openjdk to install adoptopenjdk11:
-    brew tap AdoptOpenJDK/openjdk
-    brew install --cask adoptopenjdk11
+You may want to use jenv to set this JDK as your global setting to avoid surprises.
```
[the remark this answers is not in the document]
ddietterich: You don't need to, no. The JDK17 will run the Java 11 code just fine. It is the other direction that doesn't work.

client/gradle/openapi.gradle
ddietterich: This is part of the workaround because the codegen doesn't follow the Java17 access rules.

zloery (approved):

integration/src/main/java/scripts/utils/ClientTestUtils.java
ddietterich: Good question. Let me check. I think this snippet may need to go into a custom modification of the generated code. Otherwise, every client would have to do this I think.

PR #645, opened 2022-05-08
[PF-1238] Remove deprecated GCP context code

ddietterich:
When I implemented PF-1207, I did a careful upgrade of GCP context that defended against any flights running during the upgrade. It is now safe to remove the old code.

When testing, I noticed I had missed fixing the list of retry steps to debug. Fixing the test showed up a bug with retrying the `CreateDbGcpCloudContextStep` so that is also fixed.


jaycarlton:
I confirmed with Greg that Docker desktop licenses are still being worked, so stay tuned.

PR #647, opened 2022-05-09
[no ticket] spotless

jaycarlton:
ran spotless with `./gradlew build -x test`.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
@@ -175,13 +175,14 @@ public class ControlledAzureResourceApiController extends ControlledResourceCont
   }
 
   @Override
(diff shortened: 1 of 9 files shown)


Qi77Qi:
think this will cause conflicts with https://github.com/DataBiosphere/terra-workspace-manager/pull/644....

jaycarlton:
That's tragic, but them's the rules.

jaycarlton:
We also have a `spotbugs` violation at HEAD.

It looks like `lint-and-static-analysis` not required anymore. That's probably the culprit.

Qi77Qi (approved):
yea, I think fixing linting will be helpful....I'll leave it up to you and Dan about the actual formatting fix (either this PR or the  other one)

jaycarlton:
Huh. It found the spotbugs violation but didn't fail the workflow.

ddietterich (approved):
This is fine, but it doesn't address whatever the root cause is that allows this to make it into the repository. Do you understand what happened?

ddietterich:
> Where do you set which workflows are required to pass? In the repo config?
Yes. I just put linting-and-static-analysis back on the list. Not sure how that got lost.

PR #648, opened 2022-05-10
TOAZ-79 [WSM] Create VM that has no public IP

sergiygetlin:
Make public ip optional for VM

https://broadworkbench.atlassian.net/browse/TOAZ-79

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
@@ -30,6 +30,7 @@ import com.azure.resourcemanager.compute.models.VirtualMachineSizeTypes;
 import com.azure.resourcemanager.network.models.Network;
 import com.azure.resourcemanager.network.models.NetworkInterface;
 import com.azure.resourcemanager.network.models.PublicIpAddress;
+import java.util.Optional;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
```diff
@@ -104,19 +110,12 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
-              .withRegion(resource.getRegion())
-              .withExistingResourceGroup(azureCloudContext.getAzureResourceGroupId())
-              .withExistingPrimaryNetwork(existingNetwork)
-              .withSubnet(networkResource.getSubnetName())
-              .withPrimaryPrivateIPAddressDynamic()
-              .withExistingPrimaryPublicIPAddress(
-                  existingAzureIp) // TODO this needs to be updated to support not exposing public
-              // IP
-              .create();
+          createNetworkInterface(
```
ddietterich, line 113: I think this step has idempotency problems. If this step is retried or there is a pod failover, then the newly created NIC will be orphaned, right?

I think you need a sequence something like this:
- step 1: Compute and store the NIC name in the working map (Or indicate "use existing ip")
- step 2:
  - do: if NIC exists, return success. Otherwise, create the NIC. 
  - undo: if NIC exists, delete it. Return success
- step 3: the CreateAzureVmStep
  - do: compute the inputs (nothing is created), then create the VM
  - undo: delete the VM if it exists

You could probably squeeze step 2 into step 3, but I think more, simpler steps make reasoning about failure and idempotency easier.

ddietterich (changes requested):
I think the step is no longer idempotent and can leak ips.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
ddietterich: After a side conversation with @sergiygetlin, I understand that this is the same issue that is already captured in the ticket TOAZ-114. I would like that work prioritized so we are not accumulating tech debt in the WSM code.

ddietterich (approved):

Qi77Qi (approved):
lgtm

PR #651, opened 2022-05-10
[WIP][WOR-228] Delete Azure resources when deleting workspace or cloud context

marctalbott:
Ticket: [WOR-228](https://broadworkbench.atlassian.net/browse/WOR-228)
Tests coming


service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledAzureResourcesStep.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+
+    List<ControlledResource> controlledResourceList =
+        resourceDao.listControlledResources(workspaceUuid, CloudPlatform.AZURE);
+
+    /**
+     * TODO: does this need to be done in any particular order? Most resources are likely fine, but
```
aherbst-broad, line 49: I think we'll need to explicitly delete the containers (first), to make sure the related SAM resources and DB entries are also cleaned up.

aherbst-broad, line 49: I think we'll need to explicitly delete the containers (first), to make sure the related SAM resources and DB entries are also cleaned up.

marctalbott: I forgot that storage containers aren't a WSM resource type yet. I added a note to https://broadworkbench.atlassian.net/browse/WOR-92 and updated the TODO here to reference that ticket

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteAzureContextFlightTest.java
```diff
@@ -0,0 +1,354 @@
...
+    assertTrue(testUtils.getAuthorizedAzureCloudContext(workspaceUuid, userRequest).isEmpty());
+    assertThrows(
+        ResourceNotFoundException.class,
+        () -> controlledResourceService.getControlledResource(workspaceUuid, relayId, userRequest));
+  }
+
+  // todo: this test would be better in the WorkspaceDeleteFlightTest, but that class extends
+  // BaseConnectedTest which doesn't have azure enabled... figure out what to do about this test
+  @Test
+  void deleteMcWorkspaceWithAzureContextAndResource() throws Exception {
```
marctalbott, line 262: Looking for input on where to put this test. 

I've thought about moving it back to `WorkspaceDeleteFlightTest` and changing that test class to extend `BaseAzureTest` instead, but I'm worried that will change too much about when and where workspace deletion tests are run in our CI pipeline. 

Also thought about making a separate `AzureWorkspaceDeleteFlightTest`, but there isn't an actual `AzureWorkspaceDeleteFlight` so that's not great either. 

I suppose we could also just leave it here for now until `BaseConnectedTest` and `BaseAzureTest` converge, assuming that that will one day happen?

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledAzureResourcesStep.java
```diff
@@ -0,0 +1,67 @@
...
+
+    List<ControlledResource> controlledResourceList =
+        resourceDao.listControlledResources(workspaceUuid, CloudPlatform.AZURE);
+
+    /**
+     * TODO: https://broadworkbench.atlassian.net/browse/WOR-92 delete Azure storage containers
+     * first before deleting storage account to ensure Sam and WSM DBs are cleaned up
+     */
+    for (ControlledResource resource : controlledResourceList) {
+      controlledResourceService.deleteControlledResourceSync(
```
ddietterich, line 53: This will throw an exception if the delete fails. That will cause this flight to fail. Is that the behavior you want?
That failure will not be retried. Stairway will undo and fail the flight.

An alternative would be to catch and track the failures, but continue deleting as many resources as you can. If the failures are retry-able, then you could return Retry. That would maybe succeed in the face of transient errors.

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledAzureResourcesStep.java
```diff
@@ -0,0 +1,67 @@
...
+    this.workspaceUuid = workspaceUuid;
+    this.userRequest = userRequest;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+
+    List<ControlledResource> controlledResourceList =
+        resourceDao.listControlledResources(workspaceUuid, CloudPlatform.AZURE);
```
ddietterich, line 46: Are there any dependencies between resources that need to be considered?

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteAzureContextFlightTest.java
ddietterich: Good question! I'll be interested in what other people think. My take is to leave it here for now.

We should have a larger conversation about how we want to structure WSM tests. Some of the factors I am aware of are the conflating of tags with the base classes; what tests should be integration tests vs connected tests; and what the needs are of the Verily deployment vs the Broad deployment. There are probably more things to think about.

ddietterich (commented):
This is looking good. I have a question about error handling I'd like to explore before Approve.

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteAzureContextFlightTest.java
aherbst-broad: I think leaving it here is fine for now. It feels like a better fit for the `WorkspaceDeleteFlightTest` in the long run, where maybe we conditionally run it in the broad deployment, but leave it off in the verily?

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteAzureContextFlightTest.java
```diff
@@ -0,0 +1,356 @@
...
+
+    AzureCloudContext azureCloudContext =
+        workspaceService.getAuthorizedAzureCloudContext(workspaceUuid, userRequest).orElse(null);
+    assertNotNull(azureCloudContext);
+
+    // Create Azure relay namespace
+    UUID relayId = UUID.randomUUID();
+    ApiAzureRelayNamespaceCreationParameters creationParameters =
+        ControlledResourceFixtures.getAzureRelayNamespaceCreationParameters();
+    ControlledAzureRelayNamespaceResource relayNamespaceResource =
```
aherbst-broad, line 209: 🤔 wondering if there is a less time consuming resource we can spin up here, given our knowledge that this burns ~40. seconds for relay namespace creation.

aherbst-broad (approved):
lgtm

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledAzureResourcesStep.java
marctalbott: The only dependency I'm aware of is a Storage container on a Storage account, but I'll look more closely at the existing Azure resources to be more certain

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteAzureContextFlightTest.java
marctalbott: Good call, I started with a relay namespace since that's what we're creating in Rawls at the moment, but you're right that we shouldn't wait for 40 seconds here if we don't have to.

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledAzureResourcesStep.java
ddietterich: Hmm... I thought the disk delete would fail if it were mounted on a running VM, but you would know better than me.

marctalbott: Yeah that's a great point. Also not sure what happens if we delete a VM out from under an Azure IP for example. I misspoke on my first comment -- the only dependency I was aware of \*off the top of my head\* was the container -> account dependency. There are other dependencies that will need to be considered

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteAzureContextFlightTest.java
```diff
@@ -0,0 +1,356 @@
...
+    workspaceService.deleteWorkspace(workspaceUuid, userAccessUtils.defaultUserAuthRequest());
+  }
+
+  @Test
+  void deleteContextDo() throws Exception {
+    AuthenticatedUserRequest userRequest = userAccessUtils.defaultUserAuthRequest();
+    FlightMap createParameters =
+        azureTestUtils.createAzureContextInputParameters(workspaceUuid, userRequest);
+
+    // Create the azure context.
```
cahrens, line 89: Nit: the different test cases have blocks of common setup that could be put into helper methods (creating context, creating Azure IP resource).

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteAzureContextFlightTest.java
```diff
@@ -0,0 +1,356 @@
...
+        StairwayTestUtils.blockUntilFlightCompletes(
+            jobService.getStairway(),
+            CreateControlledResourceFlight.class,
+            azureTestUtils.createControlledResourceInputParameters(
+                mcWorkspaceUuid, userRequest, ipResource),
+            CREATION_FLIGHT_TIMEOUT,
+            null);
+    assertEquals(FlightStatus.SUCCESS, flightState.getFlightStatus());
+
+    // Run the delete flight, retrying every step once
```
cahrens, line 322: Not ever step is retried-- `DeleteControlledAzureResourcesStep.class.getName()` specifically is not.

cahrens (approved):

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteAzureContextFlightTest.java
cahrens: Thanks for doing this cleanup!

service/src/main/java/bio/terra/workspace/service/workspace/flight/DeleteControlledAzureResourcesStep.java
marctalbott: We opted to remove the retry rule here so that if deleting a resource fails the flight will fail. The exception handling when deleting Azure resources just retries all failures at the moment. We are hesitant to do too much retrying here as a result since we expect that simple errors like a resource not existing will lead to it retrying unnecessarily. We've filed [this ticket](https://broadworkbench.atlassian.net/browse/WOR-327) to improve the Azure resource deletion error handling, but we believe this is sufficient for now

PR #652, opened 2022-05-11
[PF-1636] validate cloning instructions on update paths

jaycarlton:
I had a leak in my validation that let non-applicable states into the DB. This poisons the whole row, as we re-validate on the way out of the DB.

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
@@ -71,7 +71,6 @@ import bio.terra.workspace.service.resource.controlled.model.ControlledResource;
 import bio.terra.workspace.service.resource.controlled.model.ControlledResourceFields;
 import bio.terra.workspace.service.resource.exception.DuplicateResourceException;
 import bio.terra.workspace.service.resource.exception.ResourceNotFoundException;
-import bio.terra.workspace.service.resource.model.CloningInstructions;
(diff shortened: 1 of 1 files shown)


service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -601,6 +611,8 @@ void createGetUpdateDeleteBqDataset() throws Exception {
     features.setAlpha1Enabled(true);
     StairwayTestUtils.enumerateJobsDump(
         alpha1Service, workspace.getWorkspaceId(), user.getAuthenticatedRequest());
+
+
```
yuhuyoyo, line 615: nit: remove extra lines

yuhuyoyo (approved):

zloery (approved):
lgtm, are there any poisoned rows in live DBs we need to clean up?

ddietterich:
Wait, what? Why are we merging this before fixing up the databases?

jaycarlton:
> Wait, what? Why are we merging this before fixing up the databases?

The problem comes when pulling things out of the DB that got in earlier, since we validate on build. This fix only prevents the DB from getting worse.

ddietterich:
I understand the situation to be that when users touch "poisoned" resources, they will get failures. Right?
How do we make sure that does not happen?

jaycarlton:
> I understand the situation to be that when users touch "poisoned" resources, they will get failures. Right? How do we make sure that does not happen?

We can turn off the enforcement in `validate()` until we clean up the databases, then turn it back on.

PR #654, opened 2022-05-13
[PF-1641] Bump the test time for EnablePet

ddietterich:
Increase the run time for EnablePet to allow for slow permission propagation.

### DIFF
--- a/integration/src/main/resources/configs/integration/EnablePet.json
+++ b/integration/src/main/resources/configs/integration/EnablePet.json
@@ -9,7 +9,7 @@
       "name": "EnablePet",
       "numberOfUserJourneyThreadsToRun": 1,
       "userJourneyThreadPoolSize": 1,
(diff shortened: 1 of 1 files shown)


integration/src/main/resources/configs/integration/EnablePet.json
```diff
@@ -9,7 +9,7 @@
       "name": "EnablePet",
       "numberOfUserJourneyThreadsToRun": 1,
       "userJourneyThreadPoolSize": 1,
-      "expectedTimeForEach": 3,
+      "expectedTimeForEach": 12,
```
yuhuyoyo, line 12: hmm seems like this doesn't help as we just run out of retries.

ddietterich:
Yes, this doesn’t fix anything. Zach will provide the right fix. Closing the PR.

PR #656, opened 2022-05-16
[PF-1583] Make workspace.user_facing_id required in DB schema

melissachang:
This PR depends on #638. Before #638, when you cloned a workspace, `user_facing_id` was empty. After #638, `user_facing_id` is set.

- #638 was pushed to Broad production on 5/12. #638 was pushed to Verily production on 5/12. It's been two business days; I assume #638 won't be rolled back.
- For both Broad (PF-1582) and Verily (PF-1580), I confirmed that `user_facing_id` is set for all envs. For Verily, I had to manually backfill some cloned workspaces that were missing `user_facing_id`; see PF-1580.


DEVELOPMENT.md
```diff
@@ -91,13 +91,18 @@ You can install the JDK from IntelliJ::
...
+
+    # If ~/.jenv/versions directory doesn't exist:
+    mkdir -p ~/.jenv/versions
     jenv add /Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home
  ```
 
 You may want to use jenv to set this JDK as your global setting to avoid surprises.
 
+To switch between jdk versions, modify `.bash_profile` and run `jenv global` as [described here](https://stackoverflow.com/a/43900696/6447189).
+
```
jaycarlton, line 105: Mention `jenv shell`. I have to use this currently to build the CLI in Java 11.

jaycarlton (approved):

DEVELOPMENT.md
melissachang: To switch back and forth for CLI and WSM, I do `jenv global 17` and `jenv global 11`. What commands do you run?

jaycarlton: that works. I do `jenv shell openjdk64-11.0.9.1`, so it only applies to the current shell.

ddietterich: Maybe point to jenv help instead of documenting here.

ddietterich (approved):
The code change looks fine. I have a suggestion on the docs, but it is fine with me as is.

DEVELOPMENT.md
melissachang: Done. I also changed `jenv global` to generic `jenv`, so people can use whichever variant that want.

PR #657, opened 2022-05-16
[PF-1642] Increase timeout for WorkspaceLifecycle integration test

melissachang:
I'm confused at how this test passed. It takes more than 5 seconds to create a workspace...

### DIFF
--- a/integration/src/main/resources/configs/integration/WorkspaceLifecycle.json
+++ b/integration/src/main/resources/configs/integration/WorkspaceLifecycle.json
@@ -9,8 +9,8 @@
       "name": "WorkspaceLifecycle",
       "numberOfUserJourneyThreadsToRun": 1,
       "userJourneyThreadPoolSize": 1,
-      "expectedTimeForEach": 1,
-      "expectedTimeForEachUnit": "MINUTES"
+      "expectedTimeForEach": 10,
+      "expectedTimeForEachUnit": "SECONDS"
     }
   ],
   "testUserFiles": ["liam.json"]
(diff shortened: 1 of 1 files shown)


ddietterich (approved):
Good catch.

zloery (approved):
From looking at previous test runs, it looks like this actually does run in ~3 seconds. The test makes 3 sync calls to WSM (create, patch, delete) and the workspace it creates doesn't have any cloud context attached (which is normally where most of the time is spent). But 1m is probably a better limit, I don't know why the old one was so low

melissachang:
> From looking at previous test runs, it looks like this actually does run in ~3 seconds. The test makes 3 sync calls to WSM (create, patch, delete) and the workspace it creates doesn't have any cloud context attached (which is normally where most of the time is spent). But 1m is probably a better limit, I don't know why the old one was so low

Ahh thanks. There's no spend profile in the request, so cloud context isn't created.

Now that I understand there's no cloud context created, I changed to 10 seconds, that should be enough. Thanks.

zloery:
> Ahh thanks. There's no spend profile in the request, so cloud context isn't created.
> 
> Now that I understand there's no cloud context created, I changed to 10 seconds, that should be enough. Thanks.

It's not just the spend profile, at the WSM level creating a cloud context is actually a separate (and async) API call (`POST /api/workspaces/v1/{workspaceId}/cloudcontexts`), we just hide that at the UI/CLI layer.

10 seconds sgtm, thanks for this!

melissachang:
> It's not just the spend profile, at the WSM level creating a cloud context is actually a separate (and async) API call (`POST /api/workspaces/v1/{workspaceId}/cloudcontexts`), we just hide that at the UI/CLI layer.

I did not know that. TIL. Thanks!

PR #658, opened 2022-05-16
[PF-1643] Fix CreateWorkspaceAuthzStep flakiness

melissachang:
See [PF-1643](https://broadworkbench.atlassian.net/browse/PF-1643) for analysis. Test failed due to `CreateWorkspaceAuthzStep` being retried. However, `CreateWorkspaceAuthzStep` already checks for retries:

```
38    // Even though WSM should own this resource, Stairway steps can run multiple times, so it's
39    // possible this step already created the resource. If WSM can either read the existing Sam
40    // resource or create a new one, this is considered successful.
41    if (!canReadExistingWorkspace(workspace.getWorkspaceId())) {
42      samService.createWorkspaceWithDefaults(userRequest, workspace.getWorkspaceId());
43    }
```

My theory is that there was a race condition. SAM resource was created **between** line 41 and `samService.createWorkspaceWithDefaults` SAM-checking-if-resource-exists.

I restructured things to avoid this race condition. Rather than call `canReadExistingWorkspace` **and** catch `SamConflictException`, I just did the latter, to make code simpler.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceAuthzStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceAuthzStep.java
@@ -1,5 +1,6 @@
 package bio.terra.workspace.service.workspace.flight;
 
+import bio.terra.common.sam.exception.SamConflictException;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
@@ -35,11 +36,11 @@ public class CreateWorkspaceAuthzStep implements Step {
   public StepResult doStep(FlightContext flightContext)
       throws RetryException, InterruptedException {
 
-    // Even though WSM should own this resource, Stairway steps can run multiple times, so it's
-    // possible this step already created the resource. If WSM can either read the existing Sam
-    // resource or create a new one, this is considered successful.
-    if (!canReadExistingWorkspace(workspace.getWorkspaceId())) {
+    try {
       samService.createWorkspaceWithDefaults(userRequest, workspace.getWorkspaceId());
+    } catch (SamConflictException e) {
+      // Stairway steps can run multiple times. This step must have run before.
+      return StepResult.getStepResultSuccess();
     }
     return StepResult.getStepResultSuccess();
   }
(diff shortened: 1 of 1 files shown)


service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceAuthzStep.java
```diff
@@ -35,11 +36,11 @@ public CreateWorkspaceAuthzStep(
   public StepResult doStep(FlightContext flightContext)
       throws RetryException, InterruptedException {
 
-    // Even though WSM should own this resource, Stairway steps can run multiple times, so it's
-    // possible this step already created the resource. If WSM can either read the existing Sam
-    // resource or create a new one, this is considered successful.
-    if (!canReadExistingWorkspace(workspace.getWorkspaceId())) {
+    try {
       samService.createWorkspaceWithDefaults(userRequest, workspace.getWorkspaceId());
+    } catch (SamConflictException e) {
```
ddietterich, line 41: I think your analysis is right, but there is another bug that this fix will mask. Actually, it was masked by the previous code also.

The workspace id is supplied by the client, so we are not guaranteed it is unique. A CONFLICT exception from Sam could mean this is a retry or it could mean the client passed in a duplicate workspace id. (A NOT FOUND also could be thrown due to a duplicate that the client does not have access to).

We need to do a uniqueness check before this step, so this step can have a better guarantee that CONFLICT means it has retried. I think the only practical uniqueness check we can do is to inspect the WSM database for a matching workspace id and throw `DuplicateWorkspaceException`. That is not a 100% guarantee, because Rawls creates Sam workspace resources as well; we would not detect a Rawls <--> WSM id conflict.

One option is to move the `CreateWorkspaceMetadataStep` to the beginning. It would implicitly do the duplicate check. That might work. I am a little worried that `validateWorkspaceAndAction` would make 500 errors if a parallel workspace operation started after the `CreateWorkspaceMetadataStep` and before the `CreateWorkspaceAuthzStep`. Today a parallel workspace operation gets a 404 up until the `CreateWorkspaceMetadataStep` is complete.

The other option is to make a new step that would lookup the workspaceId via the DAO and throw if it existed. That leaves a small window where two parallel create requests could both see not found from the DAO, but one would fail at the metadata step. So far, we have not worried about those windows.

The third option is to split the DAO create workspace into two parts the way `createCloudContextStart` and `createCloudContextFinish` works. By recording the flight id in the row when it is in the process of being created, we prevent parallel requests and can know that it is our flight that is supposed to be doing state transitions on the DAO. That is the most complete solution. It would require changing other code that does `getWorkspace` to behave properly when the workspace is under construction.

ddietterich (changes requested):
Turns out there is a larger problem here that we should fix.

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceAuthzStep.java
melissachang: Thanks! Since this is more complicated that I thought, and this flakiness is rare, closing for now. I made a note on PF-1643. Please reopen if you think it's urgent.

PR #660, opened 2022-05-23
[PF-1658] Store UFID as notebook metadata instead of uuid

zloery:
The default notebook startup script currently retrieves the workspace ID from the notebook metadata key `instance/attributes/terra-workspace-id` and attempts to run the command `terra workspace set --id=${TERRA_WORKSPACE}`. 

WSM is still setting the UUID workspace ID as the value for this key, but terra workspace set now requires the user-facing ID instead of the UUID. This breaks the default startup script (and indirectly breaks git repo cloning because the script errors out before it gets there). I think the right behavior here is to store the UFID as the metadata key

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookInstanceResource.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookInstanceResource.java
@@ -124,7 +124,7 @@ public class ControlledAiNotebookInstanceResource extends ControlledResource {
       FlightBeanBag flightBeanBag) {
 
     WorkspaceDao workspaceDao = flightBeanBag.getWorkspaceDao();
-    String userFacingWorkspaceId = workspaceDao.getWorkspace(getWorkspaceId()).getUserFacingId();
+    String workspaceUserFacingId = workspaceDao.getWorkspace(getWorkspaceId()).getUserFacingId();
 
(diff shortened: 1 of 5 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookInstanceResource.java
```diff
@@ -136,7 +140,7 @@ public void addCreateSteps(
         gcpRetryRule);
     flight.addStep(
         new CreateAiNotebookInstanceStep(
-            this, petSaEmail, flightBeanBag.getCrlService(), flightBeanBag.getCliConfiguration()),
+            this, petSaEmail, userFacingWorkspaceId, flightBeanBag.getCrlService(), flightBeanBag.getCliConfiguration()),
```
zloery, line 143: I *think* passing the value to this step is the right approach for a couple reasons:

- This value is mutable, so I don't think we should store it as a property of a resource like we do for workspace UUID. If we did, we would need to update all resources every time we change the UFID.
- As far as I know this is the only step that will need the user facing ID, so I don't think we should move this to a flight-level parameter or change the `addCreateSteps` signature that we use for all resource types.

I'm not huge on putting too much logic here though, open to other opinions on this part!

melissachang (approved):
Thanks for fixing! We should have a test for cloning git repos in a notebook. Filed PF-1660.

jmczerk (approved):

PR #661, opened 2022-05-23
Wor 336

cahrens:


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
```diff
@@ -0,0 +1,65 @@
...
+    StorageManager manager = crlService.getStorageManager(azureCloudContext, azureConfig);
+    var azureResourceId =
+        String.format(
+            "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Storage/storageAccounts/%s",
+            azureCloudContext.getAzureSubscriptionId(),
+            azureCloudContext.getAzureResourceGroupId(),
+            resource.getStorageAccountName());
+    try {
+      logger.info("Attempting to delete Azure Storage Account " + azureResourceId);
+      manager.storageAccounts().deleteById(azureResourceId);
```
cahrens, line 47: I followed the pattern of the other delete steps by using`deleteById`, but it looks like we could also use `deleteByResourceGroup` like we do if the [creation step fails](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/CreateAzureStorageStep.java#L103). I don't know if one is better than the other.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
```diff
@@ -0,0 +1,65 @@
...
+import bio.terra.workspace.service.workspace.model.AzureCloudContext;
+import com.azure.resourcemanager.storage.StorageManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A step for deleting a controlled Azure Storage Account resource.
+ */
+public class DeleteAzureStorageStep implements Step {
+  private static final Logger logger = LoggerFactory.getLogger(CreateAzureStorageStep.class);
```
aherbst-broad, line 19:
```suggestion
  private static final Logger logger = LoggerFactory.getLogger(DeleteAzureStorageStep.class);
```

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    final AzureCloudContext azureCloudContext =
+        context
+            .getWorkingMap()
+            .get(ControlledResourceKeys.AZURE_CLOUD_CONTEXT, AzureCloudContext.class);
+
+    StorageManager manager = crlService.getStorageManager(azureCloudContext, azureConfig);
```
aherbst-broad, line 38: Nitpick: this can be `final` as well

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
aherbst-broad: `deleteByResourceGroup` might be _slightly_ cleaner so you don't have to do the string interpolation above and we can potentially reuse the deletion code between the two steps.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
```diff
@@ -0,0 +1,65 @@
...
+        String.format(
+            "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Storage/storageAccounts/%s",
+            azureCloudContext.getAzureSubscriptionId(),
+            azureCloudContext.getAzureResourceGroupId(),
+            resource.getStorageAccountName());
+    try {
+      logger.info("Attempting to delete Azure Storage Account " + azureResourceId);
+      manager.storageAccounts().deleteById(azureResourceId);
+      return StepResult.getStepResultSuccess();
+    } catch (Exception ex) {
```
aherbst-broad, line 49: 🤔 `Exception` seems overly broad here -- I wonder if we can limit this to just the [Azure exception hierarchy](https://docs.microsoft.com/en-us/java/api/overview/azure/core-readme?view=azure-java-stable#exception-hierarchy-with-azureexception)

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
```diff
@@ -208,22 +193,63 @@ public void createAndDeleteAzureRelayNamespaceControlledResource() throws Interr
...
+    StorageManager manager = azureTestUtils.getStorageManager();
+    ManagementException exception =
+            assertThrows(
+                    ManagementException.class,
+                    () ->
+                            manager.storageAccounts().getByResourceGroup(
+                                    azureTestUtils.getAzureCloudContext().getAzureResourceGroupId(),
+                                    resource.getStorageAccountName()));
+    // We see both ResourceNotFound and NotFound in the code field
+    assertTrue(exception.getValue().getCode().contains("NotFound"));
```
aherbst-broad, line 245: ```exception.response.getStatusCode() == 404``` might be clearer (assuming both flavors of error alluded to in the comment on L244 are both HTTP 404s.

aherbst-broad (approved):
lgtm, had a few comments.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
cahrens: I followed the pattern of other Delete steps, which are using the Create class for logger factory. For example:

```
public class DeleteAzureIpStep implements Step {
  private static final Logger logger = LoggerFactory.getLogger(CreateAzureIpStep.class);
```

@ddietterich was this intentional?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
cahrens: I agree that `deleteByResourceGroup` seems friendlier, but again I was following the pattern of other delete steps.

@ddietterich do you know the history of using `deleteById` vs. `deleteByResourceGroup`?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
cahrens: @ddietterich what do you think about this one? This is another place where I was just following the example of other deletion steps.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
cahrens: Done (and the variable below it).

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
```diff
@@ -203,27 +187,66 @@ public void createAndDeleteAzureRelayNamespaceControlledResource() throws Interr
...
+            StairwayTestUtils.blockUntilFlightCompletes(
+                    jobService.getStairway(),
+                    DeleteControlledResourceFlight.class,
+                    azureTestUtils.deleteControlledResourceInputParameters(
+                            workspaceUuid, resourceId, userRequest, resource),
+                    STAIRWAY_FLIGHT_TIMEOUT,
+                    null);
+    assertEquals(FlightStatus.SUCCESS, deleteFlightState.getFlightStatus());
+
+    TimeUnit.SECONDS.sleep(5);
```
marctalbott, line 233: Is this sleep required for the resource deletion to fully propagate through Azure?

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
```diff
@@ -61,23 +55,61 @@ public class CreateAndDeleteAzureControlledResourceFlightTest extends BaseAzureT
...
 
-  @Test
-  public void createAzureIpControlledResource() throws InterruptedException {
-    UUID workspaceUuid = azureTestUtils.createWorkspace(workspaceService);
-    AuthenticatedUserRequest userRequest = userAccessUtils.defaultUserAuthRequest();
-
-    // Cloud context needs to be created first
-    FlightState createAzureContextFlightState =
-        StairwayTestUtils.blockUntilFlightCompletes(
+  private void createCloudContext(UUID workspaceUuid, AuthenticatedUserRequest userRequest)
```
marctalbott, line 58: Thanks for abstracting all of this out! 🎉

marctalbott (approved):

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
cahrens: It's what was in the relay test that I used as the basis for this new test… I do generally hate sleeps and could be convinced to rework it to poll.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
cahrens: I felt the need to reduce my guilt for copy/pasting so much in this PR. :)

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
cahrens: I switched to `deleteByResourceGroup` as it seems less brittle. If there is a good reason to use `deleteById`, I can change it back in my PR where I add storage containers.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/DeleteAzureStorageStep.java
cahrens: I'm going to leave this as-is unless I hear back that I should make the exception more targeted.

cahrens:
jenkins retest

PR #662, opened 2022-05-24
[PF-1657] workspaceId on clone subflights

jaycarlton:
These flights were launched without the job builder, and left out the allegedly required parameter workspaceId. This should help identify these jobs in during a clone, and perhaps the UI can make use of this information. Also added descriptions including source workspace IDs, as I don't know the destination workspace IDs at that point.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCloneControlledGcpBigQueryDatasetResourceFlightStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCloneControlledGcpBigQueryDatasetResourceFlightStep.java
@@ -15,6 +15,7 @@ import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.job.JobMapKeys;
 import bio.terra.workspace.service.resource.controlled.cloud.gcp.bqdataset.ControlledBigQueryDatasetResource;
 import bio.terra.workspace.service.resource.controlled.flight.clone.dataset.CloneControlledGcpBigQueryDatasetResourceFlight;
+import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys;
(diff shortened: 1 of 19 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCloneControlledGcpBigQueryDatasetResourceFlightStep.java
```diff
@@ -61,6 +62,11 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
         ControlledResourceKeys.CLONING_INSTRUCTIONS, resource.getCloningInstructions());
     subflightInputParameters.put(ResourceKeys.RESOURCE, resource);
     subflightInputParameters.put(ControlledResourceKeys.LOCATION, location);
+    subflightInputParameters.put(WorkspaceFlightMapKeys.WORKSPACE_ID, destinationWorkspaceId);
+    subflightInputParameters.put(
+        JobMapKeys.DESCRIPTION.getKeyName(),
+        String.format("Clone BigQuery Dataset %s", resource.getResourceId().toString()));
```
zloery, line 68: Would it be helpful for this (and other clone resource flight) description to also include the destination workspace ID? maybe a question for @cbookg

zloery (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCloneControlledGcpBigQueryDatasetResourceFlightStep.java
jaycarlton: I figured the destination workspace ID is how they'll be found to begin with. I'm also wondering if I shouldn't have a more regular format for the descriptions that just plain English, though I believe that's what Chris is parsing today.

PR #663, opened 2022-05-24
[PF-1519] Create a single combined jacoco report

zloery:
Currently, we only generate a Jacoco report showing coverage for the most recently run task. Instead, when running multiple tasks (such as `unitTest` + `connectedTest` like we do for PRs), this should now give us the combined coverage from all test types we've run.

This also creates and documents a new `jacocoBootRun` task which runs the server instrumented by Jacoco. This is useful for integration tests which run against a local server, but I think there's a moderate performance hit so it's better to keep using `bootRun` for live environments.

Incidentally I think removing `dependsOn test` from the jacoco test report task also fixes the bug where running `unitTest` would run the unit tests twice! Previously the `unitTest` task triggered the `jacocoTestReport` which depended on the `test` task (which also ran unit tests), so `unitTest` indirectly depended on `test`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -141,7 +141,7 @@ jobs:
       run: ./gradlew :service:${{ matrix.gradleTask }} --scan
 
     - name: Codacy Coverage Reporter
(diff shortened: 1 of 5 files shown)


ddietterich (approved):
LGTM - thanks for doing this. Very cool!

PR #664, opened 2022-05-25
TOAZ-114 Abandoned network interface in case of VM creation failure

sergiygetlin:
https://broadworkbench.atlassian.net/browse/TOAZ-114

Initial assumption was that there might be abandoned network interface due to retry issue in VM create step (in case of VM creation failure). The real behavior is that every time we try to create network interface, and we try to create it with the same name. Azure handles this situation good and doesn’t return any errors. So, even in case of multiple sequential calls we will have only one network interface created. But we can't rely on this behavior now because it might be changed in future. To prevent situation when Azure fails in this case and also to prevent WSM from doing necessary redundant call to create network interface in case of multiple retries we need to adjust VM creation step (especially do step). The undo step already has logic which removes network interface.

PR details:
-Introduce standalone steps for network creation/deletion.
-remove redundant azure call to create network interface in case of retries


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
```diff
@@ -261,9 +263,27 @@ private NetworkInterface createNetworkInterface(
...
+      Optional<PublicIpAddress> existingAzureIp) {
+    NetworkInterface networkInterface;
+    if (context.getWorkingMap().containsKey(WORKING_MAP_NETWORK_INTERFACE_KEY)) {
+      String nicId = context.getWorkingMap().get(WORKING_MAP_NETWORK_INTERFACE_KEY, String.class);
+      networkInterface = computeManager.networkManager().networkInterfaces().getById(nicId);
+    } else {
+      networkInterface =
+          createNetworkInterface(
+              computeManager, azureCloudContext, existingNetwork, subnetName, existingAzureIp);
+      context.getWorkingMap().put(WORKING_MAP_NETWORK_INTERFACE_KEY, networkInterface.id());
```
ddietterich, line 285: I do not recommend this technique, although it probably works most of the time.

The workingMap is only serialized at a step boundary, so it is not persisted here on retry. You would do just as well with a local variable. You can see, though, that on a pod failure or failover, the working map would be state is lost.

I think a better approach would be to make a new step that just handles the create of the network interface. When that step completes, you would know you had created it and that the working map item would be persisted.

ddietterich (changes requested):
I am suggesting adding a step to handle the network interface

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
sergiygetlin: Thank you for the feedback. I probably missed that part. Let me check it.

sergiygetlin: New step for network interface added

ddietterich (approved):
lgtm

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
```diff
@@ -109,18 +109,32 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
                   azureCloudContext.getAzureResourceGroupId(), networkResource.getNetworkName());
 
-      var createNic =
-          createNetworkInterface(
-              computeManager,
-              azureCloudContext,
-              existingNetwork,
-              networkResource.getSubnetName(),
-              existingAzureIp);
+      if (!context.getWorkingMap().containsKey(AzureVmHelper.WORKING_MAP_NETWORK_INTERFACE_KEY)) {
```
Qi77Qi, line 112: what happens if there're multiple create VM requests at the same time? how does the code know which nic is for which VM?

Qi77Qi: @ddietterich is that the case?

sergiygetlin: From my understanding each flight isolated. So the context is also per flight.

Qi77Qi (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
ddietterich: Yes, each flight is isolated. Multiple VM creates would be multiple flights.

We have not implemented much concurrency control in WSM. In general, we rely on the metadata transaction to add metadata that will conflict with a parallel operation on the same object. That covers many cases. If we discover a need to cover more cases (like concurrent update and delete operations), then we will need to add some kind of busy state to the metadata in the WSM database.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
```diff
@@ -43,793 +56,941 @@
 import org.springframework.beans.factory.annotation.Autowired;
 
 public class CreateAndDeleteAzureControlledResourceFlightTest extends BaseAzureTest {
```
cahrens, line 58: @sergiygetlin FYI that this repo uses indenting of 2 spaces (`google-java-format`). I am re-indenting this file in my open PR, but I wanted to let you know for future commits. Indenting changes like this cause merge conflicts, which can be painful, and they make the commit history more difficult to follow.

Thanks!

cahrens, line 58: @sergiygetlin FYI that this repo uses indenting of 2 spaces (`google-java-format`). I am re-indenting this file in my open PR, but I wanted to let you know for future commits. Indenting changes like this cause merge conflicts, which can be painful, and they make the commit history more difficult to follow.

Thanks!

PR #665, opened 2022-05-25
[PF-1659] Add an update notebook resource endpoint.

yuhuyoyo:
Currently we can update name, description and metadata.


service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -396,6 +397,39 @@ public String createAiNotebookInstance(
     return jobId;
   }
 
+  public ControlledAiNotebookInstanceResource updateAiNotebookInstance(
+      ControlledAiNotebookInstanceResource resource,
+      ApiGcpAiNotebookUpdateParameters updateParameters,
+      Optional<String> newName,
```
zloery, line 403: generally `Optional` should only be used for return values. I think these should be `@Nullable String` instead

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
```diff
@@ -42,6 +45,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
+import scala.Predef.StringFormat;
```
zloery, line 48: nit: unused imports?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/UpdateAiNotebookAttributesStep.java
```diff
@@ -0,0 +1,108 @@
...
+    final ApiGcpAiNotebookUpdateParameters prevParameters =
+        workingMap.get(PREVIOUS_UPDATE_PARAMETERS, ApiGcpAiNotebookUpdateParameters.class);
+    var projectId =
+        cloudContextService.getRequiredGcpProject(resource.getWorkspaceId());
+    InstanceName instanceName = resource.toInstanceName(projectId);
+    AIPlatformNotebooksCow notebooks = crlService.getAIPlatformNotebooksCow();
+    try {
+      var currentMetadata = notebooks.instances().get(instanceName).execute().getMetadata();
+      for (var entry: currentMetadata.entrySet()) {
+        if (!prevParameters.getMetadata().containsKey(entry.getKey())) {
```
zloery, line 66: iiuc this will delete metadata values for any newly-introduced keys.
Do we also need to reset values for any metadata keys that were modified (i.e. the key is in `prevParameters.getMetadata()` but the value is different)?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/UpdateAiNotebookAttributesStep.java
```diff
@@ -0,0 +1,108 @@
...
+      }
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    } catch (IOException e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
+  }
+
+  private StepResult updateAiNotebook(ApiGcpAiNotebookUpdateParameters updateParameters) {
+    var projectId =
+        cloudContextService.getRequiredGcpProject(resource.getWorkspaceId());
```
zloery, line 85: nit: pass project name to this method instead of looking it up in the DB again

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/UpdateAiNotebookAttributesStep.java
```diff
@@ -0,0 +1,108 @@
...
+      for (var entry: currentMetadata.entrySet()) {
+        if (!prevParameters.getMetadata().containsKey(entry.getKey())) {
+          // reset the new key entry to "" value because the gcp api does not allow deleting
+          // metadata item so we can't simply undo the add.
+          currentMetadata.put(entry.getKey(), "");
+        }
+      }
+      return updateAiNotebook(new ApiGcpAiNotebookUpdateParameters().metadata(currentMetadata));
+    } catch (GoogleJsonResponseException e) {
+      if (HttpStatus.BAD_REQUEST.value() == e.getStatusCode() || HttpStatus.NOT_FOUND.value() == e.getStatusCode()) {
```
zloery, line 74: I think this should throw an exception and fail the flight to be consistent with notebook creation. 

currently when creating the notebook, we throw a `ReservedMetadataKeyException` and fail the operation if the user specifies a reserved key. In this case, we're silently not including it and continuing. I think either approach is reasonable, but we should at least be consistent.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
yuhuyoyo: why not? especially we don't have a static check for nullable

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/UpdateAiNotebookAttributesStep.java
yuhuyoyo: good catch.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/UpdateAiNotebookAttributesStep.java
yuhuyoyo: i'm just forseeing that it gives more burden on the UI  to check for valid input :) but yeah i agree consistency is good.

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -297,6 +303,212 @@ void createAiNotebookInstanceDo() throws Exception {
...
+    );
+    assertEquals(
+        newDescription,
+        updatedInstance.getDescription()
+    );
+    // cloud notebook attributes are updated.
+    var updatedInstanceFromCloud = crlService.getAIPlatformNotebooksCow().instances().get(updatedInstance.toInstanceName(projectId)).execute();
+    // Add the new value to the metadata.
+    AI_NOTEBOOK_UPDATE_PARAMETERS.getMetadata().forEach(
+        (key, value) -> metadata.merge( key, value, (v1, v2) -> v1.equalsIgnoreCase(v2) ? v1 : v1 + "," + v2)
```
jaycarlton, line 365: comment on this bit of magic

zloery (approved):

PR #666, opened 2022-05-25
[PF-1519] Small tests and dead code cleanup from test coverage report

zloery:
An assortment of low-hanging fruit I found from the code coverage report. This isn't an exhaustive list of easy fixes or dead code, I expect there are probably still a few others.


service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -630,184 +630,6 @@ public boolean resourceExists(UUID workspaceUuid, UUID resourceId) {
     return (count != null && count > 0);
   }
 
-  private void validateUniqueGcsBucket(ControlledGcsBucketResource bucketResource) {
```
zloery, line 633: These methods are all unused, they were replaced by the pattern of putting `UniquenessCheckAttributes` on each resource type

service/src/main/java/bio/terra/workspace/service/buffer/BufferService.java
```diff
@@ -43,36 +43,6 @@ private BufferApi bufferApi(String instanceUrl) throws IOException {
         getApiClient(bufferServiceConfiguration.getAccessToken()).setBasePath(instanceUrl));
   }
 
-  /**
```
zloery, line 46: Unused code

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/GetGcpCloudContextStep.java
```diff
@@ -1,47 +0,0 @@
...
-import bio.terra.stairway.exception.RetryException;
-import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
-import bio.terra.workspace.service.workspace.GcpCloudContextService;
-import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.ControlledResourceKeys;
-import bio.terra.workspace.service.workspace.model.GcpCloudContext;
-import java.util.UUID;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-public class GetGcpCloudContextStep implements Step {
```
zloery, line 15: Unused step, this was replaced by `GetCloudContextStep` which also handles Azure contexts

service/src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -134,14 +134,6 @@ protected <T> T submitAndWait(
     return resultOrException.getResult();
   }
 
-  protected <T> T submitAndWait(
```
zloery, line 137: Unused overload, everything already calls the `submitAndWait(..., boolean doAccessCheck)` version instead

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
```diff
@@ -119,8 +120,8 @@ public GcpCloudContext getRequiredGcpCloudContext(
           samService.getWorkspacePolicy(workspaceUuid, WsmIamRole.READER, userRequest));
       context.setSamPolicyApplication(
           samService.getWorkspacePolicy(workspaceUuid, WsmIamRole.APPLICATION, userRequest));
+      workspaceDao.updateCloudContext(workspaceUuid, CloudPlatform.GCP, context.serialize());
     }
-    workspaceDao.updateCloudContext(workspaceUuid, CloudPlatform.GCP, context.serialize());
```
zloery, line 123: We should only be updating the cloud context if we're modifying it. Otherwise, we're just reading a DB value -> not modifying it -> writing it back to the DB

service/src/test/java/bio/terra/workspace/service/workspace/ApplicationServiceTest.java
```diff
@@ -108,6 +130,7 @@ void setup() throws Exception {
                 Mockito.any(), Mockito.eq(SamResource.WORKSPACE), Mockito.any(), Mockito.any()))
         .thenReturn(true);
 
+    appService.enableTestMode();
```
jaycarlton, line 133: I wonder if there's a Springier way of achieving this, like injecting a fake version of the service, or something. It's a bit odd having a test method in the production code.

service/src/test/java/bio/terra/workspace/service/workspace/ApplicationServiceTest.java
```diff
@@ -195,6 +218,49 @@ public void applicationEnableTest() {
...
+    decommissionApp.state(WsmApplicationState.DECOMMISSIONED);
+    appService.processApp(decommissionApp, appService.buildAppMap());
+    assertFalse(appService.getErrorList().isEmpty());
+    assertTrue(appService.getErrorList().get(0).contains("associated resources"));
+    // Delete the resource
+    resourceDao.deleteResource(workspaceUuid, resourceId);
+    // try to deprecate the app again, should succeed this time
+    appService.processApp(decommissionApp, appService.buildAppMap());
+    // should be no new errors
+    assertEquals(1, appService.getErrorList().size());
```
jaycarlton, line 251: what's the existing error expected to be here?

jaycarlton (approved):

service/src/test/java/bio/terra/workspace/service/workspace/ApplicationServiceTest.java
ddietterich: There is probably a better way. Not salient for this PR, though.

service/src/test/java/bio/terra/workspace/service/workspace/ApplicationServiceTest.java
```diff
@@ -195,6 +222,58 @@ public void applicationEnableTest() {
     wsmApp = appService.getWorkspaceApplication(USER_REQUEST, workspaceId2, LEO_ID);
     assertEquals(LEO_APP, wsmApp.getApplication());
     assertTrue(wsmApp.isEnabled());
+
+    // validate decommissioned apps can't be re-enabled
+    // Clear error list first
+    appService.enableTestMode();
```
ddietterich, line 228: Seems like you either put this here or in the `@BeforeEach`, but not both

ddietterich (approved):
LGTM

service/src/test/java/bio/terra/workspace/service/workspace/ApplicationServiceTest.java
zloery: yeah - I had been using it to clear the error list, which is a little hacky. A couple different operations (creates, enables, etc.) write to the error log so I wanted to avoid grabbing magic indices out of `appService.getErrorList()` (side note: should only errors write to this log, or can we rename it to `getLogList`?)

I think a better way is just getting the most recent log, switched to that.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -973,32 +973,6 @@ private Map<String, AccessPolicyMembershipV2> defaultWorkspacePolicies(String ow
     return policyMap;
   }
 
-  /**
-   * Add WSM's service account as the owner of a controlled resource in Sam. Used for admin
-   * reassignment of resources. This assumes samService.initialize() has already been called, which
-   * should happen on start.
-   */
-  private void addWsmResourceOwnerPolicy(CreateResourceRequestV2 request)
```
zloery, line 981: replaced by `addWsmResourceOwnerPolicy` in `ControlledResourceSamPolicyBuilder`

PR #667, opened 2022-05-26
[PF-1524] Add permissions for GAR and Cloudbuild

yuhuyoyo:
https://cloud.google.com/artifact-registry/docs/access-control

PROJECT_READER_PERMISSION maps to roles/artifactregistry.reader (view and get artifacts, view repository metadata)
PROJECT_WRITER_PERMISSION maps to roles/artifactregistry.writer (READA and write artifacts)

https://cloud.google.com/build/docs/iam-roles-permissions

PROJECT_READER_PERMISSION maps to cloudbuild.builds.viewer (can view cloud build resources)
PROJECT_WRITER_PERMISSION maps to cloudbuild.builds.editor (full control of cloud build resources)

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
@@ -28,7 +28,6 @@ public class CloudSyncRoleMapping {
           "artifactregistry.files.list",
           "artifactregistry.files.get",
           "artifactregistry.packages.list",
-          "artifactregistry.packages.listTagBindings",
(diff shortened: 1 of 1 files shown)


zloery (approved):
LGTM!

Just to confirm: setting up permissions this way means all users in a workspace can read/write GAR and cloud build artifacts according to their role, meaning we can't support "private" or "WSM app-specific" GAR or cloud build artifacts. Is that alright from a user perspective?

(IMO it seems totally reasonable to start with and much lower cost to implement than GAR/cloud build controlled resources, which we would need for "private" artifacts)

ddietterich (approved):
LGTM

PR #668, opened 2022-05-26
[WOR-92] Add the ability to create an Azure storage container

cahrens:
This adds the ability to create and delete storage containers within a workspace's storage account. Note that deletion is not exposed through the Swagger API (this is true of storage accounts as well).

I have followed https://github.com/DataBiosphere/terra-workspace-manager/blob/main/ADDING_RESOURCES.md to the best of my ability, though I did not add an integration test. To the best of my knowledge, we do not have integration tests in WSM for Azure resources (storage accounts, relays, etc.).

### DIFF
--- a/openapi/src/parts/controlled_azure_storage.yaml
+++ b/openapi/src/parts/controlled_azure_storage.yaml
@@ -31,9 +31,9 @@ components:
         Storage-specific properties to be set on creation. These are a subset of the values
         accepted by the azure resource API
       type: object
-      required: [ name, region ]
+      required: [ storageAccountName, region ]
       properties:
-        name:
+        storageAccountName:
           description: A valid storage account name per https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
           type: string
         region:
--- a/openapi/src/parts/controlled_azure_storage_container.yaml
+++ b/openapi/src/parts/controlled_azure_storage_container.yaml
@@ -1,0 +1,62 @@
+paths:
+  /api/workspaces/v1/{workspaceId}/resources/controlled/azure/storageContainer:
+    parameters:
+      - $ref: '#/components/parameters/WorkspaceId'
+    post:
+      summary: Create a new controlled Azure Storage Container
+      operationId: createAzureStorageContainer
+      tags: [ControlledAzureResource]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateControlledAzureStorageContainerRequestBody'
+      responses:
+        '200':
+          description: Response to create controlled azure storage
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreatedControlledAzureStorageContainer'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '500':
+          $ref: '#/components/responses/ServerError'
+
+components:
+  schemas:
+    AzureStorageContainerCreationParameters:
+      description: >-
+        Storage container-specific properties to be set on creation. These are a subset of the values
+        accepted by the azure resource API
+      type: object
+      required: [ name, storageAccountName ]
+      properties:
+        name:
+          description: A valid storage container name per https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
+          type: string
+        storageAccountName:
+          description: A valid storage account name per https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
+          type: string
+
+    CreateControlledAzureStorageContainerRequestBody:
+      description: Payload for requesting a new controlled Azure storage container resource.
+      type: object
+      properties:
+        common:
+          $ref: '#/components/schemas/ControlledResourceCommonFields'
+        azureStorageContainer:
+          $ref: '#/components/schemas/AzureStorageContainerCreationParameters'
+
+    CreatedControlledAzureStorageContainer:
+      description: Response payload for requesting a new Azure storage container
+      type: object
+      properties:
+        resourceId:
+          description: UUID of a newly-created resource.
+          type: string
+          format: uuid
+        azureStorageContainer:
+          $ref: '#/components/schemas/AzureStorageContainerResource'
+
--- a/openapi/src/parts/controlled_azure_storage_container.yaml
+++ b/openapi/src/parts/controlled_azure_storage_container.yaml
@@ -3,7 +3,7 @@ paths:
     parameters:
       - $ref: '#/components/parameters/WorkspaceId'
     post:
-      summary: Create a new controlled Azure Storage Container
+      summary: Create a new controlled Azure Storage Container with private access
       operationId: createAzureStorageContainer
       tags: [ControlledAzureResource]
       requestBody:
@@ -29,7 +29,7 @@ components:
     AzureStorageContainerCreationParameters:
       description: >-
         Storage container-specific properties to be set on creation. These are a subset of the values
-        accepted by the azure resource API
+        accepted by the azure resource API (note that public access is disabled).
       type: object
       required: [ name, storageAccountName ]
       properties:
@@ -59,4 +59,3 @@ components:
           format: uuid
         azureStorageContainer:
           $ref: '#/components/schemas/AzureStorageContainerResource'
-
--- a/openapi/src/parts/controlled_azure_storage_container.yaml
+++ b/openapi/src/parts/controlled_azure_storage_container.yaml
@@ -31,13 +31,14 @@ components:
         Storage container-specific properties to be set on creation. These are a subset of the values
         accepted by the azure resource API (note that public access is disabled).
       type: object
-      required: [ name, storageAccountName ]
+      required: [ storageAccountId, storageContainerName ]
       properties:
-        name:
-          description: A valid storage container name per https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
+        storageAccountId:
+          description: The resource ID of the storage account in which the container will be created.
           type: string
-        storageAccountName:
-          description: A valid storage account name per https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
+          format: uuid
+        storageContainerName:
+          description: A valid storage container name per https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
           type: string
 
     CreateControlledAzureStorageContainerRequestBody:
--- a/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
@@ -66,6 +66,16 @@ public class ResourceValidationUtils {
       Pattern.compile("^[a-z0-9]{3,24}$");
 
   /**
+   * Azure Storage Container name validation. An storage container name must be between 3-63
+   * characters in length and may contain numbers, lowercase letters, and dash (-) character only.
+   * They must start and end with a letter or number. Matching this pattern is necessary but not sufficient for a
+   * valid container name (in particular, the dash character must be immediately preceded and followed by a letter or
+   * number; consecutive dashes are not permitted).
+   */
+  public static final Pattern AZURE_STORAGE_CONTAINER_NAME_VALIDATION_PATTERN =
+          Pattern.compile("^[a-z0-9][-a-z0-9]{1,61}[a-z0-9]$");
+
+  /**
    * BigQuery datasets must be 1-1024 characters, using letters (upper or lowercase), numbers, and
    * underscores.
    */
@@ -374,6 +384,16 @@ public class ResourceValidationUtils {
       logger.warn("Invalid Storage Account name: {}", storageAccountName);
       throw new InvalidReferenceException(
           "Invalid Azure Storage Account name. The name must be 3 to 24 alphanumeric lower case characters.");
+    }
+  }
+
+  public static void validateStorageContainerName(String storageContainerName) {
+    if (!AZURE_STORAGE_CONTAINER_NAME_VALIDATION_PATTERN.matcher(storageContainerName).matches() ||
+            storageContainerName.contains("--")) {
+      logger.warn("Invalid Storage Container name: {}", storageContainerName);
+      throw new InvalidReferenceException(
+              "Invalid Azure Storage Container name. The name must be 3 to 63 alphanumeric lower case characters " +
+                      "or dashes, must start and end with a letter or number, and cannot contain consecutive dashes.");
     }
   }
 
--- a/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
@@ -66,7 +66,7 @@ public class ResourceValidationUtils {
       Pattern.compile("^[a-z0-9]{3,24}$");
 
   /**
-   * Azure Storage Container name validation. An storage container name must be between 3-63
+   * Azure Storage Container name validation. A storage container name must be between 3-63
    * characters in length and may contain numbers, lowercase letters, and dash (-) characters only.
    * It must start and end with a letter or number. Matching this pattern is necessary but not sufficient for a
    * valid container name (in particular, the dash character must be immediately preceded and followed by a letter or
--- a/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
@@ -67,13 +67,13 @@ public class ResourceValidationUtils {
 
   /**
    * Azure Storage Container name validation. An storage container name must be between 3-63
-   * characters in length and may contain numbers, lowercase letters, and dash (-) character only.
-   * They must start and end with a letter or number. Matching this pattern is necessary but not sufficient for a
+   * characters in length and may contain numbers, lowercase letters, and dash (-) characters only.
+   * It must start and end with a letter or number. Matching this pattern is necessary but not sufficient for a
    * valid container name (in particular, the dash character must be immediately preceded and followed by a letter or
    * number; consecutive dashes are not permitted).
    */
   public static final Pattern AZURE_STORAGE_CONTAINER_NAME_VALIDATION_PATTERN =
-          Pattern.compile("^[a-z0-9][-a-z0-9]{1,61}[a-z0-9]$");
+      Pattern.compile("^[a-z0-9][-a-z0-9]{1,61}[a-z0-9]$");
 
   /**
    * BigQuery datasets must be 1-1024 characters, using letters (upper or lowercase), numbers, and
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
@@ -95,7 +95,7 @@ public class CreateAzureStorageContainerStep implements Step {
               .getByResourceGroup(
                       azureCloudContext.getAzureResourceGroupId(), storageAccountName);
     } catch (ManagementException ex) {
-      if (ManagementExceptionUtils.isResourceNotFound(ex)) {
+      if (ManagementExceptionUtils.isExceptionCode(ex, ManagementExceptionUtils.RESOURCE_NOT_FOUND)) {
         logger.warn(
                 "Deletion of the storage container is not required. Parent storage account does not exist. {}",
                 storageAccountName);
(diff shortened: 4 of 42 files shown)


openapi/src/parts/controlled_azure_storage_container.yaml
```diff
@@ -0,0 +1,62 @@
...
+components:
+  schemas:
+    AzureStorageContainerCreationParameters:
+      description: >-
+        Storage container-specific properties to be set on creation. These are a subset of the values
+        accepted by the azure resource API
+      type: object
+      required: [ name, storageAccountName ]
+      properties:
+        name:
```
cahrens, line 36: In keeping with other resources, I used `name` here instead of `storageContainerName`. However, I could see going with `storageContainerName` for clarity and to be consistent with what is used for the container attributes.

Thoughts?

service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
```diff
@@ -377,6 +387,16 @@ public static void validateStorageAccountName(String storageAccountName) {
     }
   }
 
+  public static void validateStorageContainerName(String storageContainerName) {
+    if (!AZURE_STORAGE_CONTAINER_NAME_VALIDATION_PATTERN.matcher(storageContainerName).matches() ||
+            storageContainerName.contains("--")) {
```
cahrens, line 392: My regexp skills aren't strong enough to figure out how to enforce no consecutive dashes in the regexp itself.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
```diff
@@ -0,0 +1,142 @@
...
+              .storageAccounts()
+              .getByResourceGroup(
+                      azureCloudContext.getAzureResourceGroupId(), resource.getStorageAccountName());
+    } catch (ManagementException ex) {
+      logger.warn(
+              "Deletion of the storage container is not required. Parent storage account does not exist. {}",
+              resource.getStorageAccountName());
+      return StepResult.getStepResultSuccess();
+    }
+    final PagedIterable<ListContainerItemInner> existingContainers = storageManager.blobContainers().list(
```
cahrens, line 102: If the previous `getByResourceGroup` call didn't throw an exception, this list method should not throw an exception. I could be convinced though to put everything within a try/catch.

aherbst-broad, line 102: ❓ Would it make sense to use [`blobContainers.get()` ](https://docs.microsoft.com/en-us/java/api/com.azure.resourcemanager.storage.models.blobcontainers.get?view=azure-java-stable#com-azure-resourcemanager-storage-models-blobcontainers-get(java-lang-string-java-lang-string-java-lang-string)) to see if the container exists, rather than iterating over the containers in the account?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/DeleteAzureStorageContainerStep.java
```diff
@@ -0,0 +1,60 @@
...
+import bio.terra.workspace.app.configuration.external.AzureConfiguration;
+import bio.terra.workspace.service.crl.CrlService;
+import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.ControlledResourceKeys;
+import bio.terra.workspace.service.workspace.model.AzureCloudContext;
+import com.azure.resourcemanager.storage.StorageManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A step for deleting a controlled Azure Storage Account resource.
```
aherbst-broad, line 16: ⛏️ Nitpick: 
```suggestion
 * A step for deleting a controlled Azure Storage Container resource.
```

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/GetAzureStorageContainerStep.java
```diff
@@ -0,0 +1,80 @@
...
+              .storageAccounts()
+              .getByResourceGroup(
+                      azureCloudContext.getAzureResourceGroupId(), resource.getStorageAccountName());
+    } catch (ManagementException ex) {
+      return new StepResult(
+              StepStatus.STEP_RESULT_FAILURE_FATAL, new ResourceNotFoundException(
+              String.format("No Azure storage account with name '%s' exists for this workspace",
+                      resource.getStorageAccountName())));
+    }
+    final PagedIterable<ListContainerItemInner> existingContainers = storageManager.blobContainers().list(
```
aherbst-broad, line 57: ❓ Same question about using `blobContainers.get()` here.

openapi/src/parts/controlled_azure_storage_container.yaml
aherbst-broad: I'm in favor of the response + request using the same name for parameters; we should adjust the storage account parameters to do this as well.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
```diff
@@ -0,0 +1,142 @@
...
+                resource.getStorageContainerName()
+        );
+        logger.warn("Successfully deleted storage container '{}' in account '{}'",
+                resource.getStorageContainerName(),
+                resource.getStorageAccountName()
+        );
+        return StepResult.getStepResultSuccess();
+      } catch (ManagementException ex) {
+        logger.error(
+                "Attempt to delete Azure Storage Container failed on this try: " + resource.getStorageContainerName(),
```
aherbst-broad, line 135: ❓ It will be helpful for debugging to include the error code from the exception in this logging statement, will we get that by including ex as the last arg here? Or do we need to include ex.getValue().getCode() ?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
```diff
@@ -0,0 +1,142 @@
...
+                    .get(ControlledResourceKeys.AZURE_CLOUD_CONTEXT, AzureCloudContext.class);
+    final StorageManager storageManager = crlService.getStorageManager(azureCloudContext, azureConfig);
+
+    try {
+      storageManager
+              .storageAccounts()
+              .getByResourceGroup(
+                      azureCloudContext.getAzureResourceGroupId(), resource.getStorageAccountName());
+    } catch (ManagementException ex) {
+      logger.warn(
```
aherbst-broad, line 97: Is it certain that all `ManagementException`s thrown in the above statement will be of the not found variety? It seems like there may be other types of errors that this may be masking.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
```diff
@@ -0,0 +1,142 @@
...
+      }
+    }
+    if (existingContainer == null) {
+      logger.warn(
+              "Deletion of the storage container is not required. Storage container does not exist. {}",
+              resource.getStorageContainerName());
+      return StepResult.getStepResultSuccess();
+    } else {
+      try {
+        logger.warn("Attempting to delete storage container '{}' in account '{}'",
```
aherbst-broad, line 119: I like these breadcrumb logging statements.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
```diff
@@ -0,0 +1,142 @@
...
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    final AzureCloudContext azureCloudContext =
+            context
+                    .getWorkingMap()
+                    .get(ControlledResourceKeys.AZURE_CLOUD_CONTEXT, AzureCloudContext.class);
+    final StorageManager storageManager = crlService.getStorageManager(azureCloudContext, azureConfig);
+
+    try {
+      storageManager
```
aherbst-broad, line 92: 🤔 We may need to add an additional authz check here on the storage account. I'm thinking through the scenario where I nefariously pass in a storage account name that is in the MRG but not "owned" by the workspace -- will this still succeed?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/GetAzureStorageContainerStep.java
```diff
@@ -0,0 +1,80 @@
...
+            context.getWorkingMap().get(ControlledResourceKeys.AZURE_CLOUD_CONTEXT, AzureCloudContext.class);
+    final StorageManager storageManager = crlService.getStorageManager(azureCloudContext, azureConfig);
+
+    try {
+      storageManager
+              .storageAccounts()
+              .getByResourceGroup(
+                      azureCloudContext.getAzureResourceGroupId(), resource.getStorageAccountName());
+    } catch (ManagementException ex) {
+      return new StepResult(
```
aherbst-broad, line 52: Same question about `ManagementException`, does this always mean not found?

aherbst-broad (commented):
Looking good, had a couple of questions.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
ddietterich: This will get more complex when the storage account is in the landing zone.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/GetAzureStorageContainerStep.java
```diff
@@ -0,0 +1,80 @@
...
+import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.ControlledResourceKeys;
+import bio.terra.workspace.service.workspace.model.AzureCloudContext;
+import com.azure.core.http.rest.PagedIterable;
+import com.azure.core.management.exception.ManagementException;
+import com.azure.resourcemanager.storage.StorageManager;
+import com.azure.resourcemanager.storage.fluent.models.ListContainerItemInner;
+
+
+/**
+ * Gets an Azure Storage Container, and fails if it already exists or if the parent storage account does not exist.
```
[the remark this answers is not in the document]
ddietterich: True enough. A different name than "Get" would be very appropriate; e.g., `EnsureContainerDoesNotExistStep`

openapi/src/parts/controlled_azure_storage_container.yaml
cahrens: I would like the name to be the same, but it will be a breaking change for folks using the API client. Not a big deal for Rawls because I can easily update the code in my next PR. @ddietterich do you think it's OK to change the name of the request attribute? I'm not sure if anyone (outside of Rawls) is actually using this part of the API yet.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
cahrens: I guess I can check that the storage account is one of the controlled resources of the workspace. But I do wonder how all of this will change when the storage account moves to the landing zone.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
cahrens: I shied away from the `get` method because I think it throws an exception if the container doesn't exist. As far as I can tell, the `list` method is the closest to `exists`.

I can call `get` and handle the exception if folks think that is clearer (I will also double-check that it throws an exception if the container DNE-- the documentation does not say).

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/resourcemanager/data/BaseStorageContainerRequestData.java
```diff
@@ -0,0 +1,37 @@
+package bio.terra.workspace.service.resource.controlled.cloud.azure.storageContainer.resourcemanager.data;
+
+import bio.terra.cloudres.azure.resourcemanager.common.ResourceManagerRequestData;
+import com.google.gson.JsonObject;
+
+/**
+ * Extends {@link ResourceManagerRequestData} to add common fields for working with the Storage
+ * Resource Provider API.
+ *
+ * <p>Note: Azure Storage has its own resource provider. Therefore, the type structure should be
```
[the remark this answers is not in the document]
cahrens: I don't know anything beyond what is in this comment.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/GetAzureStorageContainerStep.java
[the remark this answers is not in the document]
cahrens: Will rename to `VerifyAzureStorageContainerCanBeCreatedStep` (the step both checks the storage account existence and that the container does not exist).

openapi/src/parts/controlled_azure_storage_container.yaml
ddietterich: I think the API is not used in production and only used in that one place, right? If so, it should be no problem to change both sides. We have made changes like that before.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/GetAzureStorageContainerStep.java
cahrens: You are right-- I shouldn't assume it doesn't exist. However, if we can't retrieve the storage account I think we should assume that creating the container will fail. I will change the error message to be more general.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
dvoet: I think part of the problem is that `storageAccountName` is part of the request and cannot be trusted. When there is a landing zone it will dictate which storage account to use which means it should not be supplied in the request. I guess the question for now is can the workspace infer which storage account to use before landing zones come online? If they can, then do and remove it from the request. If not, well, I would encourage you to figure out an interim solution where they can :)

cahrens: I assume there is a way that I could query the workspace resources to see if a storage account already exists for the workspace. If one doesn't exist, we could error and say that a storage account must be created first. If one does exist, we would just use the first one we find?

I agree that not passing the storage account name would help protect the API from the storage account moving into the WLZ.

ddietterich: How about we implement this current case by specifying the storage account resource id in the workspace to use for the create? We can look that up in the WSM metadata and do a Sam authz check on it to ensure the requesting user has access.

When landing zones are present, the landing zone manager will have to be able to provide the list of available landing zone storage accounts. When that is online, we can make a variant of the container create request that accepts the object id (or name) of the landing zone storage account to use. WSM can validate with LZ manager in the context of the LZ associated with the workspace.

I don't think we want to un-implement the case of a storage account allocated in the workspace.

cahrens: @ddietterich to be clear, are you suggesting that instead of the storage container create API taking the storage account name, it takes a storage account resource ID? Since the storage account name has to be unique, I think we could figure out the storage account resource ID from the name, though it would be a bit extra work.

Or are you suggesting when we create a storage account in the workspace, we store the storage account resource ID in the WSM metadata to use for future storage container create requests (and those do not specify a storage account)?

ddietterich: I think that right now we only want to allow creation of containers from storage accounts in the workspace. When we have landing zones, we will also allow creation of containers from storage accounts in the landing zone.

Today when we make a storage account, we create a controlled resource in the workspace. It is scoped to the workspace and has a unique resource id. I'm suggesting you use that resource id of the storage account instead of the storage account name. The WSM metadata is setup to do lookups by resource id. You would have to craft a custom query on the database to look up by storage account name.

In the future when we have landing zones, we can amend the API to allow passing in some id for a landing zone storage account. AFAIK, the landing zone APIs have not been designed. At least I have not seen a doc.

Happy to have a short chat if this is not clear.

cahrens: Got it, I will change to the storage account resource ID.

service/src/test/java/bio/terra/workspace/common/fixtures/ControlledResourceFixtures.java
```diff
@@ -65,7 +43,7 @@ public class ControlledResourceFixtures {
 
   public static final UUID WORKSPACE_ID = UUID.fromString("00000000-fcf0-4981-bb96-6b8dd634e7c0");
   public static final UUID RESOURCE_ID = UUID.fromString("11111111-fcf0-4981-bb96-6b8dd634e7c0");
-  public static final UUID DATA_REFERENCE_ID =
```
cahrens, line 68: This variable was unused.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
```diff
@@ -47,950 +36,1010 @@
...
 import com.azure.resourcemanager.compute.models.VirtualMachine;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
 import java.time.Duration;
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
-import org.junit.jupiter.api.Test;
```
cahrens, line 55: This file was recently reformatted to 4 spaces, which is not the google-java-format standard. I have reformatted it, and I recommend that you choose "hide whitespace" for the review.

openapi/src/parts/controlled_azure_storage_container.yaml
cahrens: I have changed both the storage account and storage container request attributes to use the same name as what is returned in the creation response.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/CreateAzureStorageContainerStep.java
cahrens: I'm no longer using the list method, and doing everything in try/catches.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/DeleteAzureStorageContainerStep.java
cahrens: Fixed.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/GetAzureStorageContainerStep.java
cahrens: Fixed.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/GetAzureStorageContainerStep.java
cahrens: I have improved the exception logging to be more specific.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/GetAzureStorageContainerStep.java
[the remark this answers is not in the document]
cahrens: Have switched to get.

openapi/src/parts/controlled_azure_storage.yaml
```diff
@@ -31,9 +31,9 @@ components:
         Storage-specific properties to be set on creation. These are a subset of the values
         accepted by the azure resource API
       type: object
-      required: [ name, region ]
+      required: [ storageAccountName, region ]
       properties:
-        name:
+        storageAccountName:
```
cahrens, line 36: Changed to be consistent with the response.

marctalbott (approved):

cahrens:
@ddietterich I'm going to go ahead and merge so I can finish up the Rawls side of things. If you see things that you think should be changed, I'm happy to address them in a follow-up PR.

PR #669, opened 2022-05-26
[PF-1672] Fix broken RevokePetUsagePermissionStep undo

zloery:
(small bug I noticed while adding test coverage)

Currently, the undo step of RevokePetUsagePermissionStep tries to re-grant the same permission it revokes in the do step. However, when this flight is triggered by one user removing another from the workspace, this step’s undo fetches the caller's pet SA, not the removed user’s pet, which breaks the undo.

This bug could affect real users, but it’s difficult to trigger, as the RemoveUserFromWorkspace flight would need to fail on or after the second-to-last step. This functionality is also not available in any WSM environments above dev, so it's unlikely this has affected any real users yet. If it affected a real user, I think they could fix it by calling the enablePet endpoint, though it would be pretty hard for them to know to do that.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
+++ b/service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
@@ -80,7 +80,9 @@ public class PetSaService {
    *
    * @param workspaceUuid ID of the workspace to enable pet SA in
    * @param userToEnableEmail The user whose proxy group will be granted permission.
-   * @param userReq Auth info for calling SAM.
+   * @param userReq Auth info for calling SAM. Do not use userReq.getEmail() here; it will return
+   *     the caller's email, but there's no guarantee whether that will be an end-user email or a
+   *     pet SA email.
    * @param eTag GCP eTag which must match the pet SA's current policy. If null, this is ignored.
    * @return The new IAM policy on the user's pet service account, or empty if the eTag value
    *     provided is non-null and does not match current IAM policy on the pet SA.
--- a/service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
+++ b/service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
@@ -54,12 +54,12 @@ public class PetSaService {
    * particular GCP eTag value.
    */
   public Policy enablePetServiceAccountImpersonation(
-      UUID workspaceUuid, String userToEnableEmail, String token) {
+      UUID workspaceUuid, String userToEnableEmail, AuthenticatedUserRequest userRequest) {
     // enablePetServiceAccountImpersonationWithEtag will only return an empty optional if the
     // provided eTag does not match current policy. Because we do not use eTag checking here, this
     // is always nonempty.
     return enablePetServiceAccountImpersonationWithEtag(
-            workspaceUuid, userToEnableEmail, token, null)
+            workspaceUuid, userToEnableEmail, userRequest, null)
         .orElseThrow(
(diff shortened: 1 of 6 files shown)


service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -79,27 +80,41 @@ public Policy enablePetServiceAccountImpersonation(
...
-            () ->
-                samService.getOrCreatePetSaEmail(
-                    gcpCloudContextService.getRequiredGcpProject(workspaceUuid), token),
-            "enablePet");
+      UUID workspaceUuid,
+      String userToEnableEmail,
+      AuthenticatedUserRequest userReq,
+      @Nullable String eTag) {
+    String projectId = gcpCloudContextService.getRequiredGcpProject(workspaceUuid);
+    ServiceAccountName petSaName = getUserPetSa(projectId, userToEnableEmail, userReq).orElse(null);
```
melissachang, line 94: Leave petSaName as Optional and on line 97, call isEmpty()?

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -79,27 +80,41 @@ public Policy enablePetServiceAccountImpersonation(
...
+    // If the pet SA does not exist, there's no way the eTag can match, so return empty optional.
```
melissachang, line 95: Should this comment be on line 108 instead of here?

melissachang (approved):

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
```diff
@@ -79,27 +80,41 @@ public Policy enablePetServiceAccountImpersonation(
...
-      UUID workspaceUuid, String userToEnableEmail, String token, @Nullable String eTag) {
-    String petSaEmail =
-        SamRethrow.onInterrupted(
-            () ->
-                samService.getOrCreatePetSaEmail(
-                    gcpCloudContextService.getRequiredGcpProject(workspaceUuid), token),
-            "enablePet");
+      UUID workspaceUuid,
+      String userToEnableEmail,
+      AuthenticatedUserRequest userReq,
```
ddietterich, line 91: nit: I can't tell if this change to passing `userReq` is aesthetic or functional. It doesn't matter; just curious. The comment on line 83 is maybe out of date as a result.

ddietterich (approved):
LGTM - nit question to satisfy my curiousity :)

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
zloery: restructured comments to clarify

service/src/main/java/bio/terra/workspace/service/petserviceaccount/PetSaService.java
zloery: mostly aesthetic, we generally pass `AuthenticatedUserReqest` objects around instead of tokens directly.

the only functional concern is avoiding potential confusion with `userReq.getEmail()`, which will be the caller's email (but might be a pet email instead of an end user email, while `userToEnableEmail` and `petSaName` are guaranteed to be a user and pet email, respectively). Added more to the javadoc to help prevent a future mistake like that

ddietterich: SGTM, thanks!

PR #671, opened 2022-06-01
[PF-1690] Handle invalid http methods better

zloery:
Currently WSM will respond with a 500 error (generated by the catchall error handler) if a user calls a valid endpoint with an invalid HTTP method (e.g. `DELETE /status`). I think a 400 is a more appropriate response here.


jaycarlton (approved):

ddietterich (approved):
LGTM

PR #673, opened 2022-06-01
[PF-1691] sort the resources to clone

jaycarlton:
UX will be improved by sorting the resources by stewardship type, with referenced resources first. This code will break as soon as we have a third stewardship type, so maybe there's a more robust way to state what I want.

I tried testing this by looking at the ordering of the output details, but it turns out these are backed by a map at runtime and the sort order is not maintained in that structure. It would be expensive to build a test harness just for this feature, so we'll need to test it manually.


jaycarlton:
@cbookg FYI

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/FindResourcesToCloneStep.java
```diff
@@ -41,12 +45,18 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
+          batch.stream().filter(FindResourcesToCloneStep::isCloneable).toList();
       cloneableResources.forEach(
           r -> result.add(new ResourceWithFlightId(r, context.getStairway().createFlightId())));
     } while (batch.size() == limit);
-    context.getWorkingMap().put(ControlledResourceKeys.RESOURCES_TO_CLONE, result);
 
+    // sort the resources by stewardship type reversed, so reference types go first
+    result.sort(
+        Comparator.comparing(
+            r -> r.getResource().getStewardshipType().toString(), Comparator.reverseOrder()));
```
cbookg, line 56: if you didn't want to rely on on the enum names / order not changing, perhaps you could make a comparator function that just returns 1 for reference, and -1 for anything else (or something similar to meet the contract)?

yuhuyoyo: does the order for other stewardship type besides referenced (if there are multiple matters)? we can also just order by enum values and add a comment that it shouldn't be swapped.

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/FindResourcesToCloneStep.java
jaycarlton: If there were multiple, that might matter. I tend to avoid dealing with enum ordinals, though that's a valid way to do it too.

Note that we'll eventually maybe do the processing in parallel, so that even if we do referenced types first, we might start the controlled types before the last ref type is finished.

zloery (approved):
LGTM, though we'll probably need to revisit this if/when we add a third stewardship type

PR #676, opened 2022-06-07
[PF-1559] fix spotless & run 🏃🏻 

jaycarlton:
Remove an over-optimization that made `spotlessApply` and its ilk rarely actually run. The first file is the only manual change.

### DIFF
--- a/service/gradle/taskDependencies.gradle
+++ b/service/gradle/taskDependencies.gradle
@@ -12,11 +12,6 @@ for (task in TASKS_ONLY_IF_COMPILE_JAVA) {
   task.dependsOn(compileJava.inputs.files)
   // If compileJava runs, this must run after it (for the following to work).
   task.mustRunAfter(compileJava)
-  // Don't run if compileJava didn't do any work. This is necessary because otherwise, these
-  // tasks always want to run.
-  task.configure {
-    onlyIf { compileJava.didWork }
-  }
 }
 
 spotbugsTest.dependsOn(compileTestJava.inputs.files)
--- a/service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/CreateAzureStorageStepTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/CreateAzureStorageStepTest.java
@@ -1,5 +1,10 @@
 package bio.terra.workspace.service.resource.controlled.cloud.azure.storage;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+
 import bio.terra.stairway.StepResult;
 import bio.terra.workspace.common.fixtures.ControlledResourceFixtures;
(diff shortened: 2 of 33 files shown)


ddietterich:
Can we make an entry in the spotless config [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/f8a77ab6e8d3de120fa243918726bf8b55697acb/buildSrc/src/main/groovy/terra-workspace-manager.java-conventions.gradle#L53) to skip the generated code?

jaycarlton:
> Can we make an entry in the spotless config [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/f8a77ab6e8d3de120fa243918726bf8b55697acb/buildSrc/src/main/groovy/terra-workspace-manager.java-conventions.gradle#L53) to skip the generated code?

Great idea. That looks like the ticket.

ddietterich (approved):
LGTM

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/CreateAzureStorageStepTest.java
```diff
@@ -1,5 +1,10 @@
 package bio.terra.workspace.service.resource.controlled.cloud.azure.storage;
 
+import static org.hamcrest.MatcherAssert.assertThat;
```
cahrens, line 3: @jaycarlton This import ordering doesn't seem to be enforced by `google-java-format`, as when I reformat wit `google-java-format` in IntelliJ I don't get these changes. Therefore it seems that I need to run spotless locally also. What is the command to do that, and do you know if there is a way to set up IntelliJ's formatting to automatically use spotless?

jaycarlton: To run spotless, do `./gradlew spotlessApply` after this is merged.

It's possible to set up IntelliJ so that it runs `spotlessApply` every time a file is saved.  @melissachang is the only one I know who's done that.

Also, import ordering seems to work for me when I let IJ handle all the imports. I.e. I never manually add an import statement, and they seem to go to the right place FWIW.

yuhuyoyo (approved):

service/gradle/taskDependencies.gradle
```diff
@@ -12,11 +12,6 @@ for (task in TASKS_ONLY_IF_COMPILE_JAVA) {
   task.dependsOn(compileJava.inputs.files)
   // If compileJava runs, this must run after it (for the following to work).
   task.mustRunAfter(compileJava)
-  // Don't run if compileJava didn't do any work. This is necessary because otherwise, these
```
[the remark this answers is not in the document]
jaycarlton: Yeah, one fix was to make `spotlessApply`, etc, actually run, and the other was to avoid running it on generated code.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/CreateAzureStorageStepTest.java
melissachang: Have Intellij run spotless on file save:

https://github.com/diffplug/spotless/issues/200#issuecomment-1063519752

PR #679, opened 2022-06-08
[PF-1706] Minimize polling time for referenced resource creation

jaycarlton:
The referenced resources were taking 10-20 seconds each, and almost all of that time was spent waiting in `waitForFlight()`, which does a sleep first and then checks the flight state. Created an exponential backoff polling function and put it in `FlightUtils`. It could be added to `Stairway` I suppose.

Also deleted an unused step class.


jaycarlton:
@cbookg FYI

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCreateReferenceResourceFlightStep.java
```diff
@@ -59,7 +60,8 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
         FlightUtils.validateRequiredEntries(
             context.getWorkingMap(), ControlledResourceKeys.DESTINATION_REFERENCED_RESOURCE);
         final FlightState subflightState =
-            context.getStairway().waitForFlight(flightId, FLIGHT_POLL_SECONDS, FLIGHT_POLL_CYCLES);
+            FlightUtils.waitForFlightExponential(context.getStairway(), flightId,
+                Duration.ofMillis(10), Duration.ofMinutes(5));
```
ddietterich, line 64: Oooh, 10ms is very short. Does it ever complete the first time through?

ddietterich (approved):
LGTM - curious about the very short initial wait

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCreateReferenceResourceFlightStep.java
jaycarlton: Probably not. The fastest I've seen this flight go is 36 ms. My requirement from product is for this to be "instantaneous", so I went as low as I figured I could go.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
```diff
@@ -4,20 +4,25 @@
...
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import org.springframework.http.HttpStatus;
 
 /** Common methods for building flights */
 public final class FlightUtils {
 
-  public static final int FLIGHT_POLL_SECONDS = 10;
+  public static final int FLIGHT_POLL_SECONDS = 1;
```
melissachang, line 25: A bunch of places uses FLIGHT_POLL_SECONDS and FLIGHT_POLL_CYCLES. If you decrease FLIGHT_POLL_SECONDS, you should increase FLIGHT_POLL_CYCLES.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
```diff
@@ -164,4 +169,27 @@ public static <T> T getRequired(FlightMap flightMap, String key, Class<T> tClass
     }
     return value;
   }
+
```
melissachang, line 172: optional - I don't think this is needed. The natural place for this is next to Stairway.waitForFlight(). PF-1706 mentions 20 seconds. If you can get it down to 1 sec, that seems good enough.

Did you test this PR? How long did cloning take?

melissachang (approved):

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
melissachang: Could you leave these and just use a different number for the one flight you care about?

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
jaycarlton: One second is too long, because it sleeps first before checking the flight state. The flight only takes 35-50 ms. We want all of the reference clones to take place virtually instantaneously b/c we don't want another spinner in the UI.
Screen Shot 2022-06-09 at 12 30 04 PM

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
jaycarlton: The polling waits the whole interval for the first go-round, and thus every sub-flight was taking a minimum of ten seconds. These add up during clone, which uses many sub-flights, so I wanted to keep this change.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCreateReferenceResourceFlightStep.java
jaycarlton: Bumping it to 50 ms.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
jaycarlton: I didn't want to poll too frequently in the case where it takes longer, so I like the fact that it spreads out over time.

I'll put this in Stairway after giving it some run time in WSM.

PR #681, opened 2022-06-10
[WOR-274] create SAS tokens

marctalbott:
Ticket: [WOR-274](https://broadworkbench.atlassian.net/browse/WOR-274)
* add endpoint to create SAS tokens for `ControlledAzureStorageContainer` resources


service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -21,36 +21,25 @@
...
+import org.broadinstitute.dsde.workbench.client.sam.model.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Component;
 
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.*;
```
ddietterich, line 40: Convention is not to have `*` imports

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -181,6 +187,84 @@ private void cleanUpSharedWorkspace() {
...
+    OffsetDateTime startTime = OffsetDateTime.now().minusMinutes(15);
+    OffsetDateTime expiryTime = OffsetDateTime.now().plusHours(1);
+    String sas =
+        controlledResourceService.createAzureStorageContainerSasToken(
+            workspaceUuid,
+            storageContainer.getResourceId(),
+            startTime,
+            expiryTime,
+            user.getAuthenticatedRequest());
+    assertThat("sas is formatted correctly", sas.startsWith("sv"));
```
marctalbott, line 255: This isn't exactly a comprehensive assertion that we're getting a valid token. Do people think this is fine? Maybe we add in a direct call to Azure to try and use it? This also might be better suited as an integration test than a connected test.

service/src/test/java/bio/terra/workspace/common/fixtures/ControlledResourceFixtures.java
```diff
@@ -29,14 +29,11 @@
...
+import javax.annotation.Nullable;
 import java.time.OffsetDateTime;
 import java.time.format.DateTimeFormatter;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.UUID;
-import javax.annotation.Nullable;
+import java.util.*;
```
ddietterich, line 36: ditto

ddietterich (commented):
Took a quick look to see how it came out. Looks good.
Suggest you set your IntelliJ so it doesn't make `*` imports.

openapi/src/parts/controlled_azure_storage.yaml
```diff
@@ -24,6 +24,28 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+
+  /api/workspaces/v1/{workspaceId}/resources/controlled/azure/storageContainer/{resourceId}/getSasToken:
```
cahrens, line 28: I would put this in controlled_azure_storage_container.yaml instead given that the URL path includes `storageContainer`.

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -167,6 +170,39 @@ public ResponseEntity<ApiAzureRelayNamespaceResource> getAzureRelayNamespace(
     return new ResponseEntity<>(resource.toApiResource(), HttpStatus.OK);
   }
 
+  @Override
+  public ResponseEntity<ApiCreatedAzureStorageContainerSasToken>
+      createAzureStorageContainerSasToken(UUID workspaceUuid, UUID storageContainerUuid) {
+    features.azureEnabledCheck();
+
+    final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    final String userEmail =
```
cahrens, line 179: Why do we get the userEmail? Is it because we really want to log the email, or because we are using the Sam call as a permissions check?

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
cahrens: Would it be worth then using the sas token to access the storage container and do some basic operation like creating a BlobClient and calling the `exists` method?

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
marctalbott: Just to get the `userEmail`. Logging the requesting user was called out in the ticket and the email wasn't available in the `AuthenticatedUserRequest` during my testing. It is a bit of a shame to have to make an extra call to Sam just to log the email so I'd be happy to pull it out if logging the user email is a soft requirement

cahrens: Hmmm, it seems worth checking into this requirement. Would it be possible to log something else available in the user request that would allow us to retrieve the email from Sam if necessary (if required to debug a user issue or investigate suspicious activity)?

aherbst-broad (approved):

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
marctalbott: Do we need additional logging that WSM's Apache proxy doesn't give us? It seems like that would log all the same information we're logging here except for expiry time on the successful requests

aherbst-broad: @marctalbott I'm not sure -- what identifier does the proxy log that will let us trace back to the calling principal?

marctalbott: @aherbst-broad I'm not sure what it looks like for Azure, but for Google users we log email and Google subject id

It would look something like this for the SAS token endpoint
```
... 106960538915967416130 "marc.testbott@gmail.com" "-" [2022-06-13T14:34:42.987-0400] 3423867 + "POST /api/workspaces/v1/{workspace_id}/resources/controlled/azure/storageContainer/{resource_id}/getSasToken HTTP/1.1" 200 ...
```

cahrens: I am fine with leaving the Sam call in for now. If it ends up causing pain we can take a look at what is available in the Apache proxy for Azure users and remove the Sam call if possible at that time.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureControlledStorageResourceServiceTest.java
```diff
@@ -0,0 +1,273 @@
...
+  @Autowired private AzureTestConfiguration azureConfiguration;
+
+  /**
+   * Retrieve or create a workspace ready for controlled Azure storage resources.
+   *
+   * <p>Reusing a workspace saves time between tests.
+   *
+   * @param user
+   */
+  private Workspace reusableWorkspace(UserAccessUtils.TestUser user) throws Exception {
```
cahrens, line 76: I think you could use the `BeforeAll` annotation to create this instance that is then shared across test cases.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureControlledStorageResourceServiceTest.java
```diff
@@ -0,0 +1,273 @@
...
+    workspace = reusableWorkspace(workspaceOwner);
+    resetSecondUserWorkspaceAccess(workspace.getWorkspaceId());
+  }
+
+  /**
+   * Reset the {@link FlightDebugInfo} on the {@link JobService} to not interfere with other tests.
+   */
+  @AfterEach
+  public void resetFlightDebugInfo() {
+    // Exercise enumeration by dumping after each
```
cahrens, line 206: I don't understand lines 207-209. Do you have any details about why we need to so these after every test?

cahrens (approved):

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureControlledStorageResourceServiceTest.java
marctalbott: Good call, that's exactly what I want. I used `BeforeEach`, `AfterEach`, and `AfterAll` and completely forgot about `BeforeAll` 🤦

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureControlledStorageResourceServiceTest.java
```diff
@@ -162,6 +80,89 @@ private void resetSecondUserWorkspaceAccess(UUID workspaceUuid) throws Exception
...
+    UUID workspaceUuid =
+        workspaceService.createWorkspace(
+            Workspace.builder()
+                .workspaceId(UUID.randomUUID())
+                .userFacingId("a" + UUID.randomUUID().toString())
+                .spendProfileId(spendUtils.defaultSpendId())
+                .workspaceStage(WorkspaceStage.MC_WORKSPACE)
+                .build(),
+            workspaceOwner.getAuthenticatedRequest());
+    AzureControlledStorageResourceServiceTest.workspace =
```
cahrens, line 101: Can this just be `workspace` in the same manner as in `AfterAll`?

PR #682, opened 2022-06-14
[PF-1738] add an activity log table and keep track of change activities in a workspace.

yuhuyoyo:
In this PR, the change activities we are logging are:
- create/update/delete workspace
- create/delete cloud context
- create/update/delete controlled and referenced resources

We do not cover yet:
- add/remove user from workspace
- enable/disable application
- clone workspace
- clone resources
- Background clean up ophan private resources

All the activity log is logged in activity_log table. Currently there's only three column: workspace_id, change_date, change_type. The change_type is generic: CREATE, UPDATE, DELETE. 

In the follow-on PR, we will add WorkspaceChangeDetails, ResourceChangeDetails, CloudContextChangeDetails columns with type jsonb so that we can distinguish what exactly the change activity is.


yuhuyoyo:
Hi Melissa, the PR still needs some clean up but I'm adding you for a first look of the table and DAO.

service/src/main/resources/db/changesets/20220609_activity_log_table.yaml
```diff
@@ -0,0 +1,25 @@
+databaseChangeLog:
+- changeSet:
+    id: workspace activity log table
+    author: yuhuyoyo
+    changes:
+    - createTable:
+        tableName: workspace_activity_log
```
melissachang, line 7: Rename file to 20220609_workspace_activity_log_table.yaml

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -112,7 +118,21 @@ public String createAzureRelayNamespace(
...
 
-    String jobId = jobBuilder.submit();
+    String jobId =
+        jobBuilder.submitWithCallback(
+            ControlledAzureRelayNamespaceResource.class,
+            new FutureCallback<>() {
+              @Override
+              public void onSuccess(
+                  JobResultOrException<ControlledAzureRelayNamespaceResource> result) {
+                if (result.getException() == null) {
```
melissachang, line 128: Instead of modifying all these Service classes, I think it's cleaner to just modify WorkspaceDao.java. In createWorkspace/updateWorkspace, you call activityLogDao.setWorkspaceUpdateDate().

[Here's an example ](https://github.com/DataBiosphere/jade-data-repo/blob/8a36c374c4b4a50f8851534d07c2ac1d5f818da8/src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java#L68)of a Dao injecting another Dao.

The service approach is fragile. If someone adds a new service, they might forget to set workspace update date.

@ddietterich thoughts?

service/src/test/java/bio/terra/workspace/db/ActivityLogDaoTest.java
```diff
@@ -0,0 +1,56 @@
...
+  public void setUpdateDateAndGet() {
+    String workspaceId = UUID.randomUUID().toString();
+    Instant now = Instant.now();
+    activityLogDao.setUpdateDate(workspaceId, now);
+    var latestDate = activityLogDao.getLastUpdateDate(workspaceId);
+    assertEquals(now, latestDate);
+  }
+
+  @Test
+  public void getLatestDate() {
```
melissachang, line 25: getLastUpdatedDate

melissachang (commented):
Nice!

In PR description, it's useful for me to see manual testing. Can you do things via Swagger, then show the contents of `workspace_activity_log` table?

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
yuhuyoyo: ah i was not sure if Dao calling Dao was an anti-pattern.

But one issue with doing it in DAO is that it's more noisy especially in the case of stairway undo. E.g. during resource creation, we first store metadata in the resource db, and then create a gcp resources. If it fails, we delete the resource in the db. If we log in the resourceDao, we will have two updates in this failure flight but the end result is that nothing is updated.  https://github.com/DataBiosphere/terra-workspace-manager/blob/6f8d928494244f5464f7c588b87429080fe9d923/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/create/StoreMetadataStep.java#L12

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -7,6 +7,7 @@
 import bio.terra.stairway.FlightMap;
 import bio.terra.workspace.common.utils.MdcHook;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
+import bio.terra.workspace.service.job.JobService.JobResultOrException;
```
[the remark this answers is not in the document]
yuhuyoyo: there are some "async" calls for create and delete that we immediately return a jobid but the job may fail later. i don't want to do it in the flight because then i need to delete the entry in undostep(). and i think it's easier to forget. 

So I added a submitWithCallback so that we can control whether to log an update depending on the job result. @ddietterich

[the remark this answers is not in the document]
ddietterich: There is nothing wrong with your submitWithCallback code, but I don't think it is the behavior we want.
If the pod fails, the operation happens, but your activity doesn't get logged.

My intended (but not implemented) solution to this was to add a special completion step to a flight. So after the flight is done - either failed or succeeded - Stairway would run the completion step. It would re-run the step until the completion step completed (including the usual retries, etc.)  That is also the key feature to enable notification instead of polling for async REST API calls.

So one option is, you could go do that! 😸

Another options would be something like writing the activity row with a start time but not an end time. Then when you get the callback from the flight, update the row with the end time (and maybe a success/fail indication).

Is there a reason you decided not to record failed activity?

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
ddietterich: The problem with Dao calling Dao is making sure transactions are handled properly. Spring injects a transaction layer around every public call with an `@*Transaction` annotation. I feel strongly the code should not nest methods with those annotations. I know that it works, but it has led to confusion in reviewing and debugging code, so I think it is best avoided.

Is it an option to make a "Dao" that just had methods on the activity table, but those were always called from a "real" Dao with the `@*Transaction` marks?

service/src/main/resources/db/changesets/20220609_activity_log_table.yaml
yuhuyoyo: since we are having one activity log table. keeping it as activity_log

ddietterich (changes requested):
There are issues.
AND I have an idea for a totally different approach. See what you think: we setup hooks for Stairway events [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/c8236b45c091bbd1b7e5258bcd891a5bba8b818c/service/src/main/java/bio/terra/workspace/service/job/JobService.java#L172). We could add an activity logging hook. Whenever a flight completed, the `endFlight` hook gets called. We would look at the flight context to see what flight it was and if it succeeded. In that case, we could write the activity log. The flight context holds the input parameters and the working map, so I think we could gather most everything we need for the logging. We would could create a static mapping of:   this.flight.class -> this.activity.type (and other info).

That approach would minimize the change to the code base, but would require ongoing additions as new things were created. The big advantage is that the hook is called for all flights. It would be one change and cover both the sync and async cases.

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -116,29 +126,80 @@ public JobBuilder addParameter(String keyName, @Nullable Object val) {
...
    */
-  public String submit() {
+  public <T> String submit(
+      @Nullable ActivityLogChangedType changeType, @Nullable Class<T> resultClass) {
     populateInputParams();
-    return jobService.submit(flightClass, jobParameterMap, jobId);
+    if (changeType == null) {
+      return jobService.submit(flightClass, jobParameterMap, jobId);
+    }
+    return jobService.submitWithCallback(
```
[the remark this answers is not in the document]
yuhuyoyo: submitwithCallback wait for the flight to complete on a separate thread though. it doesn't wait for the job to complete before returning the jobid.

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -38,11 +43,16 @@ public class JobBuilder {
   @Nullable private StewardshipType stewardshipType;
   @Nullable private OperationType operationType;
 
-  public JobBuilder(JobService jobService, StairwayComponent stairwayComponent, MdcHook mdcHook) {
+  public JobBuilder(
```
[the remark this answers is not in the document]
yuhuyoyo: i do intend to encapsulate all the activity change info into one class (which will have the ActivityloggedChangeType enum as a field as well) so we don't pass a million parameter.

yuhuyoyo:
cool, i need to go learn about stairway hook. i like it.

Perhaps can it be a generic acitivty log hook so we can avoid needing to write new hooks when new resources/other activities are added?

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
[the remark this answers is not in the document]
ddietterich: Yes, of course you are right. I went and read the documentation...

yuhuyoyo:
I checked what's in the flightcontext. 

For this PR, we can get workspaceid, operationtype so it works easily.

For future, if we want to fill in the details of the change activity: 
- For resources change, we can infer by fetching resourceid, resourcetype and other details..
- For cloud context, we can infer by fetching GCP_CLOUD_CONTEXT or AZURE_CLOUD_CONTEXT key
- otherwise, they are workspace change.

If we want to add a column for userId, we need to check AUTH_USER_INFO. I'm not sure if all the flight has this set, if not we need it to be mendatory.

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -116,29 +126,80 @@ public JobBuilder addParameter(String keyName, @Nullable Object val) {
...
    * Submit a job to stairway and return the jobID immediately.
    *
+   * @param changeType Specify the type of {@link ActivityLogChangedType} to be logged.
+   * @param resultClass type for the response of the flight. When null, nothing is returned from the
+   *     flight.
    * @return jobID of submitted flight
    */
-  public String submit() {
+  public <T> String submit(
+      @Nullable ActivityLogChangedType changeType, @Nullable Class<T> resultClass) {
```
melissachang, line 135: It looks like you're using `changeType = null` to mean "We don't need activity log".

Instead of null, can you use optional?

If changeType is null/Optional.empty(), don't call dao.setChangedDate

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -116,29 +126,80 @@ public JobBuilder addParameter(String keyName, @Nullable Object val) {
...
+        resultClass,
+        jobParameterMap,
+        jobId,
+        new FutureCallback<>() {
+          @Override
+          public void onSuccess(JobResultOrException<T> result) {
+            if (changeType.equals(ActivityLogChangedType.DELETE)) {
+              activityLogDao.setChangedDate(workspaceId, changeType);
+            } else {
+              if (result.getException() == null) {
```
melissachang, line 151: I know this code is moving, these comments apply to new place

// Only log to activity log if flight succeeded.

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -116,29 +126,80 @@ public JobBuilder addParameter(String keyName, @Nullable Object val) {
...
+   * @param activityLogChangedType type of change activity to be logged when the job finishes. When
+   *     null, nothing is logged.
    * @return Result of the finished job.
    */
   @Traced
-  public <T> T submitAndWait(Class<T> resultClass, boolean doAccessCheck) {
+  public <T> T submitAndWait(
+      Class<T> resultClass,
+      boolean doAccessCheck,
+      @Nullable ActivityLogChangedType activityLogChangedType) {
```
melissachang, line 175: activityLogChangeType

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -116,29 +126,80 @@ public JobBuilder addParameter(String keyName, @Nullable Object val) {
...
-        flightClass, jobParameterMap, resultClass, jobId, doAccessCheck);
+    try {
+      T result =
+          jobService.submitAndWait(flightClass, jobParameterMap, resultClass, jobId, doAccessCheck);
+      activityLogDao.setChangedDate(workspaceId, activityLogChangedType);
+      return result;
+    } catch (Exception e) {
+      if (activityLogChangedType != null
+          && activityLogChangedType.equals(ActivityLogChangedType.DELETE)) {
+        // DELETE job cannot be undone. so even when there is failure, we need to set a changed date
```
melissachang, line 185: If deletion flight failed, then deletion *maybe* happened, right?

In the DB, I'd like to distinguish between "Deletion definitely happened" and "Deletion maybe happened". Can you log DELETE is job succeeded, and DELETE_MAYBE if job failed?

I don't think we should log DELETE if there's a chance deletion didn't happen.

service/src/main/java/bio/terra/workspace/service/privateresource/PrivateResourceCleanupService.java
```diff
@@ -162,7 +162,7 @@ private void runCleanupFlight(WorkspaceUserPair workspaceUserPair) {
             .addParameter(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, null);
     try {
       // Skip the job access check as the "manager" role does not grant read access to a workspace.
-      userCleanupJob.submitAndWait(null, /*doAccessCheck=*/ false);
+      userCleanupJob.submitAndWait(null, /*doAccessCheck=*/ false, null);
```
melissachang, line 165: nit it's nice to have a comment, similar to the existing `/*doAccessCheck=/` comment. It's hard to tell what `null` means.

```
/* Don't need to update activity log */ null
```

service/src/main/java/bio/terra/workspace/db/ActivityLogDao.java
```diff
@@ -0,0 +1,58 @@
...
+  private final RowMapper<Instant> CHANGE_DATE_ROW_MAPPER =
+      (rs, rowNum) -> rs.getObject("change_date", OffsetDateTime.class).toInstant();
+
+  @Autowired
+  public ActivityLogDao(NamedParameterJdbcTemplate jdbcTemplate) {
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  @WriteTransaction
+  public void setChangedDate(String workspaceId, @Nullable ActivityLogChangedType changeType) {
```
melissachang, line 30: Could this take UUID instead of String? That would guarentee it's a valid UUID.

service/src/main/java/bio/terra/workspace/db/ActivityLogDao.java
```diff
@@ -0,0 +1,58 @@
...
+            + " VALUES (:workspace_id, :change_date, :change_type)";
+    final var params =
+        new MapSqlParameterSource()
+            .addValue("workspace_id", workspaceId)
+            .addValue("change_date", Instant.now().atOffset(ZoneOffset.UTC))
+            .addValue("change_type", changeType.name());
+    jdbcTemplate.update(sql, params);
+  }
+
+  @ReadTransaction
```
melissachang, line 45: // Returns null if there are no activity logs for workspaceId

Though usually we use Optional instead of null.

service/src/test/java/bio/terra/workspace/db/ActivityLogDaoTest.java
```diff
@@ -0,0 +1,65 @@
...
+  @Autowired private NamedParameterJdbcTemplate jdbcTemplate;
+
+  @Test
+  public void setWorkspaceUpdatedDateAndGet() {
+    String workspaceId = UUID.randomUUID().toString();
+    assertNull(activityLogDao.getLastChangedDate(workspaceId));
+
+    activityLogDao.setChangedDate(workspaceId, ActivityLogChangedType.CREATE);
+
+    String changeType = getChangeType(workspaceId);
```
melissachang, line 28: I don't think lines 28 and 31 are necessary. Generally we just test what methods return.

service/src/test/java/bio/terra/workspace/db/ActivityLogDaoTest.java
```diff
@@ -0,0 +1,65 @@
...
+    var latestDate = activityLogDao.getLastChangedDate(workspaceId);
+    assertNotNull(latestDate);
+    assertEquals(ActivityLogChangedType.CREATE.name(), changeType);
+  }
+
+  @Test
+  public void getLastUpdatedDate() {
```
melissachang, line 35: opt (optional) nit - I think first and second is enough, I don't think third adds anything.

melissachang (commented):
> We do not cover yet:
> 
> * add/remove user from workspace
> * enable/disable application
> * clone workspace
> * clone resources
> * Background clean up ophan private resources

You plan on doing clone resource for PF-1738, right?


> In the follow-on PR, we will add WorkspaceChangeDetails, ResourceChangeDetails, CloudContextChangeDetails columns with type jsonb so that we can distinguish what exactly the change activity is.

These could go in a single ChangeDetails column. The ChangeDetails column could have WorkspaceChangeDetails or ResourceChangeDetails, etc.

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
melissachang: It would be nice if a failed delete job could indicate if resource was deleted or not. (Do a get on the resource.) DELETE_MAYBE would be easier...

service/src/main/java/bio/terra/workspace/db/model/ActivityLogChangedType.java
```diff
@@ -0,0 +1,7 @@
+package bio.terra.workspace.db.model;
+
+public enum ActivityLogChangedType {
```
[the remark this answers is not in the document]
yuhuyoyo: irrelevant now

service/src/main/resources/db/changesets/20220609_activity_log_table.yaml
```diff
@@ -0,0 +1,31 @@
...
+          A table for tracking editing activities in the workspace
+        columns:
+        - column:
+            name: workspace_id
+            type: text
+            constraints:
+              nullable: false
+            remarks: |
+              The ID of the workspace
+        - column:
```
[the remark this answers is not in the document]
yuhuyoyo: we are going to have a detail column for different type of resources. for example, for resource change, the resourceChangeDetail (or something like that) column will contains a resource jsonb, the workspaceChangeDetail will contains a workspace jsonb, the cloudcontextChangedetail will contains a cloudcontext jsonb. 

So I don't want to add these two columns in this PR which will be repetitive later.

yuhuyoyo (commented):
the changes happen in the new PR https://github.com/DataBiosphere/terra-workspace-manager/runs/6990976338?check_suite_focus=true

PR #683, opened 2022-06-14
[PF-548] Pass request ID in header for RBS calls

zloery:
This change adds the `X-Request-Id` header in calls from WSM to RBS. Both services use the common logging code from TCL, so when RBS receives a request with this header it will associate the request ID with logs for that request. This lets us track requests across services, which is great for debugging.


ddietterich (approved):
LGTM

jaycarlton (approved):

PR #686, opened 2022-06-15
[PF-1736] Implement new TERRA_WORKSPACE resource

melissachang:
Did some manual testing via Swagger.

### DIFF
--- a/ADDING_RESOURCES.md
+++ b/ADDING_RESOURCES.md
@@ -14,24 +14,25 @@ to compile your code as you go.
      attributes objects.
    - Edit the `resource_type.yaml` file in that directory, creating an enumeration for the
      resource and adding the appropriate resource element to the union objects.
-   - Run `gradle :openapi:buildOpenApi` to make sure your syntax and references are correct.
-1. Create your Java resource package in `service/src/main/java/bio.terra/workspace/service/resources/`
+   - Run `./gradlew :openapi:buildOpenApi` to make sure your syntax and references are correct.
+2. Create your Java resource package in `service/src/main/java/bio.terra/workspace/service/resources/`
       in the appropriate package
     - `controlled/{cloud}/{your resource}`
     - `referenced/{cloud}/{your resource}`
-1. Create classes for your resource
+3. Create classes for your resource
    - Attributes class describing the fields beyond the common metadata
    - Resource class. This is the most complex class to implement and supplies the bulk of
      your resource code
-1. If you are making a controlled resource you will also need to add:
+   - Handler class
+4. If you are making a controlled resource you will also need to add:
    - Step classes as required for object create and update. These will be needed to
      complete the resource class.
    - Handler class. Used primarily to give the `ResourceDao` a way to create resource
      objects by type.
-1. Add your new resource type to `resource/model/WsmResourceFamily` and
-   `resource/model/WsmResourceType`. At this point, you should be able to cleanly build
-   the `service` Workspace Manager sub-project.
-1. Add your REST API
+5. Add your new resource type to `resource/model/WsmResourceFamily`,
+   `resource/model/WsmResourceType` and `ApiResourceType`. At this point,
+   `./gradlew :service:compileJava` should succeed.
+6. Add your REST API
    - In the `openapi` sub-project, in the `parts` directory, create a yaml file for your
      resource API. The file should contain all resource-type-specific parameters,
      responses, and schemas as well as the endpoint paths. You can find shared components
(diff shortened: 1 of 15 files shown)


integration/src/main/java/scripts/utils/TestUtils.java
```diff
@@ -0,0 +1,11 @@
+package scripts.utils;
+
+import java.util.Random;
+
+public class TestUtils {
+  private static final Random RANDOM = new Random();
+
+  public static String appendRandomNumber(String string) {
```
yuhuyoyo, line 8: since there's only one usage of it, i think we should keep this in the caller's class for now.

openapi/src/parts/referenced_terra_workspace.yaml
```diff
@@ -0,0 +1,94 @@
+paths:
+  /api/workspaces/v1/{workspaceId}/resources/referenced/terra/workspaces:
```
yuhuyoyo, line 2: curious about why we are adding the terra prefix. what workspaces are we trying to distinguish from?

integration/src/main/resources/configs/integration/ReferencedTerraWorkspaceLifecycle.json
```diff
@@ -0,0 +1,20 @@
+{
```
ddietterich, line 1: This test will not get run automatically until you add an entry into the `FullIntegration` test suite for the environments where you want it tested. I wonder if we have other "orphan" tests.

openapi/src/parts/referenced_terra_workspace.yaml
ddietterich: True. That doesn't seem necessary in the API. If we were being strict about REST, the object would be a "terraworkspace" not a "terra/workspace", since there isn't a "terra" collection.

I like it (or something!) in the code as a way to distinguish a referenced workspace from "this" workspace.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -920,6 +923,60 @@ public ResponseEntity<ApiCloneReferencedGitRepoResourceResult> cloneGitRepoRefer
     return new ResponseEntity<>(result, HttpStatus.OK);
   }
 
+  // - Terra workspace referenced resource - //
+  @Override
+  public ResponseEntity<ApiTerraWorkspaceResource> createTerraWorkspaceReference(
```
ddietterich, line 928: We really need another controller for non-cloud-based references. This will be the third, I think after TDR snapshot and git repo. I made [PF-1749](https://broadworkbench.atlassian.net/browse/PF-1749) to track.

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceType.java
```diff
@@ -113,6 +115,13 @@ public enum WsmResourceType {
       ApiResourceType.GIT_REPO,
       ReferencedGitRepoResource.class,
       ReferencedGitRepoHandler::getHandler),
+  REFERENCED_TERRA_WORKSPACE(
+      CloudPlatform.ANY,
+      StewardshipType.REFERENCED,
+      "REFERENCED_TERRA_WORKSPACE",
```
ddietterich, line 121: To follow the naming convention, this should be `REFERENCED_ANY_TERRA_WORKSPACE`

ddietterich (approved):
Nice piece of work!
I think the REST API needs a tweak and the integration test needs to be added to the test suite so it gets run.
Beyond that, LGTM

integration/src/main/java/scripts/utils/TestUtils.java
melissachang: I'd prefer to leave it here. If it exists, people are more likely to use it, rather than manually write this code themselves ([example](https://github.com/DataBiosphere/terra-cli/blob/5780f4bc06115e4768b2e9296882d9022b756485/src/test/java/unit/CloneWorkspace.java#L183)).

openapi/src/parts/referenced_terra_workspace.yaml
melissachang: No other workspace. I put this in the "terra" directory/namespace, similar to the azure namespace. There are a bunch of resources under `/api/workspaces/v1/{workspaceId}/resources/controlled/azure`. In the future, there could be other Terra resources under `/api/workspaces/v1/{workspaceId}/resources/referenced/terra`.

ADDING_RESOURCES.md
```diff
@@ -14,24 +14,25 @@ to compile your code as you go.
      attributes objects.
    - Edit the `resource_type.yaml` file in that directory, creating an enumeration for the
      resource and adding the appropriate resource element to the union objects.
-   - Run `gradle :openapi:buildOpenApi` to make sure your syntax and references are correct.
-1. Create your Java resource package in `service/src/main/java/bio.terra/workspace/service/resources/`
+   - Run `./gradlew :openapi:buildOpenApi` to make sure your syntax and references are correct.
```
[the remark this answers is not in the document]
melissachang: I prefer `1.`. Intellij chagned these unfortunately. I did some playing around. If I add a second `5.`, Intellij does the right thing. There are some situations where it doesn't do the right thing.

integration/src/main/resources/configs/integration/ReferencedTerraWorkspaceLifecycle.json
melissachang: Oops, done. :)

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceType.java
melissachang: Done

PR #687, opened 2022-06-16
[PF-1476] Include destination resource ID on job enumerate object

jaycarlton:
In order to inform the UI about the resources being created by cloning, we need to give it the destination resource ID via the Enumerate Jobs API. This involves pre-computing all the IDs and passing them through a couple levels of flights so that they're available in the flight input parameters later when the enumeration happens and we hydrate all the extra goodies like resource metadata and attributes.

My original thought was to include a key-value map instead of an individual parameter, but that's inviting abuse. We can tolerate a null value for destination resource id on the non-clone jobs.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
@@ -598,6 +598,7 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
             .cloneReferencedResource(
                 sourceReferencedResource,
                 body.getDestinationWorkspaceId(),
+                UUID.randomUUID(), // resourceId is not pre-allocated for individual clone endpoints
                 body.getName(),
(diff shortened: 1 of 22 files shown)


zloery (approved):
LGTM!

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -598,6 +598,7 @@ public ResponseEntity<ApiCloneReferencedGcpGcsObjectResourceResult> cloneGcpGcsO
             .cloneReferencedResource(
                 sourceReferencedResource,
                 body.getDestinationWorkspaceId(),
+                UUID.randomUUID(), // resourceId is not pre-allocated for individual clone endpoints
```
ddietterich, line 601: nit: Would it be worth making a variant of this endpoint that fills in the UUID, instead of generating one at every use?

ddietterich (approved):
LGTM - one nit question

PR #688, opened 2022-06-21
[Abandoned] TOAZ-118 [WSM] Add integration test for createVM API

sergiygetlin:
This PR is abandoned due to merge issues. New PR is https://github.com/DataBiosphere/terra-workspace-manager/pull/718

TOAZ-118 [WSM] Add integration test for createVM API

-create vm with private ip and custom script extension (main scenario for Jupyter VM)
-create vm with private ip
-create vm with public ip
-create vm with incorrect credentials. Expected failure.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/ControlledAzureVmNoPublicIpLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/ControlledAzureVmNoPublicIpLifecycle.java
@@ -2,7 +2,6 @@ package scripts.testscripts;
 
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
-import bio.terra.workspace.api.ControlledAzureResourceApi;
 import bio.terra.workspace.client.ApiException;
 import bio.terra.workspace.model.AzureVmCreationParameters;
 import bio.terra.workspace.model.CreateControlledAzureVmRequestBody;
@@ -16,7 +15,6 @@ import scripts.utils.ControlledAzureVmTestScriptBase;
 public class ControlledAzureVmNoPublicIpLifecycle extends ControlledAzureVmTestScriptBase {
   @Override
   protected CreatedControlledAzureVmResult createVm(
-      ControlledAzureResourceApi azureApi,
       String resourceSuffix,
       String createVmJobId,
       CreatedControlledAzureDisk disk,
--- a/integration/src/main/java/scripts/testscripts/ControlledAzureVmNoPublicIpLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/ControlledAzureVmNoPublicIpLifecycle.java
@@ -1,0 +1,44 @@
+package scripts.testscripts;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+import bio.terra.workspace.api.ControlledAzureResourceApi;
+import bio.terra.workspace.client.ApiException;
+import bio.terra.workspace.model.AzureVmCreationParameters;
+import bio.terra.workspace.model.CreateControlledAzureVmRequestBody;
+import bio.terra.workspace.model.CreatedControlledAzureDisk;
+import bio.terra.workspace.model.CreatedControlledAzureNetwork;
+import bio.terra.workspace.model.CreatedControlledAzureVmResult;
+import bio.terra.workspace.model.JobControl;
+import scripts.utils.ControlledAzureVmTestScriptBase;
+
+/** Creates VM with private IP and without Custom Script Extension. */
+public class ControlledAzureVmNoPublicIpLifecycle extends ControlledAzureVmTestScriptBase {
+  @Override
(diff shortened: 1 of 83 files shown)


integration/src/main/java/scripts/utils/ControlledAzureVmTestScriptBase.java
```diff
@@ -0,0 +1,232 @@
...
+
+  protected abstract CreatedControlledAzureVmResult createVm(
+      ControlledAzureResourceApi azureApi,
+      String resourceSuffix,
+      String createVmJobId,
+      CreatedControlledAzureDisk disk,
+      CreatedControlledAzureNetwork network)
+      throws ApiException;
+
+  protected CreatedControlledAzureIp createPublicIp(
```
zloery, line 113: Would it make sense to pull these method implementations (createPublicIp, createNetwork, etc.) into a utils class? If we're writing additional Azure tests in the future I'd imagine we'll probably want these bits of functionality, even if we aren't making a VM.

Waiting until we actually have additional uses also seems fine too

integration/src/main/resources/configs/integration/ControlledAzureVmNoPublicIpLifecycle.json
```diff
@@ -0,0 +1,23 @@
+{
```
zloery, line 1: As a heads-up, these tests will be runnable locally but won't be included as part of the automated nightly/PR integration test runs unless you add them to the FullIntegration test suite (`integration/src/main/resources/suites/FullIntegration.json`), not sure if you want these tests included there or not. I also don't know if Azure is enabled in the WSM nightly test environment (`wsmtest`), which would be another issue.

integration/src/main/java/scripts/testscripts/ControlledAzureVmNoPublicIpLifecycle.java
```diff
@@ -0,0 +1,44 @@
...
+            .vmSize("Standard_D8s_v3")
+            .diskId(disk.getResourceId())
+            .networkId(network.getResourceId())
+            .vmImage(createAzureVmImage())
+            .vmUser(createUser());
+    vmRequestBody.azureVm(vmCreationParameters);
+    CreatedControlledAzureVmResult vmCreateResult =
+        azureApi.createAzureVm(vmRequestBody, getWorkspaceId());
+    assertNotNull(vmCreateResult);
+    return vmCreateResult;
```
zloery, line 42: Is it worth validating the newly-created VM here (e.g. verifying the IP is private and there's no custom script extension for this test, other assertions for the other tests)?

integration/src/main/java/scripts/testscripts/ControlledAzureVmWithWrongCredentialsLifecycle.java
```diff
@@ -0,0 +1,67 @@
...
+    CreatedControlledAzureVmResult vmCreateResult =
+        azureApi.createAzureVm(vmRequestBody, getWorkspaceId());
+    assertNotNull(vmCreateResult);
+    return vmCreateResult;
+  }
+
+  @Override
+  protected void validateVm(CreatedControlledAzureVmResult vmCreateResult) {
+    assertEquals(JobReport.StatusEnum.FAILED, vmCreateResult.getJobReport().getStatus());
+    assertNull(vmCreateResult.getAzureVm());
```
zloery, line 57: I'm not sure what `vmCreateResult.getErrorReport` will have here, is there a way to verify this is failing for the right reason?

zloery (commented):
Sorry for the delay getting to this!

integration/src/main/java/scripts/utils/ControlledAzureVmTestScriptBase.java
sergiygetlin: I think we can wait until we see real use cases for this. Anyway, all these methods should be still in this class since it's base class for all VM scenarios, but they can re-use some default behavior from util class.

integration/src/main/resources/configs/integration/ControlledAzureVmNoPublicIpLifecycle.json
sergiygetlin: That's good point. Thank you. Let me clarify it.

integration/src/main/java/scripts/testscripts/ControlledAzureVmWithWrongCredentialsLifecycle.java
sergiygetlin: Added additional validation.

sergiygetlin:
This PR is abandoned because of some merge issues. New PR is https://github.com/DataBiosphere/terra-workspace-manager/pull/718New PR is https://github.com/DataBiosphere/terra-workspace-manager/pull/718

PR #689, opened 2022-06-21
PF-1738: create a hook for logging activities when a flight is finished

yuhuyoyo:


yuhuyoyo:
i haven't added comprehensive testing yet. sending out the draft for a first look for this approach. @ddietterich @melissachang

service/src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -171,6 +171,11 @@ public void initialize() {
             .context(flightBeanBag)
             .addHook(mdcHook)
             .addHook(new TracingHook())
+            .addHook(
+                new WorkspaceActivityLogHooks(
+                    flightBeanBag.getActivityLogDao(),
```
ddietterich, line 176: Since JobService is a `@Component`, you can `@Autowire` these instead of using the bean bag. The bag is intended for use by flights. The hook itself is a singleton.

service/src/main/java/bio/terra/workspace/service/job/WorkspaceActivityLogHooks.java
```diff
@@ -0,0 +1,124 @@
...
+    UUID workspaceUuid = UUID.fromString(workspaceId);
+    if (context.getFlightStatus() == FlightStatus.SUCCESS) {
+      activityLogDao.writeActivity(
+          workspaceUuid, new DbWorkspaceActivityLog().operationType(operationType));
+      return HookAction.CONTINUE;
+    }
+    // For DELETE flight, even if the flight failed, database deletion cannot be undone, so it's
+    // still deleted from the database and we should log that as a changed activity.
+    if (operationType == OperationType.DELETE) {
+      if (isWorkspaceDeleted(workspaceUuid)) {
```
melissachang, line 68: Let's make this future-proof. If someone adds a new Delete flight, they're not going to remember to update this.

Here's a suggestion that adds logging, and also makes this more modular.

```
if (operationType == OperationType.DELETE) {
  if flight is WorkspaceDeleteFlight
    maybeLogWorkspaceDeletion()
  else if flight is DeleteGcpContext or DeleteAzureContext
    maybeLogCloudContextDeletion
  else if flight is DeleteControlledResourceFlight
    maybeLogControlledResourceFlightDeletion
  else
    throw "Unknown deletion flight: %s"


maybeLogWorkspaceDeletion() {
  if (isWorkspaceDeleted()) {
    logger.info("Writing workspace {} deletion to activity log");
  } else {
    logger.info("Workspace {} not deleted; not writing deletion to activity log");
  }
}
```

service/src/main/resources/db/changesets/20220617_workspace_activity_log_table.yaml
```diff
@@ -0,0 +1,31 @@
...
+        remarks: |
+          A table for logging change activities in the workspace
+        columns:
+        - column:
+            name: workspace_id
+            type: text
+            constraints:
+              nullable: false
+            remarks: |
+              The ID of the workspace
```
melissachang, line 17: Workspace UUID

(Nice to mention UUID, since workspaces also have user-facing ID)

service/src/main/java/bio/terra/workspace/service/job/WorkspaceActivityLogHooks.java
```diff
@@ -0,0 +1,124 @@
+package bio.terra.workspace.service.job;
```
melissachang, line 1: This hook isn't related to jobs. How about move to bio.bio.terra.worskpace.common.utils, where MdcHook is?

service/src/main/java/bio/terra/workspace/db/WorkspaceActivityLogDao.java
```diff
@@ -0,0 +1,59 @@
...
+
+  @Autowired
+  public WorkspaceActivityLogDao(NamedParameterJdbcTemplate jdbcTemplate) {
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  @WriteTransaction
+  public void writeActivity(UUID workspaceId, DbWorkspaceActivityLog dbWorkspaceActivityLog) {
+    switch (dbWorkspaceActivityLog.getOperationType()) {
+      case UNKNOWN:
```
ddietterich, line 29: I'm confused. You went to the trouble of passing these in, but you are not using them?

service/src/test/java/bio/terra/workspace/service/job/JobServiceLogTest.java
```diff
@@ -0,0 +1,192 @@
...
+    var nullChangedDate = activityLogDao.getLastChangedDate(workspaceUuid);
+    assertNull(nullChangedDate);
+
+    runFlight("a creation flight", workspaceUuid, OperationType.UNKNOWN);
+    var changedDate = activityLogDao.getLastChangedDate(workspaceUuid);
+    assertNull(changedDate);
+  }
+
+  @Test
+  void setChangedActivityWhenFlightCreateFails_notSet() {
```
melissachang, line 112: Add test for deletion flight fails cases?

ddietterich (commented):
I like this approach. A few code comments.
I'm puzzled by passing in operation type, but not storing it.

service/src/test/java/bio/terra/workspace/service/job/JobServiceLogTest.java
yuhuyoyo: more tests are added for logging that doesn't happen in the flight.

service/src/main/java/bio/terra/workspace/common/utils/WorkspaceActivityLogHook.java
```diff
@@ -0,0 +1,155 @@
...
+
+  private boolean isControlledResourceDeleteFlight(String className) {
+    return DELETE_CONTROLLED_RESOURCE_FLIGHT.equals(className);
+  }
+
+  private void maybeLogWorkspaceDeletionFlight(String flightClassName, UUID workspaceUuid) {
+    checkArgument(isWorkspaceDeleteFlight(flightClassName));
+    try {
+      workspaceDao.getWorkspace(workspaceUuid);
+      logger.info(
```
melissachang, line 108: Let's make these warning instead of info

service/src/main/java/bio/terra/workspace/db/model/DbWorkspaceActivityLog.java
```diff
@@ -0,0 +1,23 @@
+package bio.terra.workspace.db.model;
```
melissachang, line 1: optional

Hmm I think there was an earlier thread where you created this so you don't have to pass in a bunch of parameters. But this just adds an extra step for callers. Callers have to list parameters regardless, and with DbWorkspaceActivityLog, they have to create a DbWorkspaceActivityLog. 

Without DbWorkspaceActivityLog:

Every caller passes in changeType

With DbWorkspaceActivityLog

Every caller creates DbWorkspaceActivityLog, passes in changeType
Every caller calls writeActivity with DbWorkspaceActivityLog

Removing DbWorkspaceActivityLog would also make the code easier to read.

With:

        workspaceActivityLogDao.writeActivity(
            workspaceUuid, new DbWorkspaceActivityLog().operationType(OperationType.UPDATE));

Without:

        workspaceActivityLogDao.writeActivity(workspaceUuid, OperationType.UPDATE);

If you keep DbWorkspaceActivityLog, can you take out validating OperationType, since that's validated in writeActivity()?

(Also, on line 21 it's weird to create an Optional that isn't used. You could just throw if operationType isn't set.)

service/src/test/java/bio/terra/workspace/common/utils/WorkspaceActivityLogHookTest.java
```diff
@@ -0,0 +1,161 @@
...
+    assertTrue(emptyChangedDate.isEmpty());
+
+    runFlight("a creation flight", workspaceUuid, OperationType.DELETE);
+
+    var changedDate = activityLogDao.getLastUpdatedDate(workspaceUuid);
+    assertTrue(changedDate.isPresent());
+  }
+
+  @Test
+  void unknownFlightSucceeds_activityLogNotUpdated() {
```
melissachang, line 82: I thought we throw for UNKNOWN?

flightWithUnknownOperationType_throws

opt Dao is the place that throws for UNKNOWN operationType. Test for "throw for UNKNOWN" only needs to be in Dao Test, it doesn't need to be in this test also.

service/src/test/java/bio/terra/workspace/common/utils/WorkspaceActivityLogHookTest.java
```diff
@@ -0,0 +1,161 @@
...
+    assertTrue(emptyChangedDate.isEmpty());
+
+    runFlight("Failed flight with operation type CREATE", workspaceUuid, OperationType.CREATE);
+
+    var changedDateAfterFailedFlight = activityLogDao.getLastUpdatedDate(workspaceUuid);
+    assertTrue(changedDateAfterFailedFlight.isEmpty());
+  }
+
+  @Test
+  void deleteFlightFails_unknownDeleteFlight_activityLogNotUpdated() {
```
melissachang, line 109: opt. deleteFlightFails_resourceNotDeleted_activityLogNotUpdated

service/src/test/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceServiceTest.java
```diff
@@ -127,6 +129,9 @@ void updateDataRepoReferenceTarget_updateSnapshotIdOnly() {
     ReferencedResource updatedResource =
         originalResource.toBuilder().snapshotId(newSnapshotId).build();
 
+    var lastUpdatedDate = workspaceActivityLogDao.getLastUpdatedDate(workspaceUuid);
```
melissachang, line 132: opt lastUpdatedDateBeforeResourceUpdate. (same below)

melissachang (commented):
Nice

ddietterich (commented):
I like this approach in general. I would prefer an enum lookup instead of passing OperationType into the jobs. But I'm OK with moving forward with this approach and making that change in the next pass.

service/src/test/java/bio/terra/workspace/common/utils/WorkspaceActivityLogHookTest.java
yuhuyoyo: this test uses jobService.waitForJob which doesn't re-throw flight. I made operationtype required not though so this will throw early at jobbuilder.

service/src/main/java/bio/terra/workspace/common/utils/WorkspaceActivityLogHook.java
```diff
@@ -0,0 +1,155 @@
...
+
+@Component
+public class WorkspaceActivityLogHook implements StairwayHook {
+  private static final Logger logger = LoggerFactory.getLogger(StairwayHook.class);
+
+  private final WorkspaceActivityLogDao activityLogDao;
+  private final WorkspaceDao workspaceDao;
+  private final ResourceDao resourceDao;
+
+  private static final String DELETE_WORKSPACE_FLIGHT = WorkspaceDeleteFlight.class.getName();
```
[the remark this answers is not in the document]
yuhuyoyo: will follow-up in the next iteration shortly

ddietterich (approved):
I am going to approve this.

I am uncomfortable with merging in code that we are going to fix later. It feels like a guaranteed way to leave tech debt in the code base. I'll raise this with Greg and Sean.

One thing that would help is explicitly adding TODO comments referencing Jira tickets everywhere that you have left a loose end or plan to change something. Then at least we could find these spots to fix them later and track them for sprint planning.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -426,6 +434,9 @@ public void grantWorkspaceRole(
                   email.toLowerCase()));
       logger.info(
           "Granted role {} to user {} in workspace {}", role.toSamRole(), email, workspaceUuid);
+      activityLogDao.writeActivity(
```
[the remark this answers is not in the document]
yuhuyoyo: fair point. it's currently called directly from `WorspaceApiController`. Should I log there?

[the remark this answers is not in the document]
ddietterich: I think that would be better.

service/src/test/java/bio/terra/workspace/common/utils/WorkspaceActivityLogHookTest.java
```diff
@@ -0,0 +1,169 @@
...
+          .subjectId("StairwayUnit")
+          .email("stairway@unit.com")
+          .token(Optional.of("not-a-real-token"));
+
+  @Autowired private JobService jobService;
+  @Autowired private WorkspaceDao workspaceDao;
+  @Autowired private WorkspaceActivityLogDao activityLogDao;
+  @MockBean private SamService mockSamService;
+
+  @BeforeEach
```
melissachang, line 41: Delete 41-49. I'm deleting these in JobServiceTest in #694.

45 is throwing an exception because the mock isn't set up correctly. The exception message has details on how it's incorrect.

ddietterich:
> Yay!
> 
> > I am going to approve this.
> > I am uncomfortable with merging in code that we are going to fix later. It feels like a guaranteed way to leave tech debt in the code base. I'll raise this with Greg and Sean.
> 
> You're talking about PF-1800 right? Yu is doing that in next PR>

Yes, but not specifically. I am concerned that these incremental milestones have us implementing things halfway and prototyping UI support in production code.

PR #690, opened 2022-06-22
[PF-1777] Change to use TCL database configuration

ddietterich:
This PR is preparing for integration of TPS into WSM. I want to put the database configuration for TPS into the TPS library, but there was a conflict in WSM when I included the `db` part of TCL for TPS. This changes WSM to use the TCL configuration for databases.

There are two other drive-by fixes:
1. Zach removed the slf4j dependencies from the gradle build file. Turns out that while spring dependencies control the versions, we still needed the dependencies. That removes a set of warnings from the build.
2. I bumped the version of spotbugs up to one that stops using APIs that are deprecated in Java17. That removes another set of warnings.


service/src/main/java/bio/terra/workspace/app/configuration/external/StairwayDatabaseConfiguration.java
```diff
@@ -1,14 +1,14 @@
...
-import bio.terra.workspace.app.configuration.BaseDatabaseConfiguration;
+import bio.terra.common.db.BaseDatabaseProperties;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
 @EnableConfigurationProperties
 @ConfigurationProperties(prefix = "workspace.stairway-database")
-public class StairwayDatabaseConfiguration extends BaseDatabaseConfiguration {
```
zloery, line 11: These are the only two classes using `BaseDatabaseConfiguration`, delete it?

jaycarlton (approved):

service/gradle/dependencies.gradle
```diff
@@ -43,6 +43,8 @@ dependencies {
   implementation group: "org.apache.commons", name: "commons-dbcp2"
   implementation group: "org.apache.commons", name: "commons-lang3"
   implementation group: "org.apache.commons", name: "commons-pool2"
+  implementation group: "org.apache.logging.log4j", name: "log4j-api"
+  implementation group: "org.apache.logging.log4j", name: "log4j-to-slf4j"
```
zloery, line 47: interesting - I don't see any changes to the actual log4j dependencies we're pulling in in the lockfile so I don't know what this is changing. where was the error you saw showing up?

zloery (approved):

service/src/main/java/bio/terra/workspace/app/configuration/external/StairwayDatabaseConfiguration.java
ddietterich: Yes, will do!

jaycarlton: Wait, are you saying two uses isn't enough to warrant a base class?

ddietterich: Not at all. Instead of using the WSM base class, this changes to using the Terra Common Lib base class. So the sharing is more not less.

The issue was that when I scanned the TCL `db` package, it created conflicts with the WSM base class. By using the TCL base class, I remove that conflict and am able to have a database config contained in my TPS library.

PR #695, opened 2022-06-27
[PF-1748] In ListWorkspaces, don't return data source workspaces for readers

melissachang:
Tested in swagger, confirmed that data source workspace isn't returned (when requester is reader).

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
@@ -27,6 +27,7 @@ import bio.terra.workspace.service.workspace.flight.create.azure.CreateAzureCont
 import bio.terra.workspace.service.workspace.model.AzureCloudContext;
 import bio.terra.workspace.service.workspace.model.OperationType;
 import bio.terra.workspace.service.workspace.model.Workspace;
+import bio.terra.workspace.service.workspace.model.WorkspaceConstants;
 import io.opencensus.contrib.spring.aop.Traced;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -178,16 +179,33 @@ public List<Workspace> listWorkspaces(
...
+        .filter(WorkspaceService::removeDataSourceWorkspacesIfReader)
         .collect(Collectors.toList());
   }
 
+  // WSM stores data sources (such as AMP-PD) as workspaces. (This is an implementation detail;
+  // Terra users are not aware that data sources are workspaces.) An AMP-PD researcher would not
+  // expect to see AMP-PD workspace in their workspace list. So hide data source workspaces, only
+  // for readers. Editors and owners need to see AMP-PD workspace, so they can edit it.
+  @Traced
+  private static boolean removeDataSourceWorkspacesIfReader(Workspace workspace) {
```
ddietterich, line 198: Access level should be orthogonal to visibility of data sources (which are now called datasets?). This would prevent someone with read access to the dataset from listing datasets they can read. It doesn't make sense to me.

We should be listing data sources if we are asked to list them and not list them by default.
I think we should add a query param to the listWorkspaces to control what workspace types the caller gets in the list.

service/src/main/java/bio/terra/workspace/service/workspace/model/WorkspaceConstants.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.workspace.service.workspace.model;
+
+public class WorkspaceConstants {
+  // For now, denote data source workspaces with this key/value. This may change, as product becomes
+  // more settled.
+  // "terra-" prefix means this key is used by Terra internally (as opposed to used by user).
+  public static final String DATA_SOURCE_KEY = "terra-type";
```
ddietterich, line 7: Product is now calling these datasets. I think we should follow suit in the code.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
cbookg: I generally agree with the logic, in that its probably what we want in the short term.

But I think this application-specific magic logic that WSM shouldn't be concerned with.  This seems better suited to a higher layer, that can apply filters as query params or on the result.

This change only works if we add a distinct listDatasets API, when we could instead add params to listWorkspaces.  I feel like the later is more in the spirit of keeping the WSM API agnostic to what a "workspace" contains.

melissachang: Discussed offline, agree this shouldn't be in WSM, even in short term. Closing PR.

PR #698, opened 2022-06-28
[PF-548] Add cross-service tracing for RBS and Sam calls

zloery:
(I'll hold off merging this until after RBS changes in https://github.com/DataBiosphere/terra-resource-buffer/pull/242)

This pulls in the newest TCL changes (https://github.com/DataBiosphere/terra-common-lib/pull/74) and adds cross-service tracing for calls from WSM to RBS and Sam. Additionally, this change means WSM will look for B3 trace headers in incoming requests instead of traceContext format headers, though I don't think this matters since nothing is sending these to WSM yet.

I saw newer versions of the Sam client had trace header population as well but I wasn't able to get that working here, so I went with the interceptor approach instead.

Example from my development environment:
Before cross-service tracing:
oldtrace

After:
newtrace


jaycarlton (approved):

ddietterich (approved):
Very nice

PR #700, opened 2022-06-29
Fix dependencies in new Sam client

rtitle:
This is a PR to `mariayaz-managedidentity-assign` branch, feel free to merge or not. :) 

Needed to upgrade `okhttp` dependency after upgrading the Sam client post-https://github.com/broadinstitute/sam/pull/714.


service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -254,7 +254,8 @@ public void createWorkspaceWithDefaults(AuthenticatedUserRequest userRequest, UU
     CreateResourceRequestV2 workspaceRequest =
         new CreateResourceRequestV2()
             .resourceId(uuid.toString())
-            .policies(defaultWorkspacePolicies(humanUserEmail));
+            .policies(defaultWorkspacePolicies(humanUserEmail))
+            .authDomain(List.of());
```
rtitle, line 258: authDomain is required apparently (Sam returns 400 without it).

service/gradle/dependencies.gradle
```diff
@@ -19,6 +19,9 @@ dependencies {
   }
   implementation group: "org.broadinstitute.dsde.workbench", name: "sam-client_2.13", version: "0.1-4962144-SNAP"
   implementation group: "bio.terra", name: "terra-resource-buffer-client", version: "0.4.3-SNAPSHOT"
+  // Required by Sam client
+  implementation group: "com.squareup.okhttp3", name: "okhttp", version: "4.9.1"
```
rtitle, line 23: WSM was pulling in 3.x before… not exactly sure from where.

zykonda (approved):
Thanks!

PR #705, opened 2022-07-01
PF-1791: clone workspace copy the property

ginaygoogle:


integration/src/main/java/scripts/utils/WorkspaceAllocateTestScriptBase.java
```diff
@@ -72,11 +74,18 @@ protected void doSetup(List<TestUserSpecification> testUsers, WorkspaceApi works
    */
   protected CreatedWorkspace createWorkspace(
       UUID workspaceUuid, String spendProfileId, WorkspaceApi workspaceApi) throws Exception {
+    Properties properties = new Properties();
```
yuhuyoyo, line 77: can you extract that as a public static field

yuhuyoyo: you also need to add the test in CloneWorkspace.java to assert that the properties is the same after cloning

integration/src/main/java/scripts/utils/WorkspaceAllocateTestScriptBase.java
```diff
@@ -28,6 +30,13 @@ public abstract class WorkspaceAllocateTestScriptBase extends WorkspaceApiTestSc
   private UUID workspaceUuid;
   private String spendProfileId;
 
+  public static void makeProperties(Properties properties) {
```
yuhuyoyo, line 33: sorry we have some miscommunication here.

i was thinking you have a static field DEFAULT_PROPERTIES here and you will assert that in `Cloneworkspace`. Your assertion is better though. Can you remove this makeProperties method and just set the properties in createWorkspace?

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -336,6 +336,9 @@ public String cloneWorkspace(
     String jobDescription =
         String.format("Clone workspace: name: '%s' id: '%s'  ", workspaceName, workspaceUuid);
 
+    destinationWorkspace =
```
ddietterich, line 339: Instead of doing this here, we should do this in the controller where we are constructing the original destination workspace object.

In the current code base, `WorkspaceApiController.cloneWorkspace()` does not have the source workspace object. However, after Zach's code change goes in, it will. So I would suggest either waiting until Zach's change goes in and then reworking this so it is in the controller. Alternately, you could leave this as is with a TODO and make yet another ticket to fix it after Zach's change goes in.

service/src/main/java/bio/terra/workspace/service/workspace/model/Workspace.java
```diff
@@ -118,6 +118,23 @@ public int hashCode() {
         .toHashCode();
   }
 
+  /**
```
ddietterich, line 121: This won't be needed if we build the destinationWorkspace object in the controller.

ddietterich (changes requested):
This looks good. One change I would like. We can discuss the approach to that change.

yuhuyoyo (approved):

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -691,11 +691,15 @@ void createGoogleContextRawlsStageThrows() throws Exception {
...
     // Create a workspace
+    Map<String, String> propertyMap = new HashMap<>();
+    propertyMap.put("foo", "bar");
+    propertyMap.put("xyzzy", "plohg");
     final Workspace sourceWorkspace =
         defaultRequestBuilder(UUID.randomUUID())
             .userFacingId("source-user-facing-id")
             .displayName("Source Workspace")
             .description("The original workspace.")
+            .properties(propertyMap)
```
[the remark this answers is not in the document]
ginaygoogle: as we talked, the product code which in the controller should test by mocking and place in WorkspaceApiControllerTest.java.

ddietterich (approved):
LGTM, thanks!

PR #710, opened 2022-07-06
[PF-1665] Move all authZ checks to controller layer

zloery:
(Implementation of discussion in the "Relayering WSM" document)

At a high level, all this change does is move authorization checks up from the service and flight levels into the controller level. This makes the code significantly more readable and makes it easier to verify by inspection that each endpoint is actually enforcing authz, as well as saving unnecessary work (e.g. launching flights) that we did before authz checks. Additionally, this saves a number of round-trips to Sam: previously, some operations would call multiple service endpoints (e.g. `create workspace` would end with a call to `get workspace`) each of which performed an authz check with Sam. Now, since auth is performed at the controller layer, we don't need these duplicate checks.

This does not add or remove any capabilities from users, all endpoints still require the same authorization checks, they just happen earlier in the request's lifecycle. This change should be mostly invisible to users, with the exception of a few errors which may have different codes. 

Additionally, I used this as an opportunity to implement some mockMvc-based tests. These use Spring's MockMVC object to act as if we're making HTTP calls to a real server and test the full request flow including controller code from inside JUnit tests, which we could previously only test via integration tests.

Some notes from implementation:
- `validateWorkspaceAndAction` and the newly-added `validateMcWorkspaceAndAction` return a full Workspace object. Since we're now calling these from the controller, I've changed a number of service method signatures to take a Workspace object instead of just the workspace ID to reduce the number of duplicate DB reads required.
- A number of service methods only needed the userRequest object to make a single authz check. Now that we're making these checks in the controller, I've removed the userRequest object from service method signatures where possible.
- However, most create and delete flights still need the user request. Because they're making additional calls to Sam these still get passed around, but there's now an initial auth check at the beginning of handling the request.
- For endpoints which take both a workspace ID and job ID, JobService now actually verifies that the given job is in the given workspace. This also uncovered some unexpected behavior with clone jobs: jobs to clone individual resources live in the source workspace, while jobs to clone whole workspaces live in the destination workspace, PF-1825 tracking this
- The `enablePet` endpoint previously didn't verify that the provided workspace was an MC_WORKSPACE stage workspace, it now does.
- A handful of requests now return slightly different error messages: e.g. trying to add a user to a workspace without permission now results in a 404 SamResourceNotFound exception instead of a 403, because Sam intentionally returns 404 when a user does not have any role on a resource.


service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -128,6 +121,9 @@ public UUID createWorkspace(Workspace workspace, AuthenticatedUserRequest userRe
    * <p>Throws ForbiddenException if the user is not permitted to perform the specified action on
    * the workspace in question.
    *
+   * <p>Throws StageDisabledException if assertMcStage is true and this is not an MC_WORKSPACE stage
```
melissachang, line 124: I'm confused, this method doesn't deal with MC_WORKSPACE. validateMcWorkspaceAndAction...

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -271,6 +283,7 @@ public ResponseEntity<Void> grantRole(
       throw new InvalidRoleException(
           "Users cannot grant role APPLICATION. Use application registration instead.");
     }
+    // No additional authz check as this is just a wrapper around a Sam endpoint.
```
melissachang, line 286: For consistency, I think it would make sense to move auth check from SamService to here...

Same for other SamService calls

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -937,9 +1004,13 @@ public ResponseEntity<ApiCloneReferencedGitRepoResourceResult> cloneGitRepoRefer
...
   public ResponseEntity<ApiTerraWorkspaceResource> createTerraWorkspaceReference(
       UUID workspaceUuid, @Valid ApiCreateTerraWorkspaceReferenceRequestBody body) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    workspaceService.validateWorkspaceAndAction(
+        userRequest, workspaceUuid, SamWorkspaceAction.CREATE_REFERENCE);
     UUID referencedWorkspaceId = body.getReferencedWorkspace().getReferencedWorkspaceId();
 
-    // Will throw if workspace does not exist or workspace id is null.
+    // Will throw if the referenced workspace does not exist or workspace id is null.
+    // Note that the user does not need read access in the destination workspace to reference it.
```
zloery, line 1013: @melissachang just to confirm expected behavior: WSM should throw a 404 if the referenced workspace doesn't exist, but it's fine if the caller isn't a member of the referenced workspace (i.e. don't throw a 403 in that case)?

melissachang: Yes, that's what Yu and Dan settled on. Your comment changes look good, thanks for clarifying.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
zloery: ah, I missed the `checkAuthz` calls in SamService, good catch! 

The intended pattern here is that explicit auth checks are in the controller, but any call to Sam (like adding/removing a user) implicitly also checks authz.
Those authz checks were always redundant calls - there's no need to explicitly ask Sam if user A can add user B to the workspace, since the next call is "add user B to the workspace using user A's credentials", and `addUserToPolicy` will just throw an exception if the caller doesn't have permission.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
zloery: whoops, leftover comment from a previous approach

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -536,31 +506,16 @@ public void removeResourceRole(
...
-      AuthenticatedUserRequest userRequest,
-      ControlledResourceIamRole role,
-      String email)
+      ControlledResource resource, ControlledResourceIamRole role, String email)
       throws InterruptedException {
-    // Validate that the provided user credentials can modify the owners of the resource's
-    // workspace.
-    // Although the Sam call to revoke a resource role must use WSM SA credentials instead, this
-    // is a safeguard against accidentally invoking these credentials for unauthorized users.
-    checkAuthz(
```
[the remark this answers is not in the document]
zloery: The Sam call here is made as WSM's SA, but at the controller layer before we get to this method we've already verified that the user has the required permission so we don't need the additional check here.

We could keep a duplicate check here in case of developer mistake, but part of the driver for this refactor was giving us confidence that we don't need to because it's easier to see every endpoint is doing authorization checks by inspection.

service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * A collection of utilities and constants useful for MockMVC-based tests. This style of tests lets
+ * us test controller-layer code (request/response parsing, authz, and validation) without actually
+ * spinning up a local server.
+ *
+ * <p>TODO: it's probably worth looking into whether we can automatically pull routes from the
+ * generated swagger, instead of manually wrapping them here.
+ */
+@Component
+public class MockMvcUtils {
```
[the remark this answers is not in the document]
zloery: I think some endpoints will end up needing connected tests instead of just unit tests so I'm hesitant to include that in the name, and I'd lean towards `MockMvcUtils` over something like `ControllerTestUtils` since these are really testing endpoints, not just the controller code

zloery:
> Yay for MockMvc!
> 
> If @ddietterich is ok with this:
> 
> My preference is for new endpoints, by default, to have a controller test and not service test.
> 
> So a new WorkspaceService endpoint would have a test in WorkspaceApiControllerTest.java, not WorkspaceServiceTest.java.
> 
> Could you add a proof-of-concept controller test (WorkspaceApiControllerTest.getWorkspace ?)
> 
> Could you also add to DEVELOPMENT.md that new endpoints should have controller unit tests, not service unit tests, and point to your example? At some point, I hope to add a TOC to DEVELOPMENT.md, to make it more usable. :)

I added some mockMvc testing to existing service code, but good point that it's better to have a clear example to point to. Added a new test and a section in README.md (where we currently describe testing more).
I think having these tests for every endpoint would be great, but I wonder if some endpoints (especially thin wrappers around other services/cloud stuff) will really need to be connected tests instead

melissachang (approved):

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
ddietterich: I thought it would look silly if we did not test existence within our own database! Not testing access is consistent with the other referenced resources.

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/WorkspaceApiControllerTest.java
```diff
@@ -0,0 +1,101 @@
...
+                Mockito.any(),
+                eq(SamSpendProfileAction.LINK)))
+        .thenReturn(true);
+    Mockito.when(mockSamService.listRequesterRoles(Mockito.any(), Mockito.any(), Mockito.any()))
+        .thenReturn(List.of(WsmIamRole.OWNER));
+  }
+
+  @Test
+  public void getWorkspace() throws Exception {
+    Mockito.when(
```
ddietterich, line 63: Is this needed here? Looks like it is done in the setup.

ddietterich (approved):
This is a big lift! And a great improvement! Thanks for taking it on.
I made a number of comments, but no showstoppers.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
zloery: sgtm, I just wanted to clarify what was expected after I tripped over a few tests!

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -181,31 +206,37 @@ public ResponseEntity<ApiCreateControlledAzureRelayNamespaceResult> createAzureR
...
-
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    jobService.verifyUserAccess(jobId, userRequest, workspaceUuid);
     ApiCreateControlledAzureRelayNamespaceResult result =
-        fetchCreateControlledAzureRelayNamespaceResult(jobId, userRequest);
+        fetchCreateControlledAzureRelayNamespaceResult(jobId);
     return new ResponseEntity<>(result, getAsyncResponseCode(result.getJobReport()));
   }
 
-  @Override
```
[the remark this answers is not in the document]
zloery: It's actually just moved down next to the other `get` endpoints, I noticed it was out of place while modifying the file

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -104,6 +105,8 @@ public ResponseEntity<ApiCreatedWorkspace> createWorkspace(
         body.getId(),
         userRequest.getEmail(),
         userRequest.getSubjectId());
+    // Unlike other operations, there's no Sam permission required to create a workspace. As long as
```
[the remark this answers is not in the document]
zloery: We actually don't check billing profile permission until the user tries to create a cloud context (maybe we should?)

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApplicationApiController.java
```diff
@@ -49,8 +51,11 @@ public ResponseEntity<ApiWorkspaceApplicationDescription> disableWorkspaceApplic
       @PathVariable("workspaceId") UUID workspaceUuid,
       @PathVariable("applicationId") String applicationId) {
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    Workspace workspace =
```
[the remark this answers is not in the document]
zloery: (one rabbithole later...)
we implemented disabling as "remove the application SA's role in the workspace" and "disable the app in the DB", but we don't do anything with the disabled application's resources. Not sure if/when we're planning to change that

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
zloery: I'm actually fine with having controllers call SamService directly, it seems reasonable that controllers might need to call multiple different services.
The calls going directly from controller -> SamService are generally ones where WSM is really just wrapping Sam endpoints, so internally wrapping those in another service seems like more hassle than it's worth to me

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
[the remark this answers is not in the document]
ddietterich: Oh, duh. Yeah, I knew that...

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApplicationApiController.java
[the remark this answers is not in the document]
ddietterich: In the DAO, the `disableWorkspaceApplication` verifies that the application does not own any resources in the workspace.

PR #711, opened 2022-07-06
[WOR-377] Add Azure storage SAS URL to response

aherbst-broad:


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureControlledStorageResourceService.java
```diff
@@ -34,6 +34,8 @@
 @Component
 public class AzureControlledStorageResourceService {
 
+  public record AzureSasBundle(String sasToken, String sasUrl) {}
```
aherbst-broad, line 37: This is my solution to the "I want to return two values from a method" problem, rather than a `Pair<String, String>`. Open to other suggestions.

cahrens (approved):

PR #714, opened 2022-07-08
[PF-1817] Add lastUpdatedDate to workspace description

yuhuyoyo:
i'm piggyback on which is previously `WorkspaceAndHighestRole` and added `lastUpdatedDate` there. I'm planning to add CreatedDate in `WorkspaceAndAdditionalAttributes` as well in a following pr.

I'm putting the `WorkspaecAndAdditionalattributes` construction logic in workspaceService instead of workspaceapicontroller to force caller to get those instead of `workspace` and include these additional attributes in the workspace description. 

The other way is to have the construction logic in the controller but because we don't have unit test for the controller, testing it there is rather awkward and I don't like the controller to do the complicated thing.

### DIFF
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -494,6 +494,11 @@ components:
           $ref: '#/components/schemas/AzureContext'
         properties:
           $ref: '#/components/schemas/Properties'
+        lastUpdatedDate:
+          description: Lastest timestamp when the workspace is changed.
+          type: string
+          format: date-time
+
 
     WorkspaceDescriptionList:
       type: object
--- a/integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
@@ -3,7 +3,9 @@ package scripts.testscripts;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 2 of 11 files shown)


openapi/src/parts/workspace.yaml
```diff
@@ -494,6 +494,11 @@ components:
           $ref: '#/components/schemas/AzureContext'
         properties:
           $ref: '#/components/schemas/Properties'
+        lastUpdatedDate:
+          description: Lastest timestamp when the workspace is changed.
```
ddietterich, line 498: Suggest: Timestamp of the latest workspace change

ddietterich (approved):
lgtm

ddietterich:
> I defer to Dan on where to put this logic. My natural tendency is also to put this logic in lower level, Service, so all controller calls get the new fields automatically. Dan, just curious on why you prefer logic in controller?

Here is my take: I do not want to compute highest permission and latest update every time we build a `Workspace` object.
And I don't want to have sometimes-there-sometimes-not-there fields in the `Workspace` object. So I think fetching them should be separate. They are only consumed in the API, so making a form of get/list for the controller layer makes sense to me.

But maybe I am missing something. I'm not seeing unexpected logic in the controller.

yuhuyoyo:
> > I defer to Dan on where to put this logic. My natural tendency is also to put this logic in lower level, Service, so all controller calls get the new fields automatically. Dan, just curious on why you prefer logic in controller?
> 
> Here is my take: I do not want to compute highest permission and latest update every time we build a `Workspace` object. And I don't want to have sometimes-there-sometimes-not-there fields in the `Workspace` object. So I think fetching them should be separate. They are only consumed in the API, so making a form of get/list for the controller layer makes sense to me.
> 
> But maybe I am missing something. I'm not seeing unexpected logic in the controller.

I was originally motivated to do this because in the controller it seems that we fetches the highest role when converting to workspace description every time. I must have missed the conversation about where this should live...

PR #715, opened 2022-07-11
 PF-1790: update the property with new properties endpoint by POST

ginaygoogle:
Screen Shot 2022-07-11 at 11 02 53 AM
Screen Shot 2022-07-11 at 11 03 11 AM


ginaygoogle:
I have a question about the null key failure test. 
I am not sure whether we should add this failure unit test or not. 
If we should do it, I find my product code do not required the key not null, and just limit the Map<String, String> propertyMap is not null. 

it also impacts on the exist code for create properties while creating the workspace. there is no limitation.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -268,6 +264,36 @@ public boolean updateWorkspace(
     return updated;
   }
 
+  /** Update a workspace properties */
```
yuhuyoyo, line 267: can you add a test in WorkspaceDao too, thanks

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -275,6 +266,19 @@ public ResponseEntity<ApiWorkspaceDescription> getWorkspaceByUserFacingId(
     return new ResponseEntity<>(desc, HttpStatus.OK);
   }
 
+  @Override
+  public ResponseEntity<Void> updateWorkspaceProperties(
```
yuhuyoyo, line 270: can you add a test in workspaceapicontrollertest thanks

yuhuyoyo:
your product code is fine because there is this check in controller. https://cs.github.com/DataBiosphere/terra-workspace-manager/blob/78f675bc5785d9312a2911be81473d0bbed5a742/service/src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java?q=validatePropertyKey#L75.

at workspaceservice level, the map already can't have null key any more because of this check 

i'd like there to be a unit test for that util method but there isn't any.. up to you i guess. you can add a controller test and assert that it fails if you pass in null key.

integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
```diff
@@ -40,11 +43,24 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
     String validUserFacingId = "user-facing-id-" + uuidStr;
     String validUserFacingId2 = "user-facing-id-2-" + uuidStr;
 
+    Property property1 = new Property();
```
yuhuyoyo, line 46: nit: can we do a bit refactoring to make this a bit more readable :) 

you can have a helper method `Properties buildProperties(Map<String, String> propertyMap)`. And then to initialize a Map you can do a one liner with Map.of().At line 118 we will do `buildProperties(propertyMap.putAll(updateProperties))` so it's clear that the update is following the same rule as java hashmap#putAll.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
yuhuyoyo: 1 is fine thx

it's not enforced by this repo but generally i like all the public method to be tested.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -268,6 +264,35 @@ public boolean updateWorkspace(
...
+    final String sql =
+        "UPDATE workspace SET properties = cast(:properties AS jsonb) WHERE workspace_id = :id";
+
+    var params = new MapSqlParameterSource();
+    params
+        .addValue("properties", DbSerDes.propertiesToJson(properties))
+        .addValue("id", workspaceUuid.toString());
+
+    int rowsAffected = jdbcTemplate.update(sql, params);
+    boolean updated = rowsAffected > 0;
```
ddietterich, line 288: The `queryForObject` will throw if the workspace record is not found, so you will never get here if it doesn't exist. So the update will always return 1. I think the rest of this logic is not needed. In fact, do we even need or use the return value?

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -286,6 +284,22 @@ public void deleteWorkspace(Workspace workspace, AuthenticatedUserRequest userRe
...
+  /**
+   * Update workspace properties.
+   *
+   * @param userRequest authenticated user
+   * @param workspaceUuid workspace of interest
+   * @param properties map of key-value properties
+   */
+  public void updateWorkspaceProperties(
+      AuthenticatedUserRequest userRequest, UUID workspaceUuid, Map<String, String> properties) {
+    validateWorkspaceAndAction(userRequest, workspaceUuid, SamConstants.SamWorkspaceAction.WRITE);
```
ddietterich, line 296: As of Zach's PR, we are moving all access checks into the controller. This should be moved there.

ddietterich (changes requested):
I marked two changes. Otherwise, looking good

integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
ginaygoogle: Hi Yu. I find the Map.of is an unmodifiable map. So it will fail while I assert the updated properties if I use putAll to combine. 
I just modify the all first propertymap to modifiable map. Just a head up.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -268,6 +264,40 @@ public boolean updateWorkspace(
...
+    // get current property in this workspace id
+    String selectPropertiesSql = "SELECT properties FROM workspace WHERE workspace_id = :id";
+    MapSqlParameterSource propertiesParams =
+        new MapSqlParameterSource().addValue("id", workspaceUuid.toString());
+    int rowsAffected;
+    try {
+      String result =
+          jdbcTemplate.queryForObject(selectPropertiesSql, propertiesParams, String.class);
+      logger.info("Retrieved workspace record {}", result);
+      Map<String, String> properties = DbSerDes.jsonToProperties(result);
```
yuhuyoyo, line 279: can you move L279-289 outside the try catch. In general, we want the try catch scope tight.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
yuhuyoyo: that's a good point Dan..  We need the return value to log a change activity if an update happened. 

There are two ways to do this:
1. we compare the hashmap before and after putAll to see if the value does change
2. We assume that the properties is changed if we reached the end. Don't return anything here and just log an update in WorkspaceService.

I'm OK with just doing 2.

service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
```diff
@@ -22,6 +22,8 @@ public class MockMvcUtils {
   public static final String ADD_USER_TO_WORKSPACE_PATH_FORMAT =
       "/api/workspaces/v1/%s/roles/%s/members";
   public static final String CLONE_WORKSPACE_PATH_FORMAT = "/api/workspaces/v1/%s/clone";
+  public static final String UPDATE_WORKSPACES_V1_PROPERTIES_PATH_FORMAT =
```
yuhuyoyo, line 25: Could you rename to just WORKSPACE_V1_PROPERTIES_PATH_FORMAT so it can be shared with deleteproperties

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
ginaygoogle: good catch. I will also correct this in PF-1792

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -268,6 +264,40 @@ public boolean updateWorkspace(
...
+          "UPDATE workspace SET properties = cast(:properties AS jsonb) WHERE workspace_id = :id";
+
+      var params = new MapSqlParameterSource();
+      params
+          .addValue("properties", DbSerDes.propertiesToJson(properties))
+          .addValue("id", workspaceUuid.toString());
+
+      rowsAffected = jdbcTemplate.update(sql, params);
+    } catch (EmptyResultDataAccessException e) {
+      throw new WorkspaceNotFoundException(String.format("Workspace %s not found.", workspaceUuid));
```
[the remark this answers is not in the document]
ginaygoogle: Hi yu. This sql is for select the properties information in the specific UUID. The property attribute must exist even the value is null. The SQL can still return the null. But the only try catch fail will happens while the workspace UUID is not exist. From my understanding, this will still be workspaceNotFoundException.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
ginaygoogle: Thank you Yu. I am ok with both ways. I think once the updateProperties method was called, must have something update even if update a=b to a=b. 
I am wondering, at this moment, I added the try catch like all other methods. Does this still need to delete the return? 
If the update method should delete the return. just double check, should I follow this in Delete method? (I mean PF-1792)

service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
ginaygoogle: agree! When I write it, I also really hesitate on it due to all other names with a verb at beginning. And I supposed the UPDATE can represent both modify update and delete update. but your way is more generic. Thanks for the confirming.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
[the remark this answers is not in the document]
yuhuyoyo: i see. thanks for explaining

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
ddietterich: I'd make this method `void` and removing `rowsAffected` and related processing.
If the transaction succeeds, then the `WorkspaceService.updateWorkspaceProperties` can proceed with writing the activity. If it fails, then it will have thrown an exception that will be handled by the global exception handler.

ddietterich (commented):
I made another comment. Since I will be out for the next 4 days, you should probably find another reviewer.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
ginaygoogle: Done. Dan. Thanks for the explanation. I just correct it. Let me know if there any more ideas. 
really appreciate for the last minute before your vacation.

ddietterich (approved):
LGTM!

PR #716, opened 2022-07-11
[PF-1698] Add more git server to the allowlist

yuhuyoyo:
now support bitbucket, gitlab, azure and aws.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
@@ -109,7 +109,7 @@ public class ResourceValidationUtils {
   // An object named "." or ".." is nearly impossible for a user to delete.
   private static final ImmutableList<String> DISALLOWED_OBJECT_NAMES = ImmutableList.of(".", "..");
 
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
```diff
@@ -109,7 +109,7 @@ public class ResourceValidationUtils {
   // An object named "." or ".." is nearly impossible for a user to delete.
   private static final ImmutableList<String> DISALLOWED_OBJECT_NAMES = ImmutableList.of(".", "..");
 
-  private static final Pattern GIT_SSH_URI_PATTERN = Pattern.compile("git@(.*?)\\:(.*\\.git)$");
+  private static final Pattern GIT_SSH_URI_PATTERN = Pattern.compile("git@(.*?)\\:(.*)$");
```
ddietterich, line 112: That change allows invalid names - as in the failing unit test.
Maybe a comment on why we need to allow arbitrary file extensions?

ddietterich (approved):
I'd like to understand why you don't check the `.git` extension anymore.
Besides that, LGTM

zloery (approved):
LGTM pending Dan's question

PR #719, opened 2022-07-12
[PF-1817] add last updated date to workspace description

yuhuyoyo:


melissachang (approved):
Thanks. I was confused by last PR where last updated was set in service, and highest role was set in controller.

FYI here's the previous conversation:

https://github.com/DataBiosphere/terra-workspace-manager/pull/694#discussion_r908536005

Originally I set highest role in controller, then I moved to service.

Yay for more controller tests. :)

Do you mind adding a comment to WorkspaceService.listWorkspacesAndHighestRoles:

In general, highest role should be fetched in controller. Fetch here to save a SAM call.

melissachang:
Optional - delete any obsolete WorkspaceServiceTest, if they're covered by WorkspaceApiControllerTest

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -195,7 +196,12 @@ private ApiWorkspaceDescription buildWorkspaceDescription(
         .spendProfile(workspace.getSpendProfileId().map(SpendProfileId::getId).orElse(null))
         .stage(workspace.getWorkspaceStage().toApiModel())
         .gcpContext(gcpContext)
-        .azureContext(azureContext);
+        .azureContext(azureContext)
+        .lastUpdatedDate(
+            workspaceActivityLogDao
```
ddietterich, line 201: Is there a reason to have the timezone mapping and null-ing here instead of inside the DAO?

ddietterich (approved):
One small change, maybe. Otherwise, LGTM.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
yuhuyoyo: i think it's better to return optional than null in general but I can have it return offsetdate in the DAO instead of here.

PR #721, opened 2022-07-12
[PF-1817] Get created date of a given workspace in WorkspaceActivityLogDao 

yuhuyoyo:


service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -196,8 +197,8 @@ private ApiWorkspaceDescription buildWorkspaceDescription(
         .stage(workspace.getWorkspaceStage().toApiModel())
         .gcpContext(gcpContext)
         .azureContext(azureContext)
-        .lastUpdatedDate(
-            workspaceActivityLogDao.getLastUpdatedDate(workspace.getWorkspaceId()).orElse(null));
+        .lastUpdatedDate(workspaceActivityLogDao.getLastUpdatedDate(workspaceUuid).orElse(null))
```
ddietterich, line 200: It would be more efficient to do these calculations as a single transaction in SQL and a single request to the DAO. However, if we are going to move the created date into the workspace row eventually, it would not be worth the trouble.

ddietterich (commented):
I have a question about using a single DAO request instead of 2.

openapi/src/parts/workspace.yaml
```diff
@@ -498,6 +498,10 @@ components:
           description: Timestamp of the latest workspace change
           type: string
           format: date-time
+        createdDate:
```
melissachang, line 501: nit can this go before lastUpdateDate

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
melissachang: My understanding of end state was: we're moving this to middle layer, and not adding a new column.

WSM GetWorkspace no longer returns created on, last updated

We add WSM GetActivityLog

Middle layer calls GetWorkspace, GetActivityLog, and returns a workspace with created on, last updated

(If this turns out to be slow, we could add column in the future)

melissachang (approved):

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
yuhuyoyo: +1 for melissa's thinking for having an api to fetch the activity log per workspace since we eventually want trail. 

But even with a middle layer, maybe we don't *always* want the whole logs and singling out created on and last updated can improve performance (??). Then getting the two in one trip sounds good to me

PR #724, opened 2022-07-14
PF-1218: Disallow PrivateResourceUser on user-private resources

mikenev:
Have the controller return an error if the PrivateResourceUser is sent when creating private resources. Update tests to account for this.


zloery (approved):
Thanks for this cleanup!

yuhuyoyo (approved):
Thanks! have you checked if CLI is doing this?

mikenev:
> Thanks! have you checked if CLI is doing this?

Yes, the CLI is updated for this.

PR #728, opened 2022-07-19
PF-1218: Change the resource role list to a single role

mikenev:
Changes the input from a list of roles to a single role. UI and CLI should also be updated to account for this, but they aren't sending anything for this field now anyway.

### DIFF
--- a/openapi/src/common/schemas.yaml
+++ b/openapi/src/common/schemas.yaml
@@ -154,6 +154,14 @@ components:
       type: string
       pattern: '^[a-zA-Z0-9][-_a-zA-Z0-9]{0,1023}$'
 
+    PrivateResourceIamRoles:
+      description: >-
+        List of role(s) granted to the user of a private resource. In the current permission model
+        EDITOR includes WRITER includes READER by definition. We only grant the maximum role listed.
+      type: array
+      items:
+        $ref: '#/components/schemas/ControlledResourceIamRole'
+
     PrivateResourceUser:
       description: >-
         This text describes the target state:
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/app/controller/ControllerBase.java
```diff
@@ -143,12 +143,13 @@ public PrivateUserRole computePrivateUserRole(
...
+          // Translate the incoming role into our internal model form
           // This also validates that the incoming API model values are correct.
-          List<ControlledResourceIamRole> roles =
-              commonFields.getPrivateResourceUser().getPrivateResourceIamRoles().stream()
-                  .map(ControlledResourceIamRole::fromApiModel)
-                  .collect(Collectors.toList());
+          List<ControlledResourceIamRole> roles = new ArrayList<ControlledResourceIamRole>();
+          roles.add(
+              ControlledResourceIamRole.fromApiModel(
+                  commonFields.getPrivateResourceUser().getPrivateResourceIamRole()));
```
yuhuyoyo, line 151: why do you need a list here? i thought you changed it to singular now?

yuhuyoyo (approved):
neat

service/src/main/java/bio/terra/workspace/app/controller/ControllerBase.java
mikenev: I left it as a list since the code that follows expects a list. It would probably be good to refactor that portion too. I'll do that and send out an update.

openapi/src/common/schemas.yaml
```diff
@@ -172,13 +172,13 @@ components:
       type: object
   # TODO: PF-1218 - when CLI and UI are no longer passing this, disallow it for user-private
   #  resources, make the role List into a single role, and require both fields to be present.
-  #      required: [ userName, privateResourceIamRoles ]
+  #      required: [ userName, privateResourceIamRole ]
```
zloery, line 175: Do the UI or CLI ever include the `PrivateResourceUser` object in the request? If not, is it safe to uncomment this `required` block?

openapi/src/common/schemas.yaml
```diff
@@ -172,13 +172,13 @@ components:
...
       properties:
         userName:
           description: email of the workspace user to grant access
           type: string
-        privateResourceIamRoles:
-          $ref: '#/components/schemas/PrivateResourceIamRoles'
+        privateResourceIamRole:
+          $ref: '#/components/schemas/ControlledResourceIamRole'
```
zloery, line 181: Is anything else referencing `ControlledResourceIamRoles`, or can we delete that definition?

zloery (approved):
LGTM, though the change to computePrivateUserRole sounds good too

openapi/src/common/schemas.yaml
mikenev: Good catch, this should be safe to do now.

zloery (approved):

PR #730, opened 2022-07-19
WOR-382

blakery:
add optional ip range parameter to sas token endpoint [(WOR-382)](https://broadworkbench.atlassian.net/browse/WOR-382)


service/src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -115,4 +117,23 @@ public static String samCreateAction(ControlledResourceFields commonFields) {
...
   }
+
+  /**
+   * Validate the format of an ipAddress or range of addresses for Azure SAS tokens. We can't do
+   * this directly in the generated spring code yet, because the swagger codegen plugin doesn't
+   * support the use of oneOf in schema generation.
+   *
+   * @param ipRange a single ip address, or a range of ip addresses separated by a dash ('-')
+   */
+  public static void validateIpAddressRange(@Nullable String ipRange) {
```
blakery, line 128: There's a couple of edge cases that aren't tested here, like out of order ranges (like "168.1.5.70-168.1.5.60") and mixing ipv4 and ipv6.

Does it make sense to add those cases, or does it run too much of a risk of catching valid input?

aherbst-broad:
Cool, thanks for doing this. Please add a more descriptive title for future PRs and a bit more description for reviewers (including a link to the ticket)

service/src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
marctalbott: I guess I'm curious about what Azure will do if those edge cases are allowed through. Will Azure just return an error? Will it ignore the range entirely and return an unrestricted SAS token? Some other unexpected behavior?

openapi/src/common/parameters.yaml
```diff
@@ -80,6 +80,14 @@ components:
       schema:
         $ref: '#/components/schemas/IamRole'
 
+    SasIpRange:
+      name: sasIpRange
+      in: query
+      description: An IP address or range of addresses to restrict a SAS token to
```
marctalbott, line 86: Nit: maybe add a note in the description here about how to format a range of addresses?

marctalbott (approved):
one nit but looks good overall!

PR #735, opened 2022-07-25
[PF-1752] Log change agent email and subject id

yuhuyoyo:
{
  "workspaces": [
    {
      "id": "12a13ab2-0c55-11ed-861d-0242ac120002",
      "userFacingId": "change-by-workspace",
      "displayName": "",
      "description": "",
      "highestRole": "OWNER",
      "stage": "MC_WORKSPACE",
      "properties": [],
      "createdDate": "2022-07-25T20:05:37.99077Z",
      "createdBy": "yuhuyoyo@google.com",
      "lastUpdatedDate": "2022-07-25T20:05:37.99077Z",
      "lastUpdatedBy":"yuhuyoyo@google.com"
    }
  ]
}

In SAM, email is not nullable while subject id is nullable so I'm keeping that the same in our db. https://cs.github.com/broadinstitute/sam/blob/3081228a206fe4ad44911bb948af18bf506ab67c/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/initial_schema.xml?q=google_subject_id#L46

In the Dao, I merged change agent info and change date info together to one sql query because both are needed when constructing the  `workspaceDescription` in `WorkspaceApiController` so we need to fetch both at the same time always. 

I'm calling SAM to get the email and subject id instead of using the http request directly to make sure that we are always storing the user email not pet SA email.


yuhuyoyo:
@cbookg fyi

openapi/src/parts/workspace.yaml
```diff
@@ -545,3 +549,12 @@ components:
...
       enum: ['RAWLS_WORKSPACE', 'MC_WORKSPACE']
 
+    WorkspaceActivityChangeAgent:
+      description: Info for workspace activity change agent
+      required: [userEmail]
+      properties:
+        userEmail:
+          type: string
+          description: user email of the change agent
+        subjectId:
```
cbookg, line 559: What is the subjectId, and when could it be empty?

melissachang: Subject ID is sam ID. I don't think we should expose subject ID in API. Client just needs email.

Also, should Dan review this PR?

yuhuyoyo: actually it shouldn't be empty.

https://cs.github.com/broadinstitute/sam/blob/3081228a206fe4ad44911bb948af18bf506ab67c/src/main/resources/swagger/tos-enabled.yaml?q=userSubjectId#L95

it's user id stored in SAM per terra user. it equals to either a google subject id or a an azure b2c id (this part seems to still be some on-going work as there's no api exposing it.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -301,11 +304,22 @@ public void deleteWorkspace(Workspace workspace, AuthenticatedUserRequest userRe
...
-  public void deleteWorkspaceProperties(UUID workspaceUuid, List<String> propertyKeys) {
+  public void deleteWorkspaceProperties(
+      UUID workspaceUuid, List<String> propertyKeys, String userEmail, String subjectId) {
     workspaceDao.deleteWorkspaceProperties(workspaceUuid, propertyKeys);
     workspaceActivityLogDao.writeActivity(
-        workspaceUuid, new DbWorkspaceActivityLog().operationType(OperationType.UPDATE));
+        workspaceUuid, getDbWorkspaceActivityLog(OperationType.DELETE, userEmail, subjectId));
+  }
+
+  private DbWorkspaceActivityLog getDbWorkspaceActivityLog(
```
melissachang, line 317: You have several classes with this helper method, can you move to someplace common?

service/src/test/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHookTest.java
```diff
@@ -32,154 +38,162 @@
...
   }
 
   @Test
-  void deleteCloudContextFlightFails_cloudContextNotExist_logChangedDate() {
+  void deleteCloudContextFlightFails_cloudContextNotExist_logChangeDetails()
+      throws InterruptedException {
     var workspaceUuid = UUID.randomUUID();
-    var emptyChangedDate = activityLogDao.getLastUpdatedDate(workspaceUuid);
-    assertTrue(emptyChangedDate.isEmpty());
+    var emptychangeDetails = activityLogDao.getLastUpdateDetails(workspaceUuid);
```
melissachang, line 178: change -> Change

melissachang (commented):
Nice!

Sorry for delay, I've been learning about Apache proxy (how proxy subject id relates to SAM user subject id).

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -189,8 +196,15 @@ public ResponseEntity<Void> deleteGcsObjectReference(UUID workspaceUuid, UUID re
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
     workspaceService.validateWorkspaceAndAction(
         userRequest, workspaceUuid, SamWorkspaceAction.DELETE_REFERENCE);
+    var userStatusInfo =
```
ddietterich, line 199: In general, we don't use `var` if the type is not obvious. In this case, I do not know the type.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -189,8 +196,15 @@ public ResponseEntity<Void> deleteGcsObjectReference(UUID workspaceUuid, UUID re
...
+        SamRethrow.onInterrupted(
+            () -> samService.getUserStatusInfo(userRequest), "get user status info from SAM");
```
ddietterich, line 201: Could you DRY this out by passing the userRequest into the method and have it do the Sam call?
Like you do on the create call.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -601,8 +644,15 @@ public ResponseEntity<Void> deleteDataRepoSnapshotReference(UUID workspaceUuid,
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
     workspaceService.validateWorkspaceAndAction(
         userRequest, workspaceUuid, SamWorkspaceAction.DELETE_REFERENCE);
+    var userStatusInfo =
```
ddietterich, line 647: This is the fifth time for the same code. I think it should either move into the service or be a method in this class.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -228,10 +236,18 @@ public ResponseEntity<ApiWorkspaceDescription> updateWorkspace(
     }
     workspaceService.validateWorkspaceAndAction(
         userRequest, workspaceUuid, SamConstants.SamWorkspaceAction.WRITE);
+    var userStatusInfo =
+        SamRethrow.onInterrupted(
+            () -> samService.getUserStatusInfo(userRequest),
```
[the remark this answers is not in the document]
ddietterich: You could just move the `SamRethrow` phrase into a method in SamService; something like
`getUserStatusInfoWithRethrow()`

I would prefer pushing the whole thing into the service instead of repeatedly doing the Sam call and passing the results as params.

service/src/main/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHook.java
```diff
@@ -60,11 +69,20 @@ public HookAction endFlight(FlightContext context) {
       return HookAction.CONTINUE;
     }
 
+    var userRequest =
+        checkNotNull(
+            context
+                .getInputParameters()
+                .get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class));
+    var userStatusInfo = samService.getUserStatusInfo(userRequest);
```
ddietterich, line 77: Have you already made this call once at a higher level before calling the flight?
Should we be serializing the user status info into the input parameters instead of asking Sam again?

ddietterich (commented):
A number of comments. There is some DRYing out to do.

service/src/main/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHook.java
yuhuyoyo: No and sometimes yes. only if it's a direct write into the database, I check sam for user status info. But sometimes it's in the same method depending on whether it needs a flight or is a direct write to db. 

do you want to serialize the user status info and only if it doesn't exist ask Sam?

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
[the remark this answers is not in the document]
yuhuyoyo: I thought about extract a method. It seems to trivial/specific to me at the time. 

i'm not sure I understand this part : 
`I would prefer pushing the whole thing into the service instead of repeatedly doing the Sam call and passing the results as params.`

just to clarify, pushing the whole thing as in this entire block?

```
var userStatusInfo =
   SamRethrow.onInterrupted(
      () -> samService.getUserStatusInfo(userRequest),
      ...);
Workspace workspace =
        workspaceService.updateWorkspace(
            workspaceUuid, body.getUserFacingId(), body.getDisplayName(), body.getDescription());

            workspaceUuid,
            body.getUserFacingId(),
            body.getDisplayName(),
            body.getDescription(),
            userStatusInfo.getUserEmail(),
            userStatusInfo.getUserSubjectId());
    WsmIamRole highestRole = workspaceService.getHighestRole(workspaceUuid, userRequest);
    ApiWorkspaceDescription desc = buildWorkspaceDescription(workspace, highestRole);
```

service/src/main/resources/db/changesets/20220721_change_agent_email_and_subject_id_columns.yaml
```diff
@@ -0,0 +1,22 @@
+databaseChangeLog:
+- changeSet:
+    id: add change agent email and subject id column
+    author: yuhuyoyo
+    changes:
+    - addColumn:
+        tableName: workspace_activity_log
+        columns:
+        - column:
+            name: change_agent_email
```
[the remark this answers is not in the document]
yuhuyoyo: i was thinking it may get confused as if this is the change_agent or the change_target. But i guess we are preferring to have a json blob for the change target so it wouldn't get confused.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
yuhuyoyo: ah, i passed in the user request in the create call because i want to pass it in to the flight and we make the sam call in the hook in that case...

since Zach just moved all the sam checkAuz call to controller, I wasn't sure if I should re-introduce sam call back to the service.  but sure i can do that

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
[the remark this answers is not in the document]
yuhuyoyo: i can have a WorkspaceActivitylogService that has a logActivity(userRequest, ...) that can dry out some code
```
var userStatusInfo =
        samService.getUserStatusInfoWithRethrow(new AuthenticatedUserRequest());
    workspaceActivityLogDao.writeActivity(
        uuid,
        new DbWorkspaceActivityLog()
            .operationType(OperationType.GRANT_WORKSPACE_ROLE)
            .userEmail(userStatusInfo.getUserEmail())
            .subjectId(userStatusInfo.getUserSubjectId()));
```

service/src/main/resources/db/changesets/20220721_change_agent_email_and_subject_id_columns.yaml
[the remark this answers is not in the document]
melissachang: Good point, how about actor_email?

- I know what "actor" means, I don't know what "agent" means
- Could we take out "change"?
- Code should be consistent. Eg ActivityLogChangeDetails should match what's here

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
yuhuyoyo: the method is called getUserStatusinfo though :) i specified the type

melissachang (approved):

service/src/main/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHook.java
ddietterich: > do you want to serialize the user status info and only if it doesn't exist ask Sam?

I think we should. You could do it as a separate ticket. Sam is very heavily used, so we do a community service by avoiding extra calls.

ddietterich: > can you use WorkspaceActivityLogService instead of calling SAM yourself?

SamService is a service wrapping and handling calls to Sam, so technically this is not calling Sam directly. We typically call our SamService in cases like this.

service/src/main/java/bio/terra/workspace/db/WorkspaceActivityLogDao.java
```diff
@@ -37,18 +51,27 @@ public WorkspaceActivityLogDao(NamedParameterJdbcTemplate jdbcTemplate) {
 
   @WriteTransaction
   public void writeActivity(UUID workspaceId, DbWorkspaceActivityLog dbWorkspaceActivityLog) {
+    logger.info(
+        String.format(
+            "#writeActivity: workspaceId=%s, operationType=%s, actorEmail=%s, actorSubjectId=%s",
```
ddietterich, line 56: What does the `#` do in the format string?

service/src/main/java/bio/terra/workspace/db/model/DbWorkspaceActivityLog.java
```diff
@@ -4,10 +4,15 @@
 import bio.terra.workspace.service.workspace.model.OperationType;
 import java.util.Optional;
 import java.util.function.Supplier;
+import javax.annotation.Nullable;
 
 public class DbWorkspaceActivityLog {
 
-  private OperationType operationType;
+  private @Nullable OperationType operationType;
```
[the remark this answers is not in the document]
ddietterich: What was the resolution of this? I would not expect subjectId to be required, since (as I understand it) that will only be present for GCP logins.

ddietterich (approved):
A few questions, otherwise LGTM

service/src/main/java/bio/terra/workspace/db/WorkspaceActivityLogDao.java
yuhuyoyo: nothing. it's just i generally like to have a # in front of a method to signify that this is from this method call from this class.

yuhuyoyo:
`I think we should. You could do it as a separate ticket. Sam is very heavily used, so we do a community service by avoiding extra calls.`

It would be less dry though? I would need to do SAM call in the beginning of flight at all site instead of centralizing it here

ddietterich:
> It would be less dry though? I would need to do SAM call in the beginning of flight at all site instead of centralizing it here

Yes, you are right. I would not want that!

In the fullness of time, we will switch from `AuthenticatedUserRequest` objects to the new `SamUser` and `BearerToken` objects. I believe when we do that, the `SamUser` will have the correct user email in it and we won't have to make the extra call. For now, this is just fine.

yuhuyoyo:
`What was the resolution of this? I would not expect subjectId to be required, since (as I understand it) that will only be present for GCP logins.`

No, it's a workbench user id, different from the google subject id.

https://cs.github.com/broadinstitute/sam/blob/1607de4037370b1f1798aaf0e68d5652ffba3828/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala#L283 generated here. 

defined here: https://github.com/broadinstitute/sam/blob/7adddfb3741f4d1b01c1f32977683fdbfb60190f/src/main/resources/swagger/tos-enabled.yaml#L91

and get here: https://github.com/broadinstitute/sam/blob/1607de4037370b1f1798aaf0e68d5652ffba3828/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala#L168

ddietterich:
> No, it's a workbench user id

OK, thank you!

PR #736, opened 2022-07-25
Gradle retry for non-integration tests

melissachang:
For GHA runs, retry failing tests. If 2nd time passes, workflow will still fail and oncall will still be notified. This way, oncall doesn't have to rerun workflow in GHA. Note that this is faster because only the tests that failed are rerun.

In CLI repo, @zloery found that adding to `test` block didn't work, he had to add to `runTestsWithTag`. So unfortunately I have to duplicate. (And this can't go in terra-gradle-plugin repo.)

I didn't do integration tests, since integration tests don't use junit (QA-1764). [Retry plugin requires junit.](https://github.com/gradle/test-retry-gradle-plugin#supported-test-frameworks)

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -15,6 +15,7 @@ plugins {
     id "org.springframework.boot" version "${springBootVersion}"
     id "io.spring.dependency-management" version "1.0.10.RELEASE"
     id 'ru.vyarus.quality' version '4.5.0'
+    id 'org.gradle.test-retry' version '1.4.0'
 }
 
 // constants visible to all .gradle files in this project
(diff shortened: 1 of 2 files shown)


zloery (approved):
I think you'll also need to add the plugin itself (`id 'org.gradle.test-retry'`) to `service/build.gradle`, but LGTM otherwise

melissachang:
> I think you'll also need to add the plugin itself (`id 'org.gradle.test-retry'`) to `service/build.gradle`, but LGTM otherwise

Oops done

yuhuyoyo (approved):

PR #739, opened 2022-07-27
[TOAZ-184] Integrate Landing Zone database

zykonda:
TOAZ-172 implementation.
Unit tests pending.


scripts/connect-sql.sh
```diff
@@ -31,6 +31,11 @@ case $1 in
     dbuser=$(<"${configdir}/stairway-db-username.txt")
     dbpw=$(<"${configdir}/stairway-db-password.txt")
     ;;
+  landingzone)
+    db=$(<"${configdir}/landingzone-db-name.txt")
+    dbuser=$(<"${configdir}/landingzone-db-username.txt")
+    dbpw=$(<"${configdir}/landingzone-db-password.txt")
+    ;;
   *)
     echo "Specify wsm or stairway to choose which database to connect to"
```
sergiygetlin, line 40: should we add landingzone here as well? As you did below?

service/src/main/java/bio/terra/landingzone/db/LandingZoneDao.java
```diff
@@ -0,0 +1,152 @@
...
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Component;
+
+/** LandingZoneDao includes operations on the landing zone tables. */
+@Component
+public class LandingZoneDao {
+  /** SQL query for reading landing zone records. */
+  private static final String LANDINGZONE_SELECT_SQL =
+      "SELECT landingzone_id, spend_profile, template_id, template_version, properties"
```
sergiygetlin, line 27: Maybe definition is better than template? You proposed it initially. And for landing zone listing rest api we are using path you've suggested - /landingzones/definitions.. So, I am thinking maybe it's better to use definition here as well..

service/src/main/resources/landingzone.db/changesets/20220722_revised_initial_schema.yaml
```diff
@@ -0,0 +1,51 @@
+# The schema for a landing zone
```
sergiygetlin, line 1: Do we need tenantId, subscriptionId here?

sergiygetlin: Disregard it please

scripts/clean_landingzone.sql
```diff
@@ -0,0 +1,3 @@
+-- Remove all of the data from all tables; leave the schema intact.
+TRUNCATE landingzone CASCADE;
```
ddietterich, line 2: Is this a development convenience or is it used somewhere in the code?

ddietterich (changes requested):
Almost none of this code should not be in WSM. Please look at the structure of Terra Policy Service (TPS).

Please read [AMALGAM_README](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/amalgam/AMALGAM_README.md) and follow that for how to structure LZ integration with WSM.

scripts/clean_landingzone.sql
zykonda: it is convenience script, similar to workspace and stairway table clean up scripts.

zykonda:
The work has been restructured per suggestion.
A new repo has been created for Landing Zone component. 
Work for [TOAZ-184] made this PR irrelevant, so closing.

PR #733, opened 2022-07-22
[PF-1747] Install pandas_gbq in notebook startup script

melissachang:

PR #740, opened 2022-07-28
In post-startup script, use pip instead of conda

melissachang:
[Previous PR](https://github.com/DataBiosphere/terra-workspace-manager/pull/733) broke startup script:

image

With conda, must be `pandas-gbq` (this is why [previous PR](https://github.com/DataBiosphere/terra-workspace-manager/pull/733) broke). With pip, can be `pandas-gbq` or `pandas_gbq`.

conda took 17 minutes (!!). pip takes less than 1 minute, so changed to pip.


zloery (approved):
I'm not too familiar with either package manager, but the script lgtm

yuhuyoyo (approved):

PR #741, opened 2022-08-01
PF-1725: Return conflict when attempting to create a workspace with a duplicate id

mikenev:
Without a check, we get odd behavior. If nothing else has changed in the request JSON, then we'll return the ID as if we're creating the workspace. If something else (user facing id, description) has changed, then we return an error and remove the existing workspace.

### DIFF
--- a/service/src/test/java/bio/terra/workspace/app/configuration/external/controller/WorkspaceApiControllerTest.java
+++ b/service/src/test/java/bio/terra/workspace/app/configuration/external/controller/WorkspaceApiControllerTest.java
@@ -43,6 +43,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.http.MediaType;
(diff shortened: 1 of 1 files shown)


yuhuyoyo:
can we please have a test?

yuhuyoyo (approved):

ddietterich (commented):
This is a fine fix for the moment. I think we may have issues with some resource creation paths as well. We need to tolerate retrying the database operation on create: we can get a rerun of the database create transaction to create the same thing.

Anyway, if we do locking as suggested in [WSM Concurrency and UI Resource View](https://docs.google.com/document/d/1EhXlHEn27CrTbaOvtZwethL4re6Sw0alDgqdnIESFGw/edit?resourcekey=0-lr91gYF41AvkSL9fSE-tQg#heading=h.9pc5ed9qxb2j), then we can detect the difference between this flight retrying and a separate flight conflicting.

And as @yuhuyoyo suggests, a test would be a good idea.

mikenev:
Added a test.
This does feel like temporary fix but seemed appropriate since we're not using the async pattern for workspace create and the failure case can result in the metadata being removed. I feel like there's a larger fix coming like implementing the locking you mentioned.

ddietterich (approved):
LGTM

PR #742, opened 2022-08-01
[PF-2048] Replace DeleteGcpContextWithControlledResource with unit test

zloery:
(Output from discussion with IRWG)

While looking through WSM tests, I noticed this integration test didn’t actually call any WSM dependencies, so this feature is better tested with a unit test. This should be smaller, faster, and less flaky.


service/src/test/java/bio/terra/workspace/service/workspace/GcpCloudContextUnitTest.java
```diff
@@ -149,4 +161,65 @@ public void autoUpgradeTest() throws Exception {
     assertEquals(updatedContext.getSamPolicyReader().orElse(null), POLICY_READER);
     assertEquals(updatedContext.getSamPolicyApplication().orElse(null), POLICY_APPLICATION);
   }
+
+  @Test
+  public void deleteGcpContextDeletesControlledResources() throws Exception {
```
yuhuyoyo, line 166: nit:
deleteGcpContext_deletesControlledResourcesInDb

but how do you check if the cloud resource is deleted in the gcp project?

yuhuyoyo (approved):

service/src/test/java/bio/terra/workspace/service/workspace/GcpCloudContextUnitTest.java
zloery: this doesn't, I don't think we should be testing purely GCP functionality here (i.e. that deleting a GCP project deletes its resources). the original integration test also didn't verify that, which was the inspiration for this refactor

yuhuyoyo: I agree that we don't need to test GCP funcationality here but i think there should be test to make sure that an attempt is made to delete gcp project.

zloery: Agreed, I think that's out of scope for this test case though and is already covered by other tests which delete workspaces/contexts

PR #743, opened 2022-08-01
For PR tests, only post failures to slack

melissachang:
When I'm oncall, I enable notifications for #terra-wsm-alerts. This will reduce noise. @ddietterich said he would like success slack messages for nightly tests (to make sure they're running), but he's ok with turning off success slack messages for PR tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -158,7 +158,7 @@ jobs:
         path: service/build/reports/tests
 
     - name: "Notify QA Slack"
-      if: always() && (steps.set-env-step.outputs.test-env == 'alpha' || steps.set-env-step.outputs.test-env == 'staging')
+      if: failure() && (steps.set-env-step.outputs.test-env == 'alpha' || steps.set-env-step.outputs.test-env == 'staging')
       uses: broadinstitute/action-slack@v3.8.0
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -158,7 +158,7 @@ jobs:
         path: service/build/reports/tests
 
     - name: "Notify QA Slack"
-      if: always() && (steps.set-env-step.outputs.test-env == 'alpha' || steps.set-env-step.outputs.test-env == 'staging')
+      if: failure() && (steps.set-env-step.outputs.test-env == 'alpha' || steps.set-env-step.outputs.test-env == 'staging')
```
ddietterich, line 161: This one goes to the QA slack. We should probably not change this one, unless QA agrees.
@ungwudik is the QA person I usually work with but he is on vacation until Aug 6.

ddietterich (commented):
Suggestion on changing the QA slack and going ahead.

.github/workflows/test.yml
melissachang: Reverted

ddietterich (approved):
Thanks!

zloery (approved):
LGTM, and agreed about the QA slack

PR #744, opened 2022-08-02
PF-1877: set default value unknown to actor email and subject id

yuhuyoyo:


zloery (approved):

service/src/main/resources/db/changelog.xml
```diff
@@ -10,4 +10,5 @@
     <include file="changesets/20220516_workspace_user_facing_id_required.yaml" relativeToChangelogFile="true"/>
     <include file="changesets/20220617_workspace_activity_log_table.yaml" relativeToChangelogFile="true"/>
     <include file="changesets/20220721_actor_email_and_actor_subject_id_columns.yaml" relativeToChangelogFile="true"/>
+    <include file="changesets/20220802_default_value_to_actor_email_and_actor_subject_id_columns.yaml" relativeToChangelogFile="true"/>
```
ddietterich, line 13: I don't think this will work. I think you need to edit `20220721_actor_email_and_actor_subject_id_columns.yaml` and provide the defaults there.
In its current form it cannot be applied by liquibase if there are any existing entries in the activity log table.

ddietterich (changes requested):
You need to change the existing liquibase changeset, not add a new one.

service/src/main/resources/db/changelog.xml
yuhuyoyo: oh i thought i'm not allowed to change liquidbase file once it's in.

ddietterich: Yes, that is the general practice. But when the liquibase file doesn't work, then we need to fix it. Since it has never been deployed anywhere except local testing, it is fine to revise it.

integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
```diff
@@ -76,11 +76,12 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
     var createdDate = workspaceDescription.getCreatedDate();
     assertEquals(firstLastUpdatedDate, createdDate);
 
-    var lastUpdatedBy = workspaceDescription.getLastUpdatedBy();
-    assertEquals(USER_EMAIL, lastUpdatedBy);
-
-    var createdBy = workspaceDescription.getCreatedBy();
-    assertEquals(USER_EMAIL, createdBy);
+    // TODO(PF-1752): Change to assert lastUpdatedBy equals to user email once the code changes 735
```
ddietterich, line 79: I'm curious why you needed to comment it out. This would be for a new workspace, so it would have the activity log, wouldn't it?

ddietterich (approved):
A question about the test. The liquibase change LGTM, so I'll approve as is.

integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
yuhuyoyo: https://broadinstitute.slack.com/archives/C01JVP0MLLX/p1659370937292819 it's failing in the staging environment where my PR is not in?

ddietterich: OK, so I think if you leave the test in, then this will all be sorted as this fix works its way through the system.

yuhuyoyo: just to clarify, is this change necessary or should i undo it?

ddietterich: I don't think commenting out the test is necessary. However, it does not hurt to leave it commented out now and re-enable it later we have confidence in the liquibase change. Your choice - either is fine with me.

PR #745, opened 2022-08-02
[PF-1877] Take 2: use value instead of defaultValue

ddietterich:


yuhuyoyo (approved):

zloery (commented):
Why do we need this? It seems like the other change is working (e.g. if I call listWorkspaces on dev WSM now, I see the `lastUpdatedBy` and `createdBy` fields with values "Unknown")

I also see one other liquibase change where we use defaultValue, but none where we use value

ddietterich:
> Why do we need this? It seems like the other change is working

This change is not working. The problem is that when there are existing records in the activity log table, adding columns fills them with `null`. The new columns have a not-null constraint, so the liquibase upgrade fails. See [log explorer](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22terra-kernel-k8s%22%0Aresource.labels.location%3D%22us-central1-a%22%0Aresource.labels.cluster_name%3D%22terra-integration%22%0Aresource.labels.namespace_name%3D%22terra-wsmtest%22%0Aresource.labels.pod_name:%22workspacemanager-deployment%22%0Aresource.labels.container_name%3D%22workspacemanager%22;timeRange=2022-08-02T12:16:23.588296Z%2F2022-08-02T14:16:23.588Z;cursorTimestamp=2022-08-02T13:26:48.069Z?project=terra-kernel-k8s) for the failure details.

What we need to do is provide a value to fill in to all existing rows. Yu and I thought this was `defaultValue` and that was the first fix to PF-1877 she made this morning. We were wrong. Turns out `defaultValue` only applies to new rows inserted. You need to set a `value` to set the value on existing rows. At least that is what the documentation says.

ddietterich:
This change is not needed. The issue with the nightly tests was the sync on wsmtest.

PR #747, opened 2022-08-02
[PF-1632] Accept policy attributes on create workspace

zloery:
When creating a workspace, users can now specify policy inputs which should apply to the workspace, and WSM will pass these to TPS. WSM will also fetch the policies on a workspace when getting or listing workspaces, and will delete the policies associated with a workspace when deleting that workspace.

I made a couple assumptions here that I'd like to check with reviewers:

- ~~It seems like WSM will need to know about certain policies to enforce them (e.g. it needs to know what group-membership policies look like so that it knows when to call TPS to evaluate them), so it seems reasonable to me that WSM validates that policy inputs are properly formed (e.g. that `terra:group-constraint` policies have `group-name` in additonalData). I'm less sure that WSM should know all types of policies that exist, so I didn't do any validation against which policies are present. Is this the right level of policy validation?~~ The current plan is to centralize validation in TPS rather than spreading validation across services. We can validate known `terra` policies, even if TPS might not be able to validate unknown policies in other namespaces. 
- When returning policy information on `WorkspaceDescription`s, I'm extracting the effective policy attributes and only including those. Returning whole PAOs seemed like a lot of unnecessary information (e.g. there's no point telling a user this is a WSM-enforced policy on a WORKSPACE object, they know that already). Is this the right level of information or should I return something else?
- Right now, if TPS is enabled via config flag, WSM will create a PAO for all MC_WORKSPACE stage workspaces. That means it will create an empty PAO if a user doesn't specify any policies. I think this is better than not creating a PAO if the user doesn't specify policy (otherwise we'll need to create one the first time a user adds a policy to a workspace), but I'm not positive on this.


ddietterich:
> It seems like WSM will need to know about certain policies to enforce them

My goal is that WSM would not need to know the innards of the PAO structures. The idea is that when WSM has a question, it asks TPS and gets a nicely formatted answer. So you could ask specific questions, like what are my group constraints and get a nicely formatted answer. I'd like to try that and see if it breaks down. There are no CLI/GUI mocks or requirements yet so I expect this to jiggle.

> I'm extracting the effective policy attributes and only including those

That makes sense. The other info will be relevant when we need to describe things like:
- this policy is here because of this reference to this snapshot containing this table with this PII column (in my dreams)
- this policy is in conflict with that policy; I'm trying to figure out how to express those conflicts right now.

> if TPS is enabled via config flag, WSM will create a PAO for all MC_WORKSPACE stage workspaces

Yes, that is exactly right. That way when we do update and link-child, you won't have to remember if you created it or not. It imposes some overhead, but I think it will be simpler overall.

mikenev:
> The idea is that when WSM has a question, it asks TPS and gets a nicely formatted answer

I like this idea and the idea that WSM shouldn't need to know much about PAO innards. Do we just try to save these policies and have TPS validate them/return an error at that point? Or do we expose another Validate function in TPS?

ddietterich:
> Do we just try to save these policies and have TPS validate them/return an error at that point? Or do we expose another Validate function in TPS?

That is a really good question. The policy attribute space is open-ended, so there is not much validation that you can do. It makes sense to me to centralize validation in TPS for the known policies. That is a missing piece of the design. Do you have thoughts on where that would fit in the flow? (Maybe outside of this PR...)

service/src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
```diff
@@ -36,6 +44,9 @@ public final class ControllerValidationUtils {
   public static final Pattern USER_FACING_ID_VALIDATION_PATTERN =
       Pattern.compile("^[a-z0-9][-_a-z0-9]{2,62}$");
 
+  public static final Map<String, Set<String>> REQUIRED_POLICY_INFORMATION =
+      ImmutableMap.of("terra:group-constraint", ImmutableSet.of("group-name"));
```
ddietterich, line 48: I am changing this to "groups", because it is going to be a comma-separated list of groups.
But as Mike suggests, this validation should probably be in TPS.
Do we need a validatePao endpoint or should it be a side-effect of calling any of Pao-receiving TPS endpoints?

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/WorkspaceApiControllerTest.java
```diff
@@ -77,6 +99,13 @@ public void setup() throws InterruptedException {
             new UserStatusInfo()
                 .userEmail(USER_REQUEST.getEmail())
                 .userSubjectId(USER_REQUEST.getSubjectId()));
+
+    when(mockFeatureConfiguration.isTpsEnabled()).thenReturn(true);
```
ddietterich, line 103: Should we enable TPS in `application-test.yml`?

ddietterich (commented):
Very nice! Two things:
1. Change `group-name` to `groups`
2. We should figure out validation, but I'm not sure that should be a block for this PR.

zloery:
> That is a really good question. The policy attribute space is open-ended, so there is not much validation that you can do. It makes sense to me to centralize validation in TPS for the known policies. That is a missing piece of the design. Do you have thoughts on where that would fit in the flow? (Maybe outside of this PR...)

Agreed, at some point it seems reasonable that we'd have a policy managed by something out of Terra's control so we can't really validate or check against a known set. Taking the validation out of WSM means we'll fail slower when the user provides an invalid policy (i.e. we'll have to launch and then undo a flight instead of checking beforehand), but it also means less places to change as policies shift a bit, which seems worth it when we expect these things to change frequently

service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiController.java
```diff
@@ -51,7 +52,11 @@ public ResponseEntity<Void> deletePao(UUID objectId) {
 
   @Override
   public ResponseEntity<ApiTpsPaoGetResult> getPao(UUID objectId) {
-    ApiTpsPaoGetResult result = tpsApiDispatch.getPao(bearerTokenFactory.from(request), objectId);
+    ApiTpsPaoGetResult result =
+        tpsApiDispatch
```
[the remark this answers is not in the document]
zloery: I'm on the fence on this one. If this were an external service, I would expect `getPao` to return 404 on not found. If this was an internal wrapper (like SamService in WSM), I'd prefer returning Optional. Since `tpsApiDispatch` is kind of both, it seems like we might want both `getPao` and `getPaoIfExists`

service/src/main/java/bio/terra/workspace/common/utils/ControllerValidationUtils.java
zloery: It would require an (internal for now, external later) service call but save us from launching + undoing a flight, so it's probably a performance improvement but not required for functionality

service/src/main/resources/application.yml
```diff
@@ -165,7 +165,7 @@ feature:
   # alpha1-enabled - Controls support of the Alpha1 experimental API and service
   alpha1-enabled: false
   # tps-enabled - Controls whether Terra Policy Service is called. It is always built into WSM
-  tps-enabled: false
+  tps-enabled: true
```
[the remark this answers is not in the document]
zloery: good point! moved this to the generated `local-properties` instead

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/WorkspaceApiControllerTest.java
zloery: we probably should for the sake of reminding future test writers this is on, but for these tests I think it's valuable to verify the different behavior with it off/on, so I think the mocks help

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -81,12 +83,22 @@ public WorkspaceService(
 
   /** Create a workspace with the specified parameters. Returns workspaceID of the new workspace. */
   @Traced
-  public UUID createWorkspace(Workspace workspace, AuthenticatedUserRequest userRequest) {
+  public UUID createWorkspace(
+      Workspace workspace,
+      @Nullable ApiTpsPolicyInputs policies,
```
mikenev, line 88: Not a big issue, I'm curious why this was added in the middle.

zloery: at one point very early on, all of the methods in `WorkspaceService` and a few other places had the `AuthenticatedUserRequest` as the last argument and I think that pattern got stuck in my brain. it's no longer true in `WorkspaceService` and it's definitely not worth the churn to re-order all the signatures, but for some reason I keep defaulting to that ordering. I'm also fine moving this to the last argument if that's cleaner

mikenev: I don't think it needs to be reordered, it was just interesting to see. Makes sense with the history.

mikenev (approved):

scripts/write-config.sh
```diff
@@ -297,6 +297,8 @@ workspace:
     managed-app-tenant-id: ${tenantid}
     sas-token-start-time-minutes-offset: 15
     sas-token-expiry-time-minutes-offset: 60
+feature:
```
ddietterich, line 300: Does this work? The include of `local-properties.yml` is in `application.yml`, so there will be two definitions of `tps-enabled`.  I don't know which one wins.

openapi/src/parts/workspace.yaml
```diff
@@ -418,11 +418,7 @@ components:
           description: ID of provided spend profile
           type: string
         policies:
-          description: Policies provided by the containing folder; NOT IMPLEMENTED
-          type: array
-          items:
-            type: string
-            format: uuid
+          $ref: '#/components/schemas/TpsPolicyInputs'
```
ddietterich, line 421: Maybe we should just get rid of this until we implement it. The input might want to be the UUID of the "folder" (or whatever we call the container of workspaces.

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/WorkspaceApiControllerTest.java
```diff
@@ -60,9 +74,17 @@ public class WorkspaceApiControllerTest extends BaseConnectedTest {
   AuthenticatedUserRequest USER_REQUEST =
       new AuthenticatedUserRequest(
           "fake@email.com", "subjectId123456", Optional.of("ThisIsNotARealBearerToken"));
+  /** A fake group-constraint policy for a workspace. */
+  private static final ApiTpsPolicyInput GROUP_POLICY =
+      new ApiTpsPolicyInput()
+          .namespace("terra")
+          .name("group-constraint")
+          .addAdditionalDataItem(new ApiTpsPolicyPair().key("group-name").value("my_fake_group"));
```
ddietterich, line 82: The actual key will be "groups"

ddietterich (approved):
LGTM - a couple of minor things

scripts/write-config.sh
zloery: Yep! Imports are treated as additional documents appended to the end of the importing document ([Spring doc](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.files.importing)) so the values in `local-properties` are guaranteed to override the values in `application.yml`. I've also verified this works locally.

Added a comment above the import clarifying this as well since it's important behavior

openapi/src/parts/workspace.yaml
zloery: Right now this isn't used by callers and I think that will probably be true while we figure out the exact shape of policy. I'm inclined to leave this in as a best-guess for now and be really cautious about relying on it heavily in clients until folders/LZs are a bit clearer, but it seems reasonable to me that we would have policy on a workspace beyond its folder restrictions

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/WorkspaceApiControllerTest.java
```diff
@@ -55,14 +70,22 @@
  * tests.
  */
 @AutoConfigureMockMvc
-public class WorkspaceApiControllerTest extends BaseConnectedTest {
+public class WorkspaceApiControllerTest extends BaseUnitTest {
```
zloery, line 73: This test is already mocking all external dependencies, so moving this to be a unit test instead

PR #748, opened 2022-08-03
[PF-1882] Handle unknown roles from Sam

zloery:
Currently when WSM encounters unknown Sam roles, it throws an exception. However, this can happen when reading the roles from a RAWLS_WORKSPACE stage workspaces, as Rawls may have set a role that WSM doesn't understand. Instead of throwing an exception WSM should silently ignore these roles.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


ddietterich (approved):
LGTM

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -276,6 +277,9 @@ public void createWorkspaceWithDefaults(AuthenticatedUserRequest userRequest, UU
    * List all workspace IDs in Sam this user has access to. Note that in environments shared with
    * Rawls, some of these workspaces will be Rawls managed and WSM will not know about them.
    *
+   * <p>Additionally, Rawls may create additional roles that WSM does not know about. Those roles
```
melissachang, line 280: Can we add this somewhere

Rawls uses same SAM workspace object as WSM.

melissachang (approved):
Thanks. If I had a nickel for every bug of mine that you fixed... :)

PR #749, opened 2022-08-04
PF-1218: Stop accepting a list of roles on the private resource user

mikenev:
This is the last phase of the API update - where we stop accepting a list of roles. We also require both the user + single role to be sent if the PrivateResourceUser is sent.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
@@ -197,9 +197,7 @@ public class PrivateControlledGcsBucketLifecycle extends WorkspaceAllocateTestSc
                     ManagedBy.USER,
                     CloningInstructionsEnum.NOTHING,
                     privateUserNoEmail));
-    assertThat(
-        ex.getMessage(),
-        containsString("MethodArgumentNotValidException"));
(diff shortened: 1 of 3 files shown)


integration/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
```diff
@@ -200,7 +200,7 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
     assertThat(
         ex.getMessage(),
         containsString(
-            "PrivateResourceUser can only be specified by applications for private resources"));
+            "Request could not be parsed or was invalid: MethodArgumentNotValidException. Ensure that all types are correct and that enums have valid values."));
```
zloery, line 203: nit: checking for the whole error message is pretty brittle. Could this just assert something like `containsString("MethodArgumentNotValidException")`?

zloery (approved):
Thanks for sticking with this change the whole way through!

integration/src/main/java/scripts/testscripts/PrivateControlledGcsBucketLifecycle.java
mikenev: Good point, will do that.

melissachang (approved):

PR #750, opened 2022-08-04
[PF-1882] Handle roles that WSM doesn't know about yet

melissachang:
[Last night's nightly test failed.](https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/2794805592)

Yesterday, with a local WSM, I added discoverer policy to some workspaces in broad-dev SAM. The test's ListWorkspaces picked those up. Nightly test WSM (`https://workspace.wsmtest.integ.envs.broadinstitute.org/`) didn't know about discoverer role. It threw an exception when a workspace didn't have any SAM permissions.

In this PR, if a workspace doesn't have any SAM permission, ListWorkspaces doesn't return it.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -299,7 +299,10 @@ public class SamService {
                   .map(WsmIamRole::fromSam)
                   .filter(Objects::nonNull)
                   .collect(Collectors.toList());
-          workspacesAndRoles.put(workspaceId, WsmIamRole.getHighestRole(workspaceId, roles));
+          Optional<WsmIamRole> highestRole = WsmIamRole.getHighestRole(workspaceId, roles);
+          if (highestRole.isPresent()) {
+            workspacesAndRoles.put(workspaceId, highestRole.get());
+          }
         } catch (IllegalArgumentException e) {
           // WSM always uses UUIDs for workspace IDs, but this is not enforced in Sam and there are
           // old workspaces that don't use UUIDs. Any workspace with a non-UUID workspace ID is
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -300,8 +300,7 @@ public class SamService {
                   .filter(Objects::nonNull)
                   .collect(Collectors.toList());
           Optional<WsmIamRole> highestRole = WsmIamRole.getHighestRole(workspaceId, roles);
-          // Skip workspaces with no roles. (That means there's a role this WSM doesn't know about
-          // yet.)
+          // Skip workspaces with no roles. (That means there's a role this WSM doesn't know about.)
           if (highestRole.isPresent()) {
             workspacesAndRoles.put(workspaceId, highestRole.get());
           }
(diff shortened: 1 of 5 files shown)


service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -299,7 +299,10 @@ public Map<UUID, WsmIamRole> listWorkspaceIdsAndHighestRoles(AuthenticatedUserRe
                   .map(WsmIamRole::fromSam)
                   .filter(Objects::nonNull)
                   .collect(Collectors.toList());
-          workspacesAndRoles.put(workspaceId, WsmIamRole.getHighestRole(workspaceId, roles));
+          Optional<WsmIamRole> highestRole = WsmIamRole.getHighestRole(workspaceId, roles);
```
ddietterich, line 302: Maybe a comment here "skip workspaces with no WSM roles" or something

service/src/main/java/bio/terra/workspace/service/iam/model/WsmIamRole.java
```diff
@@ -46,20 +50,23 @@ public static WsmIamRole fromSam(String samRole) {
...
   }
 
-  public static WsmIamRole getHighestRole(UUID workspaceId, List<WsmIamRole> roles) {
+  public static Optional<WsmIamRole> getHighestRole(UUID workspaceId, List<WsmIamRole> roles) {
     if (roles.isEmpty()) {
-      throw new InternalServerErrorException(
-          String.format("Workspace %s missing roles", workspaceId.toString()));
+      // This should be extremely rare. This only happens when a WSM role has been added to SAM,
+      // but WSM doesn't know about it yet (eg a local WSM created a workspace with this role, but
+      // broad-dev WSM doesn't know about role yet).
```
ddietterich, line 57: I think the general case is that it happens when there is a workspace that has no WSM roles. Technically, it is not specific to us adding a Sam role for WSM. Suggest the warning below say "Ignoring workspace %s with no WSM roles".

ddietterich (approved):
Code looks fine. A couple of comment comments

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
melissachang: Done

service/src/main/java/bio/terra/workspace/service/iam/model/WsmIamRole.java
melissachang: SAM listResourcesAndPoliciesV2 only returns workspaces where caller is in at least one role.

The exact sequence of events:

- With Local WSM, I created a workspace with discoverer policy.
- Local WSM grantRole, added Liam test user to discoverer policy
- Nightly test calls ListWorkspaces as Liam. Workspace is returned with only discoverer role.

melissachang (commented):
Forgot to mention, I reproed failure locally. Test now passes locally.

ddietterich:
> The exact sequence of events...

My only thought was that other sequences of events would get us to the same error. RAWLs creates a workspace that uses no WSM-known roles, for example.

This also shows the problem of using a single Sam instance for many disposable environments.

melissachang:
> > The exact sequence of events...
> 
> My only thought was that other sequences of events would get us to the same error. RAWLs creates a workspace that uses no WSM-known roles, for example.
> 
> This also shows the problem of using a single Sam instance for many disposable environments.

Got it, updated both comments.

zloery (approved):

PR #755, opened 2022-08-08
[PF-1681] Add discoverer policy when creating SAM workspace resource

melissachang:
This PR:
- Changes WSM to understand SAM discoverer role
- When a workspace is created, adds empty discoverer policy. (Since policies can only be added when a SAM resource is created.)

Future PRs will modify ListWorkspaces/GetWorkspace to add a request permission for returning discoverer workspaces. For now, ListWorkspaces *does not* return a workspace if caller is only discoverer.

Manually tested:

- Created workspace. Can see empty discoverer policy with getRoles.
- Called grantRole to add second user as discoverer
- Second user calls GetWorkspace. GetWorkspace returns 403 as expected
- At first, when second user called ListWorkspaces, discoverer workspace was returned, which was wrong. I made the change to SamService.java. Now when second user calls ListWorkspace, discoverer workspace is not returned.

WorkspaceApiControllerTest.java uses mock bio.terra.workspace.service.iam.SamService. I wanted my tests to use real bio.terra.workspace.service.iam.SamService, so I created WorkspaceApiControllerConnectedTest.java.

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,3 +316,11 @@ The integration tests live in the `integration` project. Consult the integration
 
 In the early days of the project, there were JUnit-based integration tests. We are in
 process of migrating them to Test Runner.
+
+#### Cleaning up workspaces in tests
+
+For connected tests that use real SamService/Broad dev SAM, delete workspaces at
+the end of test, so Broad dev SAM isn't cluttered with test workspaces.
+
+For integration tests, don't need to delete workspace because it's taken care of
+by janitor.
--- a/README.md
+++ b/README.md
@@ -319,8 +319,11 @@ process of migrating them to Test Runner.
 
 #### Cleaning up workspaces in tests
 
+For integration tests, don't need to delete workspaces because it's taken care of
+by janitor.
+
 For connected tests that use real SamService/Broad dev SAM, delete workspaces at
 the end of test, so Broad dev SAM isn't cluttered with test workspaces.
+(Connected tests can't use janitor because the WSM is ephemeral.)
 
-For integration tests, don't need to delete workspace because it's taken care of
-by janitor.
+
--- a/README.md
+++ b/README.md
@@ -335,4 +335,5 @@ to clean up SAM workspaces. Why not just call `deleteWorkspace()` and not use
 janitor? Janitor is useful in case test fails (or `deleteWorkspace()` fails).
 
 Integration tests: Tests should call `deleteWorkspace()` because integration
-tests don't use janitor.
+tests don't use janitor. Most tests don't need to worry about this because
+`WorkspaceAllocateTestScriptBase.java` deletes the workspace it creates.
--- a/README.md
+++ b/README.md
@@ -319,11 +319,20 @@ process of migrating them to Test Runner.
 
 #### Cleaning up workspaces in tests
 
-For integration tests, don't need to delete workspaces because it's taken care of
-by janitor.
+We have 2 ways of cleaning up resources (WSM workspace, SAM workspace, GCP project):
 
-For connected tests that use real SamService/Broad dev SAM, delete workspaces at
-the end of test, so Broad dev SAM isn't cluttered with test workspaces.
-(Connected tests can't use janitor because the WSM is ephemeral.)
+1. Connected tests use Janitor. Janitor deletes GCP project and not SAM
+   workspace (see
+   [here](https://github.com/DataBiosphere/terra-workspace-manager/pull/755#discussion_r942717257) for details).
+2. Tests call WSM `deleteWorkspace()`. This deletes WSM workspace + SAM workspace +
+   GCP project.
 
+Connected tests that use mock SamService: Tests don't need to call
+`deleteWorkspace()` because there is no SAM workspace to clean up.
 
+Connected tests that use real SamService: Tests should call `deleteWorkspace()`
+to clean up SAM workspaces. Why not just call `deleteWorkspace()` and not use
+janitor? Janitor is useful in case test fails (or `deleteWorkspace()` fails).
+
+Integration tests: Tests should call `deleteWorkspace()` because integration
+tests don't use janitor.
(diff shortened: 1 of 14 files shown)


zloery (approved):

service/src/main/java/bio/terra/workspace/service/workspace/flight/GcpCloudSyncStep.java
```diff
@@ -71,7 +72,11 @@ public StepResult doStep(FlightContext flightContext)
       newBindings.addAll(currentPolicy.getBindings());
       // Add appropriate project-level roles for each WSM IAM role.
       workspaceRoleGroupsMap.forEach(
-          (role, email) -> newBindings.add(bindingForRole(role, email, gcpProjectId)));
+          (wsmRole, email) -> {
+            if (CUSTOM_GCP_PROJECT_IAM_ROLES.containsKey(wsmRole)) {
```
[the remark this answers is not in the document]
melissachang: Done

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/WorkspaceApiControllerConnectedTest.java
```diff
@@ -0,0 +1,122 @@
...
+@AutoConfigureMockMvc
+public class WorkspaceApiControllerConnectedTest extends BaseConnectedTest {
+
+  @Autowired private MockMvc mockMvc;
+  @Autowired private ObjectMapper objectMapper;
+  @Autowired private UserAccessUtils userAccessUtils;
+
+  @Test
+  public void getWorkspace_doesNotReturnWorkspaceWithOnlyDiscovererRole() throws Exception {
+    ApiCreatedWorkspace workspace = createWorkspace();
```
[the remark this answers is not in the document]
melissachang: Done, thanks. I always forget when to have tests delete workspaces, so I added a section to README, PTAL.

README.md
```diff
@@ -316,3 +316,14 @@ The integration tests live in the `integration` project. Consult the integration
 
 In the early days of the project, there were JUnit-based integration tests. We are in
 process of migrating them to Test Runner.
+
+#### Cleaning up workspaces in tests
```
ddietterich, line 320: I believe this is incorrect. Please double check my understanding @zloery 

Janitor just deletes projects, it does not do anything to clean up Sam, so we should never skip in-test cleanup.
Connected tests also use RBS and Janitor. It doesn't have anything to do with WSM being ephemeral, since Janitor is just deleting projects.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -300,8 +300,14 @@ public Map<UUID, WsmIamRole> listWorkspaceIdsAndHighestRoles(AuthenticatedUserRe
                   .filter(Objects::nonNull)
                   .collect(Collectors.toList());
           Optional<WsmIamRole> highestRole = WsmIamRole.getHighestRole(workspaceId, roles);
-          // Skip workspaces with no roles. (That means there's a role this WSM doesn't know about.)
+          // Skip workspaces with no roles. (That means there's a role this WSM doesn't know
+          // about.)
           if (highestRole.isPresent()) {
+            // TODO(PF-1875): Support requesting discoverer workspaces in ListWorkspaces.
```
ddietterich, line 306: Why not just filter in ListWorkspaces, so you don't have to remember to come back to this code?
You need the filtering there anyway.

service/src/main/java/bio/terra/workspace/service/workspace/flight/SyncSamGroupsStep.java
```diff
@@ -47,6 +47,9 @@ public StepResult doStep(FlightContext flightContext)
     workspaceRoleGroupMap.put(
         WsmIamRole.READER,
         samService.syncWorkspacePolicy(workspaceUuid, WsmIamRole.READER, userRequest));
+    workspaceRoleGroupMap.put(
```
ddietterich, line 50: I don't think we need to sync the discoverer role. That will create Google groups for discoverers. We only need that when we are going to apply that group to a cloud resource ACL; like granting a READER actual read access to a GCS bucket.

I think we only ever enforce discoverer by checking with Sam - never the cloud.

ddietterich (changes requested):
The documentation is incorrect and we don't need to sync the discoverer role

README.md
zloery: We can use Janitor in two ways:
1. the `tools` RBS instance registers every project it hands out with Janitor to be cleaned up after some time. Since all our connected tests talk to this instance, the GCP projects get cleaned up eventually
2. After PF-886, you also can register a workspace with Janitor. After some time, Janitor will call WSM to delete the workspaces, which will clean up all of the WSM DB + Sam workspace + GCP project. The CLI tests use this when they run against real WSM environments, like dev.

We can't use the second pattern for WSM connected tests because the instance is ephemeral (so there's nothing for Janitor to call later), but we can and do use the first pattern to make sure the GCP projects get cleaned up. the Sam workspaces will still be around if the test doesn't clean them up, though

README.md
```diff
@@ -316,3 +316,14 @@ The integration tests live in the `integration` project. Consult the integration
...
+
+For integration tests, don't need to delete workspaces because it's taken care of
```
zloery, line 322: TestRunner isn't integrated with Janitor so it isn't doing this cleanup (though it could if we were testing against live environments and we built that integration). The cleanup is currently part of the tests (though since it's included in the `WorkspaceAllocateTestScriptBase` base class, developers don't need to think about it while writing tests)

README.md
melissachang: Updated:

https://github.com/DataBiosphere/terra-workspace-manager/blob/17d64b6d6cf4b891977230672b6f4cc9cd592105/README.md#cleaning-up-workspaces-in-tests

This has @zloery 's stamp of approval.

ddietterich (approved):

PR #753, opened 2022-08-08
Upgrade liquibase (WOR-441).

cahrens:
Addressing security vulnerability, no breaking changes noted between 4.2.1 and 4.8.0. https://github.com/liquibase/liquibase/releases

https://broadworkbench.atlassian.net/browse/WOR-441

PR #756, opened 2022-08-09
[WOR-441, WOR-442] Upgrade terra-common-lib to correct postgres and liquibase transitive dependency versions.

cahrens:
This is a follow-up PR from #753 and #754. In those I incorrectly hand-edited the lock files, which made the versions of postgres and liquibase in the integration project incorrect (ahead) with respect to their transitive dependency via `terra-common-lib`. Therefore I have upgraded `terra-common-lib`, and am now referencing the new version here (in both integration and service for consistency).

The lock file changes are the result of running `./gradlew :service:dependencies --write-locks` and `./gradlew :integration:dependencies --write-locks`, which is the correct way to update those files. Note that liquibase and postgres versions have _not_ changed, which was the goal.


service/gradle/dependencies.gradle
```diff
@@ -20,7 +20,7 @@ dependencies {
   implementation group: "io.opencensus", name: "opencensus-contrib-http-jaxrs", version: "0.28.3"
 
   // Get stairway via TCL
-  implementation("bio.terra:terra-common-lib:0.0.61-SNAPSHOT") {
+  implementation("bio.terra:terra-common-lib:0.0.66-SNAPSHOT") {
```
cahrens, line 23: Updating for consistency with the version used by integration.

integration/build.gradle
```diff
@@ -29,9 +29,9 @@ dependencies {
         googleNotebook = "v1-rev20201110-1.30.10"
 
         datarepoClient = "1.0.155-SNAPSHOT"
-        testRunnerVersion = "0.1.3-SNAPSHOT"
+        testRunnerVersion = "0.1.4-SNAPSHOT"
```
cahrens, line 32: I was getting a ClassDefNotFound error from test runner code (after upgrading terra-common-lib version), and this fixed it.

integration/build.gradle
```diff
@@ -15,7 +15,7 @@ dependencies {
         guava = "29.0-jre"
         jackson = "2.10.2"
         jersey = "2.32"
-        kubernetesClient = "10.0.0"
+        kubernetesClient = "16.0.0"
```
cahrens, line 18: I noticed that we had a lot of compile-time dependencies on old versions of libraries (with newer versions being used as a runtime dependency). When I looked at the dependency tree, I noticed this was coming from `kubernetesClient`, which was at a version from 2020. The newer versions include a lot of dependabot upgrades (https://github.com/kubernetes-client/java/releases?page=1), and given that this is a test dependency only, upgrading seems low risk if all the tests pass.

integration/gradle.lockfile
```diff
@@ -218,51 +229,51 @@ org.ow2.asm:asm:9.3=spotbugs
 org.postgresql:postgresql:42.3.3=runtimeClasspath,testRuntimeClasspath
```
cahrens, line 229: Still at the version we want!

blakery (approved):

yunip20 (approved):

PR #758, opened 2022-08-09
[PF-1875] Add minimumHighestRole to ListWorkspaces request 

melissachang:
Also delete test from WorkspaceServiceTest.listWorkspacesAndHighestRoles_existing because same functionality is now tested in WorkspaceApiControllerConnectedTest.

Manually tested
- Configured workspace where second user only has discoverer role
- Second user runs ListWorkspaces with minimumHighestRole unset, no workspace returned
- Second user runs ListWorkspaces with minimumHighestRole set to DISCOVERER, subset of workspace is returned ([screenshot](https://screenshot.googleplex.com/6xYX7V9GiFkQz3T))
- Workspace owner calls ListWorkspace with minimumHighestRole set to DISCOVERER, full workspace is returned

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/iam/model/WsmIamRole.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/model/WsmIamRole.java
@@ -73,6 +73,22 @@ public enum WsmIamRole {
         String.format("Workspace %s has unexpected roles: %s", workspaceId.toString(), roles));
   }
 
+  public static boolean roleAtLeastAsHighAs(WsmIamRole minimumRole, WsmIamRole roleToCheck) {
+    return switch (minimumRole) {
+      case APPLICATION -> roleToCheck == WsmIamRole.APPLICATION;
+      case OWNER -> roleToCheck == WsmIamRole.APPLICATION || roleToCheck == WsmIamRole.OWNER;
+      case WRITER -> roleToCheck == WsmIamRole.APPLICATION
+          || roleToCheck == WsmIamRole.OWNER
+          || roleToCheck == WsmIamRole.WRITER;
+      case READER -> roleToCheck == WsmIamRole.APPLICATION
+          || roleToCheck == WsmIamRole.OWNER
+          || roleToCheck == WsmIamRole.WRITER
+          || roleToCheck == WsmIamRole.READER;
+      case DISCOVERER -> true;
+      case MANAGER -> throw new InternalServerErrorException("Unexpected workspace MANAGER role");
+    };
+  }
+
   public ApiIamRole toApiModel() {
     return apiRole;
   }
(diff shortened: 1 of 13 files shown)


service/src/main/java/bio/terra/workspace/service/iam/model/WsmIamRole.java
```diff
@@ -73,6 +73,22 @@ public static Optional<WsmIamRole> getHighestRole(UUID workspaceId, List<WsmIamR
         String.format("Workspace %s has unexpected roles: %s", workspaceId.toString(), roles));
   }
 
+  public static boolean roleAtLeastAsHighAs(WsmIamRole minimumRole, WsmIamRole roleToCheck) {
```
ddietterich, line 76: I suggest instead of doing this as a `switch`, you put the logic into the enumeration itself.
You could either put a comparison lambda directly in the enum or put an array of roles into the enum and test it in this method.

ddietterich (approved):
Two suggestions - otherwise, lgtm

service/src/main/java/bio/terra/workspace/service/iam/model/WsmIamRole.java
melissachang: I tried putting lambda in enum. Java said `Illegal self reference`:

image

Then I tried storing `Set<WsmIamRole> higherRoles` on enum. Java said `Illegal forward reference`.

image

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -262,6 +271,36 @@ public WsmIamRole getHighestRole(UUID uuid, AuthenticatedUserRequest userRequest
...
+    Workspace.Builder strippedWorkspace =
+        new Workspace.Builder()
+            .workspaceId(fullWorkspace.getWorkspaceId())
+            .userFacingId(fullWorkspace.getUserFacingId())
+            .workspaceStage(fullWorkspace.getWorkspaceStage());
+    if (fullWorkspace.getDisplayName().isPresent()) {
+      strippedWorkspace.displayName(fullWorkspace.getDisplayName().get());
+    }
+    if (fullWorkspace.getSpendProfileId().isPresent()) {
+      strippedWorkspace.spendProfileId(fullWorkspace.getSpendProfileId().get());
```
cbookg, line 286: this seems like a strange field to make available?

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -262,6 +271,36 @@ public WsmIamRole getHighestRole(UUID uuid, AuthenticatedUserRequest userRequest
...
+    }
+    Map<String, String> strippedProperties = new HashMap<>();
+    if (fullWorkspace.getProperties().containsKey(Properties.SHORT_DESCRIPTION)) {
```
cbookg, line 289: I think we need to expose the property that says it is a dataset?

cbookg (approved):

service/src/main/java/bio/terra/workspace/service/iam/model/WsmIamRole.java
ddietterich: TIL...

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
melissachang: Done

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
melissachang: Done, good point

PR #761, opened 2022-08-10
[PF-1875] Add minimumHighestRole to GetWorkspace/GetWorkspaceByUserFacingId requests

melissachang:
Also deleted some redundant tests from WorkspaceApiControllerTest.java

Manual testing:
- Added second user as discover
- Called getWorkspace/getWorkspaceByUserFacingId with various combinations of requester, minimumHighestRole

### DIFF
--- a/integration/src/main/java/scripts/testscripts/CloneWorkspace.java
+++ b/integration/src/main/java/scripts/testscripts/CloneWorkspace.java
@@ -269,7 +269,7 @@ public class CloneWorkspace extends WorkspaceAllocateWithPolicyTestScriptBase {
     assertNotNull(destinationWorkspaceId, "Destination workspace ID available immediately.");
     assertEquals(destinationUserFacingId, cloneResult.getWorkspace().getDestinationUserFacingId());
     final WorkspaceDescription destinationWorkspaceDescription =
-        cloningUserWorkspaceApi.getWorkspace(destinationWorkspaceId);
+        cloningUserWorkspaceApi.getWorkspace(destinationWorkspaceId, /*minimumHighestRole=*/ null);
     assertNotNull(
         destinationWorkspaceDescription,
         "Destination workspace is available in DB immediately after return from cloneWorkspace().");
@@ -300,7 +300,9 @@ public class CloneWorkspace extends WorkspaceAllocateWithPolicyTestScriptBase {
         "Source workspace ID reported accurately.");
     assertEquals(
         destinationWorkspaceDescription.getProperties(),
-        sourceOwnerWorkspaceApi.getWorkspace(getWorkspaceId()).getProperties(),
+        sourceOwnerWorkspaceApi
+            .getWorkspace(getWorkspaceId(), /*minimumHighestRole=*/ null)
+            .getProperties(),
         "Properties cloned successfully");
 
     // Verify shared GCS bucket succeeds and is populated
@@ -343,7 +345,7 @@ public class CloneWorkspace extends WorkspaceAllocateWithPolicyTestScriptBase {
 
     // We need to get the destination bucket name and project ID
     final WorkspaceDescription destinationWorkspace =
-        cloningUserWorkspaceApi.getWorkspace(destinationWorkspaceId);
+        cloningUserWorkspaceApi.getWorkspace(destinationWorkspaceId, /*minimumHighestRole=*/ null);
     assertEquals(destinationUserFacingId, destinationWorkspace.getUserFacingId());
     final String destinationProjectId = destinationWorkspace.getGcpContext().getProjectId();
     final var clonedSharedBucket =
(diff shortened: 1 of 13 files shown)


integration/src/main/java/scripts/testscripts/CloneWorkspace.java
```diff
@@ -269,7 +269,7 @@ protected void doUserJourney(
     assertNotNull(destinationWorkspaceId, "Destination workspace ID available immediately.");
     assertEquals(destinationUserFacingId, cloneResult.getWorkspace().getDestinationUserFacingId());
     final WorkspaceDescription destinationWorkspaceDescription =
-        cloningUserWorkspaceApi.getWorkspace(destinationWorkspaceId);
+        cloningUserWorkspaceApi.getWorkspace(destinationWorkspaceId, /*minimumHighestRole=*/ null);
```
ddietterich, line 272: An alternative would be to have `getWorkspace` do what it did and add a `getWorkspaceFiltered` that would take the extra arg. I would guess that 95% of our getWorkspace usage is with the current semantics.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -242,15 +242,20 @@ private ApiWorkspaceDescription buildWorkspaceDescription(
               .orElse(Collections.emptyList());
     }
 
+    // When we have another cloud context, we will need to do a similar retrieval for it.
+    var createDetailsOptional = workspaceActivityLogDao.getCreateDetails(workspaceUuid);
+    var lastChangeDetailsOptional = workspaceActivityLogDao.getLastUpdateDetails(workspaceUuid);
+
+    if (highestRole == WsmIamRole.DISCOVERER) {
+      workspace = Workspace.stripWorkspaceForRequesterWithOnlyDiscovererRole(workspace);
```
ddietterich, line 250: Self-documenting code to the max! 😄

service/src/main/java/bio/terra/workspace/service/workspace/model/Workspace.java
```diff
@@ -191,4 +192,43 @@ public Workspace build() {
...
+    if (fullWorkspace.getProperties().containsKey(Properties.SHORT_DESCRIPTION)) {
+      strippedProperties.put(
+          Properties.SHORT_DESCRIPTION,
+          fullWorkspace.getProperties().get(Properties.SHORT_DESCRIPTION));
+    }
+    if (fullWorkspace.getProperties().containsKey(Properties.VERSION)) {
+      strippedProperties.put(
+          Properties.VERSION, fullWorkspace.getProperties().get(Properties.VERSION));
+    }
+    if (strippedProperties.size() > 0) {
```
ddietterich, line 228: I don't think you need this `if`. Assigning an empty map is fine. If you don't assign it, the builder turns around and creates one.

ddietterich (approved):
Two suggestions. Otherwise lgtm

service/src/main/java/bio/terra/workspace/service/workspace/model/Workspace.java
```diff
@@ -191,4 +192,43 @@ public Workspace build() {
...
+   * If requester only has discoverer role, they can only see a subset of workspace. They cannot
+   * see:
+   *
+   * <ul>
+   *   <li>Workspace description
+   *   <li>Spend profile
+   *   <li>Can see special type, short description and version properties, and no other properties
+   * </ul>
+   */
+  public static Workspace stripWorkspaceForRequesterWithOnlyDiscovererRole(
```
cbookg, line 206: I thought you already added this function for the "list" PR?

cbookg (approved):

integration/src/main/java/scripts/testscripts/CloneWorkspace.java
melissachang: We have to pass in `minimumHighestRole` because this is the API as defined by YAML. This PR added `minimumHighestRole` as an optional query parameter.

Unfortunately, the way we generate our Java client, you pass into something for each query parameter, even if you don't want to set that parameter. I explored this and found how to generate Java clients without this behavior. Basically, you want to get your mustache from https://github.com/OpenAPITools/openapi-generator and not https://github.com/swagger-api/swagger-codegen-generators, and use `x-group-parameters` and/or `useSingleRequestParameter`.

service/src/main/java/bio/terra/workspace/service/workspace/model/Workspace.java
melissachang: Done

service/src/main/java/bio/terra/workspace/service/workspace/model/Workspace.java
melissachang: In List PR, I added this to `WorkspaceService.java`. In this PR, I moved to `model/Workspace.java`, since controller also needs it.

PR #766, opened 2022-08-12
PF-1871: Copy policy attributes on workspace clone

mikenev:
Updates for latest TPS package. Uses clone functionality from it.


service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiController.java
```diff
@@ -54,4 +54,10 @@ public ResponseEntity<ApiTpsPaoGetResult> getPao(UUID objectId) {
     ApiTpsPaoGetResult result = tpsApiDispatch.getPao(bearerTokenFactory.from(request), objectId);
     return new ResponseEntity<>(result, HttpStatus.OK);
   }
+
+  @Override
+  public ResponseEntity<Void> clonePao(UUID objectId, UUID cloneObjectId) {
```
melissachang, line 59: Basic TPS question - Does TPS have access control, or can anyone call TPS?

For WSM, we check access control (is requester allowed to perform this action) in controller ([example](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java#L281)). I don't see any access control checks, so was curious.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -228,19 +228,7 @@ private ApiWorkspaceDescription buildWorkspaceDescription(
...
-      Optional<ApiTpsPaoGetResult> workspacePao =
-          tpsApiDispatch.getPaoIfExists(
-              new BearerToken(userRequest.getRequiredToken()), workspaceUuid);
-      workspacePolicies =
-          workspacePao
-              .map(ApiTpsPaoGetResult::getEffectiveAttributes)
-              .map(ApiTpsPolicyInputs::getInputs)
-              .orElse(Collections.emptyList());
-    }
+    List<ApiTpsPolicyInput> workspacePolicies = getWorkspacePolicies(userRequest, workspaceUuid);
```
melissachang, line 231: optional

To me, this is a little cleaner, WDYT

```
List<ApiTpsPolicyInput> workspacePolicies = featureConfiguration.isTpsEnabled() ? getWorkspacePolicies(userRequest, workspaceUuid) : Collections.emptyList();
```

melissachang (commented):
Can you add a test for this? You can change WorkspaceApiControllerTest.cloneWorkspace() to assert that policies are cloned.

service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiDispatch.java
```diff
@@ -18,12 +19,9 @@
...
 import bio.terra.workspace.generated.model.ApiTpsPolicyPair;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
-import java.util.stream.Collectors;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import java.util.*;
```
zloery, line 24: nit: avoid wildcard imports

service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiDispatch.java
```diff
@@ -114,27 +117,24 @@ private static PolicyInput policyInputFromApi(ApiTpsPolicyInput apiInput) {
...
 
-    Map<String, String> data;
-    if (apiInput.getAdditionalData() == null) {
-      // Ensure we always have a map, even if it is empty.
-      data = new HashMap<>();
-    } else {
-      data =
-          apiInput.getAdditionalData().stream()
-              .collect(Collectors.toMap(ApiTpsPolicyPair::getKey, ApiTpsPolicyPair::getValue));
+    Multimap<String, String> data = ArrayListMultimap.create();
```
zloery, line 120: is it valid to have multiple `additionalData` fields with the same key?

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -34,11 +37,7 @@
 import bio.terra.workspace.service.workspace.model.WorkspaceConstants.Properties;
 import com.google.common.base.Preconditions;
 import io.opencensus.contrib.spring.aop.Traced;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
```
zloery, line 40: nit: wildcard import

service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiController.java
zloery, line 59: could you add a test for this method to `TpsBasicPaoTest` where we test other methods in this controller?

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -448,7 +450,21 @@ public String cloneWorkspace(
...
-    // TODO(PF-1871) copy policy on workspace clone
+    if (features.isTpsEnabled()) {
+      // New workspaces will always be created with empty policies, but some workspaces predate
+      // policy and so will not have associated PAOs.
+      Optional<ApiTpsPaoGetResult> workspacePao =
+          tpsApiDispatch.getPaoIfExists(
+              new BearerToken(userRequest.getRequiredToken()), sourceWorkspace.getWorkspaceId());
+
+      if (workspacePao.isPresent()) {
+        tpsApiDispatch.clonePao(
```
zloery, line 461: I think cloning the PAO needs to be a step in `CloneGcpWorkspaceFlight` rather than here own its own. 
When it's outside a flight like this it won't be undone, so if the flight in the `createWorkspace` call below fails, we will undo all the workspace creation work and return an error but won't undo the policy clone, and would have a policy object for a workspace which doesn't exist

service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiDispatch.java
melissachang: FYI - If you search Terra onboarding doc for "Java imports", it tells you what Intellij setting to change

openapi/src/parts/tps.yaml
```diff
@@ -53,6 +53,27 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
+    post:
+      summary: clone a policy attribute object
```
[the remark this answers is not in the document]
mikenev: Agreed, looks nicer.

openapi/src/parts/tps.yaml
```diff
@@ -53,6 +53,27 @@ paths:
...
+      operationId: clonePao
+      tags: [Tps]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/TpsObjectId'
```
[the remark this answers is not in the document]
mikenev: Good idea, I was following the API doc but it makes more sense to have it defined like this.

service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiController.java
mikenev: done

service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiDispatch.java
mikenev: thanks, that was annoying.

service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiDispatch.java
mikenev: yes, it switched to a multimap.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
mikenev: I could go either way. I like the ternary, but the first thing the function does is check isTpsEnabled, so that feels a little funny. I could refactor it a little more but since you marked this as optional I'll just leave it as is.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
mikenev: I switched this to a step in the Clone flight. It still feels a little funny because the Create flight adds a default policy that we need to clean later. I think that's why I had it as a sync pre workspace step before.

melissachang (approved):
Could you add a comment to WorkspaceApiController.cloneWorkspace

// Clone policies in flight instead of here. That way, if flight fails, policies won't be cloned.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/ClonePolicyAttributesStep.java
```diff
@@ -0,0 +1,77 @@
...
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    final var destinationWorkspace =
+        context.getInputParameters().get(JobMapKeys.REQUEST.getKeyName(), Workspace.class);
+    final var userRequest =
+        context
+            .getInputParameters()
+            .get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+
+    tpsApiDispatch.deletePao(
```
zloery, line 73: nit: can you add a comment that `deletePao` doesn't throw an error if the object is missing/on repeated calls? it's hard to keep track of how each service behaves and not catching errors on retry is a common bug in steps

zloery (approved):

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
zloery: ah, good point. The whole pattern of creating the destination workspace in one flight and then cloning into it in another feels pretty weird, but I think that's required in order for the "clone workspace" job to live in the destination workspace and definitely gets outside the scope of this PR

PR #771, opened 2022-08-17
PF-1660: add a new test for post-startup.sh

yuhuyoyo:
passed nightly: https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/2886005594


integration/src/main/java/scripts/testscripts/PrivateControlledAiNotebookInstancePostStartup.java
```diff
@@ -0,0 +1,132 @@
...
+            resourceUserApi,
+            testValue,
+            /*postStartupScript=*/ StringUtils.isEmpty(LOCAL_BRANCH)
+                ? null
+                : String.format(
+                    "https://raw.githubusercontent.com/DataBiosphere/terra-workspace-manager/%s/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh",
+                    LOCAL_BRANCH));
+
+    AIPlatformNotebooks userNotebooks = ClientTestUtils.getAIPlatformNotebooksClient(resourceUser);
+    var resource = getNotebookResource(resourceUserApi, creationResult);
```
ddietterich, line 65: Using var like this obscures the type of resource. I'd prefer the type specified so I can read the code without looking at the other method to see what the type is.

ddietterich, line 65: Using var like this obscures the type of resource. I'd prefer the type specified so I can read the code without looking at the other method to see what the type is.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
```diff
@@ -55,7 +56,9 @@ public class CreateAiNotebookInstanceStep implements Step {
 
   /** Default post-startup-script when starting a notebook instance. */
   protected static final String DEFAULT_POST_STARTUP_SCRIPT =
-      "https://raw.githubusercontent.com/DataBiosphere/terra-workspace-manager/main/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh";
+      "https://raw.githubusercontent.com/DataBiosphere/terra-workspace-manager/%s/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh";
```
ddietterich, line 59: This seems like a very specific thing to be implemented inside of WSM. I think it should be a property in `application.yml`. It can be set to something other than the default by setting an environment variable.

ddietterich (commented):
I am uncomfortable with the way the post-startup script is located and modified for specific branches. My intuition is that it is at the wrong level.

It seems like there are several alternatives including making the script path a property or making the script a resource packaged in the WSM jar file.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
yuhuyoyo: the branch hash is already a property in the application.yml though.

yuhuyoyo:
I don't follow how does packaging the script in resource helps with versioning?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
ddietterich: I'm asking why the path to the startup script has to be hardcoded here?

ddietterich:
> I don't follow how does packaging the script in resource helps with versioning?

If the script traveled in the WSM jar file, then the version of WSM code would be guaranteed to be in sync with the script.
If the notebook VM needs to get the script via https, we would have to expose a WSM endpoint to deliver the script. That seems more direct and more easily testable than trying to sync between the code and github. It also would seem to improve reliability, since github.com would not be on the path of launching a notebook.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
yuhuyoyo: well :) that was some tech debt.

yuhuyoyo:
> > I don't follow how does packaging the script in resource helps with versioning?
> 
> If the script traveled in the WSM jar file, then the version of WSM code would be guaranteed to be in sync with the script. If the notebook VM needs to get the script via https, we would have to expose a WSM endpoint to deliver the script. That seems more direct and more easily testable than trying to sync between the code and github. It also would seem to improve reliability, since github.com would not be on the path of launching a notebook.

i like the idea of exposing a wSM endpoint. i can move the script to resource and set up an endpoint in a separate PR

ddietterich:
> i like the idea of exposing a wSM endpoint. i can move the script to resource and set up an endpoint in a separate PR

Sounds good to me.

ddietterich (approved):
Pls make a ticket for revising how the script is delivered.

integration/README.md
```diff
@@ -168,6 +168,26 @@ export TEST_RUNNER_SERVER_SPECIFICATION_FILE="workspace-local.json"
...
+This test is to make sure that changes to the post-startup.sh does not break.
+This test can only be run on nightly against real WSM environment and cannot be
+run on local server. This is because in the script, we point Terra CLI to
+a server. Processes inside the notebook won't be able to talk to the local
+server.
+
+To run this test locally:
+1. `export TEST_RUNNER_SERVER_SPECIFICATION_FILE="workspace-dev.json"` (see [this section](#Use-a-local-Workspace-Manager-server) for details) or 
+change it to `workspace-verily-dev.json` (you do not need to clone BroadInstitutes/terra-helmfile.git in this case.)
+2. In `PrivateControlledAiNotebookInstancePostStartup.java`, change `LOCAL_BRANCH` to your branch name.
```
zloery, line 191: nit: update comment, LOCAL_BRANCH is now read from an env var

integration/src/main/resources/suites/dev/FullIntegration.json
```diff
@@ -7,6 +7,7 @@
     "integration/BasicUnauthenticated.json",
     "integration/GetRoles.json",
     "integration/ListWorkspaces.json",
-    "integration/WorkspaceLifecycle.json"
+    "integration/WorkspaceLifecycle.json",
+    "integration/PrivateControlledAiNotebookInstancePostStartup.json"
```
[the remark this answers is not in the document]
ungwudik: Good call out.

@yuhuyoyo Feel free to also post in #workspaces channel like Zach suggested. But a yardstick we used earlier to decide what tests run in dev, alpha, & staging is `does the feature end up in Terra prod?`

.github/workflows/nightly-tests.yml
```diff
@@ -129,6 +129,18 @@ jobs:
           test-server: workspace-wsmtest.json
           test: suites/FullIntegration.json
 
+      - name: Extract branch name
+        id: extract_branch
+        run: echo "TEST_LOCAL_BRANCH=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
yuhuyoyo: i triggered a manual workflow with my local branch and I do see my newly added test get run and passed. https://github.com/DataBiosphere/terra-workspace-manager/runs/7925534809?check_suite_focus=true

integration/src/main/resources/suites/dev/FullIntegration.json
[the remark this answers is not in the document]
yuhuyoyo: got it. This is not used by CWB so I removed it.

zloery (approved):

PR #772, opened 2022-08-17
[DSEC-404] fix srcclr scans by specifying scope in srcclr.yml

TomConner:
Veracode SCA, aka srcclr, is used to scan for vulnerable 3rd party components. In WSM, srcclr was failing to find any code with default scope. Adding a configuration file to specify the productionRuntimeClasspath scope fixes the scan. 

The scan will still be run in a separate appsec repo on a weekly basis for now. If the team wants to set up a github action or gradle configuration to scan it on every PR, we in appsec can help make that happen, but that is neither urgent nor required.

### DIFF
--- a/srcclr.yml
+++ b/srcclr.yml
@@ -1,0 +1,1 @@
+scope: productionRuntimeClasspath


cahrens (approved):

ddietterich:
Hi Tom, I plan to enable dependabot eventually. We need to clean up underlying libraries from other repos before we apply it here. (Yesterday I enabled it for Stairway.)

Since there are 4 subprojects in WSM, what does `productionRuntimeClasspath` refer to?

TomConner:
`productionRuntimeClasspath` is a gradle dependencies scope. The srcclr scanner is run at root level and will potentially report on anything it finds underneath. It would be possible to narrow the scope if this generates too many irrelevant findings such as test dependencies that are not present in production. But it's simplest to run it like this.

ddietterich (approved):
OK, I guess we will see what happens

TomConner:
With regard to other libraries, we are planning to set up scanning on `terra-common-lib` separately

PR #774, opened 2022-08-18
[WOR-483] Do not hardcode SAM resource type

aherbst-broad:
## Context

We should not hardcode the resource type when determining user permissions as we mint a SAS token. For controlled application private containers (such as those created by leo), this means the user will not be able to mint a token for said container. 

Instead we should get it from the container resource instance. 

## This PR
* Refactors the `AzureControlledStorageResourceService` in the following ways:
    * Renames to `AzureStorageAccessService` -- this feels like it's closer what the class actually does
    * Extracts the fetch of the storage account keys to a helper class `StorageAccountKeyProvider`. This code was responsible for pulling in several dependencies (`AzureCloudContext`, `CrlService` to name a few) that was making unit testing difficult
     * Creates a related `StorageAccountKeyProviderUnitTest` that isolates the more in-depth azure mocking
    * Pushes the endpoint string interpolation to the `ControlledAzureStorageResource` class
 * Converts the `AzureControlledStorageAccessServiceTest` integration test into a unit test (`AzureStorageAccessServiceUnitTest`) that removes the workspace and stairway dependencies and just exercises the functionality of the class


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -31,43 +24,38 @@
...
+  public AzureStorageAccessService(
       SamService samService,
-      AzureCloudContextService azureCloudContextService,
-      CrlService crlService,
-      ControlledResourceService controlledResourceService,
-      AzureConfiguration azureConfiguration,
+      StorageAccountKeyProvider storageAccountKeyProvider,
       FeatureConfiguration features) {
     this.samService = samService;
-    this.azureCloudContextService = azureCloudContextService;
```
aherbst-broad, line 55: These dependencies are no longer needed now that the key retrieval has been factored into a separate class.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -89,21 +77,21 @@ private BlobContainerSasPermission getSasTokenPermissions(
     return BlobContainerSasPermission.parse(tokenPermissions);
   }
 
-  private StorageSharedKeyCredential getStorageAccountKey(
```
aherbst-broad, line 92: Refactored into `StorageAccountKeyProvider`

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -115,12 +103,15 @@ public AzureSasBundle createAzureStorageContainerSasToken(
     features.azureEnabledCheck();
 
     BlobContainerSasPermission blobContainerSasPermission =
-        getSasTokenPermissions(userRequest, storageContainerResource.getResourceId());
+        getSasTokenPermissions(
+            userRequest,
+            storageContainerResource.getResourceId(),
+            storageContainerResource.getCategory().getSamResourceName());
```
aherbst-broad, line 109: This is the actual fix; we are threading in the resource's Sam resource name rather than relying on the previously hardcoded value.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessServiceUnitTest.java
```diff
@@ -0,0 +1,272 @@
...
+        null,
+        privateResourceState,
+        accessScopeType,
+        managedByType,
+        null,
+        UUID.randomUUID(),
+        "fake");
+  }
+
+  private void assertValidToken(String sas, BlobContainerSasPermission expectedPermissions) {
```
aherbst-broad, line 75: I pulled this in from the `AzureControlledStorageResourceServiceTest`

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureControlledStorageResourceServiceTest.java
```diff
@@ -1,290 +0,0 @@
...
-import java.util.regex.Pattern;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.TestInstance;
-import org.springframework.beans.factory.annotation.Autowired;
-
-@TestInstance(TestInstance.Lifecycle.PER_CLASS)
-public class AzureControlledStorageResourceServiceTest extends BaseAzureTest {
```
aherbst-broad, line 48: I removed this test and refactored much of it into the `AzureStorageAccessServiceUnitTest` later in this PR. I felt that this test was doing too much, especially around workspace creation and stairway management.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -31,43 +24,38 @@
...
-      AuthenticatedUserRequest userRequest, UUID storageContainerUuid) {
+      AuthenticatedUserRequest userRequest, UUID storageContainerUuid, String samResourceName) {
     final List<String> containerActions =
         SamRethrow.onInterrupted(
             () ->
                 samService.listResourceActions(
-                    userRequest,
-                    SamConstants.SamResource.CONTROLLED_USER_SHARED,
-                    storageContainerUuid.toString()),
+                    userRequest, samResourceName, storageContainerUuid.toString()),
```
aherbst-broad, line 58: This is the actual fix; we are threading in the resource's Sam resource name rather than relying on the previously hardcoded value.

blakery (approved):
Great refactor: clean abstraction, easier to understand

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessServiceUnitTest.java
```diff
@@ -0,0 +1,278 @@
...
+    assertThat("SAS is https", protocolRegex.matcher(sas).find());
+    assertThat("SAS validity starts today", startTimeRegex.matcher(sas).find());
+    assertThat("SAS validity ends today", expiryTimeRegex.matcher(sas).find());
+    assertThat("SAS is for a container resource", signedResourceRegex.matcher(sas).find());
+    assertThat("SAS grants correct permissions", permissionsRegex.matcher(sas).find());
+  }
+
+  @Test
+  public void createAzureStorageContainerSasToken_readonly() throws InterruptedException {
+    var mockSam = mock(SamService.class);
```
zloery, line 104: For mocking out Spring beans, you can use the `@MockBean` annotation to avoid repeating this setup at the start of every test (e.g. https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/test/java/bio/terra/workspace/service/workspace/GcpCloudContextUnitTest.java#L42)
It also helps by having the services your tests depends on listed at the top of the suite as either mocked (`@MockBean`) or not (`@Autowired`)

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/StorageAccountKeyProviderUnitTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentMatchers;
+
+public class StorageAccountKeyProviderUnitTest extends BaseUnitTest {
+
+  @Test
+  public void getStorageAccountKey() {
+    var azureCloudContextService = mock(AzureCloudContextService.class);
```
zloery, line 26: Same thing with these mocks - having the dependencies marked with `@MockBean` outside the test case itself and setup code in a shared `@BeforeEach` method could help clarify what's setup and what's test logic

zloery (approved):

marctalbott (approved):

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessServiceUnitTest.java
aherbst-broad: 👍 will do

PR #778, opened 2022-08-22
PF-1660: update nightly-test

yuhuyoyo:


yuhuyoyo:
> It seems like both formats should work, do they behave differently?

oh.. if so I can drop it. I just saw other GHA is using the latter. As I've only tested the change by manual trigger...

ddietterich (approved):

PR #779, opened 2022-08-23
[WOR-466] Wire up delete storage container API

aherbst-broad:
We need to be able to delete azure storage containers, this wires up the relevant API call.

### DIFF
--- a/openapi/src/parts/controlled_azure_storage_container.yaml
+++ b/openapi/src/parts/controlled_azure_storage_container.yaml
@@ -43,6 +43,8 @@ paths:
           $ref: '#/components/responses/DeleteControlledAzureResourceResponse'
         '403':
           $ref: '#/components/responses/PermissionDenied'
+        '404':
+          $ref: "#/components/responses/NotFound"
         '500':
           $ref: '#/components/responses/ServerError'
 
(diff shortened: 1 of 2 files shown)


blakery (approved):

openapi/src/parts/controlled_azure_storage_container.yaml
```diff
@@ -24,6 +24,28 @@ paths:
...
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/DeleteControlledAzureResourceRequest'
+      responses:
+        '200':
+          $ref: '#/components/responses/DeleteControlledAzureResourceResponse'
+        '403':
```
marctalbott, line 44: Should we document the 404 when resource or workspace isn't found? I see that 404 isn't documented in other similar cases, so maybe there's a reason to omit it

marctalbott (approved):

openapi/src/parts/controlled_azure_storage_container.yaml
ddietterich: We do not return 404 if the target of the delete is not found. That is on purpose; it helps with idempotency.
However, if the containing workspace is not found, I would expect we would get a 404. There probably should be documented return and it is probably incorrect in the other similar cases.

aherbst-broad: Thanks both. I've added a 404 response for this endpoint, ticketed up the remediation for the rest as [WOR-490](https://broadworkbench.atlassian.net/browse/WOR-490)

PR #780, opened 2022-08-23
add an action to run spotlessApply

yuhuyoyo:


ddietterich (approved):
This is fine and I don't mind the extra check. But I thought that since we had the lint GHA running as a required step before merge, we had solved the problem of people merging not-spotless code. Can you explain how that can happen?

yuhuyoyo:
lint check has broken before so I want to have a fallback so we never need to manually create PR for clean up lint. there's a PR not too long ago for fixing lint in CLI

ddietterich:
Zach is out this week.

yuhuyoyo:
this is a sharable GHA. it can go into a sharable place instead of here. so closing this PR.

PR #783, opened 2022-08-25
[PF-1871] Stop sending policy information for most integration tests

zloery:
Part of #766 modified `WorkspaceAllocateTestScriptBase` to include policy information, but I forgot that several tests using this base run in environments which don't have TPS enabled. As a result, the basic integration tests are broken when they run against environments > dev (as WSM rejects requests including policy information if TPS is disabled). However, the service itself is fine, so I don't think we need to roll #766 back entirely.

This splits the policy into a separate base class for now, with the expectation that we'll delete this class once TPS is enabled in all environments.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/CloneWorkspace.java
+++ b/integration/src/main/java/scripts/testscripts/CloneWorkspace.java
@@ -58,9 +58,9 @@ import scripts.utils.ClientTestUtils;
 import scripts.utils.CloudContextMaker;
 import scripts.utils.GcsBucketObjectUtils;
 import scripts.utils.GcsBucketUtils;
-import scripts.utils.WorkspaceAllocateTestScriptBase;
+import scripts.utils.WorkspaceAllocateWithPolicyTestScriptBase;
 
-public class CloneWorkspace extends WorkspaceAllocateTestScriptBase {
+public class CloneWorkspace extends WorkspaceAllocateWithPolicyTestScriptBase {
   private static final Logger logger = LoggerFactory.getLogger(CloneWorkspace.class);
   private ControlledGcpResourceApi cloningUserResourceApi;
   private CreatedControlledGcpGcsBucket copyDefinitionSourceBucket;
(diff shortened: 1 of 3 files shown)


mikenev (approved):
Thanks for fixing that up.

melissachang (approved):
Does WorkspaceAllocateTestScriptBase.java have access to policy feature? Could it say

```
if policy feature
  Add policies to workspace
```

zloery:
> Does WorkspaceAllocateTestScriptBase.java have access to policy feature? Could it say
> 
> ```
> if policy feature
>   Add policies to workspace
> ```

I don't believe so. We don't expose whether features are enabled anywhere, so the client can't tell whether TPS is enabled or not until we try to use it. We could do something like
```
try:
    create workspace with policy
catch 501 TPS disabled:
    create workspace without policy
```
that feels messier to me, but it would mean less code duplication

melissachang:
> > Does WorkspaceAllocateTestScriptBase.java have access to policy feature? Could it say
> > ```
> > if policy feature
> >   Add policies to workspace
> > ```
> 
> I don't believe so. We don't expose whether features are enabled anywhere, so the client can't tell whether TPS is enabled or not until we try to use it. We could do something like
> 
> ```
> try:
>     create workspace with policy
> catch 501 TPS disabled:
>     create workspace without policy
> ```
> 
> that feels messier to me, but it would mean less code duplication

I don't mean to block this PR, up to you. Couple random thoughts.

Eventually Verily will adopt a real feature/experiment software (eg LaunchDarkly). With that, I assume we wouldn't have this problem; tests would be able to tell if a feature is on.

I wonder if it's worth exposing features in the status endpoint, for things like this.

ddietterich:
> I wonder if it's worth exposing features in the status endpoint, for things like this.

I think that is a good idea. I think we would want it on the `/version` endpoint, since that has the other information about CLI compatibility, etc. @melissachang could you make a Jira ticket for this?

zloery:
That sounds like a better solution for this overall - filed PF-1948 and updated the TODO in this PR

PR #785, opened 2022-08-26
PF-1871 - Clone policy step will return right away if TPS is not enabled

mikenev:
Not all environments will have TPS enabled. We don't need to attempt to clone policies if the feature is not turned on.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneGcpWorkspaceFlight.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneGcpWorkspaceFlight.java
@@ -28,10 +28,9 @@ public class CloneGcpWorkspaceFlight extends Flight {
         RetryRules.cloud());
     addStep(new AwaitCreateGcpContextFlightStep(), RetryRules.cloudLongRunning());
 
-    addStep(
-        new ClonePolicyAttributesStep(
-            flightBeanBag.getTpsApiDispatch(), flightBeanBag.getFeatureConfiguration()),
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneGcpWorkspaceFlight.java
```diff
@@ -28,7 +28,10 @@ public CloneGcpWorkspaceFlight(FlightMap inputParameters, Object applicationCont
         RetryRules.cloud());
     addStep(new AwaitCreateGcpContextFlightStep(), RetryRules.cloudLongRunning());
 
-    addStep(new ClonePolicyAttributesStep(flightBeanBag.getTpsApiDispatch()), RetryRules.cloud());
+    addStep(
```
ddietterich, line 31: Might be simpler to put the feature check here and not put the step in the flight if TPS is not enabled.

That would also prevent "flip-flop" if the flight started with the feature disabled and, after a failover, tried to undo with the feature enabled.

ddietterich (commented):
I have a suggestion...

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneGcpWorkspaceFlight.java
mikenev: Makes sense, I'll move it here.

zloery (approved):

ddietterich (approved):

PR #786, opened 2022-08-29
[PF-1041] clone enabled application in workspace(product code & test)

ginaygoogle:
Firstly, we create a workspace and enable the TestWsmApp application. The workspace would contain an enabled application TestWsmApp and a disabled application Leo. 
Screen Shot 2022-08-29 at 12 25 09 PM
Then we clone the workspace, and expect that TestWsmApp will be enabled in the cloned workspace.
Screen Shot 2022-08-29 at 12 24 36 PM

Code explain:

- create a `FindEnabledApplicationStep.java`  for retrieving multiple applications that needs to be enabled (return a List of applicationId which is enabled) from the sourceworkspace under terra/workspace/service/resource/controlled/flight/clone/workspace with doStep and undoStep.
- Modify the `ApplicationAbleFlight` to support plural enabled applications. if there is only one application need to enable. the list to enable only contain one applicationId.
- Due to the changes in `ApplicationAbleFlight`,  `WsmApplicationService.java` should change following with the list. 
- No need to adding `AwaitCloneApplicationAbleFlightStep.java`  due to the enabled application does not cause time.
- add the luanchStep in the end of the `CloneGcpWorkspaceFlight.java`.
- Follow the existing code, only one test file  `FindEnabledApplicationStepTest` is needed. Do not need to test the luanchStep and awaitStep. But should have a connectedTest in the `cloneWorkspace`.


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchEnableApplicationFlightStep.java
```diff
@@ -0,0 +1,78 @@
...
+    subflightInputParameters.put(WorkspaceFlightMapKeys.APPLICATION_ID, applicationIdList);
+    subflightInputParameters.put(WsmApplicationKeys.APPLICATION_ABLE_ENUM, ableEnum);
+
+    // fields normally set by JobBuilder for identifying jobs
+    subflightInputParameters.put(
+        WorkspaceFlightMapKeys.WORKSPACE_ID, destinationWorkspace.getWorkspaceId().toString());
+    subflightInputParameters.put(
+        JobMapKeys.DESCRIPTION.getKeyName(),
+        String.format(
+            "Enable application %s in workspace %s",
```
yuhuyoyo, line 56: applications

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/FindEnabledApplicationStep.java
```diff
@@ -0,0 +1,71 @@
...
+      batch = applicationDao.listWorkspaceApplications(sourceWorkspaceId, offset, limit);
+      offset += limit;
+      final List<WsmWorkspaceApplication> enabledApplication =
+          batch.stream().filter(WsmWorkspaceApplication::isEnabled).toList();
+      enabledApplication.forEach(r -> result.add(r.getApplication().getApplicationId()));
+    } while (batch.size() == limit);
+
+    logger.info(
+        "Will enable applications {} in workspace {}",
+        result.stream().collect(Collectors.joining(", ")),
```
yuhuyoyo, line 51: why not just use the default toString?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchEnableApplicationFlightStep.java
```diff
@@ -0,0 +1,78 @@
...
+            applicationIdList.toString(), destinationWorkspace.getWorkspaceId().toString()));
```
yuhuyoyo, line 57: nit: toString is not needed.

service/src/main/java/bio/terra/workspace/service/workspace/WsmApplicationService.java
```diff
@@ -70,25 +70,29 @@ public WsmApplicationService(
...
-    return commonAbleJob(userRequest, workspace, applicationId, AbleEnum.DISABLE);
+    return commonAbleJob(
+        userRequest, workspace, new ArrayList<String>(List.of(applicationId)), AbleEnum.DISABLE);
   }
 
   public WsmWorkspaceApplication enableWorkspaceApplication(
       AuthenticatedUserRequest userRequest, Workspace workspace, String applicationId) {
-    return commonAbleJob(userRequest, workspace, applicationId, AbleEnum.ENABLE);
+    return commonAbleJob(
+        userRequest, workspace, new ArrayList<String>(List.of(applicationId)), AbleEnum.ENABLE);
```
yuhuyoyo, line 80: why wrapping in an arrayList?

service/src/main/java/bio/terra/workspace/service/workspace/flight/application/able/ApplicationAbleFlight.java
```diff
@@ -32,37 +33,42 @@ public ApplicationAbleFlight(FlightMap inputParameters, Object applicationContex
...
-    //   vis a vis the workspace. It does NOT mean the application is enabled.
-    addStep(
-        new ApplicationAblePrecheckStep(
-            beanBag.getApplicationDao(),
-            beanBag.getSamService(),
-            userRequest,
-            workspaceUuid,
-            applicationId,
-            ableEnum));
+    for (int i = 0; i < applicationIdList.size(); i++) {
```
yuhuyoyo, line 41: nit: for(String applicationId: applicationIdList):

yuhuyoyo (commented):
can we also add an integration test coverage here? https://cs.github.com/DataBiosphere/terra-workspace-manager/blob/e1ee36ef23912f43ba670e5ee5c55df125846bb5/integration/src/main/java/scripts/testscripts/CloneWorkspace.java?q=cloneworkspace#L63

just some nits.

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/FindEnabledApplicationStepTest.java
```diff
@@ -0,0 +1,93 @@
...
+            .applicationId(CARMEN_ID)
+            .displayName("Carmen")
+            .description("musical performance framework")
+            .serviceAccount("carmen@terra-dev.iam.gserviceaccount.com")
+            .state(WsmApplicationState.DEPRECATED);
+    disabledWorkspaceApplication =
+        new WsmWorkspaceApplication().application(disabledApplication).enabled(false);
+
+    findEnabledApplicationStep = new FindEnabledApplicationStep(mockApplicationDao);
+    doReturn(mockStairway).when(mockFlightContext).getStairway();
```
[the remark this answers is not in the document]
yuhuyoyo: i learned it through code reviews in my old team too. Google has a java readability process where there's a java expert review your code for some times and gives you guidance like such.

service/src/main/java/bio/terra/workspace/service/workspace/WsmApplicationService.java
yuhuyoyo: thanks for explaining. i didn't know that. can you add a comment there "wrap in arraylist for JSON serialization."

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneGcpWorkspaceFlight.java
```diff
@@ -29,5 +29,8 @@ public CloneGcpWorkspaceFlight(FlightMap inputParameters, Object applicationCont
 
     addStep(new LaunchCloneAllResourcesFlightStep(), RetryRules.cloud());
     addStep(new AwaitCloneAllResourcesFlightStep(), RetryRules.cloudLongRunning());
+
+    addStep(new FindEnabledApplicationStep(flightBeanBag.getApplicationDao()), RetryRules.cloud());
```
melissachang, line 33: Instead of adding these steps, I think it's cleaner to change `FindResourcesToCloneStep` to return enabled application resources.

melissachang (commented):
Nice! Just one comment about implementing `undo()`.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchEnableApplicationFlightStep.java
```diff
@@ -0,0 +1,76 @@
...
+      return StepResult.getStepResultSuccess();
+    } catch (StairwayException e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    return StepResult.getStepResultSuccess();
```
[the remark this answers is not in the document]
melissachang: I talked to Dan and Yu about how to test step undos. Some undo's are tested. Many aren't. But I think it's good practice to test.

Tests can call `.lastStepFailure(true)`. That will fail last step, causing all steps to undo.

Do you mind adding `WorkspaceServiceTest.cloneGcpWorkspaceUndo` ? It will call `.lastStepFailure(true)`. After the flight finishes, it should check:
- Destination workspace doesn't exist
- Cloud context (GCP project) doesn't exist

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -873,6 +873,56 @@ public void cloneGcpWorkspace() {
...
+    assertThrows(
+        InvalidResultStateException.class,
+        () ->
+            workspaceService.cloneWorkspace(
+                sourceWorkspace, USER_REQUEST, destinationLocation, destinationWorkspace));
+    assertThrows(
+        WorkspaceNotFoundException.class,
+        () -> workspaceService.getWorkspace(destinationWorkspace.getWorkspaceId()));
+
+    // Destination Workspace should have a GCP context
```
yuhuyoyo, line 916: should not have a GCP context.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -873,6 +873,56 @@ public void cloneGcpWorkspace() {
...
+    assertTrue(
+        gcpCloudContextService.getGcpCloudContext(destinationWorkspace.getWorkspaceId()).isEmpty());
+
+    // clean up
+    jobService.setFlightDebugInfoForTest(null);
```
yuhuyoyo, line 921: remove. not needed because @aftereach is doing this clean up.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -873,6 +873,56 @@ public void cloneGcpWorkspace() {
...
+    workspaceService.deleteWorkspace(sourceWorkspace, USER_REQUEST);
+    workspaceService.deleteWorkspace(destinationWorkspace, USER_REQUEST);
```
yuhuyoyo, line 923: not needed since this workspace does not exist as you asserted at line 913

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
```diff
@@ -12,7 +12,7 @@ public final class WorkspaceFlightMapKeys {
   public static final String DESCRIPTION = "descriptionId";
   public static final String USER_TO_REMOVE = "userToRemove";
   public static final String ROLE_TO_REMOVE = "roleToRemove";
-  public static final String APPLICATION_ID = "applicationId";
+  public static final String APPLICATION_IDS = "applicationIds";
```
yuhuyoyo, line 15: @ddietterich is it safe to rename key for flight? I suppose they only live as long as the flight so it's probably OK? I'd suggest we keep APPLICATION_ID and add APPLICATION_IDS.

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
ginaygoogle: Melissa comments this is not been used. And I also search in the whole repo, yes, no other place use it.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
ginaygoogle: Hi Yu, this is to remove the effect of `.lastStepFailure` (at L907) before calling deleteWorkspace in the next step. Or there will be a failure in the clean up.

yuhuyoyo: ah i see. could you add a comment? thanks

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneGcpWorkspaceFlight.java
```diff
@@ -18,6 +18,8 @@ public CloneGcpWorkspaceFlight(FlightMap inputParameters, Object applicationCont
     // 4. Clone Policy Attributes
     // 5. Launch a flight to clone all resources on the list
     // 5a. Await the clone all resources flight and build a response
+    // 6. Build a list of applications to enable
```
melissachang, line 21:
```
    // 6. Build a list of enabled applications
    // 6a. Launch a flight to enable those applications in destination workspace
```

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchEnableApplicationsFlightStep.java
```diff
@@ -0,0 +1,76 @@
...
+      return StepResult.getStepResultSuccess();
+    } catch (StairwayException e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    return StepResult.getStepResultSuccess();
```
melissachang, line 74: Hmm I had asked for this to be implemented. But maybe it doesn't need to be, since an earlier step's undo will delete destination workspace anyways. Can you add comment

```
// Don't need to disable applications, since an earlier step's undo will delete destination workspace.
```

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -823,6 +820,9 @@ public void cloneGcpWorkspace() {
         controlledResourceService.createControlledResourceSync(
             bucketResource, ControlledResourceIamRole.OWNER, USER_REQUEST, creationParameters);
 
+    // Enable an application.
+    appService.enableWorkspaceApplication(USER_REQUEST, sourceWorkspace, "leo");
```
melissachang, line 824: We usually don't hard-code strings that appear more than once. We usually make a constant. Put on line 124:

```
private static final APPLICATION = "leo";
```

One advantage - if you want to change the string, you only have to change one place.

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -865,11 +865,66 @@ public void cloneGcpWorkspace() {
...
+    assertThrows(
+        InvalidResultStateException.class,
+        () ->
+            workspaceService.cloneWorkspace(
+                sourceWorkspace, USER_REQUEST, destinationLocation, destinationWorkspace));
+    assertThrows(
+        WorkspaceNotFoundException.class,
+        () -> workspaceService.getWorkspace(destinationWorkspace.getWorkspaceId()));
+
+    // Destination Workspace should not have a GCP context
```
melissachang, line 919: Please also check that destination workspace doesn't exist. See line 853.

melissachang (approved):

PR #789, opened 2022-08-31
PF-1956: Clean up dead code-ish

yuhuyoyo:
this bucket name generation does not follow the new standard way. So removing this. at line 100, we are also checking for bucketname null and using the right way to generate one.


melissachang (approved):
Nice clean up

ginaygoogle (approved):

PR #790, opened 2022-08-31
[PF-1955] Add GOOGLE_PROJECT, OWNER_EMAIL, PET_SA_EMAIL environment variables.

mbookman:
Variables now show in Jupyter notebooks and terminals:

Here is the .bash_profile:

```
(base) jupyter@mbookman-20220831-223034:~$ cat ~/.bash_profile
export GOOGLE_PROJECT='terra-<stuff>'
export OWNER_EMAIL='<stuff>'
export PET_SA_EMAIL='pet-<stuff>@terra-<stuff>.iam.gserviceaccount.com'
eval "$(ssh-agent -s)"
```
Here's the relevant bits of the environment in a terminal in Jupyterlab:

```
(base) jupyter@mbookman-20220831-223034:~$ env | grep EMAIL
OWNER_EMAIL=<stuff>
PET_SA_EMAIL=pet-<stuff>@terra-<stuff>.iam.gserviceaccount.com
(base) jupyter@mbookman-20220831-223034:~$ env | grep GOOGLE
GOOGLE_PROJECT=terra-<stuff>
```

Added some small cleanup to the post_startup.sh where code was a little inconsistent.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -128,6 +128,9 @@ fi
 # Set variables into the .bash_profile such that they are available
 # to terminals, notebooks, and other tools
 
+# Keep in sync with terra CLI environment variables:
+# https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/app/CommandRunner.java#L88
+
 # These are environment variables that are set by Leonardo for
 # Cloud Environments (https://github.com/DataBiosphere/leonardo)
 
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -127,12 +127,18 @@ fi
 
 # Set variables into the .bash_profile such that they are available
 # to terminals, notebooks, and other tools
+#
+# We have new-style variables (eg GOOGLE_CLOUD_PROJECT) which are set here
+# and CLI (terra app execute env).
+# We also support a few variables set by Leonardo (eg GOOGLE_PROJECT).
+# Those are only set here and NOT in the CLI as they are intended just
+# to make porting existing notebooks easier.
 
 # Keep in sync with terra CLI environment variables:
-# https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/app/CommandRunner.java#L88
+# https://github.com/DataBiosphere/terra-cli/blob/14cf51dd809573c7ae9a3ef10ddd427fa057cb8f/src/main/java/bio/terra/cli/app/CommandRunner.java#L88
 
-# These are environment variables that are set by Leonardo for
-# Cloud Environments (https://github.com/DataBiosphere/leonardo)
+# *** Variables that are set by Leonardo for Cloud Environments
+# (https://github.com/DataBiosphere/leonardo)
 
 # OWNER_EMAIL is really the Terra user account email address
 
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -59,9 +59,9 @@ export NXF_MODE=google
 
 sudo apt-get update
 
-#######################################
+#########################################################
 # Install required JDK and set it as default (debian)
-#######################################
+#########################################################
 function install_java() {
   curl -Os https://download.oracle.com/java/17/latest/jdk-17_linux-x64_bin.deb
   sudo apt-get install -y ./jdk-17_linux-x64_bin.deb
@@ -72,7 +72,7 @@ function install_java() {
 if [[ -n "$(which java)" ]];
 then
   # Get the current major version of Java: "11.0.12" => "11"
-  readonly CUR_JAVA_VERSION=$(java -version 2>&1 | awk -F\" '{ split($2,a,"."); print a[1]}')
+  readonly CUR_JAVA_VERSION="$(java -version 2>&1 | awk -F\" '{ split($2,a,"."); print a[1]}')"
   if [[ "${CUR_JAVA_VERSION}" -lt 17 ]];
   then
     echo "Current Java version is ${CUR_JAVA_VERSION}, installing Java 17"
@@ -92,7 +92,7 @@ sudo mv nextflow /usr/bin/nextflow
 readonly CROMWELL_LATEST_VERSION="81"
 sudo -u "${JUPYTER_USER}" sh -c "mkdir -p /home/${JUPYTER_USER}/cromwell"
 sudo -u "${JUPYTER_USER}" sh -c "curl -LO https://github.com/broadinstitute/cromwell/releases/download/${CROMWELL_LATEST_VERSION}/cromwell-${CROMWELL_LATEST_VERSION}.jar"
-mv cromwell-${CROMWELL_LATEST_VERSION}.jar /home/${JUPYTER_USER}/cromwell/
+mv "cromwell-${CROMWELL_LATEST_VERSION}.jar" "/home/${JUPYTER_USER}/cromwell/"
 
 #Install cromshell
 sudo apt-get -y install mailutils
@@ -107,7 +107,7 @@ sudo cp terra /usr/bin/terra
 sudo -u "${JUPYTER_USER}" sh -c "terra config set browser MANUAL"
 # Set the CLI terra server based on the terra server that created the GCP notebook retrieved from
 # the VM metadata, if set.
-readonly TERRA_SERVER=$(get_metadata_value "instance/attributes/terra-cli-server")
+readonly TERRA_SERVER="$(get_metadata_value "instance/attributes/terra-cli-server")"
 if [[ -n "${TERRA_SERVER}" ]]; then
   sudo -u "${JUPYTER_USER}" sh -c "terra server set --name=${TERRA_SERVER}"
 fi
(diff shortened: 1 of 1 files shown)


melissachang (approved):
Nice! Feel free to close the ticket, if this is all you need.

I'm looking into test failures, I'm guessing due to https://github.com/DataBiosphere/terra-workspace-manager/pull/777

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -107,30 +107,56 @@ sudo cp terra /usr/bin/terra
...
-readonly TERRA_WORKSPACE=$(get_metadata_value "instance/attributes/terra-workspace-id")
+readonly TERRA_WORKSPACE="$(get_metadata_value "instance/attributes/terra-workspace-id")"
 if [[ -n "${TERRA_WORKSPACE}" ]]; then
   sudo -u "${JUPYTER_USER}" sh -c "terra workspace set --id=${TERRA_WORKSPACE}"
 fi
 
+# Set variables into the .bash_profile such that they are available
+# to terminals, notebooks, and other tools
+
+readonly GOOGLE_PROJECT="$(
```
yuhuyoyo, line 131: The TERRA env var which is used when running app command through TERRA has different variable naming from this. There we use GOOGLE_CLOUD_PROJECT, GOOGLE_SERVICE-ACCOUNT_EMAIL, (we don't have the equivalent of OWNER_EMAIL). I think we should at least keep the same naming. So the user doesn't get confused when the env is different when they are using terra CLI command vs not. 

As Melissa mentioned in the ticket, I think we should export all the TERRA env vars, not just these three. I propose we have a CLI command that returns all the env variable and we export them all to the bash_profile?

melissachang: Good point about name consistency.

To see env `terra cli` env vars, just run `terra app execute env`. I don't think we need a new command for this.

I think it's up to Matt if he wants to do the TERRA vars in this PR or not, it could always be a separate PR. :)

mbookman: Thanks Yu and Melissa. This is helpful additional context. For reference, in my notebook VM here are the relevant vars that I see:

```
(base) jupyter@<stuff>:~$ terra app execute env | sort | grep ^TERRA
TERRA_notebook_vm=projects/terra-<stuff>/locations/us-central1-a/instances/<stuff>
TERRA_vm_startup_test_bucket=gs://vm-startup-test-bucket

(base) jupyter@<stuff>:~$ terra app execute env | sort | grep ^GOOGLE
GOOGLE_CLOUD_PROJECT=terra-<stuff>
GOOGLE_PROJECT=terra-<stuff>
GOOGLE_SERVICE_ACCOUNT_EMAIL=pet-<stuff>@terra-<stuff>.iam.gserviceaccount.com
```

So addressing these as two separate variables - the TERRA variables are workspace resource references, while the GOOGLE/EMAIL variables are really Cloud-specific workspace properties.

I think exporting the TERRA variables is *likely* a nice feature to add, but I'm not sure. I would say that many notebooks would otherwise start with a series of "terra resolve" commands. But thinking this through a little more, if I'm working on a notebook, and I add a new resource, I'd like to be able to pick it up in my notebook right away. I could then use a "terra resolve" command in my notebook, but then my notebook reflects the order I was building things (sometimes I use the TERRA variable and sometimes I use the "terra resolve" command).

I want to put that aside for a moment. The specific implementation there is not obvious to me.

For the GOOGLE/EMAIL variables, things that do not change after the VM is started, and so setting the variables in the environment makes a lot of sense. It seems we have a few options for the notebook environments:

1. Keep the variable names [consistent with Leonardo](https://github.com/DataBiosphere/leonardo/blob/b1c19a89e374edcde524500e185e95080d22b657/http/src/main/resources/init-resources/startup.sh#L47)
2. Rationalize the variable naming as though Leonardo doesn't exist
3. Add both a new set of variables and support *key* Leonardo variables

I would lean towards (3) and keep a conservative eye on *key* Leonardo variables; I don't think it is a long list.

I think that the `GOOGLE_PROJECT` variable is very commonly used in app.terra.bio notebooks (any use of BigQuery, requester pays buckets, calls to `nextflow` and `dsub`, etc.)
The `OWNER_EMAIL` and `PET_SA_EMAIL` are used less frequently. The `OWNER_EMAIL` is an interesting one because it is really more of the "Terra user email" that happens today to be a Google account. The `PET_SA_EMAIL` is important functionally for things like `nextflow` and `dsub`.

I've implemented consistent with this. PTAL.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -107,30 +107,79 @@ sudo cp terra /usr/bin/terra
...
+####################################
+
 # Set the CLI terra workspace id using the VM metadata, if set.
-readonly TERRA_WORKSPACE=$(get_metadata_value "instance/attributes/terra-workspace-id")
+readonly TERRA_WORKSPACE="$(get_metadata_value "instance/attributes/terra-workspace-id")"
 if [[ -n "${TERRA_WORKSPACE}" ]]; then
   sudo -u "${JUPYTER_USER}" sh -c "terra workspace set --id=${TERRA_WORKSPACE}"
 fi
 
+# Set variables into the .bash_profile such that they are available
```
melissachang, line 128: Could it be organized like this

```
# Set variables into the .bash_profile such that they are available
# to terminals, notebooks, and other tools
# Leonardo environment variables are at # https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/startup.sh

readonly OWNER_EMAIL="$(
  sudo -u "${JUPYTER_USER}" sh -c "terra workspace describe --format=json" | \
  jq --raw-output ".userEmail")"
echo "export OWNER_EMAIL='${OWNER_EMAIL}'" >> "/home/${JUPYTER_USER}/.bash_profile"

# Leonardo sets GOOGLE_PROJECT.
readonly GOOGLE_PROJECT="$(
  sudo -u "${JUPYTER_USER}" sh -c "terra workspace describe --format=json" | \
  jq --raw-output ".googleProjectId")"
echo "export GOOGLE_PROJECT='${GOOGLE_PROJECT}'" >> "/home/${JUPYTER_USER}/.bash_profile"

# Also set GOOGLE_CLOUD_PROJECT in case people want that instead of GOOGLE_PROJECT
echo "export GOOGLE_CLOUD_PROJECT='${GOOGLE_PROJECT}'" >> "/home/${JUPYTER_USER}/.bash_profile"

# Leonardo sets PET_SA_EMAIL.
readonly PET_SA_EMAIL="$(
  sudo -u "${JUPYTER_USER}" sh -c "terra auth status --format=json" | \
  jq --raw-output ".serviceAccountEmail")"
echo "export PET_SA_EMAIL='${PET_SA_EMAIL}'" >> "/home/${JUPYTER_USER}/.bash_profile"

#  Also set GOOGLE_SERVICE_ACCOUNT_EMAIL in case people want that instead of PET_SA_EMAIL.
echo "export GOOGLE_SERVICE_ACCOUNT_EMAIL='${PET_SA_EMAIL}'" >> "/home/${JUPYTER_USER}/.bash_profile"

```

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
melissachang: I think I prefer 1. 3 works too.

Can we keep the terra cli env vars in sync with these env vars? This is where terra cli env are set:

https://cs.github.com/DataBiosphere/terra-cli/blob/70a0f6bffe2a00be51aa43f12c54631951e05b9c/src/main/java/bio/terra/cli/app/CommandRunner.java?q=GOOGLE_CLOUD_PROJECT#L89

Would you be up for doing the CLI PR too? :)

melissachang (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
melissachang: Also, can you add a comment

```
# Keep in sync with terra CLI environment variables:
https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/app/CommandRunner.java#L88
```

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
mbookman: I personally like the separation of the variables into the two separate categories (those set for consistency with Leonardo and those set for consistency with the Terra CLI).

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
mbookman: Added.

mbookman: I'm happy to do the CLI PR too.
I wasn't able to get the code to build, but will follow-up with you offline.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
melissachang: Now that I understand things better, SG.

Just checking -- if you run `terra` cli in notebook, it's ok that Leo variables aren't set, right?

Can you just clearly mark which 3 variables should be kept in sync with CLI, and which 3 should not?

```
# We have new-style variables (eg GOOGLE_CLOUD_PROJECT) which are set here and CLI (terra app execute env).
# We also support Leonardo variables (eg GOOGLE_PROJECT). Those are only set here and NOT CLI.
# Those are only for running notebooks ported from Leonardo. We encourage users to use the
# new-style variables.
```

melissachang (approved):

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -107,30 +107,82 @@ sudo cp terra /usr/bin/terra
...
+# to terminals, notebooks, and other tools
+
+# Keep in sync with terra CLI environment variables:
+# https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/app/CommandRunner.java#L88
```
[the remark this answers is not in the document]
mbookman: Good idea. Thanks.

mbookman:
@melissachang - I have updated the comments to explain the variables section a bit more.

@yuhuyoyo - the project environment variable was intentionally changed in https://github.com/DataBiosphere/terra-cli/pull/22, as nextflow (and other libraries out in the world) expect GOOGLE_CLOUD_PROJECT.

PR #791, opened 2022-09-01
PF-1956: clean up dead code.

yuhuyoyo:
Clean up an unused method `randomDatasetName`. Also if destination dataset name is specified, we should respect that.


melissachang (approved):
Cool thanks. Seems more like fixing code, rather than cleaning up dead code.

ginaygoogle (approved):

PR #792, opened 2022-09-01
[WIP] Add feature flag for azure landing zone. Update tests

sergiygetlin:
Changes:
-Add feature flag for Azure landing zone
-Update tests
-Update reference to the latest LZ library


service/local-dev/run_postgres.sh
```diff
@@ -9,7 +9,7 @@ start() {
     # start up postgres
     echo "starting up postgres container..."
     BASEDIR=$(dirname "$0")
-    docker create --name $CONTAINER --rm -e POSTGRES_PASSWORD=password -p "$POSTGRES_PORT:5432" postgres:$POSTGRES_VERSION
+    docker create --name $CONTAINER --rm -e POSTGRES_PASSWORD=password -p "$POSTGRES_PORT:5432" postgres:$POSTGRES_VERSION -c max_connections=150
```
sergiygetlin, line 12: This is to run unit tests successfully. It will be removed from here.

service/src/main/java/bio/terra/workspace/app/configuration/external/FeatureConfiguration.java
```diff
@@ -59,6 +68,12 @@ public void tpsEnabledCheck() {
     }
   }
 
+  public void azureLandingZoneEnabledCheck() {
+    if (!isAzureLandingZoneEnabled()) {
```
sergiygetlin, line 72: Should we also check here if Azure feature enabled?

sergiygetlin:
To keep it simple will reuse azure feature instead.

PR #794, opened 2022-09-01
PF-1880: Implement resource lineage and testing for controlled resources

yuhuyoyo:
I refactored the destination resource building to WorkspaceCloneUtils so that all the destination resource building logic are in one place and can use some shared methods. this has two benefits:
1. When adding another fields to resource common fields, the change can all happens in this one file. It's easy to miss when the logic are scattered in the steps in separate directories.
2. The common fields building logic is the same for cloning the controlled bq dataset and the controlled bucket. So I merged them into one common method. This avoid bugs where we set private resource state to INITALIZING for one kind of resource and forgot it for the other kind.
3. I added tests for WorkspaceCloneUtils.
4. Added connected test and integration tests as well

clone a dataset from workspace 1 to workspace 2:
8MbQoYhfSXD3Ezf

clone the dataset from workspace 2 to workspace 3:
9r8gRevVaDde96r


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CopyGcsBucketDefinitionStep.java
```diff
@@ -62,85 +61,71 @@ public CopyGcsBucketDefinitionStep(
...
-    final FlightMap inputParameters = flightContext.getInputParameters();
-    final FlightMap workingMap = flightContext.getWorkingMap();
+    FlightMap inputParameters = flightContext.getInputParameters();
+    FlightMap workingMap = flightContext.getWorkingMap();
     FlightUtils.validateRequiredEntries(
         inputParameters,
         ControlledResourceKeys.DESTINATION_WORKSPACE_ID,
         ControlledResourceKeys.DESTINATION_RESOURCE_ID);
-    final String resourceName =
+    var resourceName =
```
ddietterich, line 70: The rule for `var` is that we only use it if the type is obvious from the right-hand-side expression.
In this case, that is not the case. These should specify their types.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CopyGcsBucketDefinitionStep.java
```diff
@@ -62,85 +61,71 @@ public CopyGcsBucketDefinitionStep(
...
             ResourceKeys.RESOURCE_DESCRIPTION,
             ResourceKeys.PREVIOUS_RESOURCE_DESCRIPTION,
             String.class);
 
-    final PrivateResourceState privateResourceState =
-        sourceBucket.getAccessScope() == AccessScopeType.ACCESS_SCOPE_PRIVATE
-            ? PrivateResourceState.INITIALIZING
-            : PrivateResourceState.NOT_APPLICABLE;
-    final UUID destinationWorkspaceId =
+    var destinationWorkspaceId =
```
ddietterich, line 83: Here is a case where I do know the type (UUID), because it is on the RHS.

ddietterich (changes requested):
I want to make sure everyone is on the same page on the use of `var`.
When we are doing code reviews, it should be possible to understand the type of a variable without going and looking at another implementation. Therefore, the rule is that we only use `var` when it is obvious what the LHS type is. For example,
```
var foo = new ArrayList();
```
is fine because it is obvious what type foo is. But,
```
var bar = makeMeABar();
```
is not fine because the reviewer cannot tell what the type.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CopyGcsBucketDefinitionStep.java
yuhuyoyo: i 100% agree with you that we should only allows var when the type is obvious. In this case, the type is pretty obvious as `getInputParameterOrWorkingValue` always specified type https://cs.github.com/DataBiosphere/terra-workspace-manager/blob/9629c9fb57fd60b0ace6398fb91f117805dc8879/service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java?q=getInputParameterOrWorkingValue#L82.

melissachang: FWIW I agree with Dan. There's been times where I see `var` and I don't know what the type is. For this case, one has to be familiar with `getInputParameterOrWorkingValue` to know that it's a String.

yuhuyoyo: specified all the ones that are from custom methods.

ddietterich (approved):
Thanks, Yu!

integration/src/main/java/scripts/testscripts/ResourceLineage.java
```diff
@@ -78,24 +114,30 @@ protected void doCleanup(List<TestUserSpecification> testUsers, WorkspaceApi wor
...
             new CloneReferencedResourceRequestBody()
                 .cloningInstructions(CloningInstructionsEnum.REFERENCE)
                 .destinationWorkspaceId(workspaceId2),
             getWorkspaceId(),
-            bqDatasetWorkspace1ResourceId);
-    UUID resource2Id = resource2CloneResult.getResource().getMetadata().getResourceId();
-
+            referencedBqDatasetWorkspace1ResourceId);
+    var cloneReferencedBqDataset2ResourceId =
+        resource2CloneResult.getResource().getMetadata().getResourceId();
```
melissachang, line 132: nit blank line after 132

integration/src/main/java/scripts/testscripts/ResourceLineage.java
```diff
@@ -106,20 +148,167 @@ protected void doUserJourney(
...
+                controlledGcpResourceApi.getCloneBigQueryDatasetResult(
+                    getWorkspaceId(), cloneRequest.getJobControl().getId()),
+            CloneControlledGcpBigQueryDatasetResult::getJobReport,
+            Duration.ofSeconds(5));
+    assertJobSuccess(
+        "clone controlled BigQuery dataset 2",
+        resource2CloneResult.getJobReport(),
+        resource2CloneResult.getErrorReport());
+    var cloneControlledBqDataset2ResourceId =
+        resource2CloneResult.getDataset().getDataset().getMetadata().getResourceId();
```
melissachang, line 189: nit to make test easier to read/scan. To skim the test, I read the comments at the beginning of each block.

Newline after 189

Before 190

```
// Assert resource lineage on resource 2 in second workspace. There is one entry.
```

Same for cloneControlledGcsBucket

integration/src/main/java/scripts/testscripts/ResourceLineage.java
```diff
@@ -106,20 +148,167 @@ protected void doUserJourney(
...
+            cloneRequest2, workspaceId2, cloneControlledBqDataset2ResourceId);
+    resource3CloneResult =
+        ClientTestUtils.pollWhileRunning(
+            resource3CloneResult,
+            () ->
+                controlledGcpResourceApi.getCloneBigQueryDatasetResult(
+                    workspaceId2, cloneRequest2.getJobControl().getId()),
+            CloneControlledGcpBigQueryDatasetResult::getJobReport,
+            Duration.ofSeconds(5));
     // Assert resource lineage on resource 3.
```
melissachang, line 218: nit newline before 218, makes test easier to scan. Same for cloneControlledGcsBucket

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResource.java
```diff
@@ -222,15 +225,11 @@ public boolean equals(Object o) {
...
-    if (workspaceUuid != null
-        ? !workspaceUuid.equals(that.workspaceUuid)
-        : that.workspaceUuid != null) return false;
-    if (resourceId != null ? !resourceId.equals(that.resourceId) : that.resourceId != null)
-      return false;
-    if (name != null ? !name.equals(that.name) : that.name != null) return false;
-    if (description != null ? !description.equals(that.description) : that.description != null)
-      return false;
-    return cloningInstructions == that.cloningInstructions;
+    return Objects.equals(workspaceUuid, that.workspaceUuid)
```
melissachang, line 228: Add comment that we're not including resource lineage? Just to make it clear that it's on purpose..

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -1491,6 +1493,90 @@ void cloneGcsBucketDo() throws InterruptedException {
         expectedLifecycleRules, containsInAnyOrder(bucketInfo.getLifecycleRules().toArray()));
   }
 
+  @Test
+  void cloneGcsBucketTwice() throws InterruptedException {
+    ControlledGcsBucketResource resource =
+        ControlledResourceFixtures.makeDefaultControlledGcsBucketBuilder(workspace.getWorkspaceId())
```
melissachang, line 1499: opt nits

opt - call this `cloneGcsBucketTwice_assertResourceLineage` ?

opt - if we're already testing this in integration test we don't need this?

melissachang (approved):
Cool, lots of nice cleanups

PR #795, opened 2022-09-01
[no ticket] Minimize idle connections

ddietterich:
This change to the test configuration minimizes the number of connections held open by the connection pooling.
It seems to solve hitting the connection limit. Clearly more work is needed to manage this better in the tests, but
I think this will get us by for now.

Of course, we'll see when the GHA tests run whether it fixes things more generally.


melissachang (approved):
Yay, GHA unit tests passed. I also run unit tests locally and they passed. (I was able to repro problem locally.)

sergiygetlin (approved):

PR #797, opened 2022-09-02
Clean pools @PreDestroy; reduce variability of mocks

ddietterich:
This is a draft set of changes that further reduces the number of connections we leak when running the unit tests. There are two main changes in this code.

First, I coded `DataSourceManager`. It keeps track of the connection pool objects it initializes. When an application context is destroyed due to `@DirtiesContext`, then the manager does a `close` on the pool which closes any idle connections.

Second, the big issue is our use of `@MockBean`. Each different combination of mocked beans in a test class generates a unique Spring applciation context. Those contexts are _cached_, so the database connections do not go away.

I made a `MockBeanUnitTest` class that derives from `BaseUnitTest` and sets up almost all of the mocks used in all of the unit tests. Then I made those tests inherit from that class instead of the base class. As a result, that application context gets reused. That has the nice side-effect of speeding up the tests, since we don't have to rebuild the context and re-run liquibase on all of the databases.

Interestingly, if we find ourselves pressed for connections again, we can mark tests that generate un-reusable application contexts with `@DirtiesContext` and they will get destroyed and with my new class, the connections will be closed.

In the process of testing, I found two tests claiming to be unit tests that are, in fact, making real workspaces. I disabled them in this PR and filed PF-1962 to track.

If you all agree we should go forward with this, then I will put the `DataSourceManager` code in Terra Common Lib so it can be used by TPS, LZ, and any other TCL user.


service/src/test/java/bio/terra/workspace/app/configuration/external/controller/ControlledGcpResourceApiControllerTest.java
```diff
@@ -22,33 +10,52 @@
...
+import static bio.terra.workspace.common.utils.MockMvcUtils.WORKSPACES_V1_PATH;
+import static bio.terra.workspace.common.utils.MockMvcUtils.addAuth;
+import static bio.terra.workspace.common.utils.MockMvcUtils.addJsonContentType;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@Disabled("PF-1962 This test needs to be rewritten as a connected test since it makes a real GCP cloud context")
```
zloery, line 39: I don't think this test is making a real GCP cloud context. Tests create workspaces but don't generate contexts for them, and they're using mock output from `GcpCloudContextService` to get a project name.

It looks like this test isn't mocking Sam though, which it should be to really be a unit test

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/WorkspaceApiControllerTest.java
```diff
@@ -67,6 +68,7 @@
  * currently in WorkspaceServiceTest, it's intended as a proof-of-concept for future mockMvc-based
  * tests.
  */
+@Disabled("PF-1962 This test needs to be rewritten as a connected test since it makes a real GCP cloud context")
```
zloery, line 71: I don't think this test makes real cloud contexts either, the tests create workspaces but never calls the GCP cloud context endpoints

zloery (commented):
(still need to look through the main portion of this but the unit tests stood out to me)

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/ControlledGcpResourceApiControllerTest.java
ddietterich: OK, thanks. I misinterpreted the error. I'll update the `@Disabled` text and the ticket.

service/src/test/resources/application-test.yml
```diff
@@ -18,6 +18,7 @@ workspace:
     uri: jdbc:postgresql://127.0.0.1:5432/wsm_stairway
     username: stairwayuser
     poolMaxIdle: 1
+    jmxEnabled: false
```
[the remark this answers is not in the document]
ddietterich: John C says that when JMX is enabled, it holds a database connection open. So turning it off for testing reduces the database connection count. I am not sure it is worth it. The heavy hitters were the `@Dirties` and the `@MockBean` usage.

aherbst-broad (approved):
This looks to be a worthwhile change. I am very curious if there are any before/after metrics we can check out to see the speed gains.

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/ControlledGcpResourceApiControllerTest.java
ddietterich: Now I am confused about what error I saw. I made minor changes and these tests work.

ddietterich:
I did some rough measurements and do not see any significant speed gains. There is still some work to do to minimize the number of mock variants, so there might be some slight gains.

PR #799, opened 2022-09-06
[PF-2039] Implement the higher role concept in the `WsmIamRole` enum

pshapiro4broad:
From the discussion https://github.com/DataBiosphere/terra-workspace-manager/pull/758#discussion_r942645004. This PR is for illustrative purposes, if it's useful then someone from the WSM should take on the task of creating a ticket for the work and fixing it up for merging. If not, feel free to close the PR and delete the branch.

You can use the lambda solution, but since the enum members aren't valid at enum declaration time, you need to use a method instead. Since there aren't unit tests for this code I'm not 100% sure I copied everything over correctly.

Also there was a bug with a `log` message using `%s` when it should use `{}`. There is an IntelliJ warning that will catch this error for you.

And there were many unnecessary calls to `toString()` which I removed.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -305,7 +305,7 @@ public class SamService {
           // Skip workspaces with no roles. (That means there's a role this WSM doesn't know
           // about.)
           if (highestRole.isPresent()) {
-            if (WsmIamRole.roleAtLeastAsHighAs(minimumHighestRoleFromRequest, highestRole.get())) {
+            if (minimumHighestRoleFromRequest.roleAtLeastAsHighAs(highestRole.get())) {
               workspacesAndRoles.put(workspaceId, highestRole.get());
             }
           }
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -301,14 +301,15 @@ public class SamService {
                   .map(WsmIamRole::fromSam)
                   .filter(Objects::nonNull)
                   .collect(Collectors.toList());
-          Optional<WsmIamRole> highestRole = WsmIamRole.getHighestRole(workspaceId, roles);
           // Skip workspaces with no roles. (That means there's a role this WSM doesn't know
           // about.)
-          if (highestRole.isPresent()) {
-            if (minimumHighestRoleFromRequest.roleAtLeastAsHighAs(highestRole.get())) {
-              workspacesAndRoles.put(workspaceId, highestRole.get());
-            }
-          }
+          WsmIamRole.getHighestRole(workspaceId, roles)
+              .ifPresent(
+                  highestRole -> {
+                    if (minimumHighestRoleFromRequest.roleAtLeastAsHighAs(highestRole)) {
+                      workspacesAndRoles.put(workspaceId, highestRole);
+                    }
+                  });
         } catch (IllegalArgumentException e) {
           // WSM always uses UUIDs for workspace IDs, but this is not enforced in Sam and there are
           // old workspaces that don't use UUIDs. Any workspace with a non-UUID workspace ID is
--- a/service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
@@ -25,7 +25,8 @@ public class RemoveUserFromWorkspaceFlight extends Flight {
     String userToRemove = inputParameters.get(WorkspaceFlightMapKeys.USER_TO_REMOVE, String.class);
     Optional<WsmIamRole> roleToRemove =
         Optional.ofNullable(
-            inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, WsmIamRole.class));
+            WsmIamRole.valueOf(
+                inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, String.class)));
 
     // Flight plan:
     // 0. (Pre-flight): Validate that the user is directly granted the specified workspace role.
--- a/service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
@@ -24,9 +24,11 @@ public class RemoveUserFromWorkspaceFlight extends Flight {
         inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
     String userToRemove = inputParameters.get(WorkspaceFlightMapKeys.USER_TO_REMOVE, String.class);
     Optional<WsmIamRole> roleToRemove =
-        Optional.ofNullable(
-            WsmIamRole.valueOf(
-                inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, String.class)));
+        inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, String.class) != null
+            ? Optional.of(
+                WsmIamRole.valueOf(
+                    inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, String.class)))
+            : Optional.empty();
 
     // Flight plan:
     // 0. (Pre-flight): Validate that the user is directly granted the specified workspace role.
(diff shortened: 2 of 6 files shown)


service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -305,7 +305,7 @@ public Map<UUID, WsmIamRole> listWorkspaceIdsAndHighestRoles(
           // Skip workspaces with no roles. (That means there's a role this WSM doesn't know
           // about.)
           if (highestRole.isPresent()) {
-            if (WsmIamRole.roleAtLeastAsHighAs(minimumHighestRoleFromRequest, highestRole.get())) {
+            if (minimumHighestRoleFromRequest.roleAtLeastAsHighAs(highestRole.get())) {
               workspacesAndRoles.put(workspaceId, highestRole.get());
             }
```
pshapiro4broad, line 310: Another minor improvement here; you can use `Optional.map()` to combine `isPresent()` with `get()`.
```
WsmIamRole.getHighestRole(workspaceId, roles).map(higestRole -> { ... });
```
In general you almost never want to `get()` without calling `isPresent()`, so `map()` is usually a better way to express this.

melissachang: Got it, thanks

melissachang (approved):
Wow. Thank you so much, that was very educational.

Are you sure the logger didn't work, wasn't it calling https://www.slf4j.org/api/org/slf4j/Logger.html#warn-java.lang.String-java.lang.Object- ?

pshapiro4broad:
@melissachang 
> Are you sure the logger didn't work, wasn't it calling https://www.slf4j.org/api/org/slf4j/Logger.html#warn-java.lang.String-java.lang.Object- ?

Based on the slf4j documentation you'd think it would use `String.format()`. But it doesn't, it uses [MessageFormatter](https://www.slf4j.org/api/org/slf4j/helpers/MessageFormatter.html), which uses `{}` instead of `%s` for argument substitution. This alternate formatter is used for performance reasons.

ddietterich:
I'm not sure how to treat this PR. @pshapiro4broad do you want someone from the WSM team to take it over and finish it? Or is it just for educational purposes?

The code looks good to me. So someone could make a ticket, get it to pass the tests, and merge it.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
melissachang: Done. I ended up using `Optional.ifPresent()`, because the lambda doesn't return anything.

melissachang:
Tests pass. @ddietterich can you take a look?

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
```diff
@@ -24,8 +24,11 @@ public RemoveUserFromWorkspaceFlight(FlightMap inputParameters, Object applicati
         inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
     String userToRemove = inputParameters.get(WorkspaceFlightMapKeys.USER_TO_REMOVE, String.class);
     Optional<WsmIamRole> roleToRemove =
-        Optional.ofNullable(
-            inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, WsmIamRole.class));
+        inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, String.class) != null
```
ddietterich, line 27: I'm not seeing why you switched from WsmIamRole to the role name string. Role coming into WorkspaceService has to be a non-null, valid WsmIamRole, right?

ddietterich (commented):
I have one question; otherwise looks fine

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
melissachang: If flight key is WsmIamRole, then RemoveUserFromWorkspaceFlightTest fails with JsonConversionException

https://scans.gradle.com/s/nmjwrcsi7efm2/tests/:service:connectedTest/bio.terra.workspace.service.workspace.flight.RemoveUserFromWorkspaceFlightTest/removeUserFromWorkspaceFlightDoUndo()?top-execution=1

```
<div class="FailureException__stacktrace-summary StacktraceSummary" style="box-sizing: border-box; margin-top: 8px; align-items: stretch; color: rgb(104, 104, 104); display: flex; font: 400 14px Inconsolata, Monaco, &quot;Courier New&quot;, monospace; position: relative; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><br class="Apple-interchange-newline">

  |   | at bio.terra.stairway.impl.FlightFactory.makeFlight(FlightFactory.java:35) |  
-- | -- | -- | --
  |   | •••
  |   | at bio.terra.workspace.common.StairwayTestUtils.blockUntilFlightCompletes(StairwayTestUtils.java:42) |  
  |   | at bio.terra.workspace.service.workspace.flight.RemoveUserFromWorkspaceFlightTest.removeUserFromWorkspaceFlightDoUndo(RemoveUserFromWorkspaceFlightTest.java:171) |  
  |   | •••

</div></div></div></div></div></div></div>
```

Other enum flight keys are also stored as string, I assume for similar reason. Eg

https://cs.github.com/DataBiosphere/terra-workspace-manager/blob/db7746b1fb275333dc8747ca7e9bdb0bda8af87e/service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java?q=WorkspaceFlightMapKeys.WORKSPACE_STAGE#L25

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
```diff
@@ -24,8 +24,10 @@ public RemoveUserFromWorkspaceFlight(FlightMap inputParameters, Object applicati
         inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
     String userToRemove = inputParameters.get(WorkspaceFlightMapKeys.USER_TO_REMOVE, String.class);
     Optional<WsmIamRole> roleToRemove =
-        Optional.ofNullable(
-            inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, WsmIamRole.class));
+        inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, WsmIamRole.class) != null
```
ddietterich, line 27: This doesn't make sense to me. I don't think there is a semantic difference from what it was before. Further, it is overcomplicated. I would suggest you skip all of the Optionals and simply do:
```
WsmIamRole roleToRemove = inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, WsmIamRole.class);
```
and pass `roleRoleToRemove` into the `RemoveUserFromSamStep`. If it is null, you will pass null. If it is not, you will pass a WsmIamRole. If what is stored in the map is not a WsmIamRole, the retrieval from the map will throw.

ddietterich (approved):
One suggested change. Otherwise, lgtm

service/src/main/java/bio/terra/workspace/service/workspace/flight/RemoveUserFromWorkspaceFlight.java
melissachang: Sorry I wasn't clear. The commit you approved doesn't work. I pushed it to demonstrate the problem with storing enum -- as opposed to enum name -- in the flight.

Problem with storing enum is, Jackson says `Failed to deserialize value '"WRITER"' ` ([build scan](https://scans.gradle.com/s/q7osmkspgmaao/tests/:service:connectedTest/bio.terra.workspace.service.workspace.flight.RemoveUserFromWorkspaceFlightTest/removeUserFromWorkspaceFlightDoUndo()?top-execution=1)).

[This page](https://www.baeldung.com/jackson-serialize-enums#:~:text=By%20default%2C%20Jackson%20will%20use%20the%20Enum%20name%20to%20deserialize%20from%20JSON.) says:

> By default, Jackson will use the Enum name to deserialize from JSON.

So can I store enum name instead of enum? That's what we do for other enum flight inputs.

melissachang: Done, removed optional.

Discussed offline. Storing enum name appears to be simplest. It's unfortunate that we can't store enum as-is.

ddietterich (approved):
Looks good. Thanks for shepherding this!

melissachang:
FYI PrivateResourceCleanupServiceTest failed because PrivateResourceCleanupService set role input to null, and RemoveUserFromWorkspaceFlight `WsmIamRole.valueOf(null)` failed. To fix it, in RemoveUserFromWorkspaceFlight I changed

```
    WsmIamRole roleToRemove =
        WsmIamRole.valueOf(
            inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, String.class));
```

to

```
    WsmIamRole roleToRemove =
        inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, String.class) != null
            ? WsmIamRole.valueOf(
                inputParameters.get(WorkspaceFlightMapKeys.ROLE_TO_REMOVE, String.class))
            : null;
```
(roleToRemove is passed into RemoveUserFromSamStep, which can handle null roletoRemove.)

PR #801, opened 2022-09-06
[PF-1964] Rollback changes to CloneWorkspace test

zloery:
This rolls back the changes to the `CloneWorkspace` integration test from #786 - the TEST_WSM_APP is not enabled anywhere except locally so this test breaks. There are a few ways we could test this instead, but for now I'm rolling this change back so that tests pass again.

The test passed when run locally (on GHA) so I'm reasonably confident the code itself is OK, and we don't need to roll back the entire commit.


melissachang (approved):

ginaygoogle (approved):
Hi Zach. Thanks for letting me know this. The integration test added was following the comment. I am not sure why this code change can pass the 6 checks in the PR but fails finally. 
Also, The `TEST_WSM_APP` comes from [L101](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/integration/src/main/java/scripts/testscripts/ControlledApplicationPrivateGcsBucketLifecycle.java#L101) where I followed. not really understand why `TEST_WSM_APP` can be enabled there without problems, but fails in my code changes. Is these old code related to `TEST_WSM_APP` also should rollback too?

zloery:
there was a bit of a discussion earlier in chat, I'll tag you in that!

in terms of it running on PRs/locally, I think it worked because `TestWsmApp` gets created as part of the local app config. we don't do that for other environments though, which makes integration testing with it difficult. we don't currently run the other integration tests which use `TestWsmApp` so it's pretty misleading leaving them around, agreed we should look at either fixing + enabling them or just removing them if we aren't planning to use them

PR #764, opened 2022-08-10
PF-1891: auto generate the gcsBucketCloudName, BqDatasetCloudId, AiNotebookCloudInstanceId

ginaygoogle:
In this PR, we created 3 endpoints to generate the cloud-native name for `controlledGcpGcsBucket`, `controlledGcpBigQueryDataset`, `controlledGcpAiNotebook`. The changes include endpoint design in Yaml files, API implementation in `ControlledXXXHandler` and integration test & unit test.

We follows the generation rules based on the GCP resources name requirement as:
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-1b7b2895-7fff-0c66-b885-50e200d92bb8"><div dir="ltr" style="margin-left:0pt;" align="left">

type | name | cloud-native name | Generate format
-- | -- | -- | --
Gcs-bucket | test-clould_bucket | terra-vdevel-sunny-lemon-9339-test-clould_bucket | [resource name]-[project id]
Bq-dataset | big-query_dataset | big_query_dataset | Same as name, but replace “-” to “_”
Ai-notebook | notebook-test-gy | notebook-test-gy | same as the name

</div></b>

**These are the endpoint manually run the swagger UI** 
**controlledGcpGcsBucket**
Screen Shot 2022-08-15 at 6 54 07 PM

**controlledGcpBigQueryDataset**
Screen Shot 2022-08-15 at 6 53 40 PM

**controlledGcpAiNotebook**
Screen Shot 2022-08-15 at 6 53 16 PM

**Code explain:**
- the API request can be called before the resource create, and put actual product code changes in `ControlledXXXHandler`.  This PR also cover the case while creating the resource without cloud name, we will default use the suggested generation format to set the cloud name.
- Although only 3 resources involved in this ticket, considering the consistence, and future design. We set the `generateName` this interface in the super class `WsmResourceHandler` rather than just in the `ControlledXXXHandler` class. In this case, there are 14 more empty interfaces generates and impact on the number of file changes.
- For the test, `WsmResourceHandler` cover some actual code logic and require some corner cases, and create 3 new handler test files for unit test. We do the integration test; We also have the unit test in new created file `ControlledGcpResourceApiControllerTest` and use MockMVC to mock the API works.

PR #802, opened 2022-09-07
[PF-1954] bucket cloud names do not allow "goog" prefix; notebook cloud instance ids do not allow numbers start

ginaygoogle:
In this ticket, we are focusing correcting 2 bugs on the previous ticket for [autogeneration the cloud name](https://github.com/DataBiosphere/terra-workspace-manager/pull/764). 

For the AI notebook cloud instance id, we do not have the official document in gcp like the old code mentioned. I just add a corner case which not allow the cloud instance id starts with numbers, if it does, we strip the numbers until it satisfied the requirement.

For the controlled gcs bucket, the official doc is [here](https://cloud.google.com/storage/docs/naming-buckets). In order to improve the suggestion name, I add the bad case start with `goog` prefix, then, strip it. 

Also add the comments on bq dataset to make consistence.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
@@ -71,7 +71,7 @@ public class ControlledAiNotebookHandler implements WsmResourceHandler {
         generatedName.length() > MAX_INSTANCE_NAME_LENGTH
             ? generatedName.substring(0, MAX_INSTANCE_NAME_LENGTH)
             : generatedName;
-    generatedName = generatedName.replaceAll("[^a-zA-Z0-9-]+|^[^a-zA-Z0-9]+|[^a-zA-Z0-9]+$", "");
+    generatedName = generatedName.replaceAll("[^a-z0-9-]+|^[^a-z]+|[^a-z0-9]+$", "");
     return generatedName;
   }
 }
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
@@ -71,6 +71,12 @@ public class ControlledAiNotebookHandler implements WsmResourceHandler {
         generatedName.length() > MAX_INSTANCE_NAME_LENGTH
             ? generatedName.substring(0, MAX_INSTANCE_NAME_LENGTH)
             : generatedName;
+
+    /**
+     * The regular expression only allow legal character combinations which start with lowercase
+     * letter, lowercase letter and numbers and dash("-") in the string, and lowercase letter and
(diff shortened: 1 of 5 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gcsbucket/ControlledGcsBucketHandler.java
```diff
@@ -62,7 +73,8 @@ public String generateCloudName(@Nullable UUID workspaceUuid, String bucketName)
         generatedName.length() > MAX_BUCKET_NAME_LENGTH
             ? generatedName.substring(0, MAX_BUCKET_NAME_LENGTH)
             : generatedName;
-    generatedName = generatedName.replaceAll("[^a-zA-Z0-9-]+|^[^a-zA-Z0-9]+|[^a-zA-Z0-9]+$", "");
+    generatedName =
+        generatedName.replaceAll("google|^goog|[^a-z0-9-.]+|^[^a-z0-9]+|[^a-z0-9]+$", "");
```
ginaygoogle, line 77: As the official doc [here](https://cloud.google.com/storage/docs/naming-buckets). The last constrain for the `Bucket names cannot contain "google" or close misspellings, such as "g00gle".`. I do not do it here at this moment. Would like to know how could we define the `close misspellings`, like "go0gle", "g0Ogle", how about "goog1e" and "googIe" (the I is the capital letter of "i"). Any ideas on it?

cbookg: i wouldn't worry so much about these google variants.  the backend will just reject it and the user will see why.  its unlikely the user will be putting in a name of 'go0gle'

cbookg (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
```diff
@@ -71,7 +71,7 @@ public String generateCloudName(@Nullable UUID workspaceUuid, String aiNotebookN
         generatedName.length() > MAX_INSTANCE_NAME_LENGTH
             ? generatedName.substring(0, MAX_INSTANCE_NAME_LENGTH)
             : generatedName;
-    generatedName = generatedName.replaceAll("[^a-zA-Z0-9-]+|^[^a-zA-Z0-9]+|[^a-zA-Z0-9]+$", "");
+    generatedName = generatedName.replaceAll("[^a-z0-9-]+|^[^a-z]+|[^a-z0-9]+$", "");
```
yuhuyoyo, line 74: what happened if it leads with a dash?

Can you also add a comment here explaining the regex?

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
ginaygoogle: it will trim the start with dash. Yes. add comments to all resource reg

PR #804, opened 2022-09-07
[PF-1041] Fixes for WorkspaceServiceTest

zloery:
The local config here is configured with a service account named `leo` but Verily environments are not, which breaks connected tests. This switches the WorkspaceServiceTest to use `TestWsmApp`, which is specified in the `app-test` profile. Unlike integration tests this will only ever use local configuration, so it's safe to test with this application without worrying about what apps are enabled in live environments.

(tested via verily-src submodule)

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/privateresource/PrivateResourceCleanupServiceTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/privateresource/PrivateResourceCleanupServiceTest.java
@@ -47,8 +47,9 @@ import org.junit.jupiter.api.condition.DisabledIfEnvironmentVariable;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.context.ActiveProfiles;
 
-// Use application configuration profile in addition to the standard connected test profile.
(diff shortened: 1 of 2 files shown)


service/src/test/java/bio/terra/workspace/service/privateresource/PrivateResourceCleanupServiceTest.java
```diff
@@ -47,8 +47,9 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.context.ActiveProfiles;
 
-// Use application configuration profile in addition to the standard connected test profile.
-@ActiveProfiles({"connected-test", "app-test"})
+// Use application configuration profile in addition to the standard connected test profile
+// inherited from the base class.
+@ActiveProfiles({"app-test"})
```
zloery, line 52: I was copying this and realized the `connected-test` profile is already inherited via BaseConnectedTest

melissachang (approved):
Thank you for being so on top of things!!

ginaygoogle (approved):

PR #805, opened 2022-09-07
PF-1965: Add properties to resources

yuhuyoyo:
Also refactor to make adding new resource fields easier for referenced resources. 

I didn't add endpoint for updating/deleting resource properties in this PR because 1. it's already a big PR, 2. I don't need those endpoint for my folder work so it's not urgent to have those endpoints yet.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
@@ -97,7 +97,7 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
     // Construct a ReferenceGcsBucketResource object from the API input
     var resource =
         ReferencedGcsObjectResource.builder()
-            .wsmResourceFieldsBuilder(
+            .resourceCommonFields(
                 WsmResourceFields.builder()
                     .workspaceUuid(workspaceUuid)
                     .resourceId(UUID.randomUUID())
@@ -105,7 +105,8 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
                     .description(body.getMetadata().getDescription())
                     .cloningInstructions(
                         CloningInstructions.fromApiModel(
-                            body.getMetadata().getCloningInstructions())))
+                            body.getMetadata().getCloningInstructions()))
+                    .build())
             .bucketName(body.getFile().getBucketName())
             .objectName(body.getFile().getFileName())
             .build();
@@ -174,10 +175,10 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
         updateBucketObjectResourceBuilder.objectName(objectName);
       }
       if (cloningInstructions != null) {
-        updateBucketObjectResourceBuilder.wsmResourceFieldsBuilder(
-            referencedResource
-                .getWsmResourceFieldsBuilder()
-                .cloningInstructions(cloningInstructions));
+        updateBucketObjectResourceBuilder.resourceCommonFields(
+            referencedResource.getWsmResourceFields().toBuilder()
+                .cloningInstructions(cloningInstructions)
+                .build());
       }
       referenceResourceService.updateReferenceResource(
           workspaceUuid,
@@ -211,7 +212,7 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
     // Construct a ReferenceGcsBucketResource object from the API input
     var resource =
         ReferencedGcsBucketResource.builder()
-            .wsmResourceFieldsBuilder(
+            .resourceCommonFields(
                 WsmResourceFields.builder()
                     .workspaceUuid(workspaceUuid)
                     .resourceId(UUID.randomUUID())
@@ -219,7 +220,8 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
                     .description(body.getMetadata().getDescription())
                     .cloningInstructions(
                         CloningInstructions.fromApiModel(
-                            body.getMetadata().getCloningInstructions())))
+                            body.getMetadata().getCloningInstructions()))
+                    .build())
             .bucketName(body.getBucket().getBucketName())
             .build();
 
@@ -279,10 +281,10 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
           referencedResource.toBuilder().bucketName(bucketName);
       if (cloningInstructions != null) {
         // only overwrite if non-null
-        updateBucketResourceBuilder.wsmResourceFieldsBuilder(
-            referencedResource
-                .getWsmResourceFieldsBuilder()
-                .cloningInstructions(cloningInstructions));
+        updateBucketResourceBuilder.resourceCommonFields(
+            referencedResource.getWsmResourceFields().toBuilder()
+                .cloningInstructions(cloningInstructions)
+                .build());
       }
       referenceResourceService.updateReferenceResource(
           workspaceUuid,
@@ -315,7 +317,7 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
         userRequest, workspaceUuid, SamWorkspaceAction.CREATE_REFERENCE);
     var resource =
         ReferencedBigQueryDataTableResource.builder()
-            .wsmResourceFieldsBuilder(
+            .resourceCommonFields(
                 WsmResourceFields.builder()
                     .workspaceUuid(workspaceUuid)
                     .resourceId(UUID.randomUUID())
@@ -323,7 +325,8 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
                     .description(body.getMetadata().getDescription())
                     .cloningInstructions(
                         CloningInstructions.fromApiModel(
-                            body.getMetadata().getCloningInstructions())))
+                            body.getMetadata().getCloningInstructions()))
+                    .build())
             .projectId(body.getDataTable().getProjectId())
             .datasetId(body.getDataTable().getDatasetId())
             .dataTableId(body.getDataTable().getDataTableId())
@@ -436,7 +439,7 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
     // Construct a ReferenceBigQueryResource object from the API input
     var resource =
         ReferencedBigQueryDatasetResource.builder()
-            .wsmResourceFieldsBuilder(
+            .resourceCommonFields(
                 WsmResourceFields.builder()
                     .workspaceUuid(uuid)
                     .resourceId(UUID.randomUUID())
@@ -444,7 +447,8 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
                     .description(body.getMetadata().getDescription())
                     .cloningInstructions(
                         CloningInstructions.fromApiModel(
-                            body.getMetadata().getCloningInstructions())))
+                            body.getMetadata().getCloningInstructions()))
+                    .build())
             .projectId(body.getDataset().getProjectId())
             .datasetName(body.getDataset().getDatasetId())
             .build();
@@ -547,7 +551,7 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
 
     var resource =
         ReferencedDataRepoSnapshotResource.builder()
-            .wsmResourceFieldsBuilder(
+            .resourceCommonFields(
                 WsmResourceFields.builder()
                     .workspaceUuid(uuid)
                     .resourceId(UUID.randomUUID())
@@ -555,7 +559,8 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
                     .description(body.getMetadata().getDescription())
                     .cloningInstructions(
                         CloningInstructions.fromApiModel(
-                            body.getMetadata().getCloningInstructions())))
+                            body.getMetadata().getCloningInstructions()))
+                    .build())
             .instanceName(body.getSnapshot().getInstanceName())
             .snapshotId(body.getSnapshot().getSnapshot())
             .build();
@@ -898,7 +903,7 @@ public class ReferencedGcpResourceController implements ReferencedGcpResourceApi
         userRequest, workspaceUuid, SamWorkspaceAction.CREATE_REFERENCE);
     ReferencedGitRepoResource resource =
         ReferencedGitRepoResource.builder()
-            .wsmResourceFieldsBuilder(
+            .resourceCommonFields(
                 WsmResourceFields.builder()
                     .workspaceUuid(workspaceUuid)
                     .resourceId(UUID.randomUUID())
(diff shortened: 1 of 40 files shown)


service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -168,8 +174,11 @@ public ResponseEntity<Void> updateBucketObjectReferenceResource(
       if (!StringUtils.isEmpty(objectName)) {
         updateBucketObjectResourceBuilder.fileName(objectName);
       }
-      if (null != cloningInstructions) {
-        updateBucketObjectResourceBuilder.cloningInstructions(cloningInstructions);
+      if (cloningInstructions != null) {
```
ddietterich, line 177: I know I started this, but...
I would prefer to avoid one-off `toBuilder` uses. I wonder if we could do something like:

Have `ReferencedGcsObjectResource.Builder` have a member that is the `WsmResourceFields.Builder` instead of the `WsmResourceFields` itself. That way both builders would be "open" to take changes. Then the `build()` method of `ReferencedGcsObjectResource.Builder` would call the `build()` method of the `WsmResourceFields.Builder` and generate both objects.

Alternately, maybe the specific resource builder could derive from `WsmResourceFields.Builder` allowing the setters to be flat references, instead of having two separate builders. Since we do not serialize builders, derivation is safe.

wdyt? Maybe that is a separate refactoring ticket?

ddietterich (commented):
One thought about the builders. Otherwise, looks good.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -627,6 +631,77 @@ public boolean resourceExists(UUID workspaceUuid, UUID resourceId) {
     return (count != null && count > 0);
   }
 
+  @WriteTransaction
+  public boolean deleteResourceProperties(
```
melissachang, line 635: `deleteResourceProperties()` and `updateResourceProperties()` are not used. They should be removed from this PR.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -57,46 +60,47 @@ public class ResourceDao {
...
+                      .orElse(null))
+              .resourceLineage(
+                  Optional.ofNullable(rs.getString("resource_lineage"))
+                      .map(
+                          resourceLineage ->
+                              DbSerDes.fromJson(
+                                  resourceLineage,
+                                  new TypeReference<List<ResourceLineageEntry>>() {}))
+                      .orElse(null))
+              .properties(DbSerDes.jsonToProperties(rs.getString("properties")));
```
melissachang, line 103: There should be Optional, to be consistent with workspace properties

```
              .properties(
                  Optional.ofNullable(rs.getString("properties"))
                      .map(DbSerDes::jsonToProperties)
                      .orElse(null))
```

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtils.java
```diff
@@ -258,16 +261,37 @@ private static ReferencedResource buildDestinationGcsObjectReference(
...
 
+  private static WsmResourceFields.Builder<?> buildDestinationResourceCommonFields(
+      UUID destinationWorkspaceId,
+      UUID destinationResourceId,
+      @Nullable String name,
+      @Nullable String description,
+      List<ResourceLineageEntry> destinationResourceLineage,
+      WsmResource wsmResource) {
+    WsmResourceFields.Builder<?> sourceCommonFieldsBuilder =
+        wsmResource.getWsmResourceFields().toBuilder();
```
melissachang, line 286: nit You can combine 285 with 287

```
WsmResourceFields.Builder<?> sourceCommonFieldsBuilder =
    wsmResource.getWsmResourceFields().toBuilder()
    .workspaceUuid ...
```

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtils.java
```diff
@@ -234,15 +236,16 @@ private static ReferencedResource buildDestinationGcsBucketReference(
             sourceBucketResource.getResourceLineage(),
             sourceBucketResource.getWorkspaceId(),
             sourceBucketResource.getResourceId());
+    WsmResourceFields.Builder<?> sourceCommonFieldsBuilder =
```
melissachang, line 239: sourceCommonFieldsBuilder -> destinationCommonFieldsBuilder

Same for rest of file, including in buildDestinationResourceCommonFields

melissachang (commented):
Nice refactoring! Though it was quite large. Next time, consider separate PR. :)

I don't see tests. How about have CloneWorkspace.java assert that resource property was cloned?

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
melissachang: +1. Both suggestions seem doable in this PR (as opposed to ticket)

service/src/main/java/bio/terra/workspace/service/resource/controlled/model/ControlledResource.java
```diff
@@ -170,8 +180,7 @@ public void validate() {
     if (getResourceType() == null
         || attributesToJson() == null
         || getAccessScope() == null
-        || getManagedBy() == null
-        || getResourceId() == null) {
```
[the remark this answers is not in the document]
yuhuyoyo: resourceId is a field for all wsmResource. It is asserted at the WsmResource level already. ControlledResource is a subclass of wsmresource.

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResource.java
```diff
@@ -30,6 +34,7 @@ public abstract class WsmResource {
   private @Nullable final String description;
   private final CloningInstructions cloningInstructions;
   private final List<ResourceLineageEntry> resourceLineage;
+  private final ImmutableMap<String, String> properties;
```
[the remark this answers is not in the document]
yuhuyoyo: i made sure this field cannot be nullable in the code path. https://github.com/DataBiosphere/terra-workspace-manager/pull/805/files#diff-0a0214a663410000194a38c64de77f621609da9772780dc34be4246bb4b0ed28R104 and properties has a default value {} in the database. 
I don't want to do null check everywhere. Like lineage sometimes has nullable annotation and sometimes not in the resource classes. @Nullable should only annotated when the field can be nullable and should be consistent. i can add a comment here to clarify that the map will be empty if there's no properties. 

Workspace properties is the same: https://cs.github.com/DataBiosphere/terra-workspace-manager/blob/0b44a7fe80e37c042dcd068c8bc1a4d26c2448b1/service/src/main/java/bio/terra/workspace/service/workspace/model/Workspace.java?q=workspace+path%3Amodel#L29

service/src/main/resources/db/changesets/20220907_add_resource_properties_column.yaml
```diff
@@ -0,0 +1,16 @@
...
+    changes:
+    - addColumn:
+        tableName: resource
+        columns:
+        - column:
+            name: properties
+            type: jsonb
+            remarks: |
+              Serialized key-value map of resource properties.
+            defaultValue: "{}"
```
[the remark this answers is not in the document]
yuhuyoyo: i don't see the benefit of keeping them consistent though. I'm keeping the consistency that the properties fields cannot be null. 
workspaceDao has to do a nullcheck and create an empty hashmap if the column is null while we could have skipped that.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
yuhuyoyo: i added this because there's no way to test properties given they are always empty lol but that's not a good reason. i will merge it with my folder change.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
yuhuyoyo: they can't be null.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
yuhuyoyo: i actually don't quite follow, perhaps we can discuss offline?

I thought toBuilder is generally essential to a Builder class. How do I convert the resource to a builder without it?
re 1: how does this get rid of toBuilder()? This piece would look like this if i change the filed to WsmResourceFields.Builder

```
ReferencedGcsObjectResource.Builder updateBucketObjectResourceBuilder =
          referencedResource.toBuilder();
      if (!StringUtils.isEmpty(bucketName)) {
        updateBucketObjectResourceBuilder.bucketName(bucketName);
      }
      if (!StringUtils.isEmpty(objectName)) {
        updateBucketObjectResourceBuilder.objectName(objectName);
      }
      if (cloningInstructions != null) {
        updateBucketObjectResourceBuilder.resourceCommonFieldsBuilder(
            referencedResource.getWsmResourceFields()
                .cloningInstructions(cloningInstructions));
      }
```

`ReferencedGcsObjectResource#toBuilder()` would look like this:
```
  public Builder toBuilder() {
    return builder()
        .bucketName(getBucketName())
        .objectName(getObjectName())
        .resourceCommonFieldsBuilder(getWsmResourceFieldsBuilder());
  }
```

re 2: What do you mean by derive? `wsmResourceFields.Builder` does not build a resource so specific resource builder can't be a subclass of it.

yuhuyoyo: discussed offline with Dan. We will stick with the current way as ResourceBuilder.build() return a resource while ResourceFieldsBuilder.build() returns a resourceFields so they can't have an inheritance relationship.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
melissachang: To test, can't you set the properties field when you create a resource? And this PR has the logic for copying resource properties when you clone. (At least you did this for controlled, did you do this for referenced too?) That's why I suggested testing cloning a workspace or resource. That makes sure that 1) Property correctly set on source resource, 2) property copied over to dest resource.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
melissachang: Dan's 1st suggestion is [how proto works](https://developers.google.com/protocol-buffers/docs/reference/java-generated#sub-builders). If Foo has Bar, FooBuilder has BarBuilder.

Currently, ReferencedGcsObjectResource.Builder has WsmResourceField, which looks weird to me. It should have WsmResourceField.Builder.

Could this be

```
        updateBucketObjectResourceBuilder.resourceCommonFieldsBuilder().cloningInstructions(cloningInstructions);

  public Builder toBuilder() {
    return builder()
        .bucketName(getBucketName())
        .objectName(getObjectName())
        .resourceCommonFieldsBuilder(getResourceCommonFieldsBuilder());
  }
```

Happy to chat offline, just wanted to type the code here. :)

service/src/main/java/bio/terra/workspace/service/resource/referenced/cloud/gcp/gcsobject/ReferencedGcsObjectResource.java
```diff
@@ -161,51 +178,21 @@ public boolean checkAccess(FlightBeanBag context, AuthenticatedUserRequest userR
...
+        .resourceCommonFields(getWsmResourceFields());
   }
 
   public static Builder builder() {
     return new Builder();
   }
 
   public static class Builder {
-    private CloningInstructions cloningInstructions;
+    private WsmResourceFields resourceCommonFields;
```
melissachang, line 190: I was a bit confused that some place have `resourceCommonFields` and some have `wsmResourceFields`. Can they be the same (I don't care which one). Same for rest of PR.

service/src/main/resources/db/changesets/20220907_add_resource_properties_column.yaml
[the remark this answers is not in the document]
melissachang: I still think they should be consistent.

As Dan mentioned, ideally workspace and resource would extend a common object. Common object would have id, name, description, properties, as well as future features comments, tags. We may move to this.

For now, we're implementing separately. The separate implementations should be the same. If I make a change to properties, I don't want to have to make slightly different changes.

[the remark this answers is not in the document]
yuhuyoyo: the model object workspace and resource can still extend a common object as the nullability is consistent at that level. Only at the dao layer workspaceDao and resourceDao has different implementation.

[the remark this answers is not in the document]
melissachang: Okay, fine :)

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
yuhuyoyo: that's the thing.. i intentionally left it out of the creation path because that's what workspace properties does. It will have separate endpoint for properties update/delete. 

i added aunit test in workspacecloneutils.

yeah. it's automatically done for referenced resource through toBuilder.

melissachang: Ah... now I understand why you had the dao methods. Sorry. No tests are fine then.

Now that I understand there's no other way to test, the dao methods were fine. (Fine to leave them out too.)

melissachang (approved):

ddietterich (approved):
LGTM

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
yuhuyoyo: done.

yuhuyoyo: actually i cant.. jackson don't like it if i have a builder as a field. 

https://scans.gradle.com/s/yvs7noqtcfsaq/tests/:service:unitTest/bio.terra.workspace.service.resource.referenced.ReferencedResourceServiceTest$BigQueryReference/bigQueryDataTableReference()?top-execution=1



Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: No serializer found for class bio.terra.workspace.service.resource.model.WsmResourceFields$Builder and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) (through reference chain: bio.terra.workspace.service.resource.referenced.cloud.gcp.bqdatatable.ReferencedBigQueryDataTableResource["wsmResourceFieldsBuilder"])

PR #806, opened 2022-09-08
PF-1966: Unable to clone pre-policy workspace

mikenev:
If we're cloning a workspace that doesn't have a policy attached - IE it was created before TPS was turned on - then create an empty policy on the source workspace before cloning it.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/ClonePolicyAttributesStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/ClonePolicyAttributesStep.java
@@ -8,7 +8,11 @@ import bio.terra.stairway.StepResult;
 import bio.terra.stairway.exception.RetryException;
 import bio.terra.workspace.amalgam.tps.TpsApiDispatch;
 import bio.terra.workspace.common.utils.FlightUtils;
+import bio.terra.workspace.generated.model.ApiTpsComponent;
+import bio.terra.workspace.generated.model.ApiTpsObjectType;
+import bio.terra.workspace.generated.model.ApiTpsPaoCreateRequest;
 import bio.terra.workspace.generated.model.ApiTpsPaoGetResult;
+import bio.terra.workspace.generated.model.ApiTpsPolicyInputs;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.job.JobMapKeys;
 import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.ControlledResourceKeys;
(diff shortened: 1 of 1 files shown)


yuhuyoyo (commented):
I think TPS should also handle this more gracefully by wrapping the exception.

Here https://cs.github.com/DataBiosphere/terra-policy-service/blob/c1aa08b961e8394d712183b181964c501c087c34/service/src/main/java/bio/terra/policy/db/PaoDao.java?q=paodao#L375 this actually doesn't catch the case when no entry is found.

There should be a try catch and wrap the EmptyResultDataAccessException exception in a InternalTpsErrorException.

mikenev:
> I think TPS should also handle this more gracefully by wrapping the exception.

I agree, TPS can do some additional handling, but it is a separate project so will be a separate code review. The changes in the TPS project wouldn't change how WSM should approach this.

We do need a fix on the WSM side to start adding in PAOs to the older workspaces.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/ClonePolicyAttributesStep.java
```diff
@@ -41,11 +45,26 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
-    if (workspacePao.isPresent()) {
+    if (!sourcePao.isPresent()) {
+      // Source workspace doesn't have a PAO, so create one for it.
+      ApiTpsPaoCreateRequest request =
+          new ApiTpsPaoCreateRequest()
+              .objectId(sourceWorkspaceId)
+              .component(ApiTpsComponent.WSM)
+              .objectType(ApiTpsObjectType.WORKSPACE)
+              .attributes(new ApiTpsPolicyInputs());
+      tpsApiDispatch.createPao(new BearerToken(userRequest.getRequiredToken()), request);
```
yuhuyoyo, line 60: does this throw?

if so, return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);

same with deletePao. But perhaps that one throw FATAL instead?

mikenev: I can add that in. They don't throw anything in particular that I could look for, but they could throw something like "service unreachable". I'll have to catch the generic Exception.

ddietterich:
> There should be a try catch and wrap the EmptyResultDataAccessException exception in a InternalTpsErrorException.

It should try-catch, but I think that is not the exception in this case.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/ClonePolicyAttributesStep.java
ddietterich: Create Pao can result in TPS `DuplicateObjectException` if WSM tries to create the sam Pao twice.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/ClonePolicyAttributesStep.java
```diff
@@ -41,16 +49,44 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
             .getInputParameters()
             .get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
 
-    Optional<ApiTpsPaoGetResult> workspacePao =
+    Optional<ApiTpsPaoGetResult> sourcePao =
         tpsApiDispatch.getPaoIfExists(
             new BearerToken(userRequest.getRequiredToken()), sourceWorkspaceId);
 
-    if (workspacePao.isPresent()) {
+    if (!sourcePao.isPresent()) {
```
melissachang, line 56: // Old workspaces don't have Paos

melissachang, line 56: So is the plan to backfill paos onto old workspaces at some point?

If so, then add a TODO

// Delete this block when all workspaces have Paos

If not, would it be cleaner to:

- Don't create source pao and don't call clonePao. Just call createPao on destination.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/ClonePolicyAttributesStep.java
```diff
@@ -5,18 +5,26 @@
 import bio.terra.stairway.FlightMap;
```
melissachang, line 5: Add test for this case? Eg `cloneWorkspace_sourceDoesNotHavePao`. Could go in WorkspaceApiIControllerTest or WorkspaceApiIControllerConnectedTest, depending on if you want real or mock TPS.

ddietterich (approved):
See my note. The TPS clone implementation is wrong. I suggest making this fix to put it back to working for existing workspaces. I made two PRs for revising the TPS and then the WSM work to get a proper clone working. Sorry about that!

melissachang (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/ClonePolicyAttributesStep.java
mikenev: Going to leave this as is for now. I attempted both the ControllerTest and ConnectedControllerTest and they have separate issues for this case. Last time, I had to add an integration test. I started on that and it turned into a larger item, so I'll prepare a separate review for it.

PR #807, opened 2022-09-08
PF-1986: Add a folder table

yuhuyoyo:
In this PR, the connection between resources and folders are not established yet. folder is a stand-alone object in workspace manager. 

you can create, update, get, delete folder.

when deleting folder, subfolders are also deleted to prevent orphan folders.
When updating folder, we check for potential cycle and fail the update if the update will form a cycle.
The other operations are similar to the rest of WSM api.


service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,310 @@
...
+                                    displayName, description, UUID.randomUUID()))),
+                    USER_REQUEST)))
+        .andExpect(status().is(HttpStatus.SC_NOT_FOUND));
+  }
+
+  @Test
+  public void createFolder_duplicateFolder_throws400() throws Exception {
+    UUID workspaceId = createDefaultWorkspace(mockMvc, objectMapper).getId();
+    var displayName = "foo";
+    var description = String.format("This is folder %s", displayName);
```
melissachang, line 99: nit description isn't needed, can you delete?

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,310 @@
...
+                        .content(
+                            objectMapper.writeValueAsString(
+                                createFolderRequestBody(
+                                    displayName, description, UUID.randomUUID()))),
+                    USER_REQUEST)))
+        .andExpect(status().is(HttpStatus.SC_NOT_FOUND));
+  }
+
+  @Test
+  public void createFolder_duplicateFolder_throws400() throws Exception {
```
melissachang, line 96: Can you also test that duplicate names are allowed, in different folders? This could be an example where it's useful to test more than one thing in a test method. It's easier to read than X different methods; X different methods would have overlapping setup.

```
  @Test
  public void createFolder_subFolder() throws Exception {
    UUID workspaceId = createDefaultWorkspace(mockMvc, objectMapper).getId();
    var displayName = "foo";

    createFolder(parentFolder, displayName);

    // Can't create a second sub-folder called displayName
    createFolderExpectingError(parentFolder, displayName, HttpStatus.SC_BAD_REQUEST);

    // Can create a sub-folder called displayName in a different parent folder
    createFolder(parentFolder2, displayName);
  }
```

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,310 @@
...
+        createFolder(workspaceId, displayName, description, null);
+
+    assertEquals(displayName, folderDescription.getDisplayName());
+    assertEquals(description, folderDescription.getDescription());
+    assertNotNull(folderDescription.getId());
+    assertNull(folderDescription.getParentFolderId());
+  }
+
+  @Test
+  public void createFolder_parentFolderDoesNotExist_throws404() throws Exception {
```
melissachang, line 77: Can you have `_throws403` and `_throws404` tests for the various endpoints that throw them?

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,310 @@
...
+    var updatedFolderDescription =
+        objectMapper.readValue(serializedUpdateResponse, ApiFolderDescription.class);
+
+    assertEquals(newDisplayName, updatedFolderDescription.getDisplayName());
+    assertEquals(newDescription, updatedFolderDescription.getDescription());
+    assertNull(updatedFolderDescription.getParentFolderId());
+  }
+
+  @Test
+  public void updateFolders_invalidParentFolder_throws404() throws Exception {
```
melissachang, line 243: nit what exactly does invalid mean, does parent folder not exist? Can you be clear

updateFolders_parentFolderDoesNotExist_throws404

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,310 @@
...
+    var retrievedFolders =
+        objectMapper.readValue(foldersGetResponse, ApiFolderDescriptionsList.class);
+
+    var expectedFolders =
+        new ApiFolderDescriptionsList().addFoldersItem(firstFolder).addFoldersItem(secondFolder);
+    assertEquals(expectedFolders, retrievedFolders);
+  }
+
+  @Test
+  public void updateFolders_folderUpdated() throws Exception {
```
melissachang, line 206: Can you have

// Assert description is updated
updateFolder_moveToTopLevelFalse_folderNotMoved

// Assert description is updated and folder moved
updateFolder_moveToTopLevelTrue_folderMoved

openapi/src/parts/Folder.yaml
```diff
@@ -0,0 +1,164 @@
...
+          $ref: '#/components/responses/ServerError'
+
+components:
+  schemas:
+    FolderDescriptionsList:
+      type: object
+      required: [ folders ]
+      properties:
+        folders:
+          description: Id of the parent folder
```
melissachang, line 110: Delete

openapi/src/parts/Folder.yaml
```diff
@@ -0,0 +1,164 @@
...
+      required: [id, displayName]
+      properties:
+        id:
+          description: The ID of the folder. Immutable.
+          $ref: '#/components/schemas/FolderId'
+        displayName:
+          description: The human readable name of the folder.
+          type: string
+        description:
+          description: A description of the workspace
```
melissachang, line 125: Delete

service/src/main/java/bio/terra/workspace/service/folder/model/Folder.java
```diff
@@ -0,0 +1,23 @@
+package bio.terra.workspace.service.folder.model;
```
melissachang, line 1: Can you assert that id and name are set?

https://cs.github.com/DataBiosphere/terra-workspace-manager/blob/ea107a2c31ed014e6c333517b4e39050cea838c3/service/src/main/java/bio/terra/workspace/service/workspace/model/Workspace.java?q=missingrequiredfield#L183

Maybe not id, since it's not passed in CreateFolderRequestBody

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,310 @@
...
+    // change agents.
+    when(mockSamService.getUserStatusInfo(any()))
+        .thenReturn(
+            new UserStatusInfo()
+                .userEmail(USER_REQUEST.getEmail())
+                .userSubjectId(USER_REQUEST.getSubjectId()));
+  }
+
+  @Test
+  public void createFolder_topLevelFolderCreated() throws Exception {
```
melissachang, line 62: createFolder_requestMissingName_throws  ?

service/src/main/java/bio/terra/workspace/db/FolderDao.java
```diff
@@ -0,0 +1,254 @@
...
+import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.dao.support.DataAccessUtils;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Component;
+
+@Component
+public class FolderDao {
+  private static final String DEFAULT_ROOT = "0";
```
melissachang, line 30: First, someone might want to create a folder called `0`.

Second, why do we need to set something? Can we say, if parentFolderId is blank, folder is top-level folder?

melissachang (commented):
Nice! Though smaller cl next time please. :) go/small-cls

I didn't finish; here's what I have so far. I'll finish tomorrow.

ddietterich (commented):
That is a big piece of work. I have a small number of comments

service/src/main/java/bio/terra/workspace/service/folder/model/Folder.java
yuhuyoyo: changed to record

service/src/main/java/bio/terra/workspace/db/FolderDao.java
yuhuyoyo: i picked 0 because I want this root folder id to be a secret that only the Dao knows about because the root folder does not have a database entry. If it's a UUID, it's easier to pass it back up as a parent id all the way to the endpoint response. This way we can catch it during string to uuid conversion

service/src/main/java/bio/terra/workspace/db/FolderDao.java
```diff
@@ -0,0 +1,254 @@
...
+      int rowsAffected = jdbcTemplate.update(sb.toString(), params);
+      return rowsAffected > 0;
+    } catch (DuplicateKeyException e) {
+      if (e.getMessage() != null
+          && e.getMessage()
+              .contains(
+                  "duplicate key value violates unique constraint \"folder_display_name_parent_folder_id_workspace_id_key\"")) {
+        throw new DuplicateFolderDisplayNameException(
+            String.format(
+                "Fails to update due to duplicate display name at the same folder level"));
```
melissachang, line 163: Folder with display name %s already exists in parent folder %s

service/src/main/java/bio/terra/workspace/db/FolderDao.java
```diff
@@ -0,0 +1,254 @@
...
+      }
+      throw e;
+    }
+  }
+
+  private boolean canFormCyclicCycle(UUID workspaceId, UUID folder1, UUID folder2) {
```
melissachang, line 169: Unit test that cycle fails? Would be nice to verify cycle error message.

service/src/main/java/bio/terra/workspace/db/FolderDao.java
```diff
@@ -0,0 +1,254 @@
...
+      throw new FolderNotFoundException(
+          String.format("Cannot find folder %s in workspace %s", folderId, workspaceId));
+    }
+  }
+
+  /**
+   * Gets a list of folders.
+   *
+   * @param parentFolderId when null, get all the folders in a workspace, otherwise, get the direct
+   *     subfolders of a given parent folder.
```
melissachang, line 213: List folders in parent folder is partially implemented

Done: Dao

Not done: service, API, controller unit test

Can you either take it out, or fully implement?

service/src/main/java/bio/terra/workspace/service/folder/FolderService.java
```diff
@@ -0,0 +1,60 @@
...
+      @Nullable boolean updateParent) {
+    folderDao.updateFolder(
+        workspaceUuid, folderId, displayName, description, parentFolderId, updateParent);
+    return folderDao.getFolder(workspaceUuid, folderId);
+  }
+
+  public Folder getFolder(UUID workspaceUuid, UUID folderId) {
+    return folderDao
+        .getFolderIfExists(workspaceUuid, folderId)
+        .orElseThrow(
```
melissachang, line 39: Dao already throws FolderNotFoundException for this case, delete

Though things are bit inconsistent. For getFolder, dao returns FolderNotFoundException. For deleteFolder, service returns FolderNotFoundException. Can we be consistent?

service/src/main/resources/db/changesets/20220824_folder_table.yaml
```diff
@@ -0,0 +1,42 @@
...
+        - column:
+            name: workspace_id
+            type: text
+            constraints:
+                nullable: false
+                references: workspace(workspace_id)
+                foreignKeyName: fk_folder_wid
+                deleteCascade: true
+            remarks: The UUID of the workspace
+        - column:
```
melissachang, line 20: super nit - can this be first column :)

service/src/test/java/bio/terra/workspace/common/fixtures/WorkspaceFixtures.java
```diff
@@ -64,4 +79,22 @@ public static ApiCreateWorkspaceRequestBody createWorkspaceRequestBody() {
   public static String getUserFacingId(UUID workspaceId) {
     return String.format("user-facing-id-%s", workspaceId);
   }
+
+  public static ApiCreatedWorkspace createDefaultWorkspace(
```
melissachang, line 83: Thanks for creating. Can this go in MockMvcUtils? This is only for mock mvc tests. Also, call this createWorkspace.

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,545 @@
...
+  @Autowired MockMvc mockMvc;
+  @Autowired ObjectMapper objectMapper;
+
+  @MockBean SamService mockSamService;
+
+  @BeforeEach
+  public void setUp() throws InterruptedException {
+    // Needed for workspace creation as logging is triggered when a workspace is created in
+    // `WorkspaceActivityLogHook` where we extract the user request information and log it to
+    // change agents.
```
melissachang, line 60: nit "change agents" -> "activity log"

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,545 @@
...
+        .when(mockSamService)
+        .checkAuthz(
+            any(AuthenticatedUserRequest.class),
+            eq(SamConstants.SamResource.WORKSPACE),
+            anyString(),
+            eq(SamConstants.SamWorkspaceAction.WRITE));
+
+    createFolderExpectCode(
+        workspaceId,
+        "foo",
```
melissachang, line 95: opt nit

Thanks for adding comments for the null params

Even for not null params, comments are helpful. It's not clear `foo` is folder name and `firstFolder.getId()` is parent folder id (eg it could be folder id of the folder you're trying to create).

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,545 @@
...
+    createFolderExpectCode(
+        workspaceId,
+        displayName,
+        /*description=*/ null,
+        firstFolder.getId(),
+        HttpStatus.SC_BAD_REQUEST);
+  }
+
+  @Test
+  public void createFolder_duplicateNameAtDifferentFolderLevel_succeeds() throws Exception {
```
melissachang, line 138: Delete this test, since it's identical to lines 112-127? (This is where having one test do multiple things makes things easier. Instead of hundreds of lines of tests doing overlapping things, you have one test that's < 50 lines.)

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,545 @@
...
+        createFolder(workspaceId, displayName, description, firstFolder.getId());
+
+    assertEquals(displayName, secondFolder.getDisplayName());
+    assertEquals(description, secondFolder.getDescription());
+    assertEquals(firstFolder.getId(), secondFolder.getParentFolderId());
+    assertNotNull(secondFolder.getId());
+  }
+
+  @Test
+  public void createFolder_duplicateNameUnderDifferentParent_succeeds() throws Exception {
```
melissachang, line 157: Add test createFolder_workspaceDoesNotExist_throwsXXX ?

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,545 @@
...
+        workspaceId,
+        folder.getId(),
+        newDisplayName,
+        newDescription,
+        /*parentFolderId=*/ UUID.randomUUID(),
+        /*updateParent=*/ false,
+        HttpStatus.SC_NOT_FOUND);
+  }
+
+  private ApiFolder createFolder(
```
melissachang, line 401: opt description isn't needed for most of your tests. Create a second createFolder() that doesn't take in description (that calls this createFolder with blank description).

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,545 @@
...
+            any(AuthenticatedUserRequest.class),
+            eq(SamConstants.SamResource.WORKSPACE),
+            anyString(),
+            eq(SamConstants.SamWorkspaceAction.READ));
+
+    getFolderExpectCode(workspaceId, folder.getId(), HttpStatus.SC_FORBIDDEN);
+  }
+
+  @Test
+  public void getFolder_invalidWorkspaceAndFolder_throws404() throws Exception {
```
melissachang, line 212: getFolder_workspaceAndFolderDoNotExist_throws404

(or split up
getFolder_workspaceDoesNotExist_throws404
getFolder_folderDoesNotExist_throws404)

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,545 @@
...
+  private ResultActions getFolderExpectCode(UUID workspaceId, UUID folderId, int code)
+      throws Exception {
+    return mockMvc
+        .perform(
+            addAuth(get(String.format(FOLDER_V1_PATH_FORMAT, workspaceId, folderId)), USER_REQUEST))
+        .andExpect(status().is(code));
+  }
+
+  private ApiFoldersList listFolders(UUID workspaceId) throws Exception {
+    String foldersGetResponse =
```
melissachang, line 522: This is list response, not get response. Rather than fixing, I suggest just calling everything serializedResponse. Then you don't have to edit for each method. (And people copy-pasting this won't have to edit for each method.)

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
```diff
@@ -0,0 +1,545 @@
...
+        listFoldersExpectCode(workspaceId, HttpStatus.SC_OK)
+            .andReturn()
+            .getResponse()
+            .getContentAsString();
+    var retrievedFolders = objectMapper.readValue(foldersGetResponse, ApiFoldersList.class);
+    return retrievedFolders;
+  }
+
+  private ResultActions listFoldersExpectCode(UUID workspaceId, int code) throws Exception {
```
melissachang, line 531: Either return void, or convert ResultActions to ApiFolderList. Same for rest of this file.

service/src/test/java/bio/terra/workspace/db/FolderDaoTest.java
```diff
@@ -0,0 +1,347 @@
...
+import bio.terra.workspace.service.workspace.model.Workspace;
+import bio.terra.workspace.service.workspace.model.WorkspaceStage;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.UUID;
+import javax.annotation.Nullable;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class FolderDaoTest extends BaseUnitTest {
```
melissachang, line 26: I didn't see this file yesterday. I see you have a bunch of tests here. FYI for major functionality (create folder fails when workspace doesn't exist) I will ask for controller test. That way, it's tested regardless of whether logic lives in controller, service, or dao. I don't have to know how a particular endpoint is implemented, I'll just always look in controller test. For me, tests don't also need to be here. Ideally, we would not duplicate tests.

yuhuyoyo: Unit test should test what that corresponding class is doing. 

the controller does auth checking and convert folder to ApiFolder so that should be the thing to test in controller test. 

The implementation details like duplicate display name handling, folder not found,  cycles etc are folderDao implementation details so i prefer them to be in the folderDao. If we want to dedup, I'd prefer to remove them from the controllerTest

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
yuhuyoyo: it is used by listFolder which take in the ResultAction and convert it to an ApiFolderList.

When the code is not 200, there's no apiFolderList to convert to. And I want to re-use this code for the `listFolder` helper method.

service/src/test/java/bio/terra/workspace/db/FolderDaoTest.java
melissachang: To me, this is another rule that's not hard-and-fast.

I think of our controller tests as testing a black box. We call createFolder like a user would call it; we return what user would get. Those are the most important tests. A Dao unit test could pass, but the controller test could fail, if there's a bug in controller or service layer.

For something like cycle, sure -- that's a corner case, that could be in Dao test.

It's strange to me if "Create X fails if workspace doesn't exist" test is in controller test, service test, or dao test, depending on where code is implemented.

Interesting discussion. :) (I'm not requiring any changes.)

melissachang: From https://g3doc.corp.google.com/cloud/control/g3doc/users/style/unit_tests.md#prefer-fakes-over-mocks

black box test which tests the public interface and behaviors

Black box test is almost always preferable to white box test,

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
melissachang: Got it. Do you mind adding this comment

// Can't return ApiFolderList because when code is not 200, there is not ApiFolderList to convert to. Returns ResultActions because this is called by createFolder().

service/src/main/java/bio/terra/workspace/db/FolderDao.java
```diff
@@ -0,0 +1,254 @@
...
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Component;
+
+@Component
+public class FolderDao {
+  private static final String DEFAULT_ROOT = "0";
+  private static final RowMapper<Folder> FOLDER_ROW_MAPPER =
+      (rs, rowNum) -> {
+        var parentFolderIdString = rs.getString("parent_folder_id");
```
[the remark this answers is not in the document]
yuhuyoyo: i had it but removed it because DEFAULT_ROOT.equals(null) will still be false so it won't throw NPE like the rest.

service/src/main/java/bio/terra/workspace/service/folder/FolderService.java
yuhuyoyo: yeah.. i don't know what's the rule here. getFolder throw in dao because the sql query throws exception and we wrap it. For delete, there's no exception being thrown.

service/src/main/java/bio/terra/workspace/db/FolderDao.java
yuhuyoyo: there is a listFolders endpoint. It's in the controller unit test too. Also in the folder service

sorry there are too many new files.

service/src/main/java/bio/terra/workspace/db/FolderDao.java
yuhuyoyo: there is one too, it's called updateFolder_formCycle_throwsBadRequestException.

i dont' like asserting custom error message because they are rather fickle. The test catching it throwing the right kind of exception should be sufficient to catch bugs.

service/src/main/java/bio/terra/workspace/db/FolderDao.java
yuhuyoyo: it's because display name can be null and I dont' want to query the database for the folder display name just to how error message.

For example, there is a foo and another foo under foo.
If I update foo to be the same level as foo, the update request will have display name null but the request will fail because there is already a foo at the top level.

service/src/test/java/bio/terra/workspace/db/FolderDaoTest.java
yuhuyoyo: i don't see how the discussion for fakes vs mocks applies here.

the services are still mocked out in the folderApiControllerTest. The controller shouldn't worry about the implementation details being brokens in the folderDao. Say there are three controllers calling folderDao. I don't think all three should need to / able to cover all the edge cases as they are guaranteed through the unit test.

But i indeed caught bugs in the controller test for the controller code.

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/FolderApiControllerTest.java
yuhuyoyo: this was added by your previous suggestions tho.
I originally just have test for foo/foo. And this one is added for foo/copycat, bar/copycat. 

I changed the test to foo/foo and bar/foo to merge the two tests.

service/src/test/java/bio/terra/workspace/db/FolderDaoTest.java
melissachang: I see your point, I guess I have different paranoia level. :)

In FolderApiControllerTest, `SamService` is mocked because it talks to SAM. It uses real Folder controller, Folder service, and Folder dao. So it is mimicing what user does.

If my Controller implements CreateFoo endpoint, Controller test should have a test for createFoo(). I think we're in agreement there.

If CreateFoo throws 403 for certain case, I would add controller test for that.

If a second Controller implements CreateBar, and also uses folderDao, I would add a throws404 controller test for that second API. I see your point that it's duplicated, so fine not to add. But I would add because:

- Technically there could be a bug with CreateBar throws 404, that is not caught by Dao test. In my experience, when you have tests for bits and pieces of your stack, but you don't test the whole stack, you miss bugs. The folder code is simple, but you can imagine for other endpoints, there's complex interaction between various layers.
- To me, it's nice if there are controller tests for everything that's advertised in yaml

yuhuyoyo: i agree 403 needs to be catch at controller level, added 403 tests for all the controller test.

i think we are in general agreement that test coverage should be thorough across the stack. I like the codedacity scanner that some broad repo is using.

openapi/src/parts/folder.yaml
```diff
@@ -0,0 +1,160 @@
...
+          description: The human readable name of the folder
+          type: string
+        description:
+          description: A description of the folder
+          type: string
+        parentFolderId:
+          description: |
+            The parent folder id to update this folder to. When null and 
+            updateParent is false, do not update. When null and
+            moveToTopLevel is true, move the folder to be the top-level.
```
ddietterich, line 154: nit: missed a `moveToTopLevel`

ddietterich (approved):
LGTM

PR #808, opened 2022-09-08
[TOAZ-189] Update WSM to check LandingZoneService for the AzureRelayNamespace

jordansafer:
Call out- Leonardo currently pulls AzureRelayNamespace by calling the enumerateResources endpoint. We could execute this logic as a case in enumerateResources, or we could move Leo over to hit this endpoint instead-


service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -459,11 +479,47 @@ public ResponseEntity<ApiAzureIpResource> getAzureIp(UUID workspaceUuid, UUID re
...
       UUID workspaceId, UUID resourceId) {
     final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
     features.azureEnabledCheck();
+
+    // TODO take resource-id into consideration resourceId if there are multiple azure Relay
+    // instances
+    //  in the landing zone (no plan for this currently)
+    final AzureCloudContext azureCloudContext =
+        azureCloudContextService.getRequiredAzureCloudContext(workspaceId);
+    final bio.terra.landingzone.model.AzureCloudContext lzAzureCloudContext =
```
rtitle, line 493: Just for my own understanding -- are we inferring the landing zone from the tenant/subscription/mrg? This implies the mrg:landingZone relationship is 1:1.

Alternatively we could store the `landingZoneId` as part of the WSM `AzureCloudContext`. That would make things a little more explicit. Are there plans to do that or are we sticking with the 1:1 relationship for now?


are we storing the landingZoneId anywhere in the WSM workspace, or are we inferring the landing zone from the MRG?

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -459,11 +479,47 @@ public ResponseEntity<ApiAzureIpResource> getAzureIp(UUID workspaceUuid, UUID re
...
+    List<LandingZoneResource> resources =
+        landingZoneService.listResourcesByPurpose(
+            ResourcePurpose.SHARED_RESOURCE, lzAzureCloudContext);
+    Optional<LandingZoneResource> azureRelayOptional =
+        resources.stream()
+            .filter(
+                resource -> Objects.equals(resource.resourceType(), "Microsoft.Relay/Namespaces"))
+            .findFirst();
+
+    if (azureRelayOptional.isPresent()) {
```
rtitle, line 507: Hm, so this makes WSM first check the LZ for the resource, and fall back to the WSM resource manager if there is no LZ. 

1. I wonder if this should be inverted? Then the pattern would be to default to LZ-level resources, but be able to override them with WSM-level resources.

2. Does this logic belong in WSM at all? Perhaps WSM should only be concerned with resources it tracks. We could push the burden to Leonardo to consult the LZ service or WSM to obtain the relay namespace. I kind of like this because it keeps WSM more narrowly focused, and fits Leonardo's role as an orchestrator of downstream systems.

ddietterich: My mental model was that WSM would read a description of LZ resources and use those for its allocation and that the LZ would not be exposed to other services. It makes me kind of twitchy to have two components that are both exposing cloud resource creation to upper layers. I'll have to think about it further.

ddietterich: I thought that WSM would be allocating some thing - a node pool or a namespace - and that would be the resource handed out. But I have not looked at the details.

jordansafer: Some replies!

1.1: If we do this generically, I agree. For this case, as I understand it, there is no need for having workspace level Azure Relay Namespaces, so the workspace fallback is temporary and potentially not required.

1.2: For the current implementation, this seems cleaner to me too. I could make a quick parallel PR on Leo that shows what this would look like. For the future, we were looking at storing the LandingZone Id in WSM (this ties in to the other conversation about use AzureCloudContext to look up the LZ resources. We're doing this right now because we don't have the LZ id).

2/3. It also makes sense to me that we would read LZ resources before making allocation. I'm not sure if this would happen in WSM or Rawls. Currently, we trigger AzureRelayNamespace creation in WSM from Rawls. We should check the LZ before making resources. To Rob's point, it's definitely awkward to return LZ resources via WSM getControlledResource APIs, given that these resources are not "controlled" in that sense. I could see this coming out 2 ways:
- We call WSM to get the LZ id, and then call the LZService directly to get resources.
- We call WSM to get the resource, and WSM handles the abstraction of whether or not there is an LZ. This makes WSM a source of truth on resources associated with a workspace, not just resources controlled by WSM.

4. I'm not familiar with the AKS interactions, @giventocode is working on VM creation so we can check his thoughts

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
jordansafer: We are, I don't love it either. Right now we don't have landingZoneId stored in WSM, so this is future work we'll need.

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
giventocode: Re: relay namespace: The approach I was considering aligns with what @ddietterich mentioned: If the request does not include a namespace, then query the LZ to get the relay namespace -which will be tagged in the LZ as a shared resource. The entry point for any interaction within WSM should be via the "dispatcher," which wraps the high-level LZ service from the library - if this is not the preferred approach, any suggestions are welcome. However, the capability to query the LZ for its resources is not yet published in the LZ service component. Maria is working on these changes, and we should release a new version soon. I plan to use this version and follow this approach for the VM creation to determine the subnet where the VM needs to be created.

Re: AKS: The current approach is, as @rtitle mentioned, calling the LZ REST API to get the resources in the LZ grouped by purpose. The AKS cluster will be listed there. The expectation is that Leo can talk directly with the cluster for AKS deployments. If there's a need for ARM operation -e.g., creating a node pool - and there is no concern about making the node pool a WSM-controlled resource, then we can follow the same pattern: the AKS cluster is optional, in the WSM node pool creation request handled if no AKS cluster is provided, then look in one exists in the LZ.

rtitle: Re AKS -- one of the options [here](https://docs.google.com/document/d/1irNSwQUkSJyiutajclPRI_7oezZJFnmvI6-_fKc8TGk/edit#heading=h.6lyw39fmck9k) calls for a single shared nodepool allocated by the LZ. In that case I'm not sure we'd need to allocate any WSM resources for AKS -- Leo could just get a handle to the cluster/nodepool and start doing stuff. (I want to confirm that's still feasible though.)

Re making WSM an abstraction of landing zones vs exposing LZ Service directly -- I'm just wary of overloading WSM `GET controlled-resource` APIs to return landing zone-level resources. I think the semantics are a little different between WSM resources and LZ resources, so it feels right for them to be separate APIs. WSM will still call LZ internally when allocating controlled resources. There may be special cases where upstream systems like Rawls/Leo need to query the LZ too -- I think that's generally ok, but we should consider case by case. (For the relay example, maybe `hybrid connection pool` should really be a WSM controlled resource instead of allocated by Leo.)

rtitle: Ha we crossed, sorry

ddietterich: I think we should do some design work outside of a PR review :)

Qi77Qi: > I wonder if this should be inverted? Then the pattern would allow us to default to LZ-level resources, but be able to override them with WSM-level resources.

Doesn't feel strongly either way...But I think we should document the semantics precisely in swagger


> Does this logic belong in WSM at all? Perhaps WSM should only be concerned with resources it tracks. We could push the burden to Leonardo to consult the LZ service or WSM to obtain the relay namespace. I kind of like this because it keeps WSM more narrowly focused, and fits Leonardo's role as an orchestrator of downstream systems. But open to thoughts here..

I kinda prefer LZ stays transparent to upstream services. Or what is the benefit of having services like Leo knows about LZ vs LZ stays hidden behind WSM (either as part of it or a separate service)

jordansafer: Working on a design doc for this-

PR #814, opened 2022-09-14
PF-1956: Only return optional for fields that are nullable

yuhuyoyo:
Before this PR we are returning optional for display name assuming that it can be null. But it actually cannot be because the when building a workspace, we always set the display name to "". this is confusing to the caller of getDisplayName as it returns an optional that's never empty.

Alternatively I can just remove the optional and have getDisplayName returns a string. But I see no benefit of setting the display name to "" in this case. The database allows display name to be null, the api response also allows it to be null.


service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -118,8 +119,6 @@ public UUID createWorkspace(
             .operationType(OperationType.CREATE)
             .addParameter(
                 WorkspaceFlightMapKeys.WORKSPACE_STAGE, workspace.getWorkspaceStage().name())
-            .addParameter(WorkspaceFlightMapKeys.DISPLAY_NAME, workspaceName)
```
yuhuyoyo, line 121: those two keys are not used in the flight.

melissachang (approved):
Sorry it took me a while to understand this PR. Thanks for the cleanup.

ginaygoogle (approved):

PR #815, opened 2022-09-14
First cut at adding optional CORS support for azure storage accounts

aherbst-broad:
## Context
[Related ticket](https://broadworkbench.atlassian.net/browse/WOR-503)
We need to set CORS allowed origins, methods and headers on Azure storage containers so calls from the browser succeed.

## This PR
* Wires up the call to Azure to set the needed values on the owning storage account at the time of creation. The allowed origins are config'ed out and default to none (these will be set via a subsequent terra-helmfile PR).

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/CreateAzureStorageStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/CreateAzureStorageStep.java
@@ -46,27 +46,18 @@ public class CreateAzureStorageStep implements Step {
         context
             .getWorkingMap()
             .get(ControlledResourceKeys.AZURE_CLOUD_CONTEXT, AzureCloudContext.class);
+    if (azureCloudContext == null) {
+      logger.error(
+          "Azure cloud context is null for storage account creation. workspace_id = {}",
+          resource.getWorkspaceId());
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL);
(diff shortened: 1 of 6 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storage/CreateAzureStorageStep.java
```diff
@@ -11,25 +11,37 @@
...
+import com.azure.resourcemanager.storage.models.StorageAccount;
+import com.azure.storage.blob.BlobServiceClient;
+import com.azure.storage.blob.BlobServiceClientBuilder;
+import com.azure.storage.blob.models.BlobCorsRule;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class CreateAzureStorageStep implements Step {
   private static final Logger logger = LoggerFactory.getLogger(CreateAzureStorageStep.class);
+  private static final String CORS_ALLOWED_METHODS = "GET,HEAD,OPTIONS,PUT,PATCH,POST,MERGE,DELETE";
```
aherbst-broad, line 27: We can config these values out if need be, but I don't expect them to vary per-environment.

blakery: I don't see a ton of room for that to expand anyway

blakery (approved):

marctalbott (approved):

PR #816, opened 2022-09-14
TOAZ-209 ephemeral os disk, custom data, optional existing disk

dvoet:


service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -6,30 +6,7 @@
...
-import bio.terra.workspace.generated.model.ApiCreatedControlledAzureIp;
-import bio.terra.workspace.generated.model.ApiCreatedControlledAzureNetwork;
-import bio.terra.workspace.generated.model.ApiCreatedControlledAzureStorage;
-import bio.terra.workspace.generated.model.ApiCreatedControlledAzureStorageContainer;
-import bio.terra.workspace.generated.model.ApiCreatedControlledAzureVmResult;
-import bio.terra.workspace.generated.model.ApiDeleteControlledAzureResourceRequest;
-import bio.terra.workspace.generated.model.ApiDeleteControlledAzureResourceResult;
-import bio.terra.workspace.generated.model.ApiJobControl;
-import bio.terra.workspace.generated.model.ApiJobReport;
+import bio.terra.workspace.generated.model.*;
```
sergiygetlin, line 9: As far as I know we are trying to avoid wildcard imports.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
```diff
@@ -252,4 +243,65 @@ private VirtualMachine.DefinitionStages.WithCreate buildVmConfiguration(
...
+  private VirtualMachine.DefinitionStages.WithManagedCreate maybeAddExistingDiskStep(
+      VirtualMachine.DefinitionStages.WithFromImageCreateOptionsManaged priorSteps,
+      Optional<Disk> disk) {
+    return disk.map(priorSteps::withExistingDataDisk).orElse(priorSteps);
+  }
+
+  private VirtualMachine.DefinitionStages.WithFromImageCreateOptionsManaged maybeAddCustomDataStep(
+      VirtualMachine.DefinitionStages.WithFromImageCreateOptionsManaged priorSteps,
+      ApiAzureVmCreationParameters creationParameters) {
+    if (StringUtils.isEmpty(creationParameters.getCustomData())) {
```
sergiygetlin, line 283: Will it be easier to read with ternary operator?

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -367,6 +336,23 @@ public ResponseEntity<ApiCreatedControlledAzureVmResult> createAzureVm(
     return new ResponseEntity<>(result, HttpStatus.OK);
   }
 
+  @VisibleForTesting
+  ControlledAzureVmResource buildControlledAzureVmResource(
+      ApiAzureVmCreationParameters creationParameters, ControlledResourceFields commonFields) {
+    ControlledAzureVmResource resource =
```
sergiygetlin, line 342: Do we need local variable?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
```diff
@@ -252,4 +243,63 @@ private VirtualMachine.DefinitionStages.WithCreate buildVmConfiguration(
...
+
+  private VirtualMachine.DefinitionStages.WithManagedCreate maybeAddEphemeralDiskStep(
+      VirtualMachine.DefinitionStages.WithManagedCreate priorSteps,
+      ApiAzureVmCreationParameters creationParameters) {
+    var maybePlacement =
+        Optional.ofNullable(creationParameters.getEphemeralOSDisk())
+            .flatMap(
+                (ephemeralOSDisk) -> {
+                  final Optional<DiffDiskPlacement> diskPlacement;
+                  switch (ephemeralOSDisk) {
```
rtitle, line 255: Could this be a place for Java 17 pattern matching? 
https://docs.oracle.com/en/java/javase/17/language/pattern-matching-switch-expressions-and-statements.html

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
```diff
@@ -815,6 +815,93 @@ public void createAndDeleteAzureVmControlledResourceWithCustomScriptExtensionWit
...
+                    .accessScope(AccessScopeType.fromApi(ApiAccessScope.SHARED_ACCESS))
+                    .managedBy(ManagedByType.fromApi(ApiManagedBy.USER))
+                    .build())
+            .vmName(creationParameters.getName())
+            .vmSize(creationParameters.getVmSize())
+            .vmImage(AzureVmUtils.getImageData(creationParameters.getVmImage()))
+            .region(creationParameters.getRegion())
+            // .ipId(ipResource.getResourceId())
+            // .diskId(diskResource.getResourceId())
+            .networkId(networkResource.getResourceId())
```
rtitle, line 853: Should this add an ephemeral disk and some custom data?

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
```diff
@@ -815,6 +815,93 @@ public void createAndDeleteAzureVmControlledResourceWithCustomScriptExtensionWit
...
+    // checkForResource(resourceList, ipResource);
+    // checkForResource(resourceList, diskResource);
+    checkForResource(resourceList, networkResource);
+    checkForResource(resourceList, resource);
+
+    ComputeManager computeManager = azureTestUtils.getComputeManager();
+
+    final VirtualMachine resolvedVm = getVirtualMachine(creationParameters, computeManager);
+
+    // Submit a VM deletion flight.
```
rtitle, line 872: I was wondering if the Azure VM deletion flight needs to be updated to handle optional data disk + ephemeral OS disk.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
rtitle: Ah nevermind I guess it's in `creationParameters`

rtitle (approved):
Overall looks good, as long as the connected tests are exercising the new parameters, for create and delete.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/CreateAndDeleteAzureControlledResourceFlightTest.java
dvoet: ephemeral disk lifecycle is tied to the VM, external disks require a different flight (there are other tests that do this)

aherbst-broad (approved):

service/src/test/java/bio/terra/workspace/common/fixtures/ControlledResourceFixtures.java
```diff
@@ -183,6 +179,27 @@ public static ApiAzureVmCreationParameters getAzureVmCreationParameters() {
...
+    return new ApiAzureVmCreationParameters()
+        .name(uniqueAzureName(AZURE_VM_NAME_PREFIX))
+        .region("westcentralus")
+        .vmSize(VirtualMachineSizeTypes.STANDARD_D8S_V3.toString())
+        .vmImage(
+            new ApiAzureVmImage()
+                .publisher("microsoft-dsvm")
+                .offer("ubuntu-1804")
+                .sku("1804-gen2")
+                .version("latest"))
```
dinvlad, line 193: We should probably make it a config parameter, so that it's easier to update later.

dvoet: I am just copying what the other tests do

sergiygetlin (approved):

ddietterich:
This should have derived from BaseAzureTest or else explicitly checked the Azure-enabled feature flag so that it would not run in environments where Azure is not enabled. Like the Verily deployment for example...

ddietterich:
I filed PF-2025 and am working on a fix.

rtitle:
Thanks Dan - I always miss that on reviews. Wonder if it's worth having a GitHub action that runs all tests with Azure disabled?

ddietterich:
I'll give that some thought. I think all unit tests should always pass regardless of the feature flag settings. So maybe a run with everything `false` would do the trick.

PR #818, opened 2022-09-15
[DC-534] Handle updated TDR forbidden response

zloery:
In https://github.com/DataBiosphere/jade-data-repo/pull/1343, TDR began returning 403 responses when a user does not have permission, rather than 401, so WSM needs to handle the new response. This broke the `ReferencedDataRepoSnapshotLifecycle` integration test.

This is the only spot WSM calls TDR, so I don't think we need to update anywhere else


yuhuyoyo (approved):

ddietterich (approved):

yuhuyoyo:
Hi Zach, I'm just gonna merge this to eliminate noise on my PR.

PR #819, opened 2022-09-16
[PF-2013] Add log based alert for stale status

zloery:
This upgrades the log message for stale status to fire an alert in any environment with log-based alerting configured, instead of just logging a warning. This should also be a more informative message for oncall engineers.


jmczerk (approved):
Thanks!

yuhuyoyo (approved):

PR #820, opened 2022-09-16
PF-2011: Add a LIMIT by 1 to avoid returning multiple rows

yuhuyoyo:


yuhuyoyo:
does @writeTransaction lock the write to database? I wonder if I should just addDefaultValueCompute: now() to the change_date column.

ddietterich (approved):
How about that. We actually got duplicate results from _now()_ on separate activity log writes?!

This looks like a straightforward fix.

cbookg (approved):

PR #822, opened 2022-09-19
[TOAZ-104] Create VM operation looks for LZ VNet if a workspace VNET is not provided

giventocode:


Qi77Qi (commented):
is there any API change?
Also I think this should follow similar pattern to relayNamepsace?

Qi77Qi (approved):

Qi77Qi:
lgtm once tests pass

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -120,6 +133,19 @@ public ApiAzureLandingZoneResourcesList listAzureLandingZoneResources(String lan
...
   }
 
+  public List<ApiAzureLandingZoneDeployedResource> listSubnetsWithParentVNetByPurpose(
+      String landingZoneId, LandingZonePurpose purpose) {
+
+    if (StringUtils.isBlank(landingZoneId)) {
+      throw new LandingZoneInvalidInputException("The landing zone id can't be null or empty");
+    }
+
+    return listAzureLandingZoneResources(landingZoneId).getResources().stream()
```
rtitle, line 143: Should something in here filter for subnet resource type (or am I misunderstanding)?

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureNetworkInterfaceStepTest.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.UUID;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+import org.springframework.test.context.ActiveProfiles;
+
+@ActiveProfiles("unit")
+public class CreateAzureNetworkInterfaceStepTest extends BaseUnitTest {
```
ddietterich, line 30: Isn't this an Azure test? Should it be derived from BaseAzureTest?
If it is doing anything that requires the azure-enabled feature to be true, then it should be an Azure test.

Even though azure-enabled is true for all Broad environments it is not true for all environments.

ddietterich (changes requested):
Why is this PR in Draft state, but is being reviewed?

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -57,13 +60,23 @@ public ApiAzureLandingZoneResult createAzureLandingZone(
                     new LandingZoneInvalidInputException(
                         "LandingZoneTarget is required when creating an Azure landing zone"));
 
+    final LandingZoneTarget target = MapperUtils.LandingZoneTargetMapper.from(apiLandingZoneTarget);
+
+    // TODO: this check could be removed once WSM maintains the LZ id in the cloud context.
```
rtitle, line 65: I think this is independent of workspace/cloud context since this code relates to LZ creation. We probably still want it to fail if a LZ is already deployed in a given billing-profile (at least while we're enforcing a 1:1 relationship between LZ and billing profile).

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureNetworkInterfaceStepTest.java
rtitle, line 30: I think this should extend `BaseAzureTest` (which respects the feature flag).

rtitle (commented):
Few comments, looks good overall. I think there are some minor conflicts due to https://github.com/DataBiosphere/terra-workspace-manager/pull/816, which also touched VM creation code in WSM. Can re-review when out of draft state.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -169,4 +195,22 @@ private ApiAzureLandingZoneResult fetchCreateAzureLandingZoneResult(String jobId
...
         .landingZone(azureLandingZone);
   }
+
+  /*
+   * Note: This method is an initial implementation that must be revised,
+   * and likely be refactored once WSM stores the LZ id in the azure context.
+   * The initial assumption is that the cardinality of 1:1 between the cloud context and the LZ
+   * is enforced in the create operation, therefore more than one LZ per azure context is not allowed.
+   */
+  public String getLandingZoneId(AzureCloudContext azureCloudContext) {
```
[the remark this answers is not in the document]
giventocode: Correct.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureNetworkInterfaceStepTest.java
giventocode: Yeah, I initially had it like that, however I noticed the base test sets a Spring profile for Azure, which I don't require for this test. It is not clear to me what is the best way to describe a unit test that applies to Azure if I don't require a spring context. Happy to change it to BaseAzureTest is that is the best answer we have right now, please let me know.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -169,4 +195,22 @@ private ApiAzureLandingZoneResult fetchCreateAzureLandingZoneResult(String jobId
...
+    return getFirstLandingZoneId(MapperUtils.LandingZoneTargetMapper.from(azureCloudContext))
+        .orElseThrow(
+            () ->
+                new IllegalStateException(
```
sergiygetlin, line 209: In this particular method we are trying to find LZ. Wouldn't be LandingZoneNotFoundException more informative? Not sure if IllegalStateException is appropriate here.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureNetworkInterfaceStep.java
```diff
@@ -108,6 +114,46 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
+    ApiAzureLandingZoneDeployedResource lzResource =
+        landingZoneApiDispatch
+            .listSubnetsWithParentVNetByPurpose(
+                lzId, SubnetResourcePurpose.WORKSPACE_COMPUTE_SUBNET)
+            .stream()
+            .findFirst()
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
+                        "The landing zone does not contain the requested resource."));
```
sergiygetlin, line 151: Should we specify our expectations in the exception? It is not clear from the message which resource we are looking for.
Also, what if we use (as example) LandingZoneIllegalStateException here? IllegalStateException here is too generic from my point of view.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureNetworkInterfaceStepTest.java
```diff
@@ -0,0 +1,125 @@
...
+    networkInterfaceStep =
+        new CreateAzureNetworkInterfaceStep(
+            azureConfiguration, crlService, resource, resourceDao, landingZoneApiDispatch);
+    when(networkManager.networks()).thenReturn(networks);
+    var subnets = new HashMap<String, Subnet>();
+    subnets.put(STUB_SUBNET, armSubnet);
+    when(armNetwork.subnets()).thenReturn(subnets);
+  }
+
+  @Test
```
sergiygetlin, line 69: These two tests look almost identical except setup part. Should we somehow differentiate the asserts? Maybe to have different name/id?
The method under test is not public. Q: 1) Do we have default modifier for this method only for testing purposes? 2) Should we also test getNetworkResourcesFromLandingZone (since it's a part of the previous method)?

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
giventocode: The cause of the exception is that the creation flight is not complete (the db is not updated), hence a timing issue in which case an IllegalStateException is appropriate. I will add more details to the error message to make this clearer.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureNetworkInterfaceStep.java
giventocode: Same as above, I will add more details to the error message.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureNetworkInterfaceStepTest.java
giventocode: Mockito uses strict stubbing. If the interaction as outlined in the setup is not used, then the test will fail. The assertion confirms that through the expected calls the method returns the expected network resources. The method is scoped to the package, because it should not be public, and since there was no test for this step, this was the best way I found to test my change without writing an overall test.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureNetworkInterfaceStep.java
```diff
@@ -108,6 +116,46 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
+  }
+
+  private NetworkSubnetPair getNetworkResourcesFromLandingZone(
+      AzureCloudContext azureCloudContext, NetworkManager networkManager) {
+
+    final String lzId = landingZoneApiDispatch.getLandingZoneId(azureCloudContext);
+
+    ApiAzureLandingZoneDeployedResource lzResource =
+        landingZoneApiDispatch
+            .listSubnetsWithParentVNetByPurpose(
```
zykonda, line 146: Just a thought here: we might want to add REST API for the landing zone subnets ../{lz_id}/networkSubnets/{SubnetResourcePurpose}, or for getting resources by purpose ../{lz_id}/resources/{LandingZonePurpose} for better performance.
List all resources is an expensive API because it makes calls resource manager to retrieve by tag (N+1) times, where N is a number of purposes, and M times to get resource by Id, where M is a number of VNets in landing zone => N+M+1.
The scenario, however, is to get subnets with exact purpose that is 1 call to get VNets by tag and max(M) times (if each VNet has the subnet for a purpose specified) to get by Id => 1+max(M).

giventocode: I agree. We can optimize this as a follow up. The LZ manager already has a method for querying VNets/Subnets by purpose. We just need to expose it in the service etc..

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
```diff
@@ -130,24 +119,36 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
                       .getWorkingMap()
                       .get(AzureVmHelper.WORKING_MAP_NETWORK_INTERFACE_KEY, String.class));
 
+      // TODO:
```
rtitle, line 122: This makes sense -- wondering if we also want to put a note in the OpenAPI spec and mark the region param `deprecated`? https://github.com/DataBiosphere/terra-workspace-manager/blob/main/openapi/src/parts/controlled_azure_vm.yaml#L113

rtitle (approved):
👍

ddietterich (approved):

PR #817, opened 2022-09-15
[PF-1930] Add tests for cloning controlled GCS bucket with different cloning instructions

melissachang:
Existing "Clone controlled GCS bucket" connected/integration tests use `CloningInstructions.COPY_RESOURCE`. In `ControlledGcpResourceApiControllerConnectedTest.java`, add tests for other `CloningInstructions` enum values, in preparation for implementing `CloningInstructions.COPY_REFERENCE`.

Also make `MockMvcUtils` a spring Component. This way, users don't have to pass in `mockMvc`, `objectMapper`. In order to make `MovkMvcUtils` a spring Component, I had to add `@AutoConfigureMockMvc` to `BaseTest`. I also removed `@AutoConfigureMockMvc` from individual tests.

PR #823, opened 2022-09-19
PF-1965: Add properties fields on resource creation

yuhuyoyo:
Mostly added test coverages.

There is no unit test at the referenced resource controller level so I added referenced resource creation test for each type of resources. I also unified the commonFields construction for referenced resources; this is to eliminate risk of accidentally missing one of the referenced resources when adding new fields. 

I also added controllerConnectedTest for gcp resources creation. the code path is the same for azure and gcp so that should be somewhat sufficient. 

i didn't add much to the integration test because of the thorough test coverage on the unit test level. 

`spotbugsMain` has a couple non-blocking complains so I addresses them as well.

### DIFF
--- a/integration/src/main/java/scripts/utils/CommonResourceFieldsUtil.java
+++ b/integration/src/main/java/scripts/utils/CommonResourceFieldsUtil.java
@@ -20,22 +20,21 @@ public class CommonResourceFieldsUtil {
 
   /**
    * Makes common fields for referenced resources. Sets properties to foo -> bar.
+   *
    * @param name name of the referenced resource.
    * @param cloningInstructions when null, set to NOTHING.
    */
   public static ReferenceResourceCommonFields makeReferencedResourceCommonFields(
       String name, @Nullable CloningInstructionsEnum cloningInstructions) {
     return new ReferenceResourceCommonFields()
-        .cloningInstructions(Optional.ofNullable(cloningInstructions)
-            .orElse(CloningInstructionsEnum.NOTHING))
+        .cloningInstructions(
+            Optional.ofNullable(cloningInstructions).orElse(CloningInstructionsEnum.NOTHING))
         .description("Description of " + name)
         .name(name)
         .properties(getResourceDefaultProperties());
   }
 
-  /**
-   * Makes common fields for Controlled resource. Sets resource properties to foo->bar.
-   */
+  /** Makes common fields for Controlled resource. Sets resource properties to foo->bar. */
   public static ControlledResourceCommonFields makeControlledResourceCommonFields(
       String name,
       PrivateResourceUser privateUser,
--- a/integration/src/main/java/scripts/utils/CommonResourceFieldsUtil.java
+++ b/integration/src/main/java/scripts/utils/CommonResourceFieldsUtil.java
@@ -1,0 +1,50 @@
+package scripts.utils;
+
+import bio.terra.workspace.model.AccessScope;
+import bio.terra.workspace.model.CloningInstructionsEnum;
+import bio.terra.workspace.model.ControlledResourceCommonFields;
+import bio.terra.workspace.model.ManagedBy;
+import bio.terra.workspace.model.PrivateResourceUser;
+import bio.terra.workspace.model.Properties;
+import bio.terra.workspace.model.Property;
+import bio.terra.workspace.model.ReferenceResourceCommonFields;
+import com.google.common.collect.ImmutableMap;
+
+/** Utils to build common resource fields for both referenced resources and controlled resources. */
+public class CommonResourceFieldsUtil {
+
+  public static ImmutableMap<String, String> DEFAULT_RESOURCE_PROPERTIES =
+      ImmutableMap.of("foo", "bar");
+
+  public static ReferenceResourceCommonFields makeReferencedResourceCommonFields(
+      String name, CloningInstructionsEnum cloningInstructions) {
+    return new ReferenceResourceCommonFields()
+        .cloningInstructions(cloningInstructions)
+        .description("Description of " + name)
+        .name(name)
+        .properties(getResourceDefaultProperties());
+  }
+
+  public static ControlledResourceCommonFields makeControlledResourceCommonFields(
+      String name,
+      PrivateResourceUser privateUser,
+      CloningInstructionsEnum cloningInstructions,
+      ManagedBy managedBy,
+      AccessScope accessScope) {
+    return new ControlledResourceCommonFields()
+        .accessScope(accessScope)
+        .managedBy(managedBy)
+        .cloningInstructions(cloningInstructions)
+        .description("Description of " + name)
+        .name(name)
+        .privateResourceUser(privateUser)
+        .properties(getResourceDefaultProperties());
+  }
+
+  public static Properties getResourceDefaultProperties() {
+    Properties properties = new Properties();
+    DEFAULT_RESOURCE_PROPERTIES.forEach(
+        (key, value) -> properties.add(new Property().key(key).value(value)));
+    return properties;
+  }
+}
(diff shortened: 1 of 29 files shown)


yuhuyoyo:
this PR will need some merging once https://github.com/DataBiosphere/terra-workspace-manager/pull/817 is in.

integration/src/main/java/scripts/utils/CommonResourceFieldsUtil.java
```diff
@@ -0,0 +1,50 @@
...
+  public static ReferenceResourceCommonFields makeReferencedResourceCommonFields(
+      String name, CloningInstructionsEnum cloningInstructions) {
+    return new ReferenceResourceCommonFields()
+        .cloningInstructions(cloningInstructions)
+        .description("Description of " + name)
+        .name(name)
+        .properties(getResourceDefaultProperties());
+  }
+
+  public static ControlledResourceCommonFields makeControlledResourceCommonFields(
```
melissachang, line 28: opt a couple callers look like

```
        CommonResourceFieldsUtil.makeControlledResourceCommonFields(
            .name(resourceName)	            resourceName,
            .cloningInstructions(CloningInstructionsEnum.NOTHING)	            /*privateUser=*/ null,
            .accessScope(AccessScope.SHARED_ACCESS)	            CloningInstructionsEnum.NOTHING,
            .managedBy(ManagedBy.USER);	            ManagedBy.USER,
            AccessScope.SHARED_ACCESS);
```

Make a variant that just takes name

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -1115,4 +1054,16 @@ public ResponseEntity<Void> deleteTerraWorkspaceReference(UUID workspaceUuid, UU
         workspaceUuid, resourceId, WsmResourceType.REFERENCED_ANY_TERRA_WORKSPACE, userRequest);
     return new ResponseEntity<>(HttpStatus.OK);
   }
+
+  private static WsmResourceFields getWsmResourceFields(
+      UUID uuid, ApiReferenceResourceCommonFields metadata) {
```
melissachang, line 1059: uuid -> workspaceId. (uuid could mean resourceId)

service/src/main/java/bio/terra/workspace/service/resource/model/ResourceLineageEntry.java
```diff
@@ -52,4 +52,10 @@ public boolean equals(Object o) {
     return sourceWorkspaceId.equals(entry.sourceWorkspaceId)
         && sourceResourceId.equals(entry.sourceResourceId);
   }
+
+  @Override
+  public int hashCode() {
```
melissachang, line 57: Why `assert false`, could you implement a real hashCode? (Or delete this method?)

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/ControlledGcpResourceApiControllerConnectedTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    ApiGcpBigQueryDatasetResource retrievedResource =
+        getBigQueryDataset(
+            mockMvc,
+            objectMapper,
+            workspaceId,
+            resource.getResourceId(),
+            userAccessUtils.defaultUserAuthRequest());
+
+    assertEquals(resource.getBigQueryDataset(), retrievedResource);
```
melissachang, line 63: In general, don't use response as expectation. Say there's a bug and createBigQueryDataset returns the wrong thing, eg properties is missing. This test won't catch that.

You could write assertBigQueryDataset() and call that.

Same for other test.

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/ReferencedGcpResourceControllerTest.java
```diff
@@ -0,0 +1,217 @@
...
+    // activity log.
+    when(mockSamService.getUserStatusInfo(any()))
+        .thenReturn(
+            new UserStatusInfo()
+                .userEmail(USER_REQUEST.getEmail())
+                .userSubjectId(USER_REQUEST.getSubjectId()));
+  }
+
+  @Test
+  public void createDataRepoReferencedResource() throws Exception {
```
melissachang, line 73: nit createReferencedDataRepoResource. Same for rest of file.

You other file had "controlled" before "GcsBucket". (Putting "controlled"/"referenced" first sounds better.)

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/ReferencedGcpResourceControllerTest.java
```diff
@@ -0,0 +1,217 @@
...
+            workspaceId, request, REFERENCED_GIT_REPO_V1_PATH_FORMAT);
+
+    var retrievedResource = objectMapper.readValue(serializedResponse, ApiGitRepoResource.class);
+    assertReferenceResourceCommonFields(requestBody.getMetadata(), retrievedResource.getMetadata());
+    assertEquals(
+        requestBody.getGitrepo().getGitRepoUrl(),
+        retrievedResource.getAttributes().getGitRepoUrl());
+  }
+
+  private String createReferencedResourceAndGetSerializedResponse(
```
melissachang, line 191: It's not useful to return String serializedResponse. Caller will have to convert to ApiGitRepoResource. Do this instead

```
ApiGitRepoResource createReferencedGitRepo() {
}
```

It's better to write `createReferencedGitRepo()` once, than make callers convert to ApiGitRepoResource every time.

melissachang (commented):
Put ticket in PR description ?

service/src/main/java/bio/terra/workspace/service/resource/model/ResourceLineageEntry.java
yuhuyoyo: this  was  a complaint from spotbugs and this is its suggestions. because ResourceLineageEntry won't be put into a hashmap or hashtable, it assert false. I needed to have a hashcode because i have a  custom implementation of equals.

melissachang: ResourceLineageEntry might be put into hashmap or hashtable someday, seems safer to just implement hashCode

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/ReferencedGcpResourceControllerTest.java
yuhuyoyo: i want to have one sharable method for all referenced resource creation because the logic is are almost the same. It's up to the caller what to convert to depending on the resource. There's only one creategitrepo test so it seems not that helpful to have a separate `createReferencedGitRepo`.

integration/src/main/java/scripts/utils/CommonResourceFieldsUtil.java
yuhuyoyo: there's only one of that exact combination.

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/ControlledGcpResourceApiControllerConnectedTest.java
yuhuyoyo: why?

i don't follow. If there is a bug in create, wouldn't the get and create return the different things and get caught here.

My concern with asserting individual fields if someone adds a new field, they wouldn't know to come here to add a new assert line and it would be easily missed. And that would definitely let bug slips through.

melissachang: Say there's a bug in create and get where properties aren't returned. This `assertEquals` won't catch that, because both expected and actual are missing properties.

If someone adds a new field, it is their responsibility to have at least one test check that field. (They don't necessarily need to update all tests.)

melissachang (approved):

zloery (approved):
Please add a ticket number to the PR title

openapi/src/common/schemas.yaml
```diff
@@ -85,6 +85,8 @@ components:
         resourceId:
           type: string
           format: uuid
+        properties:
```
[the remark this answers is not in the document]
yuhuyoyo: no. response uses ResourceMetadat and properties is a non-required field in there.

PR #824, opened 2022-09-20
[PF-2025] Add azure-unit-test; remove unusaed 'azure' profile

ddietterich:
The new controller test Doug added showed up a problem with the test categories. The new test needs the MockMvc to test the REST API interface. The `azure` tag is set up for a _connected_ test, so it does not support the MockMvc.

The default for `azure-enabled` is true, but we still need support for `azure-enabled` as false. So we still need to segregate the Azure tests.

My solution in this PR is:
1. Add a new tag `azure-unit`
2. Add a new profile `azure-unit-test`
3. Add a new property file that sets `azure-enabled`:  `application-azure-unit-test`
4. Added a fourth entry to the test matrix so these tests run as part of PR qualification as well as overnight.

I removed all of the `@ActiveProfiles("azure")` from a bunch of Azure tests. There is no profile-based code or configuration. The tests run fine without it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        gradleTask: [unitTest, connectedTest, azureTest]
+        gradleTask: [unitTest, connectedTest, azureUnitTest, azureTest]
 
     services:
       postgres:
--- a/service/gradle/testing.gradle
+++ b/service/gradle/testing.gradle
@@ -105,3 +105,29 @@ task azureTest(type: Test) {
     failOnPassedAfterRetry = true
   }
 }
+
+task azureUnitTest(type: Test) {
+  useJUnitPlatform {
+    includeTags "azure-unit"
+  }
+  outputs.upToDateWhen { false }
+  finalizedBy tasks.combinedJaCoCoReport
+
+  // For GHA runs, retry failing tests. If 2nd time passes, workflow will still
+  // fail and oncall will still be notified. This way, oncall dosen't have to
+  // rerun workflow in GHA.
+  retry {
+    // This is set automatically on Github Actions runners, and never set
+    // otherwise.
+    if (System.getenv().containsKey("CI")) {
+      // Max retries per test.
+      maxRetries = 1
+      // Max total test failures.
+      // This is an estimate based on current failure rates, but if more
+      // than 5 tests fail in a run it's likely a real source of failure
+      // and we should stop retrying.
+      maxFailures = 5
+    }
+    failOnPassedAfterRetry = true
+  }
+}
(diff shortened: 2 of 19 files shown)


.github/workflows/test.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        gradleTask: [unitTest, connectedTest, azureTest]
```
Qi77Qi, line 41: jsut for my understanding...what's the difference between `azureUnitTest` and `azureTest`?

jmczerk (approved):

.github/workflows/test.yml
ddietterich: `azureTest` is set up for connected tests with Sam, RBS, etc.  But it is not set up for using the MockMvc.

service/gradle/testing.gradle
```diff
@@ -105,3 +105,29 @@ task azureTest(type: Test) {
...
+  useJUnitPlatform {
+    includeTags "azure-unit"
+  }
+  outputs.upToDateWhen { false }
+  finalizedBy tasks.combinedJaCoCoReport
+
+  // For GHA runs, retry failing tests. If 2nd time passes, workflow will still
+  // fail and oncall will still be notified. This way, oncall dosen't have to
+  // rerun workflow in GHA.
+  retry {
```
yuhuyoyo, line 119: its unit test. do we need retry?

yuhuyoyo (approved):

service/gradle/testing.gradle
ddietterich: No, I don't think so. I was just cutting and pasting. Will remove.

Qi77Qi (approved):
would be nice to have some comments around `azureTest` vs `azureUnitTest` (or rename `azureTest` to `azureConnectedTest`) ?

.github/workflows/test.yml
rtitle: Maybe we should `s/azureTest/azureConnectedTest` for clarity?

So basically:
* unitTest -- non-Azure unit tests
* connectedTest -- non-Azure connected tests
* azureUnitTest -- Azure unit tests only
* azureConnectedTest -- Azure connected tests only

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/CreateAzureDiskStepTest.java
```diff
@@ -31,9 +31,7 @@
 import org.junit.jupiter.api.Test;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
-import org.springframework.test.context.ActiveProfiles;
 
-@ActiveProfiles("azure")
 public class CreateAzureDiskStepTest extends BaseAzureTest {
```
rtitle, line 35: I _think_ all the Azure `*StepTest`s can be `BaseAzureUnitTest`. (happy to do that in a separate PR too)

rtitle (approved):
Couple optional comments, but approving as is

ddietterich:
Merging as is. I'll make a follow-up to improve documentation and rename for consistency.

PR #825, opened 2022-09-20
[PF-1987] Return resource information when updating referenced resources.

iliadsh:
Since the controlled (gcp) resources have the same behavior, it avoids having to manually retrieve the updated information when dealing with referenced resources.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/ReferencedBigQueryResourceLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/ReferencedBigQueryResourceLifecycle.java
@@ -214,17 +214,16 @@ public class ReferencedBigQueryResourceLifecycle extends WorkspaceAllocateTestSc
     // Update BQ dataset's name and description
     String newDatasetName = "newDatasetName";
     String newDatasetDescription = "newDescription";
-    BqDatasetUtils.updateBigQueryDatasetReference(
-        fullAccessApi,
-        getWorkspaceId(),
-        bqDatasetResourceId,
-        newDatasetName,
-        newDatasetDescription,
-        /*projectId=*/ null,
-        /*datasetId=*/ null,
-        CloningInstructionsEnum.NOTHING);
     GcpBigQueryDatasetResource datasetReferenceFirstUpdate =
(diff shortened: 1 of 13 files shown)


cbookg:
> Please ensure backward compatibility of the endpoint.

I believe adding a response to an API that previously returned void is backwards compatible.  Do you have concerns that it is not?

ddietterich:
I didn't look at the specific change. Just flagging since it is making an API change.

iliadsh:
FWIW, I spotted no issues with terra-cli update functionality using the added response API. The UI should also just ignore the response data if it expects a void (though I have not tested this).

ddietterich:
Maybe ask on the #terra-workspace-manager channel in Broad slack to see if anyone has a concern.

yuhuyoyo (commented):
can you also update the integration test to reflect the change?

currently the integration test is still doing an upgrade and a get call to check the updated referenced resource.

yuhuyoyo (approved):
thanks!

zloery (commented):
For consistency, could you also update referenced data repo snapshots? I believe it will follow the same pattern as GCP referenced resources.

I believe the `updateDataRepoSnapshotReferenceResource` endpoint is used by Rawls so it's probably worth a message to the #dsp-workspaces channel, but it seems like this is backwards compatible and so shouldn't cause any issues.

iliadsh:
> For consistency, could you also update referenced data repo snapshots? I believe it will follow the same pattern as GCP referenced resources.
> 
> I believe the `updateDataRepoSnapshotReferenceResource` endpoint is used by Rawls so it's probably worth a message to the #dsp-workspaces channel, but it seems like this is backwards compatible and so shouldn't cause any issues.

Thoughts on adding that in a separate PR while waiting for someone to raise an issue on the slack? Asking because the current changes can be used now in the UI.

zloery (approved):

zloery:
> Thoughts on adding that in a separate PR while waiting for someone to raise an issue on the slack? Asking because the current changes can be used now in the UI.

SGTM as long as this doesn't get lost in a backlog somewhere

PR #827, opened 2022-09-20
[PF-2016] Stop writing test user access tokens to logs

zloery:
Currently, MockMvc tests print debug information about the requests they "send", including headers with access tokens. We should stop doing this. These are short-lived access tokens for dedicated test users with no access to real data or resources so I'm not concerned about the real impact, but it's still a best practice.

Additionally, this pulls in the TCL change from https://github.com/DataBiosphere/terra-common-lib/pull/86 to stop writing access tokens to logs. This affects live environments as well as tests. Note this is for server logs (which are access controlled in live environments) unlike the above fix for Github Actions test output (which are publicly visible).

This also removes unnecessary `AutoConfigureMockMvc` annotations now that they're applied from `BaseTest`

### DIFF
--- a/service/gradle.lockfile
+++ b/service/gradle.lockfile
@@ -2,10 +2,10 @@
 # Manual edits can break the build and are not advised.
 # This file is expected to be part of source control.
 bio.terra:datarepo-client:1.41.0-SNAPSHOT=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
-bio.terra:stairway-gcp:0.0.67-SNAPSHOT=compileClasspath,productionRuntimeClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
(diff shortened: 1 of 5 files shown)


service/src/test/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHookTest.java
```diff
@@ -428,5 +429,13 @@ public String prettyStepState() {
...
       return null;
     }
+
+    @Override
+    public ProgressMeter getProgressMeter(String name) {
+      return null;
+    }
+
+    @Override
+    public void setProgressMeter(String name, long v1, long v2) throws InterruptedException {}
```
zloery, line 439: Mostly unrelated, these got pulled in with the new stairway version from the updated TCL

melissachang (approved):

yuhuyoyo (approved):

PR #828, opened 2022-09-20
[PF-1972] Call TPS to merge policy on workspace clone and resource clone

ddietterich:
This is take 2 on this PR. Since the TPS work to support replace went in, I went back and added the replace support. That let me finish up the UNDO of the clone policy step properly.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiDispatch.java
+++ b/service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiDispatch.java
@@ -30,15 +30,14 @@ import bio.terra.workspace.generated.model.ApiTpsPolicyPair;
 import bio.terra.workspace.generated.model.ApiTpsUpdateMode;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
-import org.apache.commons.lang3.StringUtils;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Component;
-
-import javax.annotation.Nullable;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import javax.annotation.Nullable;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
 
 @Component
 public class TpsApiDispatch {
@@ -158,9 +157,15 @@ public class TpsApiDispatch {
 
   private static PaoUpdateMode updateModeFromApi(ApiTpsUpdateMode apiUpdateMode) {
     switch (apiUpdateMode) {
-      case DRY_RUN -> { return PaoUpdateMode.DRY_RUN; }
-      case FAIL_ON_CONFLICT -> { return PaoUpdateMode.FAIL_ON_CONFLICT; }
-      case ENFORCE_CONFLICT -> { return PaoUpdateMode.ENFORCE_CONFLICTS; }
+      case DRY_RUN -> {
+        return PaoUpdateMode.DRY_RUN;
+      }
+      case FAIL_ON_CONFLICT -> {
+        return PaoUpdateMode.FAIL_ON_CONFLICT;
+      }
+      case ENFORCE_CONFLICT -> {
+        return PaoUpdateMode.ENFORCE_CONFLICTS;
+      }
     }
     throw new TpsInvalidInputException("Invalid update mode: " + apiUpdateMode);
   }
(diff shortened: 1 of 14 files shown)


mikenev (approved):
Looks good. You updated all the resource cloning too, nice.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/ClonePolicyAttributesStep.java
```diff
@@ -0,0 +1,105 @@
...
+    ApiTpsPaoGetResult destinationPao = tpsApiDispatch.getPao(token, destinationWorkspaceId);
+    ApiTpsPolicyInputs destinationAttributes = destinationPao.getAttributes();
+    context.getWorkingMap().put(WorkspaceFlightMapKeys.POLICIES, destinationAttributes);
+
+    ApiTpsPaoSourceRequest request =
+        new ApiTpsPaoSourceRequest()
+            .sourceObjectId(sourceWorkspaceId)
+            .updateMode(ApiTpsUpdateMode.FAIL_ON_CONFLICT);
+
+    ApiTpsPaoUpdateResult result = tpsApiDispatch.mergePao(token, destinationWorkspaceId, request);
```
zloery, line 69: I haven't looked at the TPS implmenetation but just to confirm: is it always a no-op to merge a PAO from a workspace into itself? For resource-level clone operations, I think it's possible to clone within a workspace

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CloneControlledGcpBigQueryDatasetResourceFlight.java
```diff
@@ -28,15 +31,21 @@ public CloneControlledGcpBigQueryDatasetResourceFlight(
...
+    var sourceResource = inputParameters.get(ResourceKeys.RESOURCE, ControlledResource.class);
+    var userRequest =
         inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    var destinationWorkspaceId =
+        inputParameters.get(ControlledResourceKeys.DESTINATION_WORKSPACE_ID, UUID.class);
 
-    final CloningInstructions resolvedCloningInstructions =
+    boolean mergePolicies =
+        Optional.ofNullable(
+                inputParameters.get(WorkspaceFlightMapKeys.MERGE_POLICIES, Boolean.class))
```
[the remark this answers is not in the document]
ddietterich: Good catch, thank you.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/ClonePolicyAttributesStep.java
ddietterich: Yes. There wasn't originally, but when I implemented the per-resource clone and looked at all of the tests, I realized we did it all of the time! So now it is a no-op to merge a PAO to itself.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/ClonePolicyAttributesStep.java
```diff
@@ -79,9 +81,26 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
 
   @Override
   public StepResult undoStep(FlightContext context) throws InterruptedException {
-    // var destinationAttributes =
-    //  context.getWorkingMap().get(WorkspaceFlightMapKeys.POLICIES, ApiTpsPolicyInputs.class);
-    // TODO: PF-2019 - when we have a replace method, replace the destination PAO attributes
+    BearerToken token = new BearerToken(userRequest.getRequiredToken());
+    var destinationAttributes =
```
zloery, line 86: If the `do` step fails before writing to the `POLICIES` key I think this will cause problems - it looks like this will call replacePao with a null `newAttributes` in the request, which I think will just wipe all policies.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/ClonePolicyAttributesStep.java
```diff
@@ -0,0 +1,105 @@
...
+      createPaoIfNotExist(token, destinationWorkspaceId);
+    } catch (Exception ex) {
+      logger.info("Attempt to create a PAO for workspace failed", ex);
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, ex);
+    }
+
+    // Save the destination attributes so we can restore them if the flight fails
+    ApiTpsPaoGetResult destinationPao = tpsApiDispatch.getPao(token, destinationWorkspaceId);
+    ApiTpsPolicyInputs destinationAttributes = destinationPao.getAttributes();
+    context.getWorkingMap().put(WorkspaceFlightMapKeys.POLICIES, destinationAttributes);
```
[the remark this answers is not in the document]
zloery: Ah, I didn't know working map values got persisted on the flip from do -> undo. I wonder if we can clean up some other flights where we split this into two steps

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/ClonePolicyAttributesStep.java
zloery: whoops, missed the latest commit that already fixed this!

zloery (approved):

service/src/main/java/bio/terra/workspace/amalgam/tps/TpsApiDispatch.java
```diff
@@ -174,21 +152,127 @@ public static ApiTpsPaoGetResult paoToApi(Pao pao) {
...
+  public void createPao(BearerToken bearerToken, ApiTpsPaoCreateRequest request) {
+    features.tpsEnabledCheck();
+    paoService.createPao(
+        request.getObjectId(),
+        componentFromApi(request.getComponent()),
+        objectTypeFromApi(request.getObjectType()),
+        policyInputsFromApi(request.getAttributes()));
+  }
+
+  public void deletePao(BearerToken bearerToken, UUID objectId) {
```
mikenev, line 209: thanks

PR #835, opened 2022-09-22
[PF-1930] Implement cloning controlled bucket to referenced

melissachang:
Manually tested:

```
{
  "resources": [
    {
      "metadata": {
        "workspaceId": "53f383ba-c91c-4b86-935f-b52f0bd1688b",
        "resourceId": "46bd7f0a-f4cb-4118-83ef-0b5be07343b6",
        "name": "dest-bucket",
        "resourceType": "GCS_BUCKET",
        "stewardshipType": "REFERENCED",
        "cloudPlatform": "GCP",
        "cloningInstructions": "COPY_REFERENCE",
        "resourceLineage": [
          {
            "sourceWorkspaceId": "53f383ba-c91c-4b86-935f-b52f0bd1688b",
            "sourceResourceId": "a7895e05-61ba-40a1-b235-28d52d6ac435"
          }
        ],
        "properties": []
      },
      "resourceAttributes": {
        "gcpGcsBucket": {
          "bucketName": "source-bucket-terra-wsm-t-rapid-walnut-999"
        }
      }
    },
    {
      "metadata": {
        "workspaceId": "53f383ba-c91c-4b86-935f-b52f0bd1688b",
        "resourceId": "a7895e05-61ba-40a1-b235-28d52d6ac435",
        "name": "source-bucket",
        "resourceType": "GCS_BUCKET",
        "stewardshipType": "CONTROLLED",
        "cloudPlatform": "GCP",
        "cloningInstructions": "COPY_DEFINITION",
        "controlledResourceMetadata": {
          "accessScope": "SHARED_ACCESS",
          "managedBy": "USER",
          "privateResourceUser": {},
          "privateResourceState": "NOT_APPLICABLE"
        },
        "resourceLineage": [],
        "properties": []
      },
      "resourceAttributes": {
        "gcpGcsBucket": {
          "bucketName": "source-bucket-terra-wsm-t-rapid-walnut-999"
        }
      }
    }
  ]
}
```

Also in `WorkspaceCloneUtils`:

- Move calling `buildDestinationResourceLineage` into `buildDestinationResourceCommonFields`, to avoid boilerplate
- Renamed `buildDestinationGcsBucketReference` -> `buildDestinationReferencedGcsBucket`, to be consistent with `buildDestinationControlledGcsBucket`

FYI @cbookg

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtils.java
@@ -67,37 +67,37 @@ public class WorkspaceCloneUtils {
     // ReferenceResource doesn't have Builder, only leaf resources like ReferencedGcsBucketResource
     // have Builder. So each resource type must be built separately.
     return switch (sourceReferencedResource.getResourceType()) {
-      case REFERENCED_GCP_GCS_BUCKET -> buildDestinationGcsBucketReference(
+      case REFERENCED_GCP_GCS_BUCKET -> buildDestinationReferencedGcsBucket(
           sourceReferencedResource.castByEnum(WsmResourceType.REFERENCED_GCP_GCS_BUCKET),
           destinationWorkspaceId,
           destinationResourceId,
           name,
           description);
-      case REFERENCED_GCP_GCS_OBJECT -> buildDestinationGcsObjectReference(
+      case REFERENCED_GCP_GCS_OBJECT -> buildDestinationReferencedGcsObject(
           sourceReferencedResource.castByEnum(WsmResourceType.REFERENCED_GCP_GCS_OBJECT),
           destinationWorkspaceId,
           destinationResourceId,
           name,
           description);
-      case REFERENCED_ANY_DATA_REPO_SNAPSHOT -> buildDestinationDataRepoSnapshotReference(
+      case REFERENCED_ANY_DATA_REPO_SNAPSHOT -> buildDestinationReferencedDataRepoSnapshot(
           sourceReferencedResource.castByEnum(WsmResourceType.REFERENCED_ANY_DATA_REPO_SNAPSHOT),
           destinationWorkspaceId,
           destinationResourceId,
           name,
           description);
-      case REFERENCED_GCP_BIG_QUERY_DATASET -> buildDestinationBigQueryDatasetReference(
+      case REFERENCED_GCP_BIG_QUERY_DATASET -> buildDestinationReferencedBigQueryDataset(
           sourceReferencedResource.castByEnum(WsmResourceType.REFERENCED_GCP_BIG_QUERY_DATASET),
           destinationWorkspaceId,
           destinationResourceId,
           name,
           description);
-      case REFERENCED_GCP_BIG_QUERY_DATA_TABLE -> buildDestinationBigQueryDataTableReference(
+      case REFERENCED_GCP_BIG_QUERY_DATA_TABLE -> buildDestinationReferencedBigQueryDataTable(
           sourceReferencedResource.castByEnum(WsmResourceType.REFERENCED_GCP_BIG_QUERY_DATA_TABLE),
           destinationWorkspaceId,
           destinationResourceId,
           name,
           description);
-      case REFERENCED_ANY_GIT_REPO -> buildDestinationGitHubRepoReference(
+      case REFERENCED_ANY_GIT_REPO -> buildDestinationReferencedGitHubRepo(
           sourceReferencedResource.castByEnum(WsmResourceType.REFERENCED_ANY_GIT_REPO),
           destinationWorkspaceId,
           destinationResourceId,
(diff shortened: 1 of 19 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CopyGcsBucketDefinitionToReferencedStep.java
```diff
@@ -0,0 +1,134 @@
...
+        WorkspaceCloneUtils.buildDestinationReferencedGcsBucketFromControlled(
+            sourceBucket,
+            destinationWorkspaceId,
+            destinationResourceId,
+            resourceName,
+            description,
+            sourceBucket.getBucketName());
+
+    // Launch a CreateReferenceResourceFlight to make the destination bucket
+    final ReferencedGcsBucketResource createdBucket =
```
ddietterich, line 96: This was not the right approach when Jay did it and it is still not the right approach. You cannot easily use flights as subroutines.

In this case, a failure-rerun of this step will launch a new create flight that is likely to fail because the referenced resource will already exist.

I probably shouldn't have let him off the hook with an "I'll fix it later". Let's not compound the problem.

I think the better approach would be to make shared code that generates the steps for creating a referenced resource. Use that step generator in this flight and do the referenced resource creation in line. Also use it in the main createReferencedResource method.

That is, rather than trying to share the flight, share the code generation for the steps.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CopyGcsBucketDefinitionToControlledStep.java
```diff
@@ -39,15 +41,17 @@
  * CLONED_RESOURCE_DEFINITION is put into the working map for future steps. A
  * CLONE_DEFINITION_RESULT object is put into the working map, and if the cloning instructions are
  * COPY_DEFINITION, the response is set on the flight.
+ *
+ * <p>Keep in sync with CopyGcsBucketDefinitionToReferencedStep.
```
ddietterich, line 45: Always a red flag when you have to add a comment like that. Can we figure out a way to DRY this out?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtils.java
```diff
@@ -176,7 +155,11 @@ private static ControlledResourceFields getControlledResourceCommonFields(
         .description(description)
         .workspaceUuid(destinationWorkspaceId)
         .resourceId(destinationResourceId)
-        .resourceLineage(destinationResourceLineage)
+        .resourceLineage(
+            buildDestinationResourceLineage(
```
ddietterich, line 159: Nice tidying up!

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CopyGcsBucketDefinitionToReferencedStep.java
melissachang: Done.

Old: Flight had CopyGcsBucketDefinitionStep, which started a sub-flight

New: Flight has SetReferencedDestinationGcsBucketWorkingMapStep, CreateReferenceMetadataStep, SetReferencedDestinationGcsBucketResponseStep. No sub-flight

Added tests to ControlledResourceServiceTest:

- cloneGcsBucket_copyResource_do
- cloneGcsBucket_copyReference_do
- cloneGcsBucket_copyResource_undo
- cloneGcsBucket_copyReference_undo

For the do tests, I set above three steps to STEP_RESULT_FAILURE_RETRY.

(The undo tests were in CloneControlledGcsBucketResourceFlightTest; I moved them to ControlledResourceServiceTest and deleted CloneControlledGcsBucketResourceFlightTest.)

service/src/test/java/bio/terra/workspace/app/configuration/external/controller/ControlledGcpResourceApiControllerConnectedTest.java
```diff
@@ -121,124 +118,92 @@ public void cloneBucket_copyNothing() throws Exception {
...
     assertEquals(expectedBucketName, actualBucket.getAttributes().getBucketName());
+    assertEquals(
+        ControlledResourceFixtures.RESOURCE_DESCRIPTION,
+        actualBucket.getMetadata().getDescription());
+    assertEquals(ApiResourceType.GCS_BUCKET, actualBucket.getMetadata().getResourceType());
+    assertEquals(expectedStewardshipType, actualBucket.getMetadata().getStewardshipType());
+    assertEquals(ApiCloudPlatform.GCP, actualBucket.getMetadata().getCloudPlatform());
+    assertEquals(expectedCloningInstructions, actualBucket.getMetadata().getCloningInstructions());
+
+    ApiResourceLineage expectedResourceLineage = new ApiResourceLineage();
```
ddietterich, line 194: Not for this PR, but my random thought of the day. Right now we generate API objects in flights. This _feels_ wrong. Ideally, we would do all API conversion in the controllers. I think the reason we do this is to allow the JobService to have a generic response return path. I wonder if there is a structure we could implement that would remove the API generation from the flights.

ddietterich (approved):
LGTM! Thanks for the cleanups along the way.

melissachang:
@zloery do you mind reviewing, since @yuhuyoyo is OOO today (I think?)?

yuhuyoyo (approved):
i thought i sent out the review but I guess i didn't. sorry

PR #838, opened 2022-09-23
[PF-1980] Fix GrantPetUsagePermissionStep undo

zloery:
This undo step failed (causing dismal Stairway failures) if the `CreateControlledResourceFlight` creating an AI Notebook was triggered via pet SA credentials instead of end-user credentials, which frequently happened for users working from a notebook.

The root cause was that Sam's `getUserIds` endpoint throws a 404 if it's called with a pet SA email. This fix works by fetching the end-user email from Sam, similar to what we do in the `do` step. In the future (PF-1001) this would be a good spot for a `SamUser` object instead of an `AuthenticatedUserRequest` to save some extra calls to Sam.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/GrantPetUsagePermissionStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/GrantPetUsagePermissionStep.java
@@ -80,8 +80,8 @@ public class GrantPetUsagePermissionStep implements Step {
       return StepResult.getStepResultSuccess();
     }
     try {
-      // userRequest.email might be a pet SA, so we need to call Sam and determine the end-user's
-      // email directly.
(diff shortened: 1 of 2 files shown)


ddietterich (approved):
LGTM - thanks for figuring it out!

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/GrantPetUsagePermissionStep.java
```diff
@@ -80,8 +80,15 @@ public StepResult undoStep(FlightContext context) throws InterruptedException {
       return StepResult.getStepResultSuccess();
     }
     try {
+      // userRequest.email might be a pet SA, so we need to call Sam and determine the end-user's
```
melissachang, line 83: // userRequest.email might be a pet SA (eg user ran CLI in notebook instance)

melissachang (approved):
Nice

PR #839, opened 2022-09-23
PF-1965: delete folder in a flight.

yuhuyoyo:
When a folder is deleted, all the resources and sub-folders are deleted. Only the owner of the workspace can delete folders because only owner can delete all the resources (even other's private resources) in a workspace. 

Instead of launching a sub-flight, I have `FolderDeleteFlight` to share the steps with `DeleteControlledResourcesFlight` because stairway doesn't really have good support for  the sub-flight concept.

Added FolderServiceTest to test the do and undo of FolderDeleteFlight. Also added integration test.


service/src/main/java/bio/terra/workspace/app/controller/FolderApiController.java
```diff
@@ -107,8 +107,9 @@ public ResponseEntity<ApiFolderList> listFolders(UUID workspaceId) {
   @Override
   public ResponseEntity<Void> deleteFolder(UUID workspaceId, UUID folderId) {
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
-    workspaceService.validateWorkspaceAndAction(userRequest, workspaceId, SamWorkspaceAction.WRITE);
-    folderService.deleteFolder(workspaceId, folderId);
+    // Only
+    workspaceService.validateWorkspaceAndAction(userRequest, workspaceId, SamWorkspaceAction.OWN);
```
ddietterich, line 111: Is this restriction OK with Product/UX? Are WRITERS allowed to create folders? Should it be symmetric?

It would be possible to allow writers to delete, but it would require a recursive check of all resources within the folder to make sure that the operation was possible. An alternate semantic would be that the delete would succeed in deleting everything the user had access to.

nit: incomplete comment above

service/src/main/java/bio/terra/workspace/service/folder/FolderService.java
```diff
@@ -43,15 +65,29 @@ public ImmutableList<Folder> listFolders(UUID workspaceId) {
...
+            .addParameter(ControlledResourceKeys.RESOURCES_TO_DELETE, controlledResources)
+            .addParameter(ReferencedResourceKeys.RESOURCES_TO_DELETE, referencedResources)
+            .submitAndWait(Boolean.class);
     if (!deleted) {
-      throw new FolderNotFoundException(
-          String.format(
-              "Fail to delete folder %s which is not found in workspace %s",
-              folderId, workspaceUuid));
+      logger.warn(
+          String.format("Failed to delete folder %s in workspace %s", folderId, workspaceUuid));
```
ddietterich, line 89: I think you can use `{}` in the logger string for substitutions.

ddietterich (approved):
The code looks very good. I'll approve as is.

My concern is about requiring OWN to delete the folder. Have you/could you check with Product and UX about that constraint? We could do it differently. And that could be done as a separate PR, since it is incremental from this one.

service/src/main/java/bio/terra/workspace/app/controller/FolderApiController.java
melissachang: Here's what I think should happen (can be separate PR):

- Writer can delete folder
- If there's a private resource (not owned by requester), then deletion fails with something like `Cannot delete folder because XX does not have permission to delete private resource YY. Ask workspace owner or ZZ to delete YY first.`
- Otherwise, deletion succeeds.

melissachang (approved):

service/src/main/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHook.java
```diff
@@ -138,6 +145,15 @@ private void maybeLogCloudContextDeletionFlight(
     }
   }
 
+  private void maybeLogFolderDeletion(
```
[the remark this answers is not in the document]
yuhuyoyo: I will add log for other actions in another PR. This one got into the way and i need to implement it here because it's a flight.

integration/src/main/java/scripts/testscripts/FolderLifecycle.java
```diff
@@ -83,14 +153,28 @@ protected void doUserJourney(TestUserSpecification testUsers, WorkspaceApi works
...
     Folder retrievedFolder2 = folderApi.getFolder(getWorkspaceId(), folderBar.getId());
     assertEquals(updatedFolder2, retrievedFolder2);
 
+    // Delete folder bar.
     folderApi.deleteFolder(getWorkspaceId(), folderBar.getId());
 
-    var ex =
-        assertThrows(
-            ApiException.class, () -> folderApi.getFolder(getWorkspaceId(), folderBar.getId()));
+    // All the resources in bar and its sub-folder loo are deleted.
```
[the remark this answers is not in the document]
yuhuyoyo: no. right now it's in the same flight so  it will just have one entry. It can go into the delete details column later

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -560,16 +562,19 @@ private JobBuilder commonDeletionJobBuilder(
         .newJob()
         .description(jobDescription)
         .jobId(jobId)
-        .flightClass(DeleteControlledResourceFlight.class)
+        .flightClass(DeleteControlledResourcesFlight.class)
```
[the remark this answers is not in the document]
yuhuyoyo: haha i'm trying to do that but it causes some json serialization problem :(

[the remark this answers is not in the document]
melissachang: If you're not able to do this, can you add comments to DeleteControlledResourcesFlight? Eg

This flight can either be used to delete a single resource, or multiple resources.

Single resource deletion uses flight parameters X Y Z and happens in XX.

 Multiple resource deletion uses flight parameters A B C and happens in YY.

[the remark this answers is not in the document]
yuhuyoyo: ah. json cannot serialize a generic list. Changing to an arraylist works!

the resourceType, resourceName and stewardship needs to stay because UI is currently using it for filtering flight job (to track if a resource is deleted).

Jenn says that it's oK to not track the individual resource deletion progress for folder deletion but i don't want to remove the filtering yet for single resource deletion.

service/src/test/java/bio/terra/workspace/common/fixtures/WorkspaceFixtures.java
```diff
@@ -39,6 +43,35 @@ public static void createGcpCloudContextInDatabase(
...
   }
 
+  /**
+   * Creates a workspaces with a GCP cloud context and stores it in the database. Returns the
+   * workspace id.
+   *
+   * <p>The {@link ResourceDao#createControlledResource(ControlledResource)} checks that a relevant
+   * cloud context exists before storing the resource.
+   */
+  public static UUID createGcpWorkspace(WorkspaceDao workspaceDao) {
```
melissachang, line 53: Forgot to comment earlier

createGcpWorkspace -> createGcpWorkspaceWithContext

createWorkspace -> createWorkspaceWithoutContext

melissachang: Actually, there's WorkspaceConnectedTestUtils.createWorkspaceWithGcpContext(). WorkspaceConnectedTestUtils is a better place for these methods, than WorkspaceFixtures.

Use existing WorkspaceConnectedTestUtils.createWorkspaceWithGcpContext(), and add WorkspaceConnectedTestUtils.createWorkspaceWithoutGcpContext().

yuhuyoyo: i'm using this in the unit test. No real context is created by this method. 

i can rename though.

melissachang: Can you create WorkspaceUnitTestUtils and put these there? These Fixtures classes are normally for defining constants, not making calls.

yuhuyoyo:
close and reopen to retrigger test as dev sam is down

PR #842, opened 2022-09-25
[TOAZ-163] Add Sam checks to Landing Zone Service

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-163

Depends on https://github.com/DataBiosphere/terra-landing-zone-service/pull/22.

See [doc](https://docs.google.com/document/d/1pAkgwYueuUJoON0-KrWrf-eXL5-IgDAVFTiTyihf2EY/edit#) for overview of data model between Billing Profile, Landing Zone, and Workspace.

Keeping in draft state while I do some testing.


service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiController.java
```diff
@@ -37,16 +39,20 @@ public LandingZoneApiController(
   @Override
   public ResponseEntity<ApiAzureLandingZoneResult> createAzureLandingZone(
       @RequestBody ApiCreateAzureLandingZoneRequestBody body) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
```
ddietterich, line 42: Suggest using the new and improved `BearerToken` and `SamUser` classes that Doug added rather than `AuthenticatedUserRequest`. You can save calls to Sam by using those.

service/gradle/dependencies.gradle
```diff
@@ -38,7 +38,7 @@ dependencies {
   }
 
   //Terra Landing Zone Service
-  implementation ('bio.terra:terra-landing-zone-service:0.0.5-SNAPSHOT') {
+  implementation ('bio.terra:terra-landing-zone-service:0.0.8-rt-toaz-163-2-SNAPSHOT') {
```
rtitle, line 41: Will update once https://github.com/DataBiosphere/terra-landing-zone-service/pull/22 merges.

service/gradle/dependencies.gradle
```diff
@@ -38,7 +38,7 @@ dependencies {
   }
 
   //Terra Landing Zone Service
-  implementation ('bio.terra:terra-landing-zone-service:0.0.8-SNAPSHOT') {
+  implementation ('bio.terra:terra-landing-zone-service:0.0.8-rt-toaz-163-13-SNAPSHOT') {
```
rtitle, line 41: Will update once https://github.com/DataBiosphere/terra-landing-zone-service/pull/22 merges

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -76,23 +69,24 @@ public ApiAzureLandingZoneResult createAzureLandingZone(
             .version(body.getVersion())
             .parameters(
                 MapperUtils.LandingZoneMapper.landingZoneParametersFrom(body.getParameters()))
-            .landingZoneTarget(target)
+            .billingProfileId(body.getBillingProfileId())
             .build();
-    String jobId =
+    return toApiAzureLandingZoneResult(
```
rtitle, line 74: Kind of subtle, I ran into a concurrency issue here.

Before the code was doing this:
1. Call `LandingZoneService.startLandingZoneCreationJob`  --> returns a job UUID
2. Call `LandingZoneService getAsyncJobResult(jobId)` --> returns an AsyncJobResult

However now since both of these calls check Sam, step 2 would fail if the Sam resource hadn't been created yet (which happens asynchronously).

So I changed `LandingZoneService.startLandingZoneCreationJob` to just return an `AsyncJobResult` itself so we don't need the second call to fetch the async job result.

service/src/main/resources/application.yml
```diff
@@ -30,6 +30,7 @@ env:
   urls: # While we've traditionally thought of these as env specific and ok to hardcode, with kubernetes they may change
     sam: ${SAM_ADDRESS:https://sam.dsde-dev.broadinstitute.org/}
     terra-datarepo: ${TERRA_DATAREPO_URL:https://jade.datarepo-dev.broadinstitute.org}
+    bpm: ${BPM_ADDRESS:https://bpm.dsde-dev.broadinstitute.org/}
```
rtitle, line 33: cc @aherbst-broad 
WSM now knows the BPM URL, currently only used by the LZ amalgamation.
Helm support to add it per environment is here: https://github.com/broadinstitute/terra-helmfile/pull/3374/

rtitle:
Requesting some reviews, this is working well in a BEE.

openapi/src/parts/azure_landing_zone.yaml
```diff
@@ -118,8 +118,10 @@ components:
               Here is a list of some parameters - POSTGRES_SERVER_SKU, POSTGRESQL_SUBNET, VNET_ADDRESS_SPACE.
               These are example of assigned values - POSTGRES_SERVER_SKU=GP_Gen5_2, POSTGRESQL_SUBNET=10.1.0.16/29
             $ref: '#/components/schemas/AzureLandingZoneParameter'
-        landingZoneTarget:
-          $ref: '#/components/schemas/LandingZoneTarget'
+        billingProfileId:
```
rtitle, line 121: LZ creation now takes a `billingProfileId` instead of `landingZoneTarget`.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -202,15 +191,13 @@ private ApiAzureLandingZoneResult fetchCreateAzureLandingZoneResult(String jobId
    * The initial assumption is that the cardinality of 1:1 between the cloud context and the LZ
    * is enforced in the create operation, therefore more than one LZ per azure context is not allowed.
    */
-  public String getLandingZoneId(AzureCloudContext azureCloudContext) {
-    return getFirstLandingZoneId(MapperUtils.LandingZoneTargetMapper.from(azureCloudContext))
+  public UUID getLandingZoneId(AzureCloudContext azureCloudContext) {
```
rtitle, line 194: See above comment -- we're currently querying for landing zones in the WSM `CloudContext` and taking the first one. 

However we probably want to refactor this to the desired model (I'm especially worried about missing permission checks involved in Azure cloud context creation).

service/gradle/dependencies.gradle
```diff
@@ -38,7 +38,7 @@ dependencies {
   }
 
   //Terra Landing Zone Service
-  implementation ('bio.terra:terra-landing-zone-service:0.0.8-SNAPSHOT') {
+  implementation ('bio.terra:terra-landing-zone-service:0.0.9-rt-toaz-163-SNAPSHOT') {
```
rtitle, line 41: Will change this after https://github.com/DataBiosphere/terra-landing-zone-service/pull/22 merges

service/gradle.lockfile
```diff
@@ -1,12 +1,13 @@
 # This is a Gradle generated file for dependency locking.
 # Manual edits can break the build and are not advised.
 # This file is expected to be part of source control.
+bio.terra:billing-profile-manager-client:0.1.6-SNAPSHOT=productionRuntimeClasspath,runtimeClasspath,testRuntimeClasspath
```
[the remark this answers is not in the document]
rtitle: Cool, will update

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
ddietterich: Yes, this call is upside down. Please add a TODO with the Jira ticket number that tracks when this gets done properly.

openapi/src/parts/azure_landing_zone.yaml
rtitle: Yes, the LZ stores what billing profile it's in.

> Shouldn't there be an API to lookup landing zones from billing profile?

Yes, I think so. In fact the LZ service layer already has a method to lookup by billing profile with the right Sam checks. I can add a REST API too.

> Shouldn't there be an API description in the design doc?

Probably, I can write this down too

rtitle: I wrote down the intended APIs here: https://docs.google.com/document/d/1a0OeD-7hsUXYPMZ8JQ_rjMguHlwbpbUPL9TDkLygHp4/edit#heading=h.5eyy5cpubyu0

I am thinking I will ticket/PR separately adding the missing `list` APIs as it involves some more refactoring in WSM/LZ service.

rtitle: https://broadworkbench.atlassian.net/browse/TOAZ-222

aherbst-broad (approved):

PR #843, opened 2022-09-26
PF-1952: Integrate Delete PAO API into WSM and add test

mikenev:
TpsApi integration was already there, it just needed some minor updates for the deleted field and an integration test.


zloery (approved):

integration/src/main/resources/configs/integration/DeleteWorkspaceWithPolicy.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "DeleteWorkspaceWithPolicy",
```
zloery, line 2: As a heads up, this won't run as part of the PR test or nightly suites unless you add it to the FullIntegration suite (though that might be the right call until TPS is rolled out to environments > dev)

integration/src/main/resources/configs/integration/DeleteWorkspaceWithPolicy.json
```diff
@@ -0,0 +1,20 @@
+{
```
ddietterich, line 1: This test won't get run because it is not in any test suite. What that your intention?
It is tricky because we are not exposing our feature flags from WSM, so the test doesn't know if it should run or not.

integration/src/main/java/scripts/testscripts/DeleteWorkspaceWithPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+import scripts.utils.ClientTestUtils;
+import scripts.utils.WorkspaceAllocateWithPolicyTestScriptBase;
+
+public class DeleteWorkspaceWithPolicy extends WorkspaceAllocateWithPolicyTestScriptBase {
+  private static final Logger logger = LoggerFactory.getLogger(DeleteWorkspaceWithPolicy.class);
+
+  private static final String DATASET_RESOURCE_NAME = "wsmtest_dataset";
+
+  @Override
+  protected void doSetup(
```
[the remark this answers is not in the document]
mikenev: right, I'll take it out.

ddietterich (approved):
Question about running the test. But we can figure that out separately from this PR.

PR #844, opened 2022-09-27
PF-1965: check if the user request has authorization to delete resources

yuhuyoyo:
Follow-up of https://github.com/DataBiosphere/terra-workspace-manager/pull/839.

The authorization check is in the FolderService because I'm getting all the resources in the folder in the service. The rest of the authentication happens in the controller but I would need to move a lot of logic if i do that..


service/src/main/java/bio/terra/workspace/service/folder/FolderService.java
```diff
@@ -105,7 +114,8 @@ private void getResourcesInFolder(
       UUID workspaceId,
```
melissachang, line 114: Can you add javadoc

/** Populates parameters controlledResources, referencedResources with resources in specified folder. */

I was confused because `getResourceInFolder` sounds like it returns resources. Maybe rename to `collectResourcesInFolder`.

service/src/main/java/bio/terra/workspace/app/controller/FolderApiController.java
```diff
@@ -107,9 +107,9 @@ public ResponseEntity<ApiFolderList> listFolders(UUID workspaceId) {
   @Override
   public ResponseEntity<Void> deleteFolder(UUID workspaceId, UUID folderId) {
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
-    // Only owner should be able to perform the delete.
-    // TODO(PF-1965): Allow writer to delete too but have special handling for private resources.
-    workspaceService.validateWorkspaceAndAction(userRequest, workspaceId, SamWorkspaceAction.OWN);
+
+    workspaceService.validateWorkspaceAndAction(userRequest, workspaceId, SamWorkspaceAction.WRITE);
```
melissachang, line 111: What error does the user see?

I think it's worth it to provide a more helpful error message for this case. Otherwise, user will be confused. WDYT

- Here call FolderService.checkDeleteAction(folderId);
- FolderService.checkDeleteAction {
    if requester is owner, return
    // requester must be writer
    List<ControlledResource> controlledResourcesInFolder = ...
    if any controlled resource is private and requester isn't owner:
       throw new ForbiddenException("Unable to delete private resource %s in folder %s. Only workspace owners or resource owner can delete it.", resource.getName(), folder.getName());
 }

Also, I think it's nice to do access control in one place.

integration/src/main/java/scripts/testscripts/FolderLifecycle.java
```diff
@@ -154,27 +195,60 @@ protected void doUserJourney(TestUserSpecification testUsers, WorkspaceApi works
     assertEquals(newDescription, updatedFolder2.getDescription());
 
     // Get folder bar.
-    Folder retrievedFolder2 = folderApi.getFolder(getWorkspaceId(), folderBar.getId());
+    Folder retrievedFolder2 = folderOwnerApi.getFolder(getWorkspaceId(), folderBar.getId());
     assertEquals(updatedFolder2, retrievedFolder2);
 
-    // Delete folder bar.
-    folderApi.deleteFolder(getWorkspaceId(), folderBar.getId());
+    // Second user tries to delete folder Loo
```
melissachang, line 201: // Second user unable to delete folder Loo. Loo has a private notebook owned by first user. Second user is only workspace writer.

service/src/main/java/bio/terra/workspace/service/folder/FolderService.java
```diff
@@ -105,7 +114,8 @@ private void getResourcesInFolder(
...
       UUID folderId,
       List<WsmResource> controlledResources,
-      List<WsmResource> referencedResources) {
+      List<WsmResource> referencedResources,
+      AuthenticatedUserRequest userRequest) {
     if (folderDao.getFolderIfExists(workspaceId, folderId).isEmpty()) {
       throw new FolderNotFoundException(
           String.format("Folder %s is not found in workspace %s", folderId, workspaceId));
```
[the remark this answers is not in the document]
yuhuyoyo: it does contains the root folder ids too.

service/src/main/java/bio/terra/workspace/app/controller/FolderApiController.java
ddietterich: This check is just to see if the user has WRITE action at all on the workspace. It is not saying anything about the folder. Maybe the comment applies later in the PR?

integration/src/main/java/scripts/testscripts/FolderLifecycle.java
```diff
@@ -111,7 +126,7 @@ protected void doUserJourney(TestUserSpecification testUsers, WorkspaceApi works
     var displayNameLoo = TestUtils.appendRandomNumber("Loo");
     var descriptionLoo = String.format("This is a third-level folder %s", displayNameLoo);
     Folder folderLoo =
-        folderApi.createFolder(
+        folderOwnerApi.createFolder(
```
ddietterich, line 129: Do you need to cover cases where the folder writer creates a folder? (I think they can) or is that elsewhere?

service/src/main/java/bio/terra/workspace/service/folder/FolderService.java
```diff
@@ -125,6 +139,8 @@ private void getResourcesInFolder(
                 if (StewardshipType.REFERENCED == resource.getStewardshipType()) {
                   referencedResources.add(resource);
                 } else if (StewardshipType.CONTROLLED == resource.getStewardshipType()) {
+                  controlledResourceMetadataManager.validateControlledResourceAndAction(
```
ddietterich, line 142: I think this is where we will get a generic "Forbidden" error and we could make it more precise.
One thought: there could be more than one resource that the user cannot delete. You could collect a list of the names of the failures and pass that in the "details" of an exception. The details data is returned through the API in the `ErrorReport` object as a list of strings.

ddietterich (approved):
I agree with Melissa that a better error as you are checking the resource permission would be useful.

integration/src/main/java/scripts/testscripts/FolderLifecycle.java
yuhuyoyo: sure. added a case where wrtier create a folder and a case where writer update a folder.

service/src/main/java/bio/terra/workspace/app/controller/FolderApiController.java
yuhuyoyo: i will wrap the private resource error with something more actionable.

But i think it makes sense to throw early here if the person is just a reader. And the error you don't have write permission on this workspace makes sense for a reader.

Do you want to fetch the resources and perform the check in the controller to have the all the auth in controller or is it OK to do the private resource check in the service? For the first case, the FolderService would have
`deleteFolder(workspaceId, folderId, controlledResources, referencedResources, userRequest)`

melissachang: Sorry, I meant to keep the current check and add a call to some method in folderService.

```
// Validates requester has at least writer role on workspace
workspaceService.validateWorkspaceAndAction(userRequest, workspaceId, SamWorkspaceAction.WRITE);

// Validates requester has permission to delete every resource in folder
folderService.validateDeleteAction();
```

(And the error message should print resource names, that's more helpful, and resource name is unique anyways.)

melissachang (approved):

service/src/main/java/bio/terra/workspace/app/controller/FolderApiController.java
yuhuyoyo: folderservice.validateDeleteaction() would requires making multiple calls to folder and resource db to fetch all the resources. And folderService.deleteFolder would need to do that again, because i'm passing resources into the flight. so I don't want to do this action twice.

melissachang: Then just add a comment here. Normally access checking is done in the controller beginning, so it's nice to comment when that's not the case.

// If requester is writer and folder has private resources (not owned by requester), requester won't have permission to delete private resources. That access control check is done in folderService.XXX.

PR #845, opened 2022-09-28
[WOR-502] Allow specification of SAS token expiration.

cahrens:
Allows specification of the SAS token duration (not including the padding we add to the beginning of the time to account for clock drift) in seconds. Maximum allowable value is set in configuration, currently 24 hours per devsecops recommendation.

### DIFF
--- a/openapi/src/common/parameters.yaml
+++ b/openapi/src/common/parameters.yaml
@@ -10,7 +10,7 @@ components:
       required: true
       schema:
         $ref: '#/components/schemas/CloudPlatform'
-    
+
     JobId:
       name: jobId
       in: path
@@ -18,7 +18,7 @@ components:
       required: true
       schema:
         type: string
-    
+
     Limit:
       name: limit
       in: query
@@ -28,7 +28,7 @@ components:
         type: integer
         minimum: 1
         default: 10
-    
+
     MemberEmail:
       name: memberEmail
       in: path
@@ -57,7 +57,7 @@ components:
       required: true
       schema:
         $ref: '#/components/schemas/Name'
-  
+
     Offset:
       name: offset
       in: query
@@ -67,7 +67,7 @@ components:
         type: integer
         minimum: 0
         default: 0
-  
+
     ResourceId:
       name: resourceId
       in: path
@@ -76,7 +76,7 @@ components:
       schema:
         type: string
         format: uuid
-  
+
     ResourceType:
       name: resource
       in: query
@@ -84,7 +84,7 @@ components:
       required: false
       schema:
         $ref: '#/components/schemas/ResourceType'
-  
+
     Role:
       name: role
       in: path
@@ -101,6 +101,17 @@ components:
       schema:
         type: string
 
+    SasExpirationDuration:
+      name: sasExpirationDuration
+      in: query
+      description: |
+        The number of seconds until the SAS token should expire (optional). Default and maximum allowed values are
+        set in configuration.
+      required: false
+      schema:
+        type: integer
+        format: int64
+
     StewardshipType:
       name: stewardship
       in: query
@@ -108,7 +119,7 @@ components:
       required: false
       schema:
         $ref: '#/components/schemas/StewardshipType'
-  
+
     WorkspaceId:
       name: workspaceId
       in: path
--- a/scripts/write-config.sh
+++ b/scripts/write-config.sh
@@ -297,6 +297,7 @@ workspace:
     managed-app-tenant-id: ${tenantid}
     sas-token-start-time-minutes-offset: 15
     sas-token-expiry-time-minutes-offset: 60
+    sas-token-expiry-time-maximum-minutes-offset: 1440
 feature:
   tps-enabled: true
 EOF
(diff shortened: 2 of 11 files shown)


scripts/write-config.sh
```diff
@@ -297,6 +297,7 @@ workspace:
     managed-app-tenant-id: ${tenantid}
     sas-token-start-time-minutes-offset: 15
     sas-token-expiry-time-minutes-offset: 60
+    sas-token-expiry-time-maximum-minutes-offset: 1440
```
cahrens, line 300: Set to 24 hours per devsecops recommendation.

service/src/main/resources/application.yml
```diff
@@ -125,6 +125,7 @@ workspace:
   azure:
     sas-token-start-time-minutes-offset: 15
     sas-token-expiry-time-minutes-offset: 60
+    sas-token-expiry-time-maximum-minutes-offset: 1440
```
cahrens, line 128: Set to 24 hours per devsecops recommendation.

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -209,10 +209,15 @@ public ResponseEntity<ApiCreateControlledAzureRelayNamespaceResult> createAzureR
   @Override
   public ResponseEntity<ApiCreatedAzureStorageContainerSasToken>
       createAzureStorageContainerSasToken(
-          UUID workspaceUuid, UUID storageContainerUuid, String sasIPRange) {
+          UUID workspaceUuid,
+          UUID storageContainerUuid,
+          String sasIpRange,
```
cahrens, line 214: Renamed to be consistent with Swagger documentation (when I wrote the unit test I first tried specifying the URL parameter as `sasIPRange` and was confused why it didn't work).

openapi/src/common/parameters.yaml
```diff
@@ -101,14 +101,25 @@ components:
...
+    SasExpirationDuration:
+      name: sasExpirationDuration
+      in: query
+      description: |
+        The number of seconds until the SAS token should expire (optional). Default and maximum allowed values are
+        set in configuration.
+      required: false
+      schema:
+        type: integer
+        format: int64
```
cahrens, line 113: int64 for long

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -87,7 +87,7 @@ private BlobContainerSasPermission getSasTokenPermissions(
    * @param startTime Time at which the SAS will become functional
    * @param expiryTime Time at which the SAS will expire
    * @param userRequest The authenticated user's request
-   * @param sasIPRange (optional) IP address or range of IPs from which the Azure APIs will accept
+   * @param sasIpRange (optional) IP address or range of IPs from which the Azure APIs will accept
```
cahrens, line 90: More case consistency.

service/src/test/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiControllerTest.java
```diff
@@ -93,3 +106,201 @@ public void createAzureVmWithoutDisk() throws Exception {
         .andExpect(status().is(HttpStatus.SC_OK));
   }
 }
+
+class CreateAzureStorageContainerSasTokenTest extends BaseAzureUnitTest {
```
aherbst-broad, line 110: Thanks for adding this 👍

aherbst-broad (approved):

openapi/src/common/parameters.yaml
```diff
@@ -101,14 +101,25 @@ components:
...
         type: string
 
+    SasExpirationDuration:
+      name: sasExpirationDuration
+      in: query
+      description: |
+        The number of seconds until the SAS token should expire (optional). Default and maximum allowed values are
+        set in configuration.
+      required: false
+      schema:
```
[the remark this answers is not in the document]
cahrens: It does, and I started out with that. But then since I had to add validation anywhere to support the fact that the max is set in configuration, I decided to validate both things together. Do you have a strong feeling that I should handle them differently?

openapi/src/common/parameters.yaml
cahrens: I was conforming to the fact that the duration set in configuration is a Long (even though it is in minutes). I figured if we were concerned about that value potentially not fitting in a 32-bit int than we should certainly be concerned about seconds.

marctalbott (approved):

openapi/src/common/parameters.yaml
cahrens: @marctalbott do you have any insight into the choice of a Long for configuration values? If I recall correctly you wrote that code.

marctalbott: Hmm, can't say I remember unfortunately. The only thing I can think of is that the `OffsetDateTime` methods seem to expect longs over ints, but that doesn't actually restrict us from using ints here

PR #847, opened 2022-09-30
Toaz 105 retrofit storage container operations in wsm with lz integration

sergiygetlin:
This PR addresses new requirements for Azure storage container creation process. Now it is possible to reuse shared storage account from landing zone to create associated container.
Changes:
-Storage container REST API: storageAccountId is not required
-Update logic for Verify step within azure storage container creation flight
-Update logic for Delete step within azure storage container deletion flight


sergiygetlin:
Close due to merge conflicts

PR #848, opened 2022-09-30
PF-2060: Add missing permission that has broken notebook creation.

yuhuyoyo:
example log for the cloud console: 
```
Sep 29 18:22:46 testnotebook4 c2d-startup[431]: Registering with Proxy...
Sep 29 18:22:46 testnotebook4 c2d-startup[431]: Proxy URL from the config: https://us-central1.notebooks.cloud.google.com/tun/m/4592f092208ecc84946b8f8f8016274df1b36a14
Sep 29 18:23:08 testnotebook4 c2d-startup[431]: ERROR: (gcloud.compute.instances.describe) Could not fetch resource:
Sep 29 18:23:08 testnotebook4 c2d-startup[431]:  - Required 'compute.instances.get' permission for 'projects/terra-vdevel-clean-celery-1754/zones/us-central1-a/instances/testnotebook4'
Sep 29 18:23:08 testnotebook4 c2d-startup[431]: Proxy Registration failed. Unable to get Service Account from VM
```

Missing permission `compute.instances.get`. I tried adding it at the resource level but it didn't work so I added the permission at the project level.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
@@ -76,7 +76,6 @@ public class CloudSyncRoleMapping {
               "iam.serviceAccounts.list",
               "lifesciences.operations.cancel",
               "lifesciences.workflows.run",
-              "notebooks.instances.get",
               "notebooks.operations.cancel",
               "notebooks.operations.delete",
               "notebooks.operations.get",
--- a/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
@@ -76,6 +76,7 @@ public class CloudSyncRoleMapping {
               "iam.serviceAccounts.list",
               "lifesciences.operations.cancel",
               "lifesciences.workflows.run",
+              "notebooks.instances.get",
               "notebooks.operations.cancel",
               "notebooks.operations.delete",
               "notebooks.operations.get",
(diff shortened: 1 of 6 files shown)


ddietterich:
Good find! Did that get dropped or is it a change?

yuhuyoyo:
> Good find! Did that get dropped or is it a change?

it's a change. 

but I'm slightly confused. I tested it out in the verily devel env as an admin by adding this permission to the custom role and it works. But the nightly is still failing running against this branch so I still need to confirm if this is the only one missing.

yuhuyoyo:
oh nvm i tested it by adding to the project level.

cbookg:
> > Good find! Did that get dropped or is it a change?
> 
> it's a change.
> 
> but I'm slightly confused. I tested it out in the verily devel env as an admin by adding this permission to the custom role and it works. But the nightly is still failing running against this branch so I still need to confirm if this is the only one missing.

I don't understand how this isn't an API change by Cloud  that would break many customers.

yuhuyoyo:
> > > Good find! Did that get dropped or is it a change?
> > 
> > 
> > it's a change.
> > but I'm slightly confused. I tested it out in the verily devel env as an admin by adding this permission to the custom role and it works. But the nightly is still failing running against this branch so I still need to confirm if this is the only one missing.
> 
> I don't understand how this isn't an API change by Cloud that would break many customers.

normally if you just have a pre-defined gcp iam role assigned to, you won't get affected. so i guess not many people have custom role like us. There is a similar bug when i filed it though so I don't think we are the only one

zloery (approved):

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -71,6 +71,7 @@ public class CloudSyncRoleMapping {
               "artifactregistry.tags.update",
               "cloudbuild.builds.create",
               "cloudbuild.builds.update",
+              "compute.instances.get",
```
zloery, line 74: I think this just gives read access to metadata and not instance data which is fine, but we should test once this is merged

PR #850, opened 2022-10-03
PF-1977: add policy update ability to workspace update api

mikenev:
I didn't add a test here since there isn't a great way to determine whether one should be run and the environment has TPS enabled.
Probably should add one once TPS is split into its own service and the feature flag work is available.

### DIFF
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -171,6 +171,37 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/workspaces/v1/{workspaceId}/policies:
+    parameters:
+      - $ref: '#/components/parameters/WorkspaceId'
+    patch:
+      summary: Update policies for a workspace
+      operationId: updatePolicies
+      tags: [Workspace]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/TpsPaoUpdateRequest'
+      responses:
+        '200':
+          description: Update successfully processed. The update itself may or may not have succeeded.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TpsPaoUpdateResult'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '409':
+          $ref: '#/components/responses/Conflict'
+        '500':
+          $ref: '#/components/responses/ServerError'
+
   /api/workspaces/v1/{workspaceId}/roles:
     parameters:
       - $ref: '#/components/parameters/WorkspaceId'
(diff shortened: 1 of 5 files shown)


openapi/src/parts/workspace.yaml
```diff
@@ -479,6 +479,12 @@ components:
         description:
           description: A description of the workspace
           type: string
+        removeAttributes:
```
ddietterich, line 482: I am thinking we should make a separate endpoint for the policy update. If we stick with this approach, we need a way to communicate that the update mode only applies to the policy inputs. It seems weird that you could update the description, but DRY_RUN the policy change.

Also, how does the caller get the DRY_RUN results back? The UI will want to tell the user, you cannot change that because policy XX from object UUID says YY.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -287,8 +290,20 @@ public Workspace updateWorkspace(
       @Nullable String userFacingId,
       @Nullable String name,
       @Nullable String description,
+      @Nullable ApiTpsPolicyInputs addAttributes,
+      @Nullable ApiTpsPolicyInputs removeAttributes,
+      @Nullable ApiTpsUpdateMode updateMode,
       AuthenticatedUserRequest userRequest) {
     if (workspaceDao.updateWorkspace(workspaceUuid, userFacingId, name, description)) {
+      if (addAttributes != null || removeAttributes != null) {
```
ddietterich, line 298: This ends up being two transactions, so it is possible for the update part to succeed, the TPS part to fail. The response is failure, but the workspace part of the update still happened.

I think the choice is:
1. Make a separate endpoint and you do not need to use flights
2. Use the combined endpoint, but make this a two step flight.

ddietterich (changes requested):
A few problems...

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
mikenev: Yeah, that gets to the question I had of should this be a flight or not. I think the best approach is to separate it out, then it really will mirror the tps update endpoint...

ddietterich: sgtm

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -346,6 +349,28 @@ public ResponseEntity<ApiWorkspaceDescription> updateWorkspace(
...
+  @Override
+  public ResponseEntity<ApiTpsPaoUpdateResult> updatePolicies(
+      @PathVariable("workspaceId") UUID workspaceId, @RequestBody ApiTpsPaoUpdateRequest body) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    logger.info("Updating workspace policies {} for {}", workspaceId, userRequest.getEmail());
+
+    workspaceService.validateWorkspaceAndAction(
+        userRequest, workspaceId, SamConstants.SamWorkspaceAction.WRITE);
+
+    if (!featureConfiguration.isTpsEnabled()) {
```
ddietterich, line 361: You can use `tpsEnableCheck()` for this, but your error message is better.

ddietterich (approved):
LGTM - thanks!

zloery (approved):

PR #851, opened 2022-10-04
[WOR-515] Add BPM check when creating an azure workspace

aherbst-broad:
## Why
BPM is now live in the Terra production environment. We should check that a user has the appropriate action against a workspace's billing profile at creation time.

Note that this does **not** move the billing profile ID as an argument to the cloud context creation as has been discussed in various channels. That will be follow-on work as captured [in this ticket](https://broadworkbench.atlassian.net/browse/WOR-563).

### Note about nomenclature
`Billing Profile` is the term for entities that contain billing information the Terra ecosystem. WSM refers to these as `Spend Profile` (as does TDR). `Billing profile` is used only in the context of interactions with the BPM service. We may want to do a global find and replace of spend profile -> billing profile, but I felt that would explode the size of this diff. 

## This PR
* Adds a `bpmEnabled` flag that gates off the new behavior. When enabled, WSM will:
     * Check that the user has `LINK` against the needed `spendProfile` resource as stored on the Workspace.
     * Pulls in the managed resource group ID, tenant ID and subscription ID for the cloud context from the billing profile data (as relayed from BPM) rather than from the cloud context call. I have marked those arguments as deprecated on the `AzureCloudContext` object but they are being retained for backwards compatibility when the flag is off. 
 * Shifts the `SpendProfileServiceTest` to be a unit rather than a connected test.


service/src/main/java/bio/terra/workspace/app/configuration/external/FeatureConfiguration.java
```diff
@@ -41,6 +42,14 @@ public void setTpsEnabled(boolean tpsEnabled) {
     this.tpsEnabled = tpsEnabled;
   }
 
+  public boolean isBpmEnabled() {
```
aherbst-broad, line 45: I have flagged out the use of BPM so we are backwards compatible in environments where BPM is not deployed.

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/azure/CreateDbAzureCloudContextFinishStep.java
```diff
@@ -15,20 +20,34 @@
...
   @Override
   public StepResult doStep(FlightContext flightContext) throws InterruptedException {
-    AzureCloudContext azureCloudContext =
-        flightContext
-            .getInputParameters()
-            .get(JobMapKeys.REQUEST.getKeyName(), AzureCloudContext.class);
-
+    AzureCloudContext azureCloudContext;
+    if (featureConfiguration.isBpmEnabled()) {
+      azureCloudContext =
```
aherbst-broad, line 38: Wiring in the values from BPM as the cloud context values for azure managed resource groups. I've marked the azure cloud context tenant Id, sub ID and resource group ID as deprecated on the cloud context creation call due to this change.

service/src/test/java/bio/terra/workspace/service/spendprofile/SpendProfileServiceTest.java
```diff
@@ -2,61 +2,186 @@
...
 import org.junit.jupiter.api.Test;
-import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
 
-class SpendProfileServiceTest extends BaseConnectedTest {
-  @Autowired SamService samService;
-  @Autowired SpendProfileConfiguration spendProfileConfiguration;
-  @Autowired SpendConnectedTestUtils spendUtils;
-  @Autowired UserAccessUtils userAccessUtils;
+class SpendProfileServiceTest extends BaseUnitTest {
```
aherbst-broad, line 28: I changed this over to a unit test (vs. connected) to ease testing. That is, this test was making a live call to Sam and expecting spend profile resources to be present. Updating this to mock out a sam response seemed like the better way to ease development.

service/src/test/java/bio/terra/workspace/service/spendprofile/SpendProfileServiceTest.java
```diff
@@ -2,61 +2,186 @@
...
+  AuthenticatedUserRequest userRequest =
+      new AuthenticatedUserRequest(
+          "example@example.com", "fake-subject-id", Optional.of("fake-token"));
 
   @Test
-  void authorizeLinkingSuccess() {
-    SpendProfileId id = spendUtils.defaultSpendId();
+  void authorizeLinkingSuccess() throws InterruptedException {
+    var bpmClientProvider = mock(BpmClientProvider.class);
```
aherbst-broad, line 35: I am unsure if this should be in one of the `BaseMockXYZ` classes. Given its use is limited to this service and associated test my instinct is "no", but would love to hear other opinions.

service/src/main/java/bio/terra/workspace/service/spendprofile/client/BpmClientProvider.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.workspace.service.spendprofile.client;
+
+import bio.terra.profile.api.ProfileApi;
+import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
+
+/** Knows how to instantiate Billing Profile Manager clients */
+public interface BpmClientProvider {
```
rtitle, line 7: Nice pattern -- is this so the client can be more easily mocked for tests?

Wonder if other clients are following this approach..

service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfileService.java
```diff
@@ -79,4 +107,33 @@ private static List<SpendProfile> parse(
                     .build())
         .collect(Collectors.toList());
   }
+
+  private SpendProfile getSpendProfileFromBpm(
```
rtitle, line 111: Wonder if it's worth adding `@Traced` for the external service call (ex: https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/iam/SamService.java#L348)

service/src/main/java/bio/terra/workspace/service/workspace/flight/CheckSpendProfileStep.java
```diff
@@ -46,10 +53,23 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
+    if (cloudPlatform == CloudPlatform.GCP) {
+      if (spendProfile.billingAccountId().isEmpty()) {
+        throw NoBillingAccountException.forSpendProfile(spendProfileId);
+      }
+      workingMap.put(BILLING_ACCOUNT_ID, spendProfile.billingAccountId());
+    } else if (cloudPlatform == CloudPlatform.AZURE) {
+      if (spendProfile.managedResourceGroupId().isEmpty()
+          || spendProfile.subscriptionId().isEmpty()
+          || spendProfile.tenantId().isEmpty()) {
+        throw NoBillingAccountException.forSpendProfile(spendProfileId);
```
rtitle, line 66: Maybe want a different exception type for the azure case?

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/azure/ValidateMRGStep.java
```diff
@@ -26,10 +29,33 @@ public ValidateMRGStep(CrlService crlService, AzureConfiguration azureConfig) {
...
   @Override
   public StepResult doStep(FlightContext flightContext) throws InterruptedException {
-    AzureCloudContext azureCloudContext =
-        flightContext
-            .getInputParameters()
-            .get(JobMapKeys.REQUEST.getKeyName(), AzureCloudContext.class);
+    FlightBeanBag appContext = FlightBeanBag.getFromObject(flightContext.getApplicationContext());
+
+    AzureCloudContext azureCloudContext;
+    if (appContext.getFeatureConfiguration().isBpmEnabled()) {
```
rtitle, line 35: Do we need this step at all if BPM is enabled? The original intent was to validate the input parameters were a valid MRG. Can we just assume it's valid if it came from BPM?

rtitle (commented):
A few comments, overall looks good.

If I understand correctly the semantics to obtain a billing profile are:
```
if GCP:
  if bpmEnabled call BPM
  else lookup in-memory map
else if Azure:
  if bpmEnabled call BPM
  else use createCloudContext input parameters
```

So the current behavior doesn't change but we start adding support for BPM in both clouds.

As I'm typing this I'm wondering: do we need to ability to enable BPM for Azure but not GCP? Right now if `bpmEnabled == true` it would turn it on for both clouds, right?

service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfileService.java
```diff
@@ -55,7 +76,14 @@ public SpendProfile authorizeLinking(
         "isAuthorized")) {
       throw SpendUnauthorizedException.linkUnauthorized(spendProfileId);
     }
-    SpendProfile spend = spendProfiles.get(spendProfileId);
+
+    SpendProfile spend;
+    if (bpmEnabled) {
```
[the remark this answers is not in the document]
aherbst-broad: 🤔 Good point. `LINK` is implied by a billing profile coming back from the service.

service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfileService.java
aherbst-broad: 👍  and likely on the calling `authorizeLinking` method as well, will add.

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/azure/ValidateMRGStep.java
aherbst-broad: We may want to retain this to ensure that the WSM app registration has access to the MRG. I'm 51/49 in favor of keeping.

rtitle: That's true, technically WSM has a different service principal than BPM. Fine with me to keep, doesn't harm anything.

service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfileService.java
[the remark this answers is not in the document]
aherbst-broad: That being said, there is no explicit check for `LINK` in BPM, it's implied since all owners and users have `LINK`.  If the SAM model changes that could open a hole; i.e., I'm a billing administrator and want to view the billing profiles but can not spend against them.

I think I agree that we want BPM to be making these checks rather than leaking them to external services, perhaps we can have BPM make an explicit LINK check and if the need arises for readonly profiles we can fashion a new endpoint.

aherbst-broad:
@rtitle Yes, flipping the flag enables it for both clouds. I _think_ we only use GCP functionality in CWB in prod for snapshot workspaces, which have no cloud contexts and therefore undergo no spend profile checks. 

That being said, this may be a better safe than sorry situation -- perhaps a cloud specific flag for each would make sense, I'll look into it.

rtitle (approved):
LGTM after latest changes

service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfile.java
```diff
@@ -18,6 +19,12 @@ public abstract class SpendProfile {
   /** The id of the Google Billing Account associated with the SpendProfile, if there is one. */
   public abstract Optional<String> billingAccountId();
 
+  public abstract Optional<UUID> tenantId();
```
ddietterich, line 22: Should we have something in this object that says what cloud it is for? That would at least allow verification in other code. Alternately, perhaps we should have a base class and per-cloud subclasses?

If we refactor, I would love to get rid of AutoValue and use a record instead.

service/src/main/java/bio/terra/workspace/service/spendprofile/client/BpmClientProvider.java
ddietterich: One of the goals of the _____Service modules is to wrap the interaction with the external service, so this seems like an unnecessary layer.

service/src/main/java/bio/terra/workspace/service/workspace/flight/CheckSpendProfileStep.java
```diff
@@ -1,5 +1,8 @@
 package bio.terra.workspace.service.workspace.flight;
 
+import static bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.AZURE_BILLING_MANAGED_RESOURCE_GROUP_ID;
```
ddietterich, line 3: I think we should pass the whole SpendProfile object instead of each part.
That might require changing SpendProfile.java to a record so serdes will work.

service/src/main/java/bio/terra/workspace/service/workspace/flight/CheckSpendProfileStep.java
```diff
@@ -45,11 +56,25 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
             .getSpendProfileId()
             .orElseThrow(() -> MissingSpendProfileException.forWorkspace(workspaceUuid));
 
-    SpendProfile spendProfile = spendProfileService.authorizeLinking(spendProfileId, userRequest);
-    if (spendProfile.billingAccountId().isEmpty()) {
-      throw NoBillingAccountException.forSpendProfile(spendProfileId);
+    SpendProfile spendProfile =
+        spendProfileService.authorizeLinking(spendProfileId, bpmEnabled, userRequest);
+
+    if (cloudPlatform == CloudPlatform.GCP) {
```
ddietterich, line 62: Seems to me these checks should be done when we are minting the spend profile and not here in the step.
This is likely not the only place we will call `authorizeLinking`

service/src/main/java/bio/terra/workspace/service/workspace/flight/CheckSpendProfileStep.java
```diff
@@ -45,11 +56,25 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
             .getSpendProfileId()
             .orElseThrow(() -> MissingSpendProfileException.forWorkspace(workspaceUuid));
 
-    SpendProfile spendProfile = spendProfileService.authorizeLinking(spendProfileId, userRequest);
-    if (spendProfile.billingAccountId().isEmpty()) {
-      throw NoBillingAccountException.forSpendProfile(spendProfileId);
+    SpendProfile spendProfile =
```
ddietterich, line 59: I think there are errors from `authorizeLinking` that should be retried. For example, network or service unavailable kind of errors that show up during service upgrade. A single `BillingProfileManagerServiceApiException` isn't sufficient to distinguish the retry-able cases from the fatal cases.

Suggest distinguishing between those cases in the service and adding retry to the uses of this step.

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/azure/CreateDbAzureCloudContextFinishStep.java
ddietterich: You can see that if we serialized the spend profile, we could evolve it without changing this flight.

service/src/test/java/bio/terra/workspace/service/spendprofile/SpendProfileServiceTest.java
ddietterich: No one will ever autowire BPM, so it doesn't belong in the BaseUnitTestMocks.

Currently, no unit tests touch SpendProfiileService; probably because they mock the SamService authorize call. That will have to change and we will need to add SpendProfileService to the BaseUnitTestMocks.

ddietterich (changes requested):
I think SpendProfile class needs changes and should be passed directly in the working map. It should also contain its own validation, rather than have it in the step code.

I'm not sure it is worth the "provider" stuff for the purposes of a single unit test.
I would like to have a connected or integration test that talks to BPM.

aherbst-broad:
@ddietterich fair enough, I'll look into a connected test and the other changes.

marctalbott (approved):

service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfileService.java
```diff
@@ -79,4 +115,45 @@ private static List<SpendProfile> parse(
...
+      spend =
+          SpendProfile.builder()
+              .id(spendProfileId)
+              .billingAccountId(profile.getBillingAccountId())
+              .managedResourceGroupId(profile.getManagedResourceGroupId())
+              .tenantId(profile.getTenantId())
+              .subscriptionId(profile.getSubscriptionId())
+              .build();
+    } catch (ApiException ex) {
+      if (ex.getCode() == HttpStatus.NOT_FOUND.value()
```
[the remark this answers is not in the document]
aherbst-broad: Sensible, done.

service/src/test/resources/application-connected-test.yml
```diff
@@ -6,6 +6,7 @@ workspace:
     user-delegated-service-account-path: ../config/user-delegated-sa.json
     default-user-email: elijah.thunderlord@test.firecloud.org
     second-user-email: liam.dragonmaw@test.firecloud.org
+    billing-user-email: hermione.owner@test.firecloud.org
```
aherbst-broad, line 9: I'm looking into introducing a dedicated billing user for this purpose.

service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfile.java
```diff
@@ -1,34 +1,17 @@
...
-  @AutoValue.Builder
-  public abstract static class Builder {
-    public abstract Builder id(SpendProfileId id);
-
-    public abstract Builder billingAccountId(@Nullable String billingAccountId);
-
-    public abstract SpendProfile build();
-  }
-}
+public record SpendProfile(
```
aherbst-broad, line 12: This (de)serializes ok in my testing and in CI.

service/src/test/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlightV2Test.java
```diff
@@ -143,29 +142,6 @@ void successCreatesProjectAndContext() throws Exception {
     assertPolicyGroupsSynced(workspaceUuid, project);
   }
 
-  @Test
-  @DisabledIfEnvironmentVariable(named = "TEST_ENV", matches = BUFFER_SERVICE_DISABLED_ENVS_REG_EX)
-  void createsProjectAndContext_noBillingAccount_flightFailsAndGcpProjectNotCreated()
```
aherbst-broad, line 148: I do not believe this test asserts a realistic use case anymore. Given the validation code in the compact constructor for SpendProfile, it should not be possible to instantiate a GCP spend profile that has no billing account ID:
* If there is a bad profile in the config such as this, the system will not boot while wiring up the config fixture data.
* If there is a bad profile coming back from BPM, we will fail to instantiate the SpendProfile model and raise an exception

aherbst-broad:
@ddietterich:
* I've converted the spend profile to a record type and refactored the validation to take place in its compact constructor [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/arh-wor-515-azure-bpm/service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfile.java). The spend profile is now serialized and put in the working map at the `SPEND_PROFILE` key and captures a cloud platform as a member field. 
* I've added a connected test [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/arh-wor-515-azure-bpm/service/src/test/java/bio/terra/workspace/service/spendprofile/SpendProfileBpmConnectedTest.java). 
* I've updated the error handling to return a FATAL step result in the event a non-5xx exception is received from BPM, otherwise it retries.

service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfile.java
```diff
@@ -1,34 +1,43 @@
...
  * also have Terra IAM to manage who is allowed to modify and link them.
  */
-@AutoValue
-public abstract class SpendProfile {
-  /** The unique identifier of the SpendProfile. */
-  public abstract SpendProfileId id();
+public record SpendProfile(
+    SpendProfileId id,
+    CloudPlatform cloudPlatform,
+    Optional<String> billingAccountId,
```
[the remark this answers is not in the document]
aherbst-broad: Yeah, good call, fixed.

service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfileService.java
```diff
@@ -68,15 +109,105 @@ public SpendProfile authorizeLinking(
...
+              "Invalid cloud platform for billing profile id %s: %s ",
+              profile.getId().toString(), profile.getCloudPlatform().getValue()));
+    }
+  }
+
+  /**
+   * Creates a spend profile via the billing profile manager service, intended for usage in tests
+   * only.
+   */
+  public SpendProfile createGcpSpendProfile(
```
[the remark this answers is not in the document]
aherbst-broad: I've marked these as `VisibleForTesting`. Pulling these into a fixture would require exposing the `getProfileApi` method which feels counterproductive.

aherbst-broad:
@ddietterich `Optional`s have been removed from the `SpendProfile` fields 👍

ddietterich (approved):
LGTM - thanks!

PR #854, opened 2022-10-06
[PF-1955] Add setting TERRA_USER_EMAIL to .bash_profile (notebook VMs).

mbookman:


yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -165,6 +165,10 @@ echo "export PET_SA_EMAIL='${PET_SA_EMAIL}'" >> "/home/${JUPYTER_USER}/.bash_pro
 # These are equivalent environment variables which are set for a
```
melissachang, line 165:
```
# These are equivalent environment variables which are set for a
# command when calling "terra app execute <command>". Keep in sync with "terra app execute <command"
# environment variables. See https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/app/CommandRunner.java#L81
```

melissachang (approved):

PR #861, opened 2022-10-13
[PF-2096] Enable clone of Rawls workspaces

ddietterich:
This change set allows Rawls stage workspaces to be cloned.

- The clone method now accepts a destination workspace id. That id will be used to create the workspace, in the same way the regular create workspace endpoint works.
- The restriction against cloning Rawls workspaces is removed.
- We take the workspace stage from the source and use that as the destination stage.
- We only create a GCP cloud context in the destination if there was a context in the source workspace
- I did not wire in policy service interaction for Rawls workspaces

The hardest part of the change was writing the test :)

----
Based on Melissa's comments I did some test cleanup:
1. Make a common create workspace path for MC and RAWLS
2. Make a common clone workspace method for MC and RAWLS
3. Do not always run a `getWorkspace` on every create. Moved gets to the specific userFacingId tests that needed them.
4. Moved the `destinationWorkspaceId` field as suggested. Noticed and fixed the repetitive specification of `userFacingId`

### DIFF
--- a/service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerTest.java
+++ b/service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerTest.java
@@ -8,6 +8,8 @@ import static bio.terra.workspace.common.utils.MockMvcUtils.WORKSPACES_V1_BY_UUI
 import static bio.terra.workspace.common.utils.MockMvcUtils.WORKSPACES_V1_PATH;
 import static bio.terra.workspace.common.utils.MockMvcUtils.addAuth;
 import static bio.terra.workspace.common.utils.MockMvcUtils.addJsonContentType;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
@@ -21,11 +23,16 @@ import static org.springframework.test.web.servlet.request.MockMvcRequestBuilder
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 import bio.terra.workspace.common.BaseUnitTest;
+import bio.terra.workspace.common.StairwayTestUtils;
 import bio.terra.workspace.common.fixtures.WorkspaceFixtures;
 import bio.terra.workspace.common.utils.MockMvcUtils;
 import bio.terra.workspace.generated.model.ApiCloneWorkspaceRequest;
 import bio.terra.workspace.generated.model.ApiCloneWorkspaceResult;
+import bio.terra.workspace.generated.model.ApiCreateWorkspaceRequestBody;
+import bio.terra.workspace.generated.model.ApiDataRepoSnapshotResource;
 import bio.terra.workspace.generated.model.ApiErrorReport;
+import bio.terra.workspace.generated.model.ApiJobReport;
+import bio.terra.workspace.generated.model.ApiResourceCloneDetails;
 import bio.terra.workspace.generated.model.ApiTpsComponent;
 import bio.terra.workspace.generated.model.ApiTpsObjectType;
 import bio.terra.workspace.generated.model.ApiTpsPaoGetResult;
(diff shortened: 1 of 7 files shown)


calypsomatic (approved):
Thank you so much!

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerTest.java
```diff
@@ -282,6 +303,76 @@ public void cloneWorkspace() throws Exception {
         sourceWorkspace.getDisplayName() + " (Copy)", destinationWorkspace.getDisplayName());
   }
 
+  @Test
+  public void cloneRawlsWorkspace() throws Exception {
```
zloery, line 307: Now that this test class is working with snapshots, can it use `BaseUnitTestMockDataRepoService` instead of `BaseUnitTest`? As-is, I think this would call real TDR if someone called `checkReferenceAccess` on these snapshots

zloery (approved):

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerTest.java
ddietterich: Good point. That would be a surprise if someone added such a test.

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerTest.java
```diff
@@ -282,6 +303,76 @@ public void cloneWorkspace() throws Exception {
...
+    UUID sourceWorkspaceId = UUID.randomUUID();
```
melissachang, line 308: I think it makes tests easier to read if you use helper methods. Also, future tests can reuse those helper methods.

```
  @Test
  public void cloneRawlsWorkspace() throws Exception {
    // Create source RAWLS workspace
        mockMvcUtils.createWorkspace(RAWLS);

    // Create some data repo references
    ApiDataRepoSnapshotResource snap1 =
        mockMvcUtils.createDataRepoSnapshotReference(USER_REQUEST, sourceWorkspaceId);
    ApiDataRepoSnapshotResource snap2 =
        mockMvcUtils.createDataRepoSnapshotReference(USER_REQUEST, sourceWorkspaceId);

    // Clone the rawls workspace into a destination rawls workspace
    // This relies on the mocked Sam check
    UUID destinationWorkspaceId = UUID.randomUUID();
    ApiCloneWorkspaceResult clonedWorkspace = mockMvcUtils.cloneWorkspace();

    List<ApiResourceCloneDetails> cloneDetails = cloneWorkspace.getWorkspace().getResources();
    assertEquals(2, cloneDetails.size());
    List<UUID> cloneIdList =
        cloneDetails.stream().map(ApiResourceCloneDetails::getSourceResourceId).toList();
    assertThat(cloneIdList, hasItem(snap1.getMetadata().getResourceId()));
    assertThat(cloneIdList, hasItem(snap2.getMetadata().getResourceId()));
  }
```

mockMvcUtils already has create createWorkspaceWithCloudContext.

I see you added mockMvcUtils.getCloneWorkspaceResult(). It would remove more boilerplate if the helper include the entire clone call -- everything from initial request to the stuff in getCloneWorkspaceResult().

melissachang (approved):

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerTest.java
melissachang: I'm going to send a PR to clean up all tests in this file. If you want, I can do this in my PR.

ddietterich: I'm in the process of cleaning up the mock as you suggested. I'll send it past you for a check before merging.

service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
```diff
@@ -188,12 +244,7 @@ public ApiWorkspaceDescription createWorkspaceWithoutCloudContext(
             .andReturn()
             .getResponse()
             .getContentAsString();
-
-    // Return ApiWorkspaceDescription instead of ApiCreatedWorkspace, since former has
```
melissachang, line 192: Hmm I added this because a test needed userFacingId, which wasn't in ApiCreatedWorkspace. If this is no longer needed, fine to remove -- I can always add it back later if a test needs userFacingId.

melissachang (approved):
Thanks!

service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
ddietterich: What I found was that the majority of tests did not need the description. Only the tests of userFacingId.
So I added the explicit `getWorkspace` into those tests. There are 4 and they are all very similar, so there may be a benefit in factoring part of the test out rather than putting the `getWorkspace` into the common `createWorkspace`.

melissachang: Ah, my thinking was that `getWorkspace` is fast so it was ok for every `createWorkspace`, but totally fine with this approach as well.

PR #863, opened 2022-10-13
[PF-2091] sync role permisson on existing project

yuhuyoyo:
Introduce an adminApi in WSM.

## Authorization check
It check SAM by calling sam adminApi#getUserStatus to see it the userrequest has admin access. If it does, fetch all the cloudcontext in the database and  update the custom role permission. 

## logging
If the workspace has a gcp project and the project's custom role permission requires changes, we log an activity entry for that workspace. If multiple roles in the same project requires changes, we only create one entry for that workspace. 

## testing
Because the current test users are not admin, I have the integration test in the AdminService level.


service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -114,6 +115,23 @@ public AzureApi samAzureApi(String accessToken) {
     return new AzureApi(getApiClient(accessToken));
   }
 
+  public AdminApi samAdminApi(String accessToken) {
+    return new AdminApi(getApiClient(accessToken));
+  }
+
+  @Traced
+  public boolean isAdmin(AuthenticatedUserRequest userRequest) throws InterruptedException {
+    try {
```
yuhuyoyo, line 124: @zloery it's unclear to me how to check the verilygroup access. the api is not exposed. https://g3doc.corp.google.com/company/teams/verily/platform_eng/verilygroups/guides/api_guide.md?cl=head

i'd like your early thought on just querying the sam admin api to check for admin access..

zloery: I actually didn't realize they exposed that much of the API - the RPC endpoints aren't exposed but it looks like the HTTP ones are, and `GetMember` seems like the method we're looking for. That would be a pretty clean solution within Verily (especially because it would let us validate breakglass access). It does seem like it wouldn't mesh well with the Broad's admin approach, though.

Does Sam have an endpoint for checking admin access? Generally I think that's granted via membership in the `fc-admins` group but I'm not sure if that's a Sam group or just a googlegroup. If there's something like that available that SGTM

yuhuyoyo: > I actually didn't realize they exposed that much of the API - the RPC endpoints aren't exposed but it looks like the HTTP ones are, and `GetMember` seems like the method we're looking for. That would be a pretty clean solution within Verily (especially because it would let us validate breakglass access). It does seem like it wouldn't mesh well with the Broad's admin approach, though.

where do you see the http endpoint? yeah we can flag guard it so we need breakglass access for verily env and use sam for broad.
> 
> Does Sam have an endpoint for checking admin access? Generally I think that's granted via membership in the `fc-admins` group but I'm not sure if that's a Sam group or just a googlegroup. If there's something like that available that SGTM

i didn't see a sam endpoint to directly verify if a user is admin. so I'm just calling a sam AdminApi and see if I have permissions...https://github.com/broadinstitute/sam/blob/03ffedde48a9fd810cc1086429f60675064310b4/src/test/scala/org/broadinstitute/dsde/workbench/sam/api/AdminUserRoutesSpec.scala#L32

zloery: That page links to their published API (though all their line number links are broken 😬). In this case, the endpoint is `GET /v1/groups/{group_id}/members/{member_id}`. They also have some example code for dealing with IAP, though I haven't done that before.

I would double check with the Identiteam before building something like that or if they have a recommended way to check admin status. It seems like it should work, but it's definitely an unexpected way to use that Sam endpoint

yuhuyoyo: confirmed with Broad (dvoet) that this is an OK way of checking admin. calling this api check for terra level admin (sam has some different admin types but this is the one we needed)

i can pass through userRequest through iamCow in the flight that effectively check for breakglass. But that makes testing impossible as the test user is not admin.

yuhuyoyo:
rm Dan because he's OoO

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/GcpIamCustomRolePatchStep.java
```diff
@@ -0,0 +1,106 @@
...
+            .execute();
+      } catch (IOException e) {
+        handleIOException(e);
+      }
+    } else {
+      try {
+        iamCow
+            .projects()
+            .roles()
+            .delete(customGcpIamRoles.getFullyQualifiedRoleName(projectId))
```
zloery, line 89: Will the `patch` call in the do step create the role if it doesn't exist? If not, I think undo shouldn't delete it, since this flight couldn't have created it.

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/GcpIamCustomRolePatchStep.java
```diff
@@ -0,0 +1,106 @@
...
+        handleIOException(e);
+      }
+    }
+  }
+
+  private void handleIOException(IOException e) {
+    if (e instanceof GoogleJsonResponseException googleEx) {
+      // If receives client error, do not retry
+      if (googleEx.getStatusCode() >= 400 && googleEx.getStatusCode() < 500) {
+        return;
```
zloery, line 101: It seems like this will make the step fail silently. That makes sense if you don't want to fail the whole flight if any one project fails, but it also means we might have some projects in a bad state. It would be helpful for debugging if this also logged an error message with the error + project id + role id

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
```diff
@@ -141,10 +141,6 @@ private void populateInputParams() {
       throw new MissingRequiredFieldException("Missing flight class: flightClass");
     }
 
-    if (workspaceId == null) {
```
zloery, line 144: this seems unrelated, is this supposed to be in this PR?

service/src/test/java/bio/terra/workspace/service/admin/AdminServiceTest.java
```diff
@@ -0,0 +1,160 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class AdminServiceTest extends BaseConnectedTest {
+
+  private static final List<String> INCOMPLETE_READER_PERMISSIONS =
+      ImmutableList.of(
+          "serviceusage.services.get", "serviceusage.services.list", "storage.buckets.list");
+  private static final CustomGcpIamRole PROJECT_READER =
```
zloery, line 42: nit: rename this var something like `INCOMPLETE_PROJECT_READER` for clarity

melissachang (commented):
Nice!

Did you manually test this? If not, can you? :). You could manually edit role in GCP for a workspace.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -569,12 +570,31 @@ private Optional<String> getCloudContextWorker(UUID workspaceUuid, CloudPlatform
...
             .addValue("cloud_platform", cloudPlatform.toSql());
-
     return Optional.ofNullable(
         DataAccessUtils.singleResult(
             jdbcTemplate.query(sql, params, (rs, rowNum) -> rs.getString("context"))));
   }
 
+  @ReadTransaction
+  public ImmutableSet<String> listCloudContexts(CloudPlatform cloudPlatform) {
+    return listCloudContextsWorker(cloudPlatform);
```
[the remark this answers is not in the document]
yuhuyoyo: good question. this is just. a copy paste from the existing pattern. i removed it

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -569,12 +570,31 @@ private Optional<String> getCloudContextWorker(UUID workspaceUuid, CloudPlatform
...
+   * context that is done being created.
+   *
+   * @param cloudPlatform platform context to retrieve
+   */
+  private ImmutableSet<String> listCloudContextsWorker(CloudPlatform cloudPlatform) {
+    String sql = "SELECT context FROM cloud_context" + " WHERE cloud_platform = :cloud_platform";
+    MapSqlParameterSource params =
+        new MapSqlParameterSource().addValue("cloud_platform", cloudPlatform.toSql());
+
+    return ImmutableSet.copyOf(
```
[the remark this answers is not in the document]
yuhuyoyo: it can be null per workspace since i'm specifying cloudplatform. 

i throw a 500 if this happens.

service/src/main/java/bio/terra/workspace/service/job/JobBuilder.java
yuhuyoyo: no workspace id needs to be set for this flight.

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/GcpIamCustomRolePatchStep.java
yuhuyoyo: oh good catch.. i was mid-implementing this part and forgot..

i think we should create a new role if it didn't exist.

yuhuyoyo:
> Nice!
> 
> Did you manually test this? If not, can you? :). You could manually edit role in GCP for a workspace.

one problem is I'm not admin in the dev environment?

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -569,12 +570,31 @@ private Optional<String> getCloudContextWorker(UUID workspaceUuid, CloudPlatform
...
             jdbcTemplate.query(sql, params, (rs, rowNum) -> rs.getString("context"))));
   }
 
+  @ReadTransaction
+  public ImmutableSet<String> listCloudContexts(CloudPlatform cloudPlatform) {
+    return listCloudContextsWorker(cloudPlatform);
+  }
+
+  /**
+   * Retrieve all the serialized cloud contexts of an unlocked cloud context. That is, a cloud
```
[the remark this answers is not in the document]
yuhuyoyo: also copied from getCloudContextWorker. I removed it because I dont' know the answer to that.

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/GcpIamCustomRolePatchStep.java
```diff
@@ -0,0 +1,185 @@
...
+  private void updateCustomRole(CustomGcpIamRole customRole) throws RetryException {
+    try {
+      // Only assigned field will be updated.
+      Role role = new Role().setIncludedPermissions(customRole.getIncludedPermissions());
+      iamCow
+          .projects()
+          .roles()
+          .patch(customRole.getFullyQualifiedRoleName(projectId), role)
+          .execute();
+      logger.debug(
```
melissachang, line 75: logger.info, same for rest of file

I think we're not configured to show debug logs in GCP Logging

melissachang, line 75: Someone looking at this log should be able to figure out what the new permissions are. Either:

- Log only new permissions (or only which permissions are removed)
- Log all old permissions and all new permissions.

Same for line 149.

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/RetrieveGcpIamCustomRoleStep.java
```diff
@@ -0,0 +1,64 @@
...
+          // If receives client error, do not retry
+          if (googleEx.getStatusCode() >= 400 && googleEx.getStatusCode() < 500) {
+            return;
+          }
+        }
+        throw new RetryException(e);
+      }
+      // Do not put the role in the map if it is already deleted.
+      if (role.getDeleted() == null || !role.getDeleted()) {
+        context.getWorkingMap().put(fullyQualifiedRoleName, role);
```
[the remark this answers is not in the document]
melissachang: Did you try `TypeReference`? Here's a map flight key, it uses `TypeReference`.

https://cs.github.com/DataBiosphere/terra-workspace-manager/blob/9629c9fb57fd60b0ace6398fb91f117805dc8879/service/src/main/java/bio/terra/workspace/service/workspace/flight/GcpCloudSyncStep.java?q=gcpcloudsyncstep#L61

https://www.baeldung.com/jackson-map also mentions `TypeReference`.

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/SetupWorkingMapForUpdatedWorkspacesStep.java
```diff
@@ -0,0 +1,27 @@
...
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.HashSet;
+
+/**
+ * Add an empty hashset to the working map that is to put all the updated workspace id in. This is
+ * for activity logging to record all the updated workspaces.
+ */
+public class SetupWorkingMapForUpdatedWorkspacesStep implements Step {
```
melissachang, line 15: Why do you need a step for this? Instead, can activity log skip writing activity log if UPDATED_WORKSPACES isn't in working map?

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/SyncGcpIamRolesFlight.java
```diff
@@ -0,0 +1,45 @@
...
+
+    RetryRule cloudRetryRule = RetryRules.cloud();
+    addStep(new SetupWorkingMapForUpdatedWorkspacesStep());
+    for (Map.Entry<UUID, String> projectId : projectIds.entrySet()) {
+      // Wrap IAM with WSM service account.
+      addStep(
+          new RetrieveGcpIamCustomRoleStep(crl.getIamCow(), projectId.getValue()), cloudRetryRule);
+      if (FlightUtils.getRequired(inputParameters, IS_WET_RUN, Boolean.class)) {
+        // Update permissions for real only when it's not dry run.
+        addStep(
```
melissachang, line 38: The purpose of dry run is to see what would change. The only place we log this is GcpIamCustomRolePatchStep, right? If so, pass IS_WET_RUN to GcpIamCustomRolePatchStep. In GcpIamCustomRolePatchStep, log regardless of IS_WET_RUN.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -114,6 +114,27 @@ public AzureApi samAzureApi(String accessToken) {
...
+  private boolean isAdmin(AuthenticatedUserRequest userRequest) throws InterruptedException {
+    try {
+      // If the user can successfully call sam admin api, the user has terra level admin access.
+      SamRetry.retry(
+          () ->
+              samAdminApi(userRequest.getRequiredToken())
+                  .adminGetUserByEmail(getUserEmailFromSam(userRequest)));
+      return true;
+    } catch (ApiException apiException) {
+      logger.debug(
```
melissachang, line 131: logger.info ?

service/src/test/java/bio/terra/workspace/service/admin/AdminServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+    String jobId =
+        adminService.syncIamRoleForAllGcpProjects(
+            userAccessUtils.defaultUserAuthRequest(), /*wetRun=*/ true);
+    jobService.waitForJob(jobId);
+    for (String projectId : projectIds) {
+      assertProjectReaderRoleIsUpdated(
+          projectId, CUSTOM_GCP_PROJECT_IAM_ROLES.get(WsmIamRole.READER).getIncludedPermissions());
+    }
+    assertTrue(
```
melissachang, line 119: Should there be a test for:

- No new permissions, workspace is not updated
- No activity log

Maybe add to this test

melissachang (approved):

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/SyncGcpIamRolesFlight.java
melissachang: Could you add a section to DEVELOPMENT.md?

Only SAM admins can run this action.

- Dev dry run, check GCP logs to confirm new permissions/GCP projects as expected
- Dev wet run
- Autopush dry run
- Autopush wet run

Etc

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/GcpIamCustomRolePatchStep.java
yuhuyoyo: why are we configured to not show debug level?

i feel this is too noisy and i think debug is more like an on-demand logging. especially if there are a lot of workspaces

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/SetupWorkingMapForUpdatedWorkspacesStep.java
yuhuyoyo: so I can get it in the next step and not need to check whether the set exists everytime.

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/GcpIamCustomRolePatchStep.java
melissachang: The whole point of dry run is to sanity check these logs. So these logs need to be visible in GCP logging. (Unless there's some other way to confirm dry run, that I missed?)

Running this action is not common.

openapi/src/parts/admin.yaml
```diff
@@ -0,0 +1,36 @@
+## admin
+paths:
+  /api/admin/v1/workspaces/cloudcontexts/syncIamRoles:
+    patch:
+      summary: Sync iam roles of all workspaces.
+      operationId: syncIamRoles
+      tags: [Admin]
+      responses:
+        '200':
+          $ref: '#/components/responses/SyncIamRolesResultResponse'
```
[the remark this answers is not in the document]
zloery: ah right - we don't really need a result endpoint for this

yuhuyoyo:
re: typereference/


yes i tried with typereference and i'm still getting json deserialization error. i don't know why role is different from WsmIamRole. 

 I also noticed that json can't deserialize set of UUID and I have to use string intead. Also not know why.

service/src/test/java/bio/terra/workspace/service/admin/AdminServiceTest.java
yuhuyoyo: done

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/GcpIamCustomRolePatchStep.java
melissachang: I changed to Debug, I don't see any WSM debug logs in past 7 days

image

https://github.com/DataBiosphere/terra-workspace-manager/blob/main/DEVELOPMENT.md#controlling-log-level says "By default logging at INFO level and above are displayed"

openapi/src/parts/admin.yaml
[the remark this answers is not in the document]
yuhuyoyo: can you clarify?

you would call this endpoint, get a jobid. and then calls the jobs endpoint to check for completion

service/src/main/java/bio/terra/workspace/service/admin/flights/cloudcontexts/gcp/GcpIamCustomRolePatchStep.java
yuhuyoyo: odd to disable it. i usually find debug log uesful as info starts becoming spammy as the code base grows.

anyway, i will change them to info

openapi/src/parts/admin.yaml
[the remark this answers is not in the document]
zloery: The shared job endpoint gives job status in a `JobReport` object, but doesn't provide a way to return an endpoint-specific result object. For other async endpoints, we provide one endpoint to start the job (like `Workspace/createCloudContext`) and then another endpoint to poll/fetch the result object (like `Workspace/getCreateCloudContextResult`).

I don't see any step in this flight setting output (via `FlightUtils.setResponse`) - since this is an admin only endpoint, it seems like we're just expecting the caller to look at the service logs for results instead.

zloery (approved):

PR #864, opened 2022-10-14
[PF-2104] Add workspace_id index to workspace_activity_log

melissachang:
I didn't include change_date in index. I figure, having 2 rows with same change_date is extremely rare, so it's ok to sort in those cases.

Thanks for your help Dan, I haven't had much experience with DB slowness. :)

### DIFF
--- a/service/src/main/resources/db/changesets/20221014_workspace_activity_log_add_index.yaml
+++ b/service/src/main/resources/db/changesets/20221014_workspace_activity_log_add_index.yaml
@@ -1,0 +1,12 @@
+databaseChangeLog:
+- changeSet:
+    id: workspace_activity_log_add_index
+    author: melchang
+    changes:
+    - createIndex:
+        tableName:  workspace_activity_log
+        indexName: workspace_id_index
+        unique:  false
+        columns:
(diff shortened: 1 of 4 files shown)


service/src/main/resources/db/changesets/20221014_workspace_activity_log_add_index.yaml
```diff
@@ -0,0 +1,12 @@
+databaseChangeLog:
+- changeSet:
+    id: workspace_activity_log_add_index
+    author: melchang
+    changes:
+    - createIndex:
```
yuhuyoyo, line 6: why do we not need this for the other table?

yuhuyoyo (approved):

service/src/main/resources/db/changesets/20221014_workspace_activity_log_add_index.yaml
ddietterich: Most of our other tables have primary keys, so they don't have a separate index. We do have a few where we have specific indexes.

ddietterich (approved):
LGTM

PR #866, opened 2022-10-18
[PF-2108] Separate Job API support from Job Service

ddietterich:
We want to use JobService for top-level flights and for flights of flights. However, the service mixed two tasks:
- Wrapping of Stairway with WSM specific requirements
- Converting flight status and results into API format

In WSM, we are trying to keep almost all API conversions in the controller layer.

This refactor moves the API-related methods into a new controller support class, JobApiUtils. That leaves JobService only concerned with the internals of flights.

I also removed the unused `enumerateJobs` code from `JobService` and moved the in-use `enumerateJobs` code from `AlphaV1Service` to `JobService`. I updated the associated test.


service/src/main/java/bio/terra/workspace/service/job/JobService.java
```diff
@@ -465,6 +395,23 @@ public void setFlightDebugInfoForTest(FlightDebugInfo flightDebugInfo) {
     this.flightDebugInfo = flightDebugInfo;
   }
 
+  private void logAlert(String msg, String flightId) {
```
zloery, line 398: there's also a `LoggingUtils.logAlert` method that might be useful here. It only takes a String so you'd need to format/concat the flight ID into the message, but it saves a bit of duplication

zloery (approved):
Thanks for this cleanup!

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/job/JobService.java
ddietterich: That is a good idea. Since I am out of time, I'm going to merge as is and clean that up in another PR.

PR #867, opened 2022-10-18
Clean up WorkspaceApiControllerTest

melissachang:
This way, if people copy WorkspaceApiControllerTest, they'll copy the cleaned-up version.

The workspace property tests weren't working as expected. Changed them to actually test workspace properties.


ddietterich (approved):
Very nice, thanks!

zloery (approved):

PR #869, opened 2022-10-20
[WOR-515] Check for empty spend models received from configuration and filter out

aherbst-broad:
We may receive "empty" spend models injected via environment variables (i.e., `WORKSPACE_SPEND_SPENDPROFILES_0_ID=;WORKSPACE_SPEND_SPENDPROFILES_0_BILLINGACCOUNTID=;') will render empty strings and cause the spend profile service to barf when it's instantiated at startup. Rather than set bogus values for those environment variables in all of our environments where they are not needed right now (i.e., prod), we should just check for the empty strings and filter out the models at bootstrap.


cahrens (approved):

rtitle (approved):
👍

PR #853, opened 2022-10-05
[PF-2059] Preserve new folders/ new resources structure in the new clone workspace

ginaygoogle:
This PR comes from the ticket [PF-2059](https://broadworkbench.atlassian.net/browse/PF-2059). 

This ticket is for completing resource/folder structure in the workspace clone process. It improved 2 process which did not do before.
- Before this ticket, the workspace clone does not include the folders cloning. either does not include the folder parent relationship. In this ticket,  we create a step for the folder cloning with the parent relations in cloning the workspace process.
- Before this ticket, the workspace clone includes all resources cloning. but the resources do not have properties with the folderId information. In this ticket,  we improve the Controlled resources cloning step with the parent relationship in cloning the workspace process.

The manually test shows below:
```
ginay-macbookpro:terra-cli ginay$ terra resource add-ref bq-dataset --name=refdataset --project-id=projectid --dataset-id=datasetid
Successfully added referenced BigQuery dataset.
Name:         refdataset
Description:  
Type:         BQ_DATASET
Stewardship:  REFERENCED
Cloning:      COPY_REFERENCE
Properties:   class Properties {
    []
}
GCP project id: projectid
BigQuery dataset id: datasetid
Location: (undefined)
# Tables: (unknown)
ginay-macbookpro:terra-cli ginay$ terra workspace clone --new-id=cloneworkspace
Workspace successfully cloned.
Source Workspace:
ID:                1028test
Name:              
Description:       
Google project:    terra-wsm-t-speedy-orange-3015
Cloud console:     https://console.cloud.google.com/home/dashboard?project=terra-wsm-t-speedy-orange-3015
Properties:
Created:           2022-10-30
Last updated:      2022-10-30
# Resources:       2

Destination Workspace:
ID:                cloneworkspace
Name:              1028test (Copy)
Description:       
Google project:    terra-wsm-t-glossy-kale-1261
Cloud console:     https://console.cloud.google.com/home/dashboard?project=terra-wsm-t-glossy-kale-1261
Properties:
Created:           2022-10-30
Last updated:      2022-10-30
# Resources:       2

Resources:
--------------------------------
Source resource:
    Name:         bucket
    Description:  
    Type:         GCS_BUCKET
    Stewardship:  CONTROLLED
    Cloning:      COPY_RESOURCE
    Access scope: SHARED_ACCESS
    Managed by:   USER
    Properties:   class Properties {
    [class Property {
        key: terra-folder-id
[...]

PR #873, opened 2022-10-21
[PF-2059] add deleteFolders to FolderDao

yuhuyoyo:
needed as an undo step for https://github.com/DataBiosphere/terra-workspace-manager/pull/853.

Without this method, we need to iterate through all the top level folders.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/db/FolderDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/FolderDao.java
@@ -266,7 +266,7 @@ public class FolderDao {
 
   /** Delete all folders of a given workspaceId. */
   @WriteTransaction
-  public boolean deleteFolders(UUID workspaceId) {
+  public boolean deleteAllFolders(UUID workspaceId) {
     final String sql = "DELETE FROM folder WHERE workspace_id = :workspace_id";
     MapSqlParameterSource params =
         new MapSqlParameterSource().addValue("workspace_id", workspaceId.toString());
--- a/service/src/main/java/bio/terra/workspace/db/FolderDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/FolderDao.java
@@ -264,6 +264,23 @@ public class FolderDao {
     return ImmutableList.copyOf(jdbcTemplate.query(sql, params, FOLDER_ROW_MAPPER));
   }
 
+  /** Delete all folders of a given workspaceId. */
+  @WriteTransaction
+  public boolean deleteFolders(UUID workspaceId) {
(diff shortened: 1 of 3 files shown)


ginaygoogle (approved):

melissachang (approved):
I assume this is for CloneAllFoldersStep. CloneAllFoldersStep currently does:

folderDao.deleteFolderRecursive(destinationWorkspaceID, /*folderId=*/ null);

After this PR, I assume this will change to

deleteAllFolders();

I actually like `folderDao.deleteFolderRecursive(destinationWorkspaceID, /*folderId=*/ null);` better ... I don't see the point of adding a new method, if `folderDao.deleteFolderRecursive(destinationWorkspaceID, /*folderId=*/ null);` works.

So my preference is to revert this PR, up to you.

yuhuyoyo:
deleteFolderRecursive(workspsaceId,null) wouldn't work because the sql query will have the base case of folder_id = null.

service/src/main/java/bio/terra/workspace/db/FolderDao.java
```diff
@@ -274,7 +291,7 @@ WITH RECURSIVE subfolders AS (
    * @return true if folder(s) are deleted.
    */
   @WriteTransaction
-  public boolean deleteFolderRecursive(UUID workspaceUuid, UUID folderId) {
+  public boolean deleteFoldersRecursive(UUID workspaceUuid, UUID folderId) {
```
melissachang, line 294: Can you add a comment like This cannot be used for root folder. For root folder, use deleteAllFolders().

yuhuyoyo: there is no root folder. It's a secret that FolderDao hold. also this folderId is nonNull.

melissachang: What do we store in DB for root folder, `0`? (Ie what is `parent_folder_id` for a top-level folder?) It's reasonable that someone could try to call this method for that, it doesn't hurt to add a comment and be clear.

yuhuyoyo: it's not stored. because 0 is not a uuid.

melissachang: Doesn't hurt to add a comment, in case someone's wondering how to recursively delete folders for root folder. :)

yuhuyoyo: to clarify, it's not stored as a row

PR #874, opened 2022-10-21
[PF-2074] Disable BPM connected tests if BPM is not configured

zloery:
BPM isn't deployed at Verily yet, so these tests (which assume a valid BPM server is available) break, regardless of flag values. We should remove these annotations once one is available, which should be within a week or two.


service/src/test/java/bio/terra/workspace/service/spendprofile/SpendProfileBpmConnectedTest.java
```diff
@@ -27,6 +29,11 @@ public class SpendProfileBpmConnectedTest extends BaseConnectedTest {
 
   SpendProfile profile;
 
+  /** Condition used to disable these tests if run in a deployment where BPM is not configured. */
+  public boolean bpmUnavailable() {
```
zloery, line 33: I couldn't find an easy way to cut out this duplication - using the `DisabledIf` annotation at the class level or using a function defined outside the class requires that this condition be static, which won't work here since we need to read config values.

aherbst-broad:
Good catch, thanks for doing this.

yuhuyoyo (approved):

yuhuyoyo:
side note, can we similarly turn off all the azure test if azure is not on? lol they takes so long

PR #876, opened 2022-10-25
[PF-2074] Fix disabled SpendProfileBpmConnectedTest

zloery:
In #874 I disabled `SpendProfileBpmConnectedTest` methods individually in environments where BPM is disabled, but I forgot that `@Before/AfterAll` will still run (and make calls to BPM) even if no test methods are enabled.


yuhuyoyo (approved):
how did this get in

yuhuyoyo:
> how did this get in

nvm looking at the diff

aherbst-broad (approved):

PR #881, opened 2022-10-27
[PF-2158] Use workspace region for notebook instances when available

iliadsh:


ddietterich:
Why is zone mapping needed for notebooks? We do not use that anywhere else and it appears to work properly without the zone.

It is a maintenance headache to have a flat list of locations inline in the code. The list is fairly dynamic.
We are planning to have a multi-cloud ontology of geographic regions in TPS. We will maintain the list in a yaml file and expose APIs for operating on it. That would be a better source for this information if it is needed.

cbookg:
> Why is zone mapping needed for notebooks? We do not use that anywhere else and it appears to work properly without the zone.
> 
> It is a maintenance headache to have a flat list of locations inline in the code. The list is fairly dynamic. We are planning to have a multi-cloud ontology of geographic regions in TPS. We will maintain the list in a yaml file and expose APIs for operating on it. That would be a better source for this information if it is needed.

There is a config property for a workspace's default region.  ie the user could specify they want the data in europe-west2, and we should create notebooks in that region.

cbookg:
> if no region is specified in the request, use the workspace default region

The notebook API takes a zone though, not a region?

cbookg:
So our "instance" API took a "location" param which was treated as a GCP zone.  Are you suggesting we change the meaning of this to be a region, and then always map the user-provided value?

ddietterich:
> The notebook API takes a zone though, not a region?

I have been hunting in the documentation to see if that is the case. I can find no evidence that it requires a zone.
Maybe it did at one point. We were implementing on it very early in its life.

cbookg:
> > The notebook API takes a zone though, not a region?
> 
> I have been hunting in the documentation to see if that is the case. I can find no evidence that it requires a zone. Maybe it did at one point. We were implementing on it very early in its life.

So yeah, its super weird they use "location" in this context, which is unclear.  I found this page that seems to imply it is a zone:
https://cloud.google.com/vertex-ai/docs/general/locations#vertex-ai-workbench-locations

We can certainly try to use a region and see what happens...

ddietterich:
OK, the notebook create does require a specific zone. If a location is specified on input, I think we should require that it has a zone. I believe right now we do no validation.

This is going to take a little thought. Some ideas:
1. Have both a default workspace region and a default zone. If the default zone is not specified, we query GCP for the available zones using [this endpoint](https://cloud.google.com/compute/docs/reference/rest/v1/zones/list) and randomly pick one.
2. Use the same logic if the notebook input location does not have a zone specified.

Allow the notebook creation to fail if a location is supplied that does not support notebooks. That is how it works with all other resources.

cbookg:
We previously discussed this and we don't want to present the user with a separate 'default zone'.

Using the API you refer to instead of the static map is reasonable.  I think 'random' isn't ideal.  We perhaps want to make consistent choice in the workspace.

ddietterich:
How about this: the workspace default region will accept either:
- nothing
- a region
- a region + zone

If the customer cares about zones, they can specify one.

If the customer specifies a region without a zone, we will randomly pick a zone _and remember the choice_ so that everything that wants a zone will use that one.

If the customer makes no setting, we will use us-central1-a.
In all cases, WSM can produce a consistent default region and default region+zone.

ddietterich:
This does raise another question: what if the default resource location is gobbledygook?
Since it is a property, we have no validation. That means someone can get a failure due to a bad setting by someone else.
(This is the second time that has come up. Folder ids was the first. Maybe we need validators on properties?)

iliadsh:
> This does raise another question: what if the default resource location is gobbledygook? Since it is a property, we have no validation. That means someone can get a failure due to a bad setting by someone else. (This is the second time that has come up. Folder ids was the first. Maybe we need validators on properties?)

This change will fall back to the default zone (us-central1-a) if the workspace region property is not recognized.

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -115,6 +115,60 @@ private String getResourceLocation(Workspace workspace, String requestedLocation
         : requestedLocation;
   }
 
+  private String getResourceZone(Workspace workspace, String requestedZone) {
```
yuhuyoyo, line 118: annotate return type @nullable

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -115,6 +115,60 @@ private String getResourceLocation(Workspace workspace, String requestedLocation
...
+    if (Strings.isNullOrEmpty(requestedZone)) {
+      String location =
+          workspace
+              .getProperties()
+              .getOrDefault(
+                  WorkspaceConstants.Properties.DEFAULT_RESOURCE_LOCATION,
+                  GcpResourceConstant.DEFAULT_REGION);
+      switch (location) {
+        case "asia-east1":
```
yuhuyoyo, line 127: do we even support all these regions?

yuhuyoyo: https://cs.github.com/verily-src/terra-service-resource-buffer/blob/2fbba28d70e761ad56564ff8a48f2180e080cbe9/deployment/resource-buffer/configs/prod/resource-config/resource_verilyprod_v4.yml#L42
looks like these are blocked?

makes me nervous that we are not reading from the source. random thought: there is this https://cloud.google.com/compute/docs/reference/rest/v1/regions/list api that we can use to validate if the user's regional preference is even feasible.

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -115,6 +115,60 @@ private String getResourceLocation(Workspace workspace, String requestedLocation
...
+        case "us-west1":
+        case "us-west2":
+        case "us-west3":
+        case "us-west4":
+          return location + "-a";
+        case "europe-west1":
+        case "us-east1":
+          return location + "-b";
+        default:
+          return null;
```
yuhuyoyo, line 165: return us-central-1?

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -359,14 +413,15 @@ public ResponseEntity<ApiCreatedControlledGcpAiNotebookInstanceResult> createAiN
...
+    Workspace workspace =
+        workspaceService.validateWorkspaceAndAction(
+            userRequest, workspaceUuid, ControllerValidationUtils.samCreateAction(commonFields));
     String projectId = gcpCloudContextService.getRequiredGcpProject(workspaceUuid);
 
     ControlledAiNotebookInstanceResource resource =
         ControlledAiNotebookInstanceResource.builder()
             .common(commonFields)
-            .location(body.getAiNotebookInstance().getLocation())
+            .location(getResourceZone(workspace, body.getAiNotebookInstance().getLocation()))
```
yuhuyoyo, line 424: i think there's a validator util class that you can put the helper method in.

cbookg:
@ddietterich How about:
- We keep the single "region" setting.
- When we need a zone (for instance creation), we query the API to find the list of available zones.
- We take the first (alphabetical) zone matching the region.

Thoughts?

ddietterich:
> Thoughts?
That works for me.

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
iliadsh: Using null here since the builder will use its own DEFAULT_ZONE (which does happen to be us-central-1) if it receives null

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -115,6 +123,36 @@ private String getResourceLocation(Workspace workspace, String requestedLocation
         : requestedLocation;
   }
 
+  private @Nullable String getResourceZone(Workspace workspace, String requestedZone) {
```
ddietterich, line 126: This computation should be done as part of the createNotebook flight, not in the controller. We try to avoid cloud calls and external service calls from the controller. Plus by doing them in a Stairway flight, we get to easily retry on cloud errors.

I think you should use `getResourceLocation` above to fill in the location param in the object.
Then do the part from line 135 on down in the `RetrieveNetworkNameStep`. That isolates the zone concerns to that one place.

ddietterich (changes requested):
I think you do not need the change in the controller and should isolate the zone computation to the `RetrieveNetworkNameStep` of the ainotebook create flight.

I like the algorithm for zone selection.

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
iliadsh: Hey @ddietterich , sorry for the late reply. The issue I'm running into when trying this is that there doesn't seem to be a good way to store back the computed zone in the step. Since the API allows you to specify a zone, using getResourceLocation would mean mixing zones/regions in the location param and while we could store the real zone in the flight's working map, the other operations like delete or update don't have access to this info.

ddietterich: I see, you need to populate the resource object with the corrected location. And that you need to do in the controller.

ddietterich: You can store back the computed zone by calling the resourceDao with updated attributes.
I would suggest the following structure:
- In `RetrieveNetworkNameStep` remember the complete region including zone in the working map.
- Add a step `UpdateMetadataRegionStep`. The constructor would take the AI Notebook resource object (this is an input parameter) and the `ResourceDao` bean. The do() code would call `updateResource` providing a revised set of attributes that include the region.

This is not a perfect solution, but it is closer to the way I am hoping to move the other resources. So it will become more perfect over time :)

ddietterich: To expand a bit. We have a general design problem with handling the per-resource attributes. It is done inconsistently. Having to have completely built resources in the controller in the create case makes things like this messy. It also makes clone messy, since we are building the resources from several different pieces.

So sorry you plunged into one of the messier parts of WSM...

iliadsh: TYSM for the pointers.

yuhuyoyo (changes requested):
lgtm. just have some nits

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
```diff
@@ -103,7 +104,9 @@ public StepResult doStep(FlightContext flightContext)
             .getWorkingMap()
             .get(ControlledResourceKeys.GCP_CLOUD_CONTEXT, GcpCloudContext.class);
     String projectId = gcpCloudContext.getGcpProjectId();
-    InstanceName instanceName = resource.toInstanceName(projectId);
+    final String requestedLocation =
```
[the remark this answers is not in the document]
ddietterich: I agree. That said, our code is inconsistent. I've been removing local finals as I find them.

service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
```diff
@@ -112,4 +113,18 @@ public static String getWsmSaEmail() {
           "Unable to find WSM service account credentials. Ensure WSM is actually running as a service account");
     }
   }
+
+  /**
+   * Extract the name from a network URL like
+   * "https://www.googleapis.com/compute/v1/projects/{PROJECT_ID}/global/networks/{NAME}" or route
+   * URL like "https://www.googleapis.com/compute/v1/projects/{PROJECT_ID}/regions/{REGION_NAME}"
+   */
+  public static String extractNetworkNameFromUrl(String url) {
```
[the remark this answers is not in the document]
ddietterich: I agree, since that is now the only place it is used.

ddietterich (approved):
LGTM with Yu's comments addressed.
Thanks for working through this.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/RetrieveNetworkNameStep.java
```diff
@@ -92,7 +106,7 @@ private void pickAndStoreNetwork(SubnetworkList subnetworks, FlightMap workingMa
    * "https://www.googleapis.com/compute/v1/projects/{PROJECT_ID}/global/networks/{NAME}" or route
    * URL like "https://www.googleapis.com/compute/v1/projects/{PROJECT_ID}/regions/{REGION_NAME}"
    */
-  private static String extractNameFromUrl(String url) {
+  public static String extractNameFromUrl(String url) {
```
yuhuyoyo, line 109: private

yuhuyoyo (approved):

PR #883, opened 2022-11-01
[WOR-530] Clone azure storage container definition

aherbst-broad:
## Context
This is the first step in getting azure storage container cloning up and running. This PR is focused on the basics of wiring up a clone container flight that clones a container definition, deferring the actual copying of the constituent files to a subsequent [ticket](https://broadworkbench.atlassian.net/browse/WOR-591). There is also a follow-on [task](https://broadworkbench.atlassian.net/browse/WOR-592) to create an integration test that exercises this flight.

## This PR
* Creates a new `CloneControlledAzureStorageContainerResourceFlight` that copies a container's definition. Copying referenced containers is not supported, nor is the actual copying of files. I've stubbed out the step that will house the file copy operations for now.
* To determine the destination container's parent storage account, I've implemented a step called `RetrieveDestinationStorageAccountResourceIdStep`. This step determines the storage account to use for the container, either the destination workspace's own storage account (if present), or the destination workspace's parent landing zone (again, if present).
* Adds a test for the retrieval of the storage account, as well as the container definition creation. As noted, we'll implement an integration test once this flight is fully built.


openapi/src/parts/controlled_azure_storage_container.yaml
```diff
@@ -75,7 +101,65 @@ paths:
...
+        and description null to take the source object's values.
+      type: object
+      required: [ destinationWorkspaceId, jobControl ]
+      properties:
+        destinationWorkspaceId:
+          description: ID of workspace to be cloned into
+          type: string
+          format: uuid
+        cloningInstructions:
+          $ref: "#/components/schemas/CloningInstructionsEnum"
```
cahrens, line 137: I'm confused about how the cloning instructions work-- the cloning instructions are set on a container at the time that it is created. So are these the cloning instructions for the newly created (via cloning) destination container? Or are these supposed to control what is cloned from the source container?

cahrens: Is it possible to add a description to this in addition to the $ref?

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -235,6 +236,56 @@ public String cloneGcsBucket(
...
+
+    // Write access to the target workspace will be established in the create flight
+    final String jobDescription =
+        String.format(
+            "Clone controlled resource %s; id %s; name %s",
+            sourceContainer.getResourceType(),
+            sourceContainer.getResourceId(),
+            sourceContainer.getName());
+
+    // If TPS is enabled, then we want to merge policies when cloning a container
```
cahrens, line 261: Nit: could you replace TPS in the comment with Terra Policy Service? It took me some time to figure out what TPS stands for.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
```diff
@@ -0,0 +1,125 @@
...
+    var destinationWorkspaceId =
+        inputParameters.get(
+            WorkspaceFlightMapKeys.ControlledResourceKeys.DESTINATION_WORKSPACE_ID, UUID.class);
+
+    boolean mergePolicies =
+        Optional.ofNullable(
+                inputParameters.get(WorkspaceFlightMapKeys.MERGE_POLICIES, Boolean.class))
+            .orElse(false);
+    var cloningInstructions =
+        Optional.ofNullable(
```
cahrens, line 51: At this point it shouldn't be nullable because of the similar conditional code at https://github.com/DataBiosphere/terra-workspace-manager/pull/883/files#diff-3aaa5bf7189ebec04c5d21fb75fdaed7df9f73f342118e76ecde18b2e1f69b92R281, correct?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    // Get the cloud context and store it in the working map
+    addStep(
+        new GetCloudContextStep(
+            destinationWorkspaceId,
+            CloudPlatform.AZURE,
+            flightBeanBag.getGcpCloudContextService(),
+            flightBeanBag.getAzureCloudContextService()),
+        RetryRules.shortDatabase());
```
cahrens, line 71: How does one decide which RetryRules to use?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
```diff
@@ -0,0 +1,125 @@
...
+      addStep(
+          new ClonePolicyAttributesStep(
+              sourceResource.getWorkspaceId(),
+              destinationWorkspaceId,
+              userRequest,
+              flightBeanBag.getTpsApiDispatch()));
+    }
+
+    addStep(
+        new RetrieveControlledResourceMetadataStep(
```
cahrens, line 94: Where does the retrieved metadata get stored?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
```diff
@@ -0,0 +1,125 @@
...
+            flightBeanBag.getResourceDao(),
+            sourceResource.getWorkspaceId(),
+            sourceResource.getResourceId()));
+
+    // TODO WOR-590 add step to copy source container metadata + attributes
+
+    if (CloningInstructions.COPY_REFERENCE == cloningInstructions) {
```
cahrens, line 101: Nit: it seems like we should validate this much earlier since there is no need to do anything else in this case. For instance, at input argument validation time. And it would be good to state in the Swagger documentation that COPY_REFERENCE is not supported.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
cahrens: Ah, I think in the working map?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/RetrieveDestinationStorageAccountResourceIdStep.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.UUID;
+
+/**
+ * Attempts to fetch the storage account ID for a new container in the destination workspace
+ * following process:
+ *
+ * <p>1. Checks the destination workspace for a single storage account; if more than one is present
+ * we are in an unsupported state and fail out.
+ *
+ * <p>2. If no storage accounts are present, attempts to check the owning Landing Zone. If present,
```
cahrens, line 30: Should we prefer the landing zone over the workspace's storage account (given that is the direction we are headed)?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtils.java
```diff
@@ -164,6 +165,28 @@ public static ControlledGcsBucketResource buildDestinationControlledGcsBucket(
...
   }
 
+  public static ControlledAzureStorageContainerResource buildDestinationControlledAzureContainer(
+      ControlledAzureStorageContainerResource sourceContainer,
+      UUID storageAccountId,
+      UUID destinationWorkspaceId,
+      UUID destinationResourceId,
+      String name,
+      @Nullable String description,
+      String cloudInstanceName) {
```
cahrens, line 175: Why `cloudInstanceName` for the storage container name?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
cahrens: Also, there should be some test coverage of this and the COPY_NOTHING path.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
cahrens: There should be some test coverage of the override behavior.

openapi/src/parts/controlled_azure_storage_container.yaml
```diff
@@ -24,6 +24,32 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/workspaces/v1/{workspaceId}/resources/controlled/azure/storageContainer/{resourceId}/clone:
```
ddietterich, line 27: I would expect to see a `.../clone/result` endpoint to complete the async endpoint implementation.

openapi/src/parts/controlled_azure_storage_container.yaml
ddietterich: The cloning instructions are telling WSM how to do the clone: make a reference, make a new object like the old one, make a new object and copy the contents of the old one.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
ddietterich: See the documentation in the code [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/6c85db7b50ddf77de641710d6e8cda6437f6c99b/service/src/main/java/bio/terra/workspace/common/utils/RetryRules.java#L20)

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
```diff
@@ -0,0 +1,125 @@
...
+            flightBeanBag.getGcpCloudContextService(),
+            flightBeanBag.getAzureCloudContextService()),
+        RetryRules.shortDatabase());
+
+    // Flight plan
+    // 1. Check user has read access to source container
+    // 2. Gather controlled resource metadata for source object
+    // 3. Copy container definition to new container resource
+    // 3. If referenced, bail out as unsupported.
+    // 4. Copy files from source container to destination container (TODO)
```
ddietterich, line 78: Good practice is to cite the PR on a TODO.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
```diff
@@ -0,0 +1,155 @@
...
+            destinationContainerName);
+    ApiAzureStorageContainerCreationParameters destinationCreationParameters =
+        new ApiAzureStorageContainerCreationParameters()
+            .storageContainerName(destinationContainerName)
+            .storageAccountId(destStorageAccountId);
+    ControlledResourceIamRole iamRole =
+        IamRoleUtils.getIamRoleForAccessScope(sourceContainer.getAccessScope());
+
+    // create the container
+    ControlledAzureStorageContainerResource clonedContainer =
```
ddietterich, line 101: What happens if this is the second execution of this step? E.g., the pod is restarted on an upgrade.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
```diff
@@ -0,0 +1,155 @@
...
+    workingMap.put(
+        WorkspaceFlightMapKeys.ControlledResourceKeys.CLONED_RESOURCE_DEFINITION, clonedContainer);
+
+    var apiCreatedContainer =
+        new ApiCreatedControlledAzureStorageContainer()
+            .azureStorageContainer(clonedContainer.toApiResource())
+            .resourceId(destinationContainerResource.getResourceId());
+
+    var apiContainerResult =
+        new ApiClonedControlledAzureStorageContainer()
```
ddietterich, line 117: I know it is common practice to build and return the API form of the result here. I want to move us to returning the internal form of the structure and doing the conversion from internal to API form in the controller. We may end up with some additional `record` objects to move the result around, but it will help us better insulate the implementation code and what we serialize in the database from API structures.

This will be part of the clone refactor work, so asking you to change it here is entirely self-serving! :)

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
```diff
@@ -0,0 +1,155 @@
...
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    var clonedContainer =
+        context
+            .getWorkingMap()
+            .get(
+                WorkspaceFlightMapKeys.ControlledResourceKeys.CLONED_RESOURCE_DEFINITION,
+                ControlledAzureStorageContainerResource.class);
+    if (clonedContainer != null) {
```
ddietterich, line 149: Because you set the working map after the create, you will miss deleting this if the pod is stopped at line 106.
Ideally, you can save state in the map and then try the create.

Also, I think the undo needs to handle the not-found case, because it can be retried/rerun so the object might already be gone. That should not be an error. (Maybe double check that the create code does that right as well?)

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/RetrieveDestinationStorageAccountResourceIdStep.java
```diff
@@ -0,0 +1,124 @@
...
+import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys;
+import bio.terra.workspace.service.workspace.model.AzureCloudContext;
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * Attempts to fetch the storage account ID for a new container in the destination workspace
+ * following process:
+ *
+ * <p>1. Checks the destination workspace for a single storage account; if more than one is present
```
ddietterich, line 27: Nittiest nit: you can use `<ol>` for an ordered list here

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtils.java
ddietterich: It would be good to have javadoc on this and explain the params.

ddietterich (changes requested):
Some idempotency issues in the flight

openapi/src/parts/controlled_azure_storage_container.yaml
cahrens: Right, but the container had cloning instructions set on it at creation time. Having gotten through the rest of the PR now, I see that this is an optional override-- it would be helpful to document that.

openapi/src/parts/controlled_azure_storage_container.yaml
cahrens: There is also a general-purpose job result endpoint. Why do we need endpoints specific to each resource type?

`/api/job/v1/jobs/{jobId}`.

ddietterich: The requirements for async endpoints are defined in this doc: [MC Terra - Async REST API Specification](https://docs.google.com/document/d/1PTd4xvmV9xnEkWaIgFc6d3VUyJwwymQTFdhpKrdsUKw/edit).

The specific result endpoint provides a type-safe interface. The `jobs` API is to let us write generic waiting code. It does not return the result object. It refers to the resultURL, which is the type-safe endpoint, for retrieving the actual result body.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
aherbst-broad: Correct

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
aherbst-broad: A good question, the name will collide when WSM goes to create the resource in the DB and we'll get a DuplicateResourceException. I'm not sure how we can reliably determine if the duplicate is due to retry vs a genuine naming collision at this point in the flight -- looking around a bit it seems like we are susceptible to this same issue in our other cloning flights as well (i.e, this [step](https://github.com/DataBiosphere/terra-workspace-manager/blob/7f51c97637563277723d882a256ada141f2e85e8/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/CopyGcsBucketDefinitionStep.java#L119) for gcs buckets and this [step](https://github.com/DataBiosphere/terra-workspace-manager/blob/7f51c97637563277723d882a256ada141f2e85e8/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CopyBigQueryDatasetDefinitionStep.java#L111-L115) for bq datasets).

It may make sense to handle that exception here and move on to blob cloning -- thoughts @ddietterich ?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
aherbst-broad: Fair. I'll leave this check here and add one at the controller level (belts and suspenders, etc.)

openapi/src/parts/controlled_azure_storage_container.yaml
aherbst-broad: Will add.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
ddietterich: I think I don't have the whole picture in my head. If you can assure uniqueness of the resource by inspecting the WSM metadata, then you don't have a problem here. I'm not sure how you do that, but if all containers are allocated through WSM and you scan across all workspaces, it seems possible. If that is so, you'd just need to tolerate the duplicate error because you could be sure ahead of time that the name was unique and this flight was responsible for creating it.

Sometimes it is not possible to completely eliminate the window. One technique to reduce the window is to have an existence check step ahead of the create step. If the existence check step fails with a duplicate, we fail the flight with "duplicate". If the create step fails with a duplicate, we know with high probability that it was us retrying and we continue.

The bad case is when the duplicate entry exists before this flight, the create fails with duplicate, and then the create-undo deletes the object during cleanup. We just deleted someone else's cloud resource!

aherbst-broad: If we check for the resource name within the workspace prior to the creation step, I think we should be able to reliably retry the copy definition step. If the resource already exists prior to the start of this flight (i.e,. someone else's container), the flight will fail out prior to the copy definition step. I'll add a step to do this.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
aherbst-broad: I've added a [ClonedAzureStorageContainer](https://github.com/DataBiosphere/terra-workspace-manager/blob/arh-wor-532-clone-container-flight-mk2/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/ClonedAzureStorageContainer.java) intermediate model that is set as the flight result here, which the API then adapts to an API result.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtils.java
aherbst-broad: See update doc [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/1b725dff35e3e1667f36eb1e7aad95f19269907f/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtils.java#L168-L180)

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/RetrieveDestinationStorageAccountResourceIdStep.java
aherbst-broad: [Switched](https://github.com/DataBiosphere/terra-workspace-manager/blob/1b725dff35e3e1667f36eb1e7aad95f19269907f/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/RetrieveDestinationStorageAccountResourceIdStep.java#L68-L92) around the landing zone to be the first place we check

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java
aherbst-broad: See this [unit test](https://github.com/DataBiosphere/terra-workspace-manager/blob/1b725dff35e3e1667f36eb1e7aad95f19269907f/service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlightTest.java#L38)

openapi/src/parts/controlled_azure_storage_container.yaml
aherbst-broad: See this [update](https://github.com/DataBiosphere/terra-workspace-manager/blob/1b725dff35e3e1667f36eb1e7aad95f19269907f/openapi/src/parts/controlled_azure_storage_container.yaml#L145-L147) (note I didn't update the description on the cloning instructions enum that is being referenced here as this behavior may not be universally applicable).

openapi/src/parts/controlled_azure_storage_container.yaml
```diff
@@ -24,6 +24,32 @@ paths:
...
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CloneControlledAzureStorageContainerRequest'
+      responses:
+        '200':
+          $ref: '#/components/responses/ClonedControlledAzureStorageContainerResponse'
+        '202':
+          $ref: '#/components/responses/ClonedControlledAzureStorageContainerResponse'
```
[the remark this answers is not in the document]
aherbst-broad: I don't actually know, this is something that gets copy-pasta'ed on all WSM responses, @ddietterich may have thoughts here.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
aherbst-broad: I've added a check for duplicate step [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/1b725dff35e3e1667f36eb1e7aad95f19269907f/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CloneControlledAzureStorageContainerResourceFlight.java#L102) and now catch the duplicate resource exception [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/1b725dff35e3e1667f36eb1e7aad95f19269907f/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java#L112-L123)

openapi/src/parts/controlled_azure_storage_container.yaml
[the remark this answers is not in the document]
ddietterich: See https://docs.google.com/document/d/1PTd4xvmV9xnEkWaIgFc6d3VUyJwwymQTFdhpKrdsUKw/edit#heading=h.ol9mx3vfhjjj

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
```diff
@@ -0,0 +1,185 @@
...
+        WorkspaceFlightMapKeys.ControlledResourceKeys.CLONED_RESOURCE_DEFINITION,
+        destinationContainerResource);
+
+    // create the container
+    try {
+      controlledResourceService.createControlledResourceSync(
+          destinationContainerResource, iamRole, userRequest, destinationCreationParameters);
+    } catch (DuplicateResourceException e) {
+      // We are catching DuplicateResourceException here since we check for the container's presence
+      // earlier in the parent flight and bail out if it already exists.
```
ddietterich, line 117: What is the parent flight of this flight?

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
cahrens: Nit: would still appreciate this, if you end up needing to make more changes.

ddietterich (approved):
LGTM

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/VerifyContainerResourceDoesNotExist.java
```diff
@@ -0,0 +1,53 @@
...
+import bio.terra.stairway.StepStatus;
+import bio.terra.stairway.exception.RetryException;
+import bio.terra.workspace.common.utils.FlightUtils;
+import bio.terra.workspace.db.ResourceDao;
+import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class VerifyContainerResourceDoesNotExist implements Step {
```
cahrens, line 16: I don't see any direct test coverage for this. Probably worth testing that it does detect if a container already exists.

cahrens:
@aherbst-broad just a couple things left, otherwise LGTM

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
aherbst-broad: This is meant to be the "parent flight of this step", will reword.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/VerifyContainerResourceDoesNotExist.java
aherbst-broad: Sure, good call: https://github.com/DataBiosphere/terra-workspace-manager/pull/883/commits/49ad9f3e1092c778b585ec6f03aa19565a6ac417

aherbst-broad:
@cahrens test added [here ](https://github.com/DataBiosphere/terra-workspace-manager/pull/883/commits/49ad9f3e1092c778b585ec6f03aa19565a6ac417)

cahrens (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/RetrieveDestinationStorageAccountResourceIdStep.java
```diff
@@ -0,0 +1,127 @@
...
+ * following this process:
+ *
+ * <ol>
+ *   <li>Checks the destination workspace for a single storage account; if more than one is present
+ *       we are in an unsupported state and fail out.
+ *   <li>If no storage accounts are present, attempts to check the owning Landing Zone. If present,
+ *       returns the storage account from the LZ, otherwise we fail out.
+ * </ol>
+ */
+public class RetrieveDestinationStorageAccountResourceIdStep implements Step {
```
cahrens, line 35: I feel like this logic is duplicated in the VerifyAzureStorageContainerCanBeCreated step.https://github.com/DataBiosphere/terra-workspace-manager/blob/688c3f42db57fcb1206f2a3c3a45ff6d1d12f85e/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/VerifyAzureStorageContainerCanBeCreatedStep.java#L140

PR #885, opened 2022-11-02
PF-2179: user jupyter own ~/.bash_profile

yuhuyoyo:
tested the script with a newly created notebook:
1. edit .bash_profile and added a line echo "hello"
2. source ~/.bash_profile and confirmed i can execute it.

```
(base) jupyter@testnotebook1:~$ whoami
jupyter
(base) jupyter@testnotebook1:~$ source ~/.bash_profile
Agent pid 19187
HELLO
```

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -125,6 +125,10 @@ if [[ -n "${TERRA_WORKSPACE}" ]]; then
   sudo -u "${JUPYTER_USER}" sh -c "terra workspace set --id=${TERRA_WORKSPACE}"
 fi
 
+# transfer ownership of .bash_profile to user jupyter.
+chown ${JUPYTER_USER}: /home/${JUPYTER_USER}/.bash_profile
+chmod u+w /home/${JUPYTER_USER}/.bash_profile
+
 # Set variables into the .bash_profile such that they are available
 # to terminals, notebooks, and other tools
 #


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -125,6 +125,10 @@ if [[ -n "${TERRA_WORKSPACE}" ]]; then
   sudo -u "${JUPYTER_USER}" sh -c "terra workspace set --id=${TERRA_WORKSPACE}"
 fi
 
+# transfer ownership of .bash_profile to user jupyter.
```
cbookg, line 128: it seems the problem is that the above code (perhaps line 95) is writing to this file as root.  should we not fix that instead, such that it writes the file correctly to begin with?

yuhuyoyo: I can't write as jupyter without granting it write access tho.

but i don't think it's a problem to edit it as root in this script? as long as the user can edit that file from the instance.

cbookg: I don't quite follow.  Why would the jupyter user not have permission to edit their own .bash_profile?

yuhuyoyo: it's just what the current behavior is

melissachang: +1, this seems a little cleaner. Delete 128-131 and change bash_profile lines to (would need to fix the quotes):

```
sudo -u "${JUPYTER_USER}" sh -c "echo "export OWNER_EMAIL='${OWNER_EMAIL}'" >> "/home/${JUPYTER_USER}/.bash_profile""
```

melissachang (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
yuhuyoyo: i don't understand how that solves the problem. user jupyter doesn't have write access.

(base) jupyter@testnotebook:~$ whoami
jupyter
(base) jupyter@testnotebook:~$ echo "export hello=world" >> "/home/jupyter/.bash_profile"
bash: /home/jupyter/.bash_profile: Permission denied

melissachang: That's because .bash_profile already exists, owned by root. If you change *all* bash_profile lines to run as jupyter, then it will be owned by jupyter.

I tested it, it works:

https://github.com/DataBiosphere/terra-workspace-manager/compare/mc/bash_profile?expand=1

image

yuhuyoyo:
created another PR. closing this one.

PR #886, opened 2022-11-03
[TOAZ-250] LZ is returned when create operation is called, and new return types for delete and create APIs

giventocode:
- APIs that start create and delete jobs for landing zones return new object types with data populated in the response. 
 - The responses include the following, in addition to the job control information:
   - Create operation: landing zone id, requested definition and version. 
   - Delete operation: landing zone id. 
 - The job status APIs return the list of resources when the job is successful for both operations, in addition to the job control information. 

Please note that this is a breaking change in the API contract (response objects).


cahrens (approved):
Thanks @giventocode . I noticed some commented out lines that were accidentally left, but otherwise LGTM.

rtitle (approved):
Looks good, thanks

openapi/src/parts/azure_landing_zone.yaml
```diff
@@ -27,14 +27,14 @@ paths:
     parameters:
       - $ref: '#/components/parameters/JobId'
     get:
-      summary: Get the result of a async job to create an Azure Landing Zone
+      summary: Get the status of a async job to create an Azure Landing Zone
       operationId: getCreateAzureLandingZoneResult
       tags: [ LandingZones ]
       responses:
         '200':
-          $ref: '#/components/responses/AzureLandingZoneResponse'
```
cahrens, line 35: @giventocode I am not seeing this change in Swagger nor when I import the new client (I do see the above change for `CreateLandingZoneResponse`).

image

PR #888, opened 2022-11-04
[TOAZ-222] Add APIs to get, list landing zones

zykonda:


service/src/test/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatchTest.java
```diff
@@ -120,21 +85,110 @@ public void listAzureLandingZoneResourcesByPurpose_Success_NoResults() throws Ex
   }
 
   @Test
-  public void listAzureLandingZoneResourcesByPurpose_ResourcePurpose_Success() throws Exception {
```
zykonda, line 123: Will restore the test:  listAzureLandingZoneResourcesByPurpose_ResourcePurpose_Success() - it is still valid.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -280,4 +283,45 @@ private ApiDeleteAzureLandingZoneJobResult toApiDeleteAzureLandingZoneJobResult(
...
+    ApiAzureLandingZoneList result = new ApiAzureLandingZoneList();
+    List<LandingZone> landingZones =
+        landingZoneService.getLandingZonesByBillingProfile(bearerToken, billingProfileId);
+    if (landingZones.size() > 0) {
+      // The enforced logic is 1:1 relation between Billing Profile and a Landing Zone.
+      // The landing zone service returns one record in the list if landing zone exists
+      // for a given billing profile.
+      result.addLandingzonesItem(toApiAzureLandingZone(landingZones.get(0)));
+    }
+    return result;
```
jordansafer, line 316: Since we are returning either 0 or 1 landing zones, should we return an Optional result instead of a list? I see right now this is sharing the API with listLandingZones, and I do think it saves complexity this way, so I'm ok with that

jordansafer (approved):

openapi/src/parts/azure_landing_zone.yaml
```diff
@@ -101,6 +116,25 @@ paths:
           $ref: '#/components/responses/PermissionDenied'
         '500':
           $ref: '#/components/responses/ServerError'
+    get:
+      parameters:
```
rtitle, line 120: Looks like the `LandingZoneId ` parameter is already defined on L97 so maybe not needed here?

service/src/test/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatchTest.java
zykonda: Also, I will have to merge in changes from TOAZ-221 first to fix the failing tests.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
rtitle: I think using List as the return type is fine -- we enforce 1:1 for now but this may change in the future to a 1:many model.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -280,4 +283,45 @@ private ApiDeleteAzureLandingZoneJobResult toApiDeleteAzureLandingZoneJobResult(
...
+  private ApiAzureLandingZoneList getAzureLandingZonesByBillingProfile(
+      BearerToken bearerToken, UUID billingProfileId) {
+    ApiAzureLandingZoneList result = new ApiAzureLandingZoneList();
+    List<LandingZone> landingZones =
+        landingZoneService.getLandingZonesByBillingProfile(bearerToken, billingProfileId);
+    if (landingZones.size() > 0) {
+      // The enforced logic is 1:1 relation between Billing Profile and a Landing Zone.
+      // The landing zone service returns one record in the list if landing zone exists
+      // for a given billing profile.
+      result.addLandingzonesItem(toApiAzureLandingZone(landingZones.get(0)));
```
rtitle, line 314: What if instead we returned 409 conflict if there are >1 landing zones for a given billing profile? My reasoning is this is an exceptional case which shouldn't happen right now -- better to fail than silently swallow the error.

rtitle (approved):
Couple minor comments, overall looks good.

openapi/src/parts/azure_landing_zone.yaml
zykonda: It actually worked - good to know that parameters are shared despite the syntaxis.

ddietterich: Yes, the resulting `openapi.yaml` file is flat, so anything can reference anything. I had some ideas for scoping access so that we wouldn't end up with a bowl of spaghetti of references, but it got too complicated.

If you have ideas for improving the structure of that source, I'd be interested.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
zykonda: Implemented

PR #889, opened 2022-11-06
[PF-2188] Move building clone resource objects into the resource class hierarchy

ddietterich:
Working on the prototype of the clone refactor I noticed a `switch` statement on resource type. Those should never be needed since we have a class hierarchy based on the resource type.

This work moves most of `WorkspaceCloneUtils` into `WsmResource` and the resource classes. Now whenever we need to build a reference to an existing resource, we can call a method on the existing resource.

I noticed the `ReferencedResource` and `ReferencedResourceService` were under `cloud/gcp`, even though they are independent of cloud, so I moved those.

Also the `ReferencedDataRepoSnapshot` was under `cloud/gcp`. I moved it under `cloud/any`.

I expect to refactor the controlled resources as part of the clone refactor, but those require a bigger change; not a simple code rearrangement.


service/src/main/java/bio/terra/workspace/service/resource/model/WsmResource.java
```diff
@@ -195,6 +200,88 @@ public ImmutableMap<String, String> getProperties() {
...
+      @Nullable UUID destinationResourceId,
+      @Nullable UUID destinationFolderId,
+      @Nullable String name,
+      @Nullable String description) {
+    throw new CloneInstructionNotSupportedException(
+        String.format(
+            "You cannot make a reference clone of a %s resource", getResourceType().name()));
+  }
+
+  protected WsmResourceFields buildCloneResourceCommonFields(
```
melissachang, line 229: This is only used for referenced dest resources (not controlled dest resources), right?

If so, put `referenced` somewhere in name.

Line 242-243 only makes sense for referenced dest resource. (For controlled dest resource, it would change COPY_RESOURCE to COPY_REFERENCE.)

service/src/main/java/bio/terra/workspace/service/resource/controlled/model/ControlledResource.java
```diff
@@ -201,6 +201,38 @@ public void validate() {
     }
   }
 
+  public ControlledResourceFields buildControlledResourceCommonFields(
```
melissachang, line 204: Put `clone` in method same somewhere?

service/src/main/java/bio/terra/workspace/service/resource/referenced/cloud/any/datareposnapshot/ReferencedDataRepoSnapshotResource.java
```diff
@@ -164,6 +165,25 @@ public boolean checkAccess(FlightBeanBag context, AuthenticatedUserRequest userR
...
   }
 
+  @Override
+  public WsmResource buildReferencedClone(
+      UUID destinationWorkspaceUuid,
+      @org.jetbrains.annotations.Nullable UUID destinationResourceId,
+      @org.jetbrains.annotations.Nullable UUID destinationFolderId,
+      @org.jetbrains.annotations.Nullable String name,
+      @org.jetbrains.annotations.Nullable String description) {
+    ReferencedDataRepoSnapshotResource.Builder resultBuilder =
```
melissachang, line 175: 176-184 look the same for all resource types, should this go in a common place/super class?

melissachang (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/bqdataset/ControlledBigQueryDatasetResource.java
```diff
@@ -153,6 +157,30 @@ public ApiGcpBigQueryDatasetResource toApiResource() {
         .attributes(toApiAttributes());
   }
 
+  @Override
+  public WsmResource buildReferencedClone(
+      UUID destinationWorkspaceUuid,
+      @Nullable UUID destinationResourceId,
```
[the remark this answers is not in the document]
ddietterich: Good catch! And oddly half of the Nullable came from jetbrains.

service/src/main/java/bio/terra/workspace/service/resource/referenced/cloud/any/datareposnapshot/ReferencedDataRepoSnapshotResource.java
ddietterich: It is not. Someone wrote these `toBuilder` classes that hide the resource-specific settings.
I am not a fan of that approach. Look here: https://github.com/DataBiosphere/terra-workspace-manager/blob/63f703fa396948e4d79cef44cf93aac9bcdcded0/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gcsbucket/ControlledGcsBucketResource.java#L151

That is my preference and it does not require a `toBuilder`. It constructs the new object.

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtilsTest.java
```diff
@@ -316,33 +311,4 @@ private static void assertControlledResourceCommonField(
     assertEquals(sourceResource.getAccessScope(), resourceToClone.getAccessScope());
     assertEquals(sourceResource.getAssignedUser(), resourceToClone.getAssignedUser());
   }
-
-  @Test
-  public void createDestinationResourceLineage_sourceLineageIsNull() {
```
yuhuyoyo, line 321: can we preserve the test in the wsmResource test?

yuhuyoyo (approved):

ddietterich:
Oops! Sorry! Missed @yuhuyoyo comments. I am posting up another PR today - follow on to this. I will make those updates there.

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResource.java
```diff
@@ -122,7 +127,7 @@ public CloningInstructions getCloningInstructions() {
     return cloningInstructions;
   }
 
-  public List<ResourceLineageEntry> getResourceLineage() {
+  public @Nullable List<ResourceLineageEntry> getResourceLineage() {
```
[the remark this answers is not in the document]
ddietterich: In PF-2196, I removed @Nullable from the getter and from the member declaration. By inspection, it looks like all constructor paths setup an empty array if the input is null.

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtilsTest.java
ddietterich: In PF-2196 I added a test to construct the resource object by two paths with null and non-null lineage. I think that covers the cases.

PR #891, opened 2022-11-08
[WOR-591] Build azure blob copying for cloning operations

aherbst-broad:
## Context
In [this PR](https://github.com/DataBiosphere/terra-workspace-manager/commits/main), we built out the Azure container cloning flight, stubbing out the copying of blobs for a separate effort. This PR builds out the blob copying step.

## This PR
* Builds out the `CopyAzureStorageContainerBlobsStep`. The step delegates mostly to a `BlobCopier` class that handles the mechanics of kicking off the blob copies and polling on their completion.
* The `BlobCopier` starts the copy operations and polls on them to completion. For now, this is a serial operation. In the future we may consider a more sophisticated fork-join model similar to what Rawls does but I'd like to vet out this approach before building out more complex multithreaded code. 
* Adds a connected test that:
   * Creates two containers, uploads small test blobs to them, initiates a copy and verifies they were successfully copied.


service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -257,39 +256,16 @@ public ResponseEntity<ApiCreateControlledAzureRelayNamespaceResult> createAzureR
         SamRethrow.onInterrupted(
             () -> getSamService().getUserEmailFromSam(userRequest), "getUserEmailFromSam");
 
-    logger.info(
```
aherbst-broad, line 260: This has been refactored into the service-level SAS token generation method so we have centralized logging (i.e, I want to log when the blob copy flight mints a token as well and not have to dupe logging statements at that call site).

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -96,6 +103,44 @@ private BlobContainerSasPermission getSasTokenPermissions(
...
+   * @param workspaceUuid id of the workspace that owns the container resource
+   * @param storageContainerResource storage container object we want to mint a SAS for
+   * @param storageAccountResource storage account which owns the storage container object
+   * @param userRequest The authenticated user's request
+   * @param sasIpRange (optional) IP address or range of IPs from which the Azure APIs will accept
+   *     requests for the token being minted.
+   * @return A bundle of 1) a full Azure SAS URL, including the storage account hostname and 2) the
+   *     token query param fragment
+   */
+  public AzureSasBundle createAzureStorageContainerSasToken(
```
aherbst-broad, line 118: I've added a convenience overload for creating a sas token with the default start + expiry times to avoid having to wire in the azure configuration to the blob copying code and passing in the config'ed values there.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
```diff
@@ -0,0 +1,109 @@
...
+import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BlobCopier {
+  private final Logger logger = LoggerFactory.getLogger(BlobCopier.class);
+  private final AzureStorageAccessService storageAccessService;
+  private final AuthenticatedUserRequest userRequest;
+
+  private static final Duration MAX_POLL_TIMEOUT = Duration.ofMinutes(10);
```
aherbst-broad, line 22: We can tweak this as needed, 10 minutes seems generous but I'm happy to bump it up / down if folks have thoughts.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
```diff
@@ -0,0 +1,109 @@
...
+                    sourceBlobContainerClient,
+                    destinationBlobCOntainerClient));
+
+    logger.info(
+        "Copying blobs [source_container_id = {}, source_workspace_id = {}, destination_container_id = {}, destination_workspace_id={}]",
+        sourceContainer.getResourceId(),
+        sourceContainer.getWorkspaceId(),
+        destinationContainer.getResourceId(),
+        destinationContainer.getWorkspaceId());
+    var pollResults = blobPollers.map(blobPoller -> blobPoller.waitForCompletion(MAX_POLL_TIMEOUT));
```
aherbst-broad, line 74: Worth restating from the PR description -- this is a serial operation and a potential candidate for optimization by dispatching these operations in parallel via a fork-join pool. I wanted to keep the first cut simple and iterate from there.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
```diff
@@ -0,0 +1,109 @@
...
+            sourceContainer.getWorkspaceId(),
+            sourceContainer,
+            sourceStorageAccount,
+            userRequest,
+            null,
+            sourceBlobItem.getName(),
+            "r");
+    var sourceBlobClient = sourceBlobContainerClient.getBlobClient(sourceBlobItem.getName());
+    var destinationBlobClient = destBlobContainerClient.getBlobClient(sourceBlobItem.getName());
+    var sourceBlobUrl = sourceBlobClient.getBlobUrl() + "?" + sasBundle.sasToken();
```
aherbst-broad, line 105: Concat'ing strings to form a usable source URL is awkward, I was unable to find a better approach to copying a blob from a private container using a SAS in any docs.

blakery (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
```diff
@@ -0,0 +1,109 @@
...
+      ControlledAzureStorageContainerResource sourceContainer,
+      ControlledAzureStorageContainerResource destinationContainer) {
+    var sourceBlobContainerClient =
+        storageAccessService.buildBlobContainerClient(sourceContainer, sourceStorageAccount);
+    var destinationBlobCOntainerClient =
+        storageAccessService.buildBlobContainerClient(
+            destinationContainer, destinationStorageAccount);
+
+    // filter out any zero-length blobs, these are not copyable
+    var blobItems =
```
ddietterich, line 52: This seems like an odd semantic. I would have expected you to create an empty blob on the target in this case. Isn't an empty file significant to some systems?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
ddietterich: Do I understand you are allowing only 10 minutes for copying all of the blobs? (Seems short) Or is it copying a single blob? (Not sure)

On GCP when we copy across regions, the copy times could be very long. And a local copy was not an actual copy - just some GCS backend operation. So that was very fast. Do you know what this copy is doing under the covers?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private SyncPoller<BlobCopyInfo, Void> beginCopyBlob(
+      AzureStorageAccessService storageAccessService,
+      ControlledAzureStorageContainerResource sourceContainer,
+      ControlledAzureStorageResource sourceStorageAccount,
+      AuthenticatedUserRequest userRequest,
+      BlobItem sourceBlobItem,
+      BlobContainerClient sourceBlobContainerClient,
+      BlobContainerClient destBlobContainerClient) {
+    var sasBundle =
```
ddietterich, line 94: I think these types should be spelled out. Our rule of thumb is: only use `var` when the rhs tells you what the type is.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerBlobsStep.java
```diff
@@ -3,13 +3,90 @@
...
+
+    ControlledAzureStorageContainerResource destinationContainer =
+        flightContext
+            .getWorkingMap()
+            .get(
+                WorkspaceFlightMapKeys.ControlledResourceKeys.CLONED_RESOURCE_DEFINITION,
+                ControlledAzureStorageContainerResource.class);
+
+    var results =
+        blobCopier.copyBlobs(
```
ddietterich, line 82: I don't think this is idempotent. If I read the code right, this will create a potentially duplicate set of blob copies. Some of which may fail because they already have been copied by a previous run.

ddietterich (commented):
I think there are idempotency issues with this code.

Other minor comments

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
marctalbott: feels like a reasonable place to start to me

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
aherbst-broad: Directories are presented as zero length blobs in azure, I'll update the comment to that effect.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerBlobsStep.java
```diff
@@ -3,13 +3,90 @@
...
 public class CopyAzureStorageContainerBlobsStep implements Step {
+  private final ResourceDao resourceDao;
+  private final ControlledAzureStorageContainerResource sourceContainer;
+  private final AzureStorageAccessService azureStorageAccessService;
+  private final AuthenticatedUserRequest userRequest;
+  private final BlobCopier blobCopier;
+
+  public CopyAzureStorageContainerBlobsStep(
+      AzureStorageAccessService azureStorageAccessService,
+      ControlledAzureStorageContainerResource sourceContainer,
```
marctalbott, line 29: For my own learning, how do you determine how you're going to pass parameters to a step (ie via `flightContext` vs step constructor)? Just based on whether the parameter is already present in the flight context when you create the step?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerBlobsStep.java
```diff
@@ -3,13 +3,90 @@
...
+    ControlledAzureStorageContainerResource destinationContainer =
+        flightContext
+            .getWorkingMap()
+            .get(
+                WorkspaceFlightMapKeys.ControlledResourceKeys.CLONED_RESOURCE_DEFINITION,
+                ControlledAzureStorageContainerResource.class);
+
+    var results =
+        blobCopier.copyBlobs(
+            sourceStorageAccount, destinationStorageAccount, sourceContainer, destinationContainer);
```
marctalbott, line 83: What happens if WSM needs to rerun this step? Will the blob copier be able to handle re-copying over the same files if some files have already been copied over on a previous run?

aherbst-broad:
@ddietterich that's for a single copy, I expect azure is similar to GCP in this respect wrt backend operations. Benchmarking between two containers cross region just now yielded around ~2Gbps, so max of 1.2TB in the happy case. I'll bump the timeout.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerBlobsStep.java
aherbst-broad: Yes, the destination blob will be overwritten, see these [remarks](https://learn.microsoft.com/en-us/rest/api/storageservices/copy-blob#remarks).

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerBlobsStep.java
aherbst-broad: Good question. Constructor parameters are only available at flight creation time, i.e., things like spring beans etc. Items created at runtime are passed via the working map.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerBlobsStep.java
aherbst-broad: The blobs will be overwritten (see these [remarks](https://learn.microsoft.com/en-us/rest/api/storageservices/copy-blob#remarks) ).

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
aherbst-broad: Will do.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -123,6 +168,12 @@ public AzureSasBundle createAzureStorageContainerSasToken(
       String sasPermissions) {
     features.azureEnabledCheck();
 
+    logger.info(
+        "user {} requesting SAS token for Azure storage container {} in workspace {}",
+        userRequest.getEmail(),
```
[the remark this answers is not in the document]
aherbst-broad: That's exciting...I'll test it out w/an azure user, thanks for the heads up.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
ddietterich: How do directories get created on the destination? If I have an empty directory in my source, it will disappear in my destination?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerBlobsStep.java
ddietterich: Got it. If there is an interruption, you start over from scratch. We can hope interruptions are very rare!

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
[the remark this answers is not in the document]
aherbst-broad: Yeah, `null` is definitely a thing that can happen. I've added a note to the next WSM checkin to talk about this , it looks like it's endemic to the WSM codebase as a whole. For now I think this is out of scope for this PR.

marctalbott (approved):

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
cahrens: @aherbst-broad it would be good to remove the userEmail variable ^

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -168,4 +224,27 @@ public AzureSasBundle createAzureStorageContainerSasToken(
...
+   * Returns an Azure container client suitable for interacting with a storage container and it's
+   * constiuent blobs.
+   *
+   * @param containerResource The WSM container resource the client will operate on
+   * @param storageAccountResource The parent storage account for the WSM container resource
+   * @return An Azure blob container client
+   */
+  public BlobContainerClient buildBlobContainerClient(
+      ControlledAzureStorageContainerResource containerResource,
+      ControlledAzureStorageResource storageAccountResource) {
```
cahrens, line 238: As we've talked about, we won't have a parent controlled storageAccountResource in the case of a landing zone storage account. It would be good to have a ticket for the subsequent work that will need to be done once TOAZ-264 is settled.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -168,4 +224,27 @@ public AzureSasBundle createAzureStorageContainerSasToken(
             storageContainerResource.getStorageContainerName(),
             token));
   }
+
+  /**
+   * Returns an Azure container client suitable for interacting with a storage container and it's
```
cahrens, line 229: Grammar nit: "its".

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
```diff
@@ -0,0 +1,110 @@
...
+
+  public BlobCopier(
+      AzureStorageAccessService storageAccessService, AuthenticatedUserRequest userRequest) {
+    this.storageAccessService = storageAccessService;
+    this.userRequest = userRequest;
+  }
+
+  /**
+   * Attempts to copy all blobs from the source container to the destination container. The calling
+   * principal must have READ-level access on the storage container resource.
```
cahrens, line 32: READ on source and WRITE on destination?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
ddietterich: Wow, OK.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
```diff
@@ -0,0 +1,110 @@
...
+            sourceStorageAccount,
+            userRequest,
+            null,
+            sourceBlobItem.getName(),
+            "r");
+    var sourceBlobClient = sourceBlobContainerClient.getBlobClient(sourceBlobItem.getName());
+    var destinationBlobClient = destBlobContainerClient.getBlobClient(sourceBlobItem.getName());
+    var sourceBlobUrl = sourceBlobClient.getBlobUrl() + "?" + sasBundle.sasToken();
+
+    return destinationBlobClient.beginCopy(sourceBlobUrl, Duration.ofSeconds(1));
```
cahrens, line 108: Why the 1 second duration?

cahrens:
@aherbst-broad Just a couple minor things and some questions, otherwise LGTM.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
aherbst-broad: No good reason; This controls the interval between poll. According to the docs the default is 1 second anyway, so I'll remove.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
aherbst-broad: 🤔 an interesting question. The user in this context implicitly has WRITE access as they are the owner of the newly cloned workspace.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
aherbst-broad: https://broadworkbench.atlassian.net/browse/WOR-620

cahrens (approved):

PR #892, opened 2022-11-09
PF-2176: Log workspace update policies endpoint

yuhuyoyo:
Only log an activity when the update actually applied.


service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
```diff
@@ -164,7 +167,10 @@ public class MockMvcUtils {
 
   public static MockHttpServletRequestBuilder addAuth(
       MockHttpServletRequestBuilder request, AuthenticatedUserRequest userRequest) {
-    return request.header(AUTH_HEADER, "Bearer " + userRequest.getRequiredToken());
+    return request
+        .header(AUTH_HEADER, "Bearer " + userRequest.getRequiredToken())
+        .header(AuthHeaderKeys.OIDC_CLAIM_EMAIL.getKeyName(), userRequest.getEmail())
+        .header(AuthHeaderKeys.OIDC_CLAIM_USER_ID.getKeyName(), userRequest.getSubjectId());
```
zloery, line 173: I think you need to update `UserAccessUtils`, the AuthenticatedUserRequests coming from that (used for connected tests) don't have subject IDs which is breaking tests.

zloery (approved):

service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
yuhuyoyo: thanks!

mikenev (approved):

PR #893, opened 2022-11-09
PF-1968: make created_by and last_changed_by required field

yuhuyoyo:
For workspaces that are created before the logging exists, there are no entries of them in the log table. So we arbitrarily returns unknown for actor email and a MIN timestamp for the date.


ddietterich (approved):
LGTM - IIUC this will have no effect on new workspaces (e.g. real customers), because they will have their creates logged.

cbookg (approved):

PR #895, opened 2022-11-09
PF-2159: add a post endpoint for delete async.

yuhuyoyo:
@rogerwangcs pointed this out that the rest of the async endpoint are using POST.


cbookg:
This will break the UI though...

yuhuyoyo:
> This will break the UI though...

i can copy paste and wait for UI to switch over?

rogerwangcs (approved):

cbookg:
> > This will break the UI though...
> 
> i can copy paste and wait for UI to switch over?

Yeah, I think we should keep both until the UI can migrate.

ddietterich (approved):

PR #897, opened 2022-11-10
[PF-2196] enable apps on create

ddietterich:
This PR exposes application enable on workspace creation as an upward compatible change to the create workspace API.

The motivation for the change is to simplify workspace cloning. Rather than use separate steps to perform what is a metadata copy, this lets us supply the enabled applications on the workspace create.

I also made some updates from Yu's comments on PF-2188 that I had missed.


service/src/test/java/bio/terra/workspace/service/resource/ResourceLineageUtilsTest.java
```diff
@@ -0,0 +1,127 @@
+package bio.terra.workspace.service.resource;
```
melissachang, line 1: Cool

service/src/test/java/bio/terra/workspace/service/resource/ResourceLineageUtilsTest.java
```diff
@@ -0,0 +1,127 @@
...
+            randomId,
+            resourceName,
+            "description of " + resourceName,
+            CloningInstructions.COPY_NOTHING,
+            "terra",
+            "polaroid",
+            /*resourceLineage=*/ null,
+            propertyMap);
+
+    assertEquals(resource.getResourceLineage(), new ArrayList<>());
```
melissachang, line 42: Technically this should be flipped

assertEquals(new ArrayList<>(), resource.getResourceLineage());

service/src/test/java/bio/terra/workspace/service/resource/ResourceLineageUtilsTest.java
```diff
@@ -0,0 +1,127 @@
...
+    String resourceName = "testdatarepo-" + randomId;
+    Map<String, String> propertyMap = new HashMap<>();
+    var lineageEntry = new ResourceLineageEntry(randomId, randomId);
+    var lineage = new ArrayList<ResourceLineageEntry>();
+    lineage.add(lineageEntry);
+
+    var attributes = new ReferencedDataRepoSnapshotAttributes("terra", "polaroid");
+    String attributesJson = DbSerDes.toJson(attributes);
+
+    var dbResource =
```
melissachang, line 109: opt create helper method createDbResource(resourceLineage)

melissachang (approved):

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -87,11 +91,20 @@ public WorkspaceService(
     this.workspaceActivityLogService = workspaceActivityLogService;
   }
 
+  /** For backward compatibility with test usage */
```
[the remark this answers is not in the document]
melissachang: We have a bunch of createWorkspace test helper methods, maybe the tests should use those

https://cs.github.com/DataBiosphere/terra-workspace-manager?q=%2Fpublic.*create.*workspace.*%5C%28%2F+AND+path%3Atest+AND+NOT+path%3ATest.java

If this is too tedious, someone could do this in a different PR

service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
```diff
@@ -67,6 +67,19 @@ public class ApplicationDao {
...
+  private static final String WORKSPACE_CLONE_QUERY =
+      """
+      SELECT A.application_id FROM enabled_application W, application A
+      WHERE
+        W.workspace_id = :workspace_id AND
+        W.application_id = A.application_id AND
+        A.state = :operating;
+      """;
+
+  private static final RowMapper<String> WORKSPACE_CLONE_ROW_MAPPER =
```
zloery, line 80: nit: APPLICATION_ID_ROW_MAPPER, or something else that indicates what this is mapping

service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
```diff
@@ -179,6 +192,19 @@ public WsmWorkspaceApplication enableWorkspaceApplication(
...
+  /**
+   * Enable a list of applications. Used as part of create workspace
+   *
+   * @param workspaceUuid workspace to enable applications in
+   * @param applicationIds list of application ids
+   */
+  @WriteTransaction
+  public void enableWorkspaceApplications(UUID workspaceUuid, List<String> applicationIds) {
+    for (String applicationId : applicationIds) {
+      enableWorkspaceApplication(workspaceUuid, applicationId);
```
zloery, line 204: How do transactions work if this method is annotated `@WriteTransaction` and `enableWorkspaceApplication` is also `@WriteTransaction`? Are they nested inside one larger transaction, or does each application end up in its own transaction?

zloery (approved):
LGTM pending open comments

service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
```diff
@@ -67,6 +67,19 @@ public class ApplicationDao {
           + "  WHERE W.application_id = A.application_id AND workspace_id = :workspace_id) AS ecount"
           + " FROM application A";
 
+  // Query for finding applications that should be enabled on the target workspace
+  private static final String WORKSPACE_CLONE_QUERY =
```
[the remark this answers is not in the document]
ddietterich: Called it WORKSPACE_APPLICATION_ID_QUERY and WORKSPACE_APPLICATION_ID_ROW_MAPPER.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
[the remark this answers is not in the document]
ddietterich: You're right. I'll take care of it.

service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
ddietterich: `@___Transaction` annotations are set to propagate-always. The transaction manager layer sees the existing outer transaction and uses it instead of starting a new transaction.

PR #904, opened 2022-11-15
[PF-2221] Update github action verions to latest

cbookg:


zloery (approved):

ddietterich (commented):
This looks fine, but it does not resolve our use of the deprecated `save-state` and `set-output`. You can see those warnings [here](https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/3474666845).

If you like, we can merge this and do those changes separately.

cbookg:
> This looks fine, but it does not resolve our use of the deprecated `save-state` and `set-output`. You can see those warnings [here](https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/3474666845).
> 
> If you like, we can merge this and do those changes separately.

will change seperately

PR #905, opened 2022-11-15
[TOAZ-277] Handles exception when the LZ is not found while creating a LZ

giventocode:


zykonda (approved):
LGTM

rtitle (approved):
Looks good, thanks for fixing -- agree returning an empty list in this case would be better than throwing `LandingZoneInvalidInputException`

PR #908, opened 2022-11-16
[PR-2221] Replace deprecated set-output command

cbookg:


.github/actions/bump-skip/action.yml
```diff
@@ -24,5 +24,5 @@ runs:
           IS_BUMP=yes
         fi
         echo "IS_BUMP=$IS_BUMP"
-        echo ::set-output name=is-bump::$IS_BUMP
+        echo is-bump=$IS_BUMP >> $GITHUB_OUTPUT
```
zloery, line 27: Does this need to be quoted to substitute the value? I forget the exact rules for GHA syntax

zloery (approved):

ddietterich (approved):
Thank you!!

.github/actions/bump-skip/action.yml
cbookg: Not sure, so I kept as-is.  But nothing has failed...

cbookg: a bunch of passing tests made use of non-quoted values.

PR #909, opened 2022-11-16
Fix flaky test listWorkspaces_requesterIsOwner_returnsFullWorkspace

melissachang:
Example flake: https://scans.gradle.com/s/grsyqwlai4suw

The test creates a workspace and expects `listWorkspaces()` to return exactly 1 workspace.

Now that tests run in parallel, the test sometimes fails because `listWorkspaces()` returned 2 workspaces. 1 of the workspaces was created by a different test on a different thread.

Changed test to filter `listWorkspaces()` results by workspace ID.

### DIFF
--- a/service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
+++ b/service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
@@ -8,10 +8,10 @@ import static bio.terra.workspace.common.utils.MockMvcUtils.WORKSPACES_V1_BY_UFI
 import static bio.terra.workspace.common.utils.MockMvcUtils.WORKSPACES_V1_BY_UUID_PATH_FORMAT;
 import static bio.terra.workspace.common.utils.MockMvcUtils.WORKSPACES_V1_PATH;
 import static bio.terra.workspace.common.utils.MockMvcUtils.addAuth;
+import static com.google.common.collect.MoreCollectors.toOptional;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.emptyString;
-import static org.hamcrest.Matchers.hasSize;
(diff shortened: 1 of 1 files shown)


service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
```diff
@@ -341,13 +340,21 @@ private void getWorkspaceByUserFacingIdExpectingError(
...
+        request, workspaceId, /*minimumHighestRole=*/ Optional.empty());
   }
 
-  private List<ApiWorkspaceDescription> listWorkspaces(
-      AuthenticatedUserRequest request, Optional<ApiIamRole> minimumHighestRole) throws Exception {
+  /**
+   * Call listWorkspaces and return workspace with specified workspaceId.
+   *
+   * <p>This should be used instead of just calling listWorkspaces. Tests run in parallel.
+   * listWorkspaces will return workspaces created by a different test on a different thread.
```
zloery, line 353: With the current setup, this shouldn't happen unless you're running with a single DB locally. normally, each thread is using a separate postgres DB container. it's definitely possible a different test on the same thread didn't clean up a workspace, though

zloery (approved):
Thanks! This pattern seems a lot more resilient.

I'm not sure the test parallelization is the cause - each thread should be using its own postgres container, unless you're running with a local DB. It definitely could be dependent on test ordering or something else flaky due to tests not cleaning up workspaces though, so this should help a lot!

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
melissachang: I see, good to know, thanks! I looked at build scan more closely. Turns out this is because I forgot to delete a workspace in #906, as [Dan pointed out](http s://github.com/DataBiosphere/terra-workspace-manager/pull/906#discussion_r1024269910). Deleting the workspace should make this problem go away. Closing PR.

PR #912, opened 2022-11-21
[TERRA-208] Add AWS context to WorkspaceDescription

dexamundsen:
Add AWS context to WorkspaceDescription

### DIFF
--- a/openapi/src/parts/controlled_aws_bucket.yaml
+++ b/openapi/src/parts/controlled_aws_bucket.yaml
@@ -112,6 +112,27 @@ components:
           description: A valid bucket location per https://docs.aws.amazon.com/general/latest/gr/s3.html.
           type: string
 
+    AwsBucketDefaultStorageClass:
+      description: >-
+        Type and availability of objects in a bucket, described at https://aws.amazon.com/s3/storage-classes.
+      type: string
+      enum:
+        - STANDARD
+        - INTELLIGENT_TIERING
+        - STANDARD_INFREQUENT_ACCESS
+        - ONE_ZONE_INFREQUENT_ACCESS
+        - GLACIER_INSTANT_RETRIEVAL
+        - GLACIER_FLEXIBLE_RETRIEVAL
+        - GLACIER_DEEP_ARCHIVE
+
+    AwsBucketLifecycleRuleActionType:
+      description: >-
+        Type of action to perform. Formatted like our Java enums.
+      type: string
+      enum:
+        - DELETE
+        - SET_STORAGE_CLASS
+
   responses:
     CreatedControlledAwsBucketResponse:
       description: Response to Create controlled AWS bucket
--- a/openapi/src/parts/controlled_aws_bucket.yaml
+++ b/openapi/src/parts/controlled_aws_bucket.yaml
@@ -112,27 +112,6 @@ components:
           description: A valid bucket location per https://docs.aws.amazon.com/general/latest/gr/s3.html.
           type: string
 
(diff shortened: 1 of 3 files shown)


openapi/src/parts/controlled_aws_bucket.yaml
```diff
@@ -112,6 +112,27 @@ components:
           description: A valid bucket location per https://docs.aws.amazon.com/general/latest/gr/s3.html.
           type: string
```
jmczerk, line 114: We don't yet know what the lifecycle rules are for AWS, so this might be a bit early.  AWS doesn't have a concept of "default storage class", for instance; specifying storage class must be done at object create time, which is "data path" functionality and not intermediated by Terra.

openapi/src/parts/workspace.yaml
```diff
@@ -545,6 +545,9 @@ components:
         azureContext:
           description: Azure context, if one exists
           $ref: '#/components/schemas/AzureContext'
```
jmczerk, line 547: We do not want/need this for AWS.  The AWS context creation code path gets its data from configuration.  We may want to specify a landing zone ID or description later, but we are not there yet.

openapi/src/parts/workspace_cloudcontext.yaml
```diff
@@ -84,6 +93,15 @@ components:
           description: The ID of Azure resource group associated with the workspace.
           type: string
```
jmczerk, line 95: I can get this plumbed through...  Is there anything else we want to include from the stored AWS cloud context metadata?:

https://github.com/DataBiosphere/terra-workspace-manager/blob/jczerk/aws_wlz_interface/service/src/main/java/bio/terra/workspace/service/workspace/model/AwsCloudContext.java#L16

I'm thinking most of it is probably internal for now, and we just want to return something to indicate that we have an AWS context.  Eventually we'd probably want this to be some Workspace Landing Zone identifier.

openapi/src/parts/workspace.yaml
dexamundsen: AWS account number - Added it as a placeholder, but landing zoneId makes more sense here.

openapi/src/parts/workspace_cloudcontext.yaml
dexamundsen: Landing zone Id makes more sense than the account number here. Updating

dexamundsen: in GCP context we display the projectId as part of workspace description. landingZone is unlikely familiar to the user. And they may not have access to the account for login either.

openapi/src/parts/controlled_aws_bucket.yaml
jmczerk: I really think we should leave this out for now, as we cannot meaningfully influence the storage class for AWS buckets in the same way we can for GCP buckets.  We will be able to influence lifecycle, but again this will most likely diverge from what we do with GCP semantically.  Lets just not implement this until we have a clearer picture on what we will do on the lifecycle front.

openapi/src/parts/workspace.yaml
jmczerk: `s/Azure/AWS` in description

openapi/src/parts/workspace_cloudcontext.yaml
jmczerk: 👍

jmczerk (changes requested):

jmczerk (approved):

PR #913, opened 2022-11-21
[PF-2244] Fix NPE when calling Sam createUserV2 with a null body.

jmczerk:
`body` is [now a required parameter](https://github.com/broadinstitute/sam/pull/850/files#diff-f0b79e2460badeacd16512e3022797a9598b99b4d430599411b6f05cf5b07f05R2512) on the Sam `createUserV2` API; it was not required when the currently in-use `sam-client` version was built.  Thus `SamService. registerWsmServiceAccount()` fails with an NPE when calling the API.

This is only ever called once per WSM deployment (to register the WSM SA with Sam), so has been broken for a bit.  I've targeted testing of this code change (calling `createUserV2` with an empty body), please let me know if there's any extra testing we wish to do when bumping the `sam-client` version.


zloery (approved):
LGTM. Sam API changes are (intended to be) backwards compatible and we have pretty good coverage with the connected tests on PRs, so we should have pretty reasonable protection against any other changes in this client upgrade

cbookg (approved):

PR #916, opened 2022-11-21
[PF-2150] Split up Azure connected tests

zloery:
This is a lot of LoC, but it's primarily just moving existing code around. At a high level, this covers the following changes:

- Split the existing `CreateAndDeleteAzureControlledResourceFlightTest` into separate test classes for storage, VMs, and everything else. This class alone took > 1h to run - now that WSM supports parallel test execution, this is the largest driver of slow PR tests. I also pulled a few helper methods out into their own util class since they were used across resource types.
- Merge a number of the storage container / SAS token tests which required the same setup. Previously, each test was creating and destroying a workspace, cloud context, landing zone, storage account, and storage container, which took up the majority of the test time. Combining common setup and cleanup for these tests saves a lot of time.
- Share a single workspace and cloud context across (almost) all tests within a class, with the exception of tests which are doing something non-standard. Most of these tests just need a workspace with an Azure context, so sharing one across tests eliminates unnecessary time spinning up/down.
- For VM tests, share the underlying resources (IP, network, disk) across tests rather than spinning up and tearing down new ones for each test.

With this change, I'm seeing PR times for Azure connected tests going from ~1h15m to ~35m, though one test is still disabled until `TOAZ-286` is resolved.


yuhuyoyo (approved):
thanks Zach!

melissachang:
Ping for review @aherbst-broad @sergiygetlin . After this PR, tests will take 35 mins instead of 1 hr 15 mins. :)

aherbst-broad (approved):

PR #921, opened 2022-11-23
Randomize storage account name to avoid conflicts

sergiygetlin:
Avoid name conflicts for Azure storage account name during parallel test run


rtitle (approved):

giventocode (approved):
LGTM

PR #923, opened 2022-11-28
PF-2260: undo should read WORKSPACE_ID value

yuhuyoyo:
DESTINATION_WORKSPACE_ID value is never set for CloneGcpWorkspaceFlight. So if we undo here, we will throws exception.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneAllFoldersStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneAllFoldersStep.java
@@ -1,5 +1,7 @@
 package bio.terra.workspace.service.resource.controlled.flight.clone.workspace;
 
+import static bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.WORKSPACE_ID;
+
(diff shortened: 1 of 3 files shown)


ginaygoogle (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneAllFoldersStep.java
```diff
@@ -94,9 +96,7 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
   @Override
   public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
     var destinationWorkspaceId =
-        flightContext
-            .getInputParameters()
-            .get(ControlledResourceKeys.DESTINATION_WORKSPACE_ID, UUID.class);
+        FlightUtils.getRequired(flightContext.getInputParameters(), WORKSPACE_ID, UUID.class);
```
melissachang, line 99: opt nit It's a bit strange that `doStep()` gets destinationWorkspaceId differently. Can you change `doStep()` to this?

melissachang (approved):

PR #925, opened 2022-11-29
[TOAZ 271] LZ resource quota API

giventocode:
-  Adds API to get quota information for landing zone resources.
 - Initial support for batch accounts only.


rtitle (approved):
Cool - this looks good to me 👍

zykonda (approved):
lgtm

PR #926, opened 2022-11-30
[WOR-620] Azure container cloning landing zone compatibility

aherbst-broad:
## Context
Container cloning is not compatible with Azure Landing Zones. I erroneously assumed that the resource ID we get back from the landing zone API for the destination storage account is a WSM ID; it is actually an azure storage account identifier. 

We need to be able to create the destination container in **either** the landing zone (if it exists), **or** in the destination workspace's storage account.

## This PR
* Delegates the resolution of the storage account to `AzureStorageAccessService#getStorageAccountData`, which I've made public since it is needed here. At the time of blob copying, rather than using the storage account ID as passed in the working map, we'll be using this method to resolve the storage account for the purpose of dispatching the copy requests. 
* Adjusts the `RetrieveDestinationStorageAccountResourceIdStep` to return a `DestinationStorageAccount` object, which encapsulates the semantic of either a Landing Zone storage account or a workspace storage account.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
@@ -78,7 +78,7 @@ public class CopyAzureStorageContainerDefinitionStep implements Step {
 
     var destStorageAccountInfo =
         workingMap.get(
-            WorkspaceFlightMapKeys.ControlledResourceKeys.DESTINATION_STORAGE_ACCOUNT_RESOURCE_ID,
+            WorkspaceFlightMapKeys.ControlledResourceKeys.DESTINATION_STORAGE_ACCOUNT_ID,
             DestinationStorageAccount.class);
 
     ControlledAzureStorageContainerResource destinationContainerResource =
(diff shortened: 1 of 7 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/DestinationStorageAccount.java
```diff
@@ -0,0 +1,16 @@
+package bio.terra.workspace.service.resource.controlled.flight.clone.azure.container;
+
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+/**
+ * Represents a destination storage account for the purposes of cloning.
```
aherbst-broad, line 7: This is an attempt at capturing an `Either` semantic for the destination storage account (i.e,. it's either an LZ storage account or a workspace storage account).

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/WorkspaceCloneUtils.java
```diff
@@ -51,19 +55,22 @@ public static WsmCloneResourceResult flightStatusToCloneResult(
...
-            sourceContainer.buildControlledCloneResourceCommonFields(
-                destinationWorkspaceId, destinationResourceId, null, name, description))
-        .build();
+    var builder =
+        ControlledAzureStorageContainerResource.builder()
+            .storageContainerName(cloudInstanceName)
+            .common(
+                sourceContainer.buildControlledCloneResourceCommonFields(
+                    destinationWorkspaceId, destinationResourceId, null, name, description));
+    if (StorageAccountType.WORKSPACE == storageAccountInfo.storageAccountType()) {
```
aherbst-broad, line 70: This is why we need the either semantic; we omit the storage account ID if we're creating the destination container in the landing zone.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java
```diff
@@ -79,27 +75,24 @@ public StepResult doStep(FlightContext flightContext)
...
             WorkspaceFlightMapKeys.ControlledResourceKeys.DESTINATION_RESOURCE_ID, UUID.class);
-    var destStorageAccountId =
-        flightContext
-            .getWorkingMap()
-            .get(
-                WorkspaceFlightMapKeys.ControlledResourceKeys
-                    .DESTINATION_STORAGE_ACCOUNT_RESOURCE_ID,
-                UUID.class);
+
+    var destStorageAccountInfo =
```
aherbst-broad, line 79: This is now the storage account object that contains either the landing zone resource ID or the storage account ID.

PR #929, opened 2022-11-30
[WOR-620] Azure container cloning landing zone compatibility 

aherbst-broad:
(note this is take 2 on this PR)

## Context
Container cloning is not compatible with Azure Landing Zones. I erroneously assumed that the resource ID we get back from the landing zone API for the destination storage account is a WSM ID; it is actually an azure storage account identifier.

A previous cut at this ticket had added backwards compatibility with workspace-level storage accounts. The level of complexity introduced in the flight was too high for what we're trying to achieve for Azure public preview. When the need arises, we can introduce support for workspace-level storage account container cloning, but for now we will not support cloning to containers that are not backed by a landing zone storage account.

## This PR
* Removes the `RetrieveDestinationStorageAccountResourceIdStep` step and associated tests, as we will be omitting a storage account ID for the destination container creation 
* Makes the `AzureStorageAccessService#getStorageAccountData` method public, along with the associated `StorageData` type. Blob copying now utilizes this method to get the needed storage account data


marctalbott (approved):

cahrens (approved):
Seems much simpler.

PR #922, opened 2022-11-28
PF-2258: fix flaky test ControlledGcpResourceApiControllerGcsBucketTest

yuhuyoyo:
ControlledGcpResourceApiControllerGcsBucketTest#clone_copyNothing

When I'm running this test locally, the cloning finishes so fast that it returns 200 consistently for me.

PR #833, opened 2022-09-21
[PF-1930] Fix dismal failure with CreateStorageTransferServiceJobStep undo

melissachang:
While working on implementing cloning controlled bucket -> referenced, I wrote a test for CloneControlledGcsBucketResourceFlight undo. (There is no existing test for undo.) It turns out that existing undo fails with dismal exception:

```
2022-09-21 13:33:20.472  ERROR 87902 --- [pool-8-thread-2] bio.terra.stairway.StepResult            : Step Result exception

com.google.api.client.googleapis.json.GoogleJsonResponseException: 400 Bad Request
POST https://storagetransfer.googleapis.com/v1/transferJobs/wsm-owfmSCImSDe8CFS3Aai_8w
{
  "code": 400,
  "errors": [
    {
      "domain": "global",
      "message": "Updating a deleted transfer job is not allowed.",
      "reason": "failedPrecondition"
    }
  ],
  "message": "Updating a deleted transfer job is not allowed.",
  "status": "FAILED_PRECONDITION"
}

DISMAL FAILURE: non-retry-able error during undo. Flight: owfmSCImSDe8CFS3Aai_8w(bio.terra.workspace.service.resource.controlled.flight.clone.bucket.CloneControlledGcsBucketResourceFlight) Step: 5(bio.terra.workspace.service.resource.controlled.flight.clone.bucket.CreateStorageTransferServiceJobStep)
```
Updated CloneControlledGcsBucketResourceFlight undo to 1) prevent dismal failure 2) be idempotent.

PR #914, opened 2022-11-21
PF-2246: add created_by_email and created_date columns

yuhuyoyo:
workspace create
```
Workspace successfully created.
ID:                testworkspace
Name:              
Description:       
Google project:    terra-wsm-t-silky-avocado-4340
Cloud console:     https://console.cloud.google.com/home/dashboard?project=terra-wsm-t-silky-avocado-4340
Properties:
Created:           2022-11-30
Last updated:      2022-11-30
# Resources:       0
```
resources create
```
{
  "resources": [
    {
      "metadata": {
        "workspaceId": "c24ffc1a-513c-4b4c-ac54-63ba25c90859",
        "resourceId": "686889f1-f4ee-4d94-b4a9-bf84ecddb908",
        "name": "mybucket",
        "resourceType": "GCS_BUCKET",
        "stewardshipType": "CONTROLLED",
        "cloudPlatform": "GCP",
        "cloningInstructions": "COPY_RESOURCE",
        "controlledResourceMetadata": {
          "accessScope": "SHARED_ACCESS",
          "managedBy": "USER",
          "privateResourceUser": {},
          "privateResourceState": "NOT_APPLICABLE"
        },
        "resourceLineage": [],
        "properties": [],
        "createdBy": "yuhuyoyo@google.com",
        "createdDate": "2022-11-30T02:03:46.191821Z"
      },
      "resourceAttributes": {
        "gcpGcsBucket": {
          "bucketName": "mybucket-terra-wsm-t-silky-avocado-4340"
        }
      }
    },
    {
      "metadata": {
        "workspaceId": "c24ffc1a-513c-4b4c-ac54-63ba25c90859",
        "resourceId": "8573bf5f-2797-4b5d-95ef-98c3e3ea48f9",
        "name": "mybucketreference",
        "resourceType": "GCS_BUCKET",
        "stewardshipType": "REFERENCED",
        "cloudPlatform": "GCP",
        "cloningInstructions": "COPY_REFERENCE",
        "resourceLineage": [],
        "properties": [],
        "createdBy": "yuhuyoyo@google.com",
        "createdDate": "2022-11-30T02:10:03.190935Z"
      },
      "resourceAttributes": {
        "gcpGcsBucket": {
          "bucketName": "mybucket-terra-wsm-t-silky-avocado-4340"
        }
      }
    },
    {
      "metadata": {
        "workspaceId": "c24ffc1a-513c-4b4c-ac54-63ba25c90859",
[...]

PR #930, opened 2022-12-01
[PF-819] Flesh out tests for cloning controlled GCS bucket

melissachang:
As I refine the tests, I'm keeping BQ dataset tests up-to-date, since those are the canonical tests people may copy from. 

Tests found issues:
- DeleteStorageTransferServiceJobStep failed on retry with `Updating a deleted transfer job is not allowed.`. This is the same error I saw in #833. Filed PF-2271.
- WSM does not allow BQ dataset table lifetime > 24 days. This is because [BQ APIs use Long](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#:~:text=of%20the%20dataset.-,defaultTableExpirationMs,string%20(Int64Value%20format),-Optional.%20The%20default), whereas WSM uses Integer. I fixed this by changing WSM to use Long (in this PR). Filed PF-2267 to update CLI.
- When cloning controlled BQ dataset, defaultTableLifetime/defaultPartition lifetime aren't copied over. Filed PF-2269.

Also:
- Moved `createAiNotebookInstance_correctZone()` out of ControlledGcpResourceApiControllerGcsBucketTest
- Changed `createAiNotebookInstance_correctZone()` to not create a workspace and instead use the one from `setup()`. Changed `createAiNotebookInstance_correctZone()` to delete property at end of test. For local development, it's faster to skip creating workspace. (And it's easier to skip if tests use the workspace created in `setup()`.)
- Refactored #922 to avoid code duplication. Combined success/error back into one method, and instead pass in a list of status codes.


ddietterich (approved):
Wow! Thanks for this.

service/src/test/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiControllerBqDatasetTest.java
```diff
@@ -58,6 +52,14 @@ public class ControlledGcpResourceApiControllerBqDatasetTest extends BaseConnect
   private static final Logger logger =
       LoggerFactory.getLogger(ControlledGcpResourceApiControllerBqDatasetTest.class);
 
+  // GCP default is us-central1. Use something different, so we know this is copied to clone
+  // correctly.
+  private static final String LOCATION = "us-west4";
```
yuhuyoyo, line 57: nit: rename to US_WEST4_LOCATION

yuhuyoyo (approved):

melissachang:
FYI these tests found a bug in #914. getCloneBigQueryDatasetResult() wasn't returning createdDate. I fixed it. :)

Unfortunately I can't merge because createAiNotebookInstanceDo() is failing at HEAD. You can also see the failure here.

image

(That PR didn't cause the bug; the pre-merge tests passed.)

PR #931, opened 2022-12-01
AWS Notebook - Add instanceType, validate instanceId

dexamundsen:
- Add instanceType to AWS Notebook creation parameters / attributes
- Validate instanceId
- Use a common checkFieldNonNull function

### DIFF
--- a/openapi/src/resources/aws_sagemaker_notebook.yaml
+++ b/openapi/src/resources/aws_sagemaker_notebook.yaml
@@ -11,6 +11,9 @@ components:
         region:
           description: Region in which the notebook was created
           type: string
+        instanceType:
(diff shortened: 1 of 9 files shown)


jmczerk (commented):
This looks great, thank you!

One additional thing, if we're going to add instance type as an attribute, we'll need to add it to the OpenAPI resource definition here too:

https://github.com/DataBiosphere/terra-workspace-manager/blob/jczerk/aws_wlz_interface/openapi/src/resources/aws_sagemaker_notebook.yaml#L3

And then make sure that it gets populated to/from all the places where a generated `ApiAwsSageMakerNotebookAttributes` instance is used.

jmczerk (approved):

PR #932, opened 2022-12-01
PF-2261: log change subject id and change subject type

yuhuyoyo:


service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -282,12 +284,10 @@ private ApiWorkspaceDescription buildWorkspaceDescription(
...
         .lastUpdatedDate(
             lastChangeDetailsOptional
-                .map(ActivityLogChangeDetails::getChangeDate)
+                .map(ActivityLogChangeDetails::changeDate)
                 .orElse(OffsetDateTime.MIN))
         .lastUpdatedBy(
-            lastChangeDetailsOptional
-                .map(ActivityLogChangeDetails::getActorEmail)
-                .orElse("unknown"))
+            lastChangeDetailsOptional.map(ActivityLogChangeDetails::actorEmail).orElse("unknown"))
```
melissachang, line 290: First, I thought defaulting to unknown was handled in liquibase.

Second, can we do @Nullable/null instead of "unknown"? Usually, if an optional field isn't set, it's null. Ideally the optional fields would be @Nullable in ActivityLogChnageDetails.java.

service/src/main/java/bio/terra/workspace/service/workspace/model/WsmObjectType.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.workspace.service.workspace.model;
+
+public enum WsmObjectType {
+  UNKNOW,
```
melissachang, line 4: This is mispelled.

// This row was written before change_subject_type column was added
UNKNOWN,

But as I said in another comment, if something isn't set, we usually do null instead of "unknown".

service/src/main/java/bio/terra/workspace/service/workspace/model/WsmObjectType.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.workspace.service.workspace.model;
+
+public enum WsmObjectType {
```
melissachang, line 3: This looks similar to ActivityLogChangedTarget. Can they be combined? Perhaps an enum with methods/properties.

service/src/test/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiControllerBqDatasetTest.java
```diff
@@ -128,10 +132,20 @@ public void create() throws Exception {
     assertEquals(
         expectedBqDataset.getAttributes().getProjectId(),
         actualBqDataset.getAttributes().getProjectId());
+    ActivityLogChangeDetails changeDetails =
```
melissachang, line 135: Let's not put individual field assertions in these methods. I purposely used helper methods to keep these tests short and readable.

Advantages of helper methods:

- Tests are shorter and easier to read (tens vs hundreds of lines). When there are 10 slightly different tests, easier to tell differences apart.
- Instead of adding assertions to dozens of tests, you just add to one helper method.

Let's add these assertions to assertResourceMetadata (which is now in MockMvcUtils.java).

opt Also, please assert the entire row, not just these new columns. I forgot to ask for that in the last PR.

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
```diff
@@ -247,6 +256,59 @@ public void listWorkspaces_requesterIsDiscoverer_requestMinHighestRoleNotSet_ret
...
 
+  @Test
+  public void grantRole_logsAnActivity() throws Exception {
+    mockMvcUtils.grantRole(
+        userAccessUtils.defaultUserAuthRequest(),
+        workspace.getId(),
+        WsmIamRole.DISCOVERER,
+        userAccessUtils.getSecondUserEmail());
+
+    ActivityLogChangeDetails changeDetails = getLastChangeDetails(workspace.getId());
```
melissachang, line 267: Same for test file and WorkspaceApiControllerTest, create helper methods. One of the READMEs refers to these tests, so people may copy-paste from here. In a previous PR, I went through these files and moved stuff to helper methods.

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
```diff
@@ -247,6 +256,59 @@ public void listWorkspaces_requesterIsDiscoverer_requestMinHighestRoleNotSet_ret
...
+    assertEquals(userAccessUtils.getSecondUserEmail(), changeDetails.changeSubjectId());
+    assertEquals(WsmObjectType.USER, changeDetails.changeSubjectType());
+    assertEquals(userAccessUtils.getDefaultUserEmail(), changeDetails.actorEmail());
+    assertEquals(OperationType.REMOVE_WORKSPACE_ROLE, changeDetails.operationType());
+  }
+
+  /**
```
melissachang, line 295: Are we exposing this information in the API? If so, I'd rather test the API. Our controller tests usually call the API.

If not, add comment

```
// We don't expose this in API yet, so read from DB.
// TODO(PF-XXX): Change to call API.
```

service/src/main/java/bio/terra/workspace/service/workspace/model/WsmObjectType.java
yuhuyoyo: why? i tend to avoid nullable fields just because it's easy to forget checking null

melissachang: If I see

```
public method(@Nullable WsmObjectType wsmObjectType) {
```

I know that `wsmObjectType` might be null, and I'll check for it.

If I see

```
public method(WsmObjectType wsmObjectType) {
```

For most enums, this means wsmObjectType is always set. I have to know that for this particular enum "unset means UNKNOWN". We have many enums and this is the only one that has UNKNOWN. I think it's more likely I'll remember the @Nullable case, than the UNKNOWN case. :)

yuhuyoyo: My main interaction with enums were from proto where the default is always the first enum value so I always start with an UNKNOWN. In this case, it doesn't have an apparent benefit.

OTOH, UNKNOWN.name() is still UNKNOWN which i think is a sensible output though. I will change it to null to be consistent

service/src/main/java/bio/terra/workspace/service/workspace/model/WsmObjectType.java
yuhuyoyo: good point :) will remove this class

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
yuhuyoyo: moved it to MockMvcUtil. mirror assertClonedResourceMetadata()

service/src/main/java/bio/terra/workspace/db/WorkspaceActivityLogDao.java
```diff
@@ -27,14 +31,30 @@
...
   private static final Logger logger = LoggerFactory.getLogger(WorkspaceActivityLogDao.class);
-  private static final RowMapper<ActivityLogChangeDetails> ACTIVITY_LOG_CHANGE_DETAILS_ROW_MAPPER =
-      (rs, rowNum) ->
-          new ActivityLogChangeDetails()
-              .changeDate(
-                  OffsetDateTime.ofInstant(
-                      rs.getTimestamp("change_date").toInstant(), ZoneId.of("UTC")))
-              .actorEmail(rs.getString("actor_email"))
-              .actorSubjectId(rs.getString("actor_subject_id"));
+  private static final String DEFAULT_VALUE_UNKNOWN = "unknown";
```
melissachang, line 34: // Only rows that existed before the column was added, will have "unknown"

service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
```diff
@@ -1368,6 +1372,25 @@ public void assertClonedResourceMetadata(
         expectedResourceLineage);
   }
 
+  public static void assertActivityLogChangeDetails(
+      ActivityLogChangeDetails actualChangedDetails,
```
melissachang, line 1376: actualChangeDetails

service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
```diff
@@ -1358,6 +1361,7 @@ public void assertClonedResourceMetadata(
             .sourceWorkspaceId(sourceWorkspaceId)
             .sourceResourceId(sourceResourceId));
 
+    // TODO (PF-2261): assert createdBy, lastUpdatedBy as well.
```
melissachang, line 1364: Can the TODO also include asserting that createdDate and lastUpdatedDate aren't null?

PR #933, opened 2022-12-01
PF-2246: Fix integration test in the verily-src/ repo downstream

yuhuyoyo:
https://github.com/DataBiosphere/terra-workspace-manager/compare/522b0a4f78c7b73c8e7d0dfe139d376786dcf467...b76148ce380f96e9e02c17262c8e7f4bc1cde468

I uncommented some test code that i thought would be harmless to re-enable.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
@@ -26,7 +26,6 @@ import scripts.utils.ClientTestUtils;
 import scripts.utils.WorkspaceApiTestScriptBase;
 
 public class WorkspaceLifecycle extends WorkspaceApiTestScriptBase {
-  private static final String USER_EMAIL = "liam.dragonmaw@test.firecloud.org";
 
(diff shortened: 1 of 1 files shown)


ddietterich (approved):
Thanks for noticing and taking care of this!

integration/src/main/java/scripts/testscripts/WorkspaceLifecycle.java
```diff
@@ -26,7 +26,6 @@
 import scripts.utils.WorkspaceApiTestScriptBase;
 
 public class WorkspaceLifecycle extends WorkspaceApiTestScriptBase {
-  private static final String USER_EMAIL = "liam.dragonmaw@test.firecloud.org";
```
ginaygoogle, line 29: random question need to help me understand. when you remove this hard code test user email. then the test user will not be this person, will get the email from the actual tester. is that correct?

yuhuyoyo: yes, it can be different. test user is different per environment. So the test is failing in verily-src.

PR #938, opened 2022-12-02
[WOR-531] Clone Azure workspaces

marctalbott:
Ticket: [WOR-531](https://broadworkbench.atlassian.net/browse/WOR-531)
* Add support for cloning Azure workspaces and their storage containers into existing cloneWorkspace functionality.


service/src/test/java/bio/terra/workspace/service/workspace/AzureWorkspaceTest.java
```diff
@@ -60,4 +81,64 @@ void createGetDeleteAzureContext() {
     workspaceService.deleteAzureCloudContext(workspace, userRequest);
     assertTrue(azureCloudContextService.getAzureCloudContext(workspace.getWorkspaceId()).isEmpty());
   }
+
+  @Test
```
marctalbott, line 85: I started adding to this test to have it clone a storage container as well but got stuck because there was no landing zone storage account and I don't want to have to wait to spin up a landing zone or a storage account here. Let me know if you have any suggestions, as I'd like to also test that the `cloneWorkspace` flight is correctly spinning out a flight to clone the storage container

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -500,14 +499,10 @@ public ResponseEntity<ApiCreateCloudContextResult> createCloudContext(
...
 
     if (body.getCloudPlatform() == ApiCloudPlatform.AZURE) {
-      ApiAzureContext azureContext =
-          Optional.ofNullable(body.getAzureContext())
-              .orElseThrow(
-                  () ->
-                      new CloudContextRequiredException(
-                          "AzureContext is required when creating an azure cloud context for a workspace"));
+      AzureCloudContext azureContext =
+          Optional.ofNullable(body.getAzureContext()).map(AzureCloudContext::fromApi).orElse(null);
```
cahrens, line 503: Where is the null handled now?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneControlledAzureStorageContainerResourceFlightStep.java
```diff
@@ -0,0 +1,103 @@
...
+          resultMap.get(
+              JobMapKeys.RESPONSE.getKeyName(), ApiClonedControlledAzureStorageContainer.class);
+      cloneDetails.setStewardshipType(StewardshipType.CONTROLLED);
+      cloneDetails.setResourceType(WsmResourceType.CONTROLLED_AZURE_STORAGE_CONTAINER);
+      cloneDetails.setCloningInstructions(resource.getCloningInstructions());
+      cloneDetails.setSourceResourceId(resource.getResourceId());
+      cloneDetails.setDestinationResourceId(
+          Optional.ofNullable(clonedContainer)
+              .map(ApiClonedControlledAzureStorageContainer::getStorageContainer)
+              .map(ApiCreatedControlledAzureStorageContainer::getResourceId)
```
cahrens, line 68: What does this syntax do?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneControlledAzureStorageContainerResourceFlightStep.java
```diff
@@ -0,0 +1,103 @@
...
+          context.getStairway().waitForFlight(subflightId, FLIGHT_POLL_SECONDS, FLIGHT_POLL_CYCLES);
+      WsmResourceCloneDetails cloneDetails = new WsmResourceCloneDetails();
+      WsmCloneResourceResult cloneResult =
+          WorkspaceCloneUtils.flightStatusToCloneResult(subflightState.getFlightStatus(), resource);
+      cloneDetails.setResult(cloneResult);
+
+      FlightMap resultMap = FlightUtils.getResultMapRequired(subflightState);
+      var clonedContainer =
+          resultMap.get(
+              JobMapKeys.RESPONSE.getKeyName(), ApiClonedControlledAzureStorageContainer.class);
```
aherbst-broad, line 60: I think this is the wrong type to cast to in my testing, you'll want `ClonedAzureStorageContainer`

aherbst-broad (commented):
Looking good, had some comments.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -500,14 +499,10 @@ public ResponseEntity<ApiCreateCloudContextResult> createCloudContext(
         workspaceService.validateMcWorkspaceAndAction(userRequest, uuid, SamWorkspaceAction.WRITE);
 
     if (body.getCloudPlatform() == ApiCloudPlatform.AZURE) {
-      ApiAzureContext azureContext =
```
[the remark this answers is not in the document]
marctalbott: Now that `CreateAzureContextFlight` can fetch the cloud context coordinates from BPM, the coordinates don't always need to be passed in the payload, so I don't want this to always blow up if the coordinates are omitted.

However, that flight can only get the coordinates from BPM if `bpm-azure-enabled` is true, so this needs to check that feature flag and potentially throw like it was before. If `bpm-azure-enabled` is false and cloud context coordinates are omitted, this needs to throw.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
marctalbott: The `CreateAzureContextFlight` that `workspaceService.createAzureCloudContext` spins up can handle a null cloud context IFF the `bpm-azure-enabled` feature flag is turned on. I'm going to add a check here so that we throw if `bpm-azure-enabled` is false and the cloud context is null.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneControlledAzureStorageContainerResourceFlightStep.java
marctalbott: It calls the function in the map on the object in the `Optional`. I broke it up into separate lines below in case it's helpful

```
Optional<ApiClonedControlledAzureStorageContainer> apiClonedContainerOption = Optional.ofNullable(clonedContainer)

// call getStorageContainer on the ApiClonedControlledAzureStorageContainer inside of the Optional and return result inside an Optional
Optional<ApiCreatedControlledAzureStorageContainer> apiCreatedContainerOption = apiClonedContainerOption.map(ApiClonedControlledAzureStorageContainer::getStorageContainer)

// call getResourceId on the ApiCreatedControlledAzureStorageContainer inside of the Optional and return result inside an Optional
Optional<UUID> resourceIdOption = apiCreatedContainerOption.map(ApiCreatedControlledAzureStorageContainer::getResourceId)
```

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
cahrens: Sounds good, but how does having bpm allow the cloud context to be null? Is it because it gets the cloud context from the BPM record?

marctalbott: Yeah exactly, WSM fetches the cloud context coordinates from the billing profile record referenced by the workspace's spend profile id

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneControlledAzureStorageContainerResourceFlightStep.java
```diff
@@ -0,0 +1,103 @@
...
+      ControlledAzureStorageContainerResource resource, String subflightId) {
+    this.resource = resource;
+    this.subflightId = subflightId;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    // wait for the flight
+    try {
+      FlightState subflightState =
```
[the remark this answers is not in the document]
marctalbott: Good call, this would only poll for 60 seconds, I'll have it use the same timeout as the blob copier

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCloneControlledAzureStorageContainerResourceFlightStep.java
```diff
@@ -0,0 +1,101 @@
...
+        String.format(
+            "Clone Azure Storage Container %s", sourceResource.getResourceId().toString()));
+    subflightInputParameters.put(
+        ControlledResourceKeys.DESTINATION_CONTAINER_NAME,
+        String.format(
+            "clone-%s-%s", destinationWorkspaceId, sourceResource.getStorageContainerName()));
+    subflightInputParameters.put(
+        ControlledResourceKeys.DESTINATION_RESOURCE_ID, destinationResourceId);
+    subflightInputParameters.put(ControlledResourceKeys.DESTINATION_FOLDER_ID, destinationFolderId);
+    // Do not do the policy merge on the sub-object clone
```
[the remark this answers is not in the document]
marctalbott: Someone correct me if I'm mistaken, but as I understand it, we only want to merge policies when we are cloning a resource into an existing destination workspace (ie [using the cloneContainer API](https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/ControlledAzureResource/cloneAzureStorageContainer)) because there is a possibility that the policies between the two workspaces could conflict. If we are cloning into a newly cloned workspace like we are in this case, then we don't need to worry about merging the policies on the storage container clone because the policies are [merged at the workspace level](https://github.com/DataBiosphere/terra-workspace-manager/blob/a64cc1723c496004454abb034525c5d2946ed984/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneGcpWorkspaceFlight.java#L66) in the `CloneWorkspaceFlight`

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCloneControlledAzureStorageContainerResourceFlightStep.java
```diff
@@ -0,0 +1,101 @@
...
+    } catch (DatabaseOperationException | StairwayExecutionException e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  // Nothing to undo
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    return StepResult.getStepResultSuccess();
```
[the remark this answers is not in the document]
marctalbott: This step is launching the sub-flight which should be able to handle its own rollbacks in the event of a failure. An undo of this step would mean cancelling the subflight it launched, or I suppose it could try to launch a separate `DeleteContainerFlight`. I opted to follow what we do with GCP buckets and BQ datasets, but I don't know if that could leave us vulnerable to orphaned storage container resources.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCreateAzureContextFlightStep.java
```diff
@@ -0,0 +1,84 @@
...
+    boolean flightAlreadyExists;
+    try {
+      context.getStairway().getFlightState(cloudContextJobId);
+      flightAlreadyExists = true;
+    } catch (FlightNotFoundException e) {
+      flightAlreadyExists = false;
+    } catch (DatabaseOperationException e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
+    // if we already have a flight, don't launch another one
```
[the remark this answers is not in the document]
marctalbott: There could be, but I think this check is more about Stairway not repeating itself if it is going through this step for a second time which it might do if it was interrupted the first time. As far as I know, we don't need to worry about multiple WSM instances running the same flight at the same time, just about steps in a flight getting run multiple times in a row

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCreateAzureContextFlightStep.java
```diff
@@ -0,0 +1,84 @@
...
+
+    String cloudContextJobId =
+        context
+            .getWorkingMap()
+            .get(ControlledResourceKeys.CREATE_CLOUD_CONTEXT_FLIGHT_ID, String.class);
+
+    boolean flightAlreadyExists;
+    try {
+      context.getStairway().getFlightState(cloudContextJobId);
+      flightAlreadyExists = true;
```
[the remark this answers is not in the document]
marctalbott: The flight might already exist if this step in the Stairway flight is being re-run. 

I think that would also work, although now I've gotten rid of this step and I'm just reusing the `LaunchCreateGcpContextFlight` to create our Azure context so I'm sort of inclined to leave it alone

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCreateAzureContextFlightStep.java
```diff
@@ -0,0 +1,84 @@
...
+    // if we already have a flight, don't launch another one
+    if (!flightAlreadyExists) {
+      workspaceService.createAzureCloudContext(
+          destinationWorkspace, cloudContextJobId, userRequest, null, null);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  /**
+   * Destroy the created workspace and cloud context. The only time we want to run the workspace
```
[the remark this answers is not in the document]
marctalbott: The cloud context gets deleted as part of `deleteWorkspace`

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCloneControlledAzureStorageContainerResourceFlightStep.java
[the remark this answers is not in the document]
ddietterich: I don't think you get orphans. Even if the clone flight fails later for some reason, the parts that got done stay done. Clone is not guaranteed to be atomic.

ddietterich (changes requested):
One change: support both GCP and Azure cloud context in one workspace.
Otherwise, this LGTM. Faithfully follows the pattern of existing clone. Thanks!

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCloneControlledAzureStorageContainerResourceFlightStep.java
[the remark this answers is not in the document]
cahrens: Thanks for the explanations. Can you update the comment so it includes some of this information? As it is, the comment only reiterates the line of code that follows, without explaining why.

service/src/test/java/bio/terra/workspace/service/workspace/AzureWorkspaceTest.java
```diff
@@ -60,4 +96,110 @@ void createGetDeleteAzureContext() {
...
+            landingZoneDao,
+            workspaceDao,
+            crlService,
+            azureConfig,
+            sourceWorkspace.getWorkspaceId());
+
+    testLandingZoneManager.createLandingZoneWithSharedStorageAccount(
+        landingZoneId, sourceWorkspace.getWorkspaceId(), storageAccountName, "eastus");
+
+    TimeUnit.MINUTES.sleep(1);
```
cahrens, line 146: This  1 minute sleep scares me. Can you poll for the resource to be created?

service/src/test/java/bio/terra/workspace/service/workspace/flight/create/azure/CreateAzureContextFlightTest.java
```diff
@@ -43,8 +43,8 @@ void successCreatesContext() throws Exception {
     workspaceService.createAzureCloudContext(
         workspace,
         jobId,
-        userRequest,
-        /* resultPath */ null,
+        userRequest, /* resultPath */
+        null,
```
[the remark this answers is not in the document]
marctalbott: Yeah, that was an accidental change

service/src/test/java/bio/terra/workspace/service/workspace/AzureWorkspaceTest.java
marctalbott: I pulled this from other tests that used a LZ storage account, so not sure -- I'll look into it

cahrens: Yikes, we can't sprinkle 1 minute sleeps throughout. Yes, please follow up with the author of the other tests.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
```diff
@@ -19,7 +19,7 @@ public class BlobCopier {
   private final AzureStorageAccessService storageAccessService;
   private final AuthenticatedUserRequest userRequest;
 
-  private static final Duration MAX_POLL_TIMEOUT = Duration.ofMinutes(60);
+  public static final Duration MAX_POLL_TIMEOUT = Duration.ofMinutes(60);
```
ddietterich, line 22: nit: maybe make the name more specific if it is to be global

ddietterich (approved):
LGTM

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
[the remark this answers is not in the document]
aherbst-broad: Got it; @ddietterich yeah, I think we can remove the bpm-azure-enabled flag at this point. It was added to facilitate the transition for existing API clients that were creating azure workspaces at the time (i.e., rawls). Now that we're fully shifted over, it [can go](https://broadworkbench.atlassian.net/browse/WOR-677).

@ddietterich BPM is spun up as part of BEEs and it is called [as part of this connected test](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/test/java/bio/terra/workspace/service/spendprofile/SpendProfileBpmConnectedTest.java). I believe that the test is disabled in certain environments where BPM is not available.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/SetNoOpContainerCloneResponseStep.java
```diff
@@ -21,11 +20,12 @@ public SetNoOpContainerCloneResponseStep(
   @Override
   public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
     var noopResult =
-        new ApiClonedControlledAzureStorageContainer()
-            .storageContainer(null)
-            .sourceWorkspaceId(sourceContainer.getWorkspaceId())
-            .sourceResourceId(sourceContainer.getResourceId())
-            .effectiveCloningInstructions(ApiCloningInstructionsEnum.NOTHING);
+        new ClonedAzureStorageContainer(
```
aherbst-broad, line 23: Good catch 👍

aherbst-broad (approved):

cahrens (approved):

service/src/test/java/bio/terra/workspace/service/workspace/AzureWorkspaceTest.java
cahrens: It would be good to fix the other tests also (perhaps a helper method?) to speed up all the tests.

marctalbott: I added a util and found that the offending tests have been removed in the past couple of days so we're all set

PR #940, opened 2022-12-02
PF-1868: Set NOTEBOOK_DISABLE_ROOT to true when the notebook has a cu…

yuhuyoyo:
…stom image

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
@@ -1,5 +1,6 @@
 package bio.terra.workspace.service.resource.controlled.cloud.gcp.ainotebook;
 
+import static bio.terra.workspace.service.resource.controlled.cloud.gcp.ainotebook.ControlledAiNotebookInstanceResource.NOTEBOOK_DISABLE_ROOT_METADATA_KEY;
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
```diff
@@ -223,6 +219,17 @@ static Instance setFields(
...
     }
+    // Set metadata
+    Map<String, String> metadata = new HashMap<>();
+    if (containerImageParameters != null) {
+      // User needs to run as Jupyter instead of root to have the post-startup.sh run successfully.
+      // This is not a TERRA reserved metadata key, so it is overridable. If the creationParameter
+      // says otherwise, we will set the value to what the user specified.
+      metadata.put(NOTEBOOK_DISABLE_ROOT_METADATA_KEY, "true");
+    }
+    Optional.ofNullable(creationParameters.getMetadata()).ifPresent(metadata::putAll);
```
ddietterich, line 230: What stops a user from setting NOTEBOOK_DISABLE_ROOT_METADATA_KEY="false" in their metadata? I think it would overwrite the setting above and give root access on a custom image.

ddietterich (commented):
I think there is a problem

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStep.java
ddietterich: Why add the defaulting code at all then? Just let the caller specify it if they want to.

Clearly, I don't understand the requirements or the security philosophy of this change.

ddietterich (approved):

zloery (approved):

PR #945, opened 2022-12-06
Update WSM documentation

melissachang:
- Added TOC. (I used [gh-md-toc](https://github.com/ekalinin/github-markdown-toc))
- Implemented the following split between README.md, DEVELOPMENT.md:
  - README.md - things WSM users should know
  - DEVELOPERMENT.md - things WSM developers should know
 
  @ddietterich sounds like you had in mind:
  - README.md - things WSM users + WSM developers should know - no Broad deployment-specific info
  - DEVELOPERMENT.md - Things WSM developers should know - Broad deployment-specific info
 
  I'm happy to undo this change, if you want
- Created a [Tests section](https://github.com/DataBiosphere/terra-workspace-manager/blob/dfef2f133f4a53fc3a6519020e413f83a2d2e185/DEVELOPMENT.md#tests) and moved things from various sections to there
- Wrote a new section [Making tests fast](https://github.com/DataBiosphere/terra-workspace-manager/blob/dfef2f133f4a53fc3a6519020e413f83a2d2e185/DEVELOPMENT.md#making-tests-fast) with 2 tips:

  - Create one workspace/context/resource for entire test
  - For local runs, skip workspace/context creation
         
    @zloery feedback welcome


ddietterich (commented):
This is fine. When we started, I was approaching it with the idea that we would have non-Broad-based contributors. It seems unlikely that will happen anytime soon.

Although I wonder if we should just put everything into README. There really nothing much useful left there.

DEVELOPMENT.md
```diff
@@ -310,6 +333,361 @@ Incompatible changes require incrementing the major version number. In our curre
...
+Connected tests that use mock SamService: Tests don't need to call
+`deleteWorkspace()` because there is no SAM workspace to clean up.
+
+Connected tests that use real SamService: Tests should call `deleteWorkspace()`
+to clean up SAM workspaces. Why not just call `deleteWorkspace()` and not use
+janitor? Janitor is useful in case test fails (or `deleteWorkspace()` fails).
+
+Integration tests: Tests should call `deleteWorkspace()` because integration
+tests don't use janitor. Most tests don't need to worry about this because
+`WorkspaceAllocateTestScriptBase.java` deletes the workspace it creates.
```
zloery, line 689: Could you add a sentence or two about the cases where tests do need to worry about this? In particular we usually have issues with `clone` tests, since they'll create a second workspace but not clean it up

zloery (approved):
The content lgtm, I'll defer to others on how this should be split up between README/DEVELOPMENT.md

DEVELOPMENT.md
melissachang: Done

melissachang (commented):
> Although I wonder if we should just put everything into README. There really nothing much useful left there.

Put everything into README.md. Took a stab at consolidating sections, please take a look.

ddietterich (approved):
LGTM

PR #947, opened 2022-12-06
[TERRA-246] Add awsAccountNumber to SageMakerAttributes

dexamundsen:
Add awsAccountNumber to SageMakerAttributes (open api spec)
Set awsAccountNumber in SageMakerResource return values 

Adding awsAccountNumber inline with google Project Id in GCP. 
LandingZoneId is already part of resource.metadata as workspaceId

### DIFF
--- a/openapi/src/common/parameters.yaml
+++ b/openapi/src/common/parameters.yaml
@@ -184,7 +184,7 @@ components:
     WorkspaceId:
       name: workspaceId
       in: path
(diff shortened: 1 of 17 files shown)


jmczerk (commented):
This looks OK, but why do we need it?

dexamundsen:
Closing this - the fields are not needed at this time

PR #948, opened 2022-12-06
TOAZ-227 azure vm logging

dvoet:
This PR enables logging for VMs in a workspace in azure to the landing zone's log analytics workspace via the landing zone's data collection rules. In the process the `AzureMonitorLinuxAgent` needs to be enabled on the VM using either a User or System assigned managed identity. If a UAMI is not in use a SAMI will be enabled.

Also fixed some exception handling/retry problems


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/AzureVmHelper.java
```diff
@@ -52,8 +54,7 @@ public static StepResult deleteVm(
   public static StepResult deleteNetworkInterface(
       AzureCloudContext azureCloudContext,
       ComputeManager computeManager,
-      String networkInterfaceName)
-      throws InterruptedException {
```
dvoet, line 56: handle `InterruptedException` internally to this function because I want to call it from a lambda which doesn't do checked exceptions well

service/src/main/java/bio/terra/workspace/common/utils/ManagementExceptionUtils.java
```diff
@@ -22,4 +25,19 @@ public class ManagementExceptionUtils {
   public static boolean isExceptionCode(ManagementException ex, String exceptionCode) {
     return StringUtils.equals(ex.getValue().getCode(), exceptionCode);
   }
+
+  public static Optional<HttpStatus> getHttpStatus(ManagementException me) {
+    try {
+      return Optional.of(HttpStatus.valueOf(me.getResponse().getStatusCode()));
+    } catch (Throwable t) {
```
aherbst-broad, line 32: Nit: Catching throwable here seems overly broad, I don't think there is anything that can reasonably done for `OutOfMemoryError` for example.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/AzureVmHelper.java
```diff
@@ -84,7 +85,12 @@ public static StepResult deleteNetworkInterface(
         // /subscriptions/3efc5bdf-be0e-44e7-b1d7-c08931e3c16c/resourceGroups/mrg-terra-integration-test-20211118/providers/Microsoft.Compute/virtualMachines/az-vm-b606ad7d-9b00-463d-9b6e-d70586d17eb2",
         //          "details": []
         // }
-        TimeUnit.SECONDS.sleep(NIC_RESERVED_FOR_ANOTHER_VM_ERROR_RETRY_SECONDS);
+        try {
+          TimeUnit.SECONDS.sleep(NIC_RESERVED_FOR_ANOTHER_VM_ERROR_RETRY_SECONDS);
+        } catch (InterruptedException ex) {
+          Thread.currentThread().interrupt();
+          return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
```
aherbst-broad, line 92: Won't this fail the flight and prevent retrying?

aherbst-broad (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/EnableVmLoggingStep.java
```diff
@@ -0,0 +1,187 @@
...
+  private void addMonitorAgentToVm(FlightContext context, String vmId) {
+    var virtualMachine = getComputeManager(context).virtualMachines().getById(vmId);
+
+    var extension =
+        virtualMachine
+            .update()
+            .defineNewExtension("AzureMonitorLinuxAgent")
+            .withPublisher("Microsoft.Azure.Monitor")
+            .withType("AzureMonitorLinuxAgent")
+            .withVersion("1.22")
```
giventocode, line 96: Is there a better way to keep the version information? At the very least a constant, or perhaps a configuration entry if there's a straightforward way to do it.

service/src/main/java/bio/terra/workspace/common/utils/ManagementExceptionUtils.java
dvoet: I am wary of NPE, I could just catch runtime

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/EnableVmLoggingStep.java
```diff
@@ -0,0 +1,187 @@
...
+                    "authentication",
+                    Map.of(
+                        "managedIdentity",
+                        Map.of("identifier-name", "mi_res_id", "identifier-value", petId)))),
+            (() -> enableSystemAssignedIdentity(virtualMachine)));
+
+    extension.attach().apply();
+  }
+
+  private void enableSystemAssignedIdentity(VirtualMachine virtualMachine) {
```
giventocode, line 113: With this change are we now saying that the default is system assigned identity for all VMs?  What is the case/scenario where they won't have a user assigned identity?

giventocode (approved):
LGTM, left a comment seeking further clarification on the system assigned identity scenarios.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/EnableVmLoggingStep.java
dvoet: In reality I don't know a use case when there is no user assigned identity. The code supports it however.

dvoet:
CloneWorkspace is a flakey test, ignoring the failure

PR #946, opened 2022-12-06
[PF-819] Flesh out tests for cloning referenced BQ dataset

melissachang:
Tests uncovered issue:

- We allowed cloning referenced resource where request had COPY_DEFINITION, COPY_RESOURCE. Changed to throw 400.

Also:

- Increased # retries for `createAiNotebookInstanceDo()`. After Dan's fix (#937), test still flaked after exhausting all ten retries ([build scan](https://scans.gradle.com/s/ymg4fcqexjy46/tests/:service:connectedTest/bio.terra.workspace.service.resource.controlled.ControlledResourceServiceTest/createAiNotebookInstanceDo()?top-execution=1)).

PR #950, opened 2022-12-07
Fix various test failures/flakiness

melissachang:
Something is happening with GCP, where permissions are taking longer to propagate than normal. Added retries.

#### createAiNotebookInstanceDo

Increasing timeout from 5 min -> 10 min (#946) helped, but did not totally eliminate flakiness. Test user Elijah had many workspaces. @zloery cleaned up those workspaces. However, this test still flaked. So increase to 15 min.

#### ControlledGcpResourceApiControllerBqDatasetTest

Apologies for this failing so much! The failure only happens when `setup()` creates a workspace. For local development, I comment out workspace creation.

Was able to repro by putting back workspace creation in setup(). Confirmed test passes now.

Test created a controlled BQ dataset. Test then tries to write to dataset, but gets `Access Denied: Project terra-wsm-t-tepid-berry-7944: User does not have bigquery.jobs.create permission in project terra-wsm-t-tepid-berry-7944`.
 I added a retry to test. I think proper fix is @yuhuyoyo 's suggestion to have flight block on permission propagation.

#### RemoveUser, ControlledBigQueryDatasetLifecycle integration tests

Similar, added retries

### DIFF
--- a/service/src/test/java/bio/terra/workspace/common/GcpCloudUtils.java
+++ b/service/src/test/java/bio/terra/workspace/common/GcpCloudUtils.java
@@ -81,11 +81,12 @@ public class GcpCloudUtils {
     // Add row to table
     // Retry because if project was created recently, it takes time for bigquery.jobs.create to
     // propagate
-    for (int i = 0; i < 10; i++) {
+    int retryCount = 10;
+    int retryWaitSeconds = 5;
+    for (int i = 0; i < retryCount; i++) {
       try {
         // Don't call insertAll() with InsertAllRequest. That inserts via stream. Stream buffer may
-        // not
-        // be copied for up to 90 minutes:
+        // not be copied for up to 90 minutes:
         // https://cloud.google.com/bigquery/docs/streaming-data-into-bigquery#dataavailability
         // Instead, use DDL to insert rows.
         bigQueryClient.query(
@@ -100,7 +101,7 @@ public class GcpCloudUtils {
         throw e;
       }
 
-      TimeUnit.SECONDS.sleep(5);
+      TimeUnit.SECONDS.sleep(retryWaitSeconds);
     }
   }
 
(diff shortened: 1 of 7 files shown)


zloery (approved):

service/src/test/java/bio/terra/workspace/common/GcpCloudUtils.java
```diff
@@ -77,15 +79,29 @@ public void populateBqTable(GoogleCredentials userCredential, String projectId,
...
-    // https://cloud.google.com/bigquery/docs/streaming-data-into-bigquery#dataavailability
-    // Instead, use DDL to insert rows.
-    bigQueryClient.query(
-        QueryJobConfiguration.newBuilder(
-                "INSERT INTO `%s.%s.%s` (employee_id) VALUES(%s)"
-                    .formatted(projectId, datasetId, BQ_EMPLOYEE_TABLE_NAME, BQ_EMPLOYEE_ID))
-            .build());
+    // Retry because if project was created recently, it takes time for bigquery.jobs.create to
+    // propagate
+    for (int i = 0; i < 10; i++) {
```
ddietterich, line 84: nit: small preference for putting the retry values into variables so they are easier to modify.

ddietterich (approved):
One nit. LGTM

melissachang:
I was wrong about DRS. I saw RemoveUser and thought DRS. Turns out it was the same as ControlledGcpResourceApiControllerBqDatasetTest, unable to create BQ dataset.

I fixed some integration tests. CloneWorkspace might still be failing. I'll take a look tomorrow.

ddietterich (approved):

yuhuyoyo (approved):

zloery (approved):

melissachang:
FYI CloneWorkspace integration test is failing on and off. I updated PR description with details.

PR #953, opened 2022-12-08
[DDO-2520] Report WSM versions to Broad DevOps "the new way"

jack-r-warren:
Remove usage of the old update-service from terra-helmfile and instead use our new tooling directly, courtesy of the "client" reusable workflows from Sherlock's repo.

The first one just tells us that the version exists, and lets us ingest a bunch of info about the git branch and commit. This is super useful for BEEs and our mechanisms to generate changelogs.

The second one puts the version into our dev environment. The improvement here is that it actually will directly sync the dev environment (which didn't happen in the old mechanism, the version would only get applied overnight). If WSM fails to become healthy after the change, this action will error--should help catch issues sooner.

### DIFF
--- a/.github/workflows/tag-publish.yml
+++ b/.github/workflows/tag-publish.yml
@@ -49,6 +49,9 @@ env:
 jobs:
   tag-publish-job:
     runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.tag.outputs.tag }}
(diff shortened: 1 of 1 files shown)


mflinn-broad (approved):
lgtm

ddietterich (approved):
lgtm

PR #955, opened 2022-12-08
[PF-819] Flesh out tests for cloning referenced BQ table, GCS bucket/object, git repo

melissachang:


service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -653,6 +653,11 @@ public ResponseEntity<Void> deleteDataRepoSnapshotReference(UUID workspaceUuid,
   public ResponseEntity<ApiCloneReferencedGcpGcsObjectResourceResult> cloneGcpGcsObjectReference(
       UUID workspaceUuid, UUID resourceId, @Valid ApiCloneReferencedResourceRequestBody body) {
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    if (body.getCloningInstructions() != null) {
```
ddietterich, line 656: I think it would be better to keep the permission check first and then do the other validation.
It doesn't matter for security. I'd rather show "forbidden" first, rather than "bad request".

ddietterich (approved):
One suggestion, otherwise, LGTM

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
melissachang: Done, for all resource types

melissachang (commented):
Ping @yuhuyoyo

yuhuyoyo (approved):

PR #956, opened 2022-12-09
PF-2326: add better logging when notebookUtils fails to get proxy url and retry

yuhuyoyo:


zloery (approved):

integration/src/main/java/scripts/utils/NotebookUtils.java
```diff
@@ -196,16 +199,19 @@ public static void assertInstanceHasProxyUrl(
...
                 // any other IO exception.
               } catch (GoogleJsonResponseException e) {
                 if (e.getStatusCode() == HttpStatus.SC_FORBIDDEN) {
+                  logger.info("Fails to fetch notebook proxy url due to 403, retry", e);
                   throw e;
                 }
               } catch (IOException ignored) {
+                logger.info("Fails to fetch notebook proxy url, do not retry", ignored);
               }
               return null;
```
ddietterich, line 216: There are two cases that are not logged with this code change:
1. When the GoogleJsonResponseException is not Forbidden
2. When the exception is not GoogleJsonReponseException or IOException
I would like logging of those cases.

ddietterich (changes requested):
I want the unexpected exceptions logged

integration/src/main/java/scripts/utils/NotebookUtils.java
ddietterich: Ah, I see. Great, thanks!

ddietterich (approved):

PR #959, opened 2022-12-09
add the `compute.instances.get` for notebook cpu and gpu request in UI

ginaygoogle:
Background: https://broadworkbench.atlassian.net/browse/PF-2276
https://broadworkbench.atlassian.net/browse/PF-2323

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
@@ -106,7 +106,8 @@ public class CloudSyncRoleMapping {
               "artifactregistry.repositories.getIamPolicy",
               "artifactregistry.repositories.setIamPolicy",
               "artifactregistry.repositories.update",
-              "cloudbuild.builds.approve")
+              "cloudbuild.builds.approve",
+              "compute.instances.get")
           .build();
 
   private static final CustomGcpIamRole PROJECT_READER =


service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -106,7 +106,8 @@ public class CloudSyncRoleMapping {
               "artifactregistry.repositories.getIamPolicy",
               "artifactregistry.repositories.setIamPolicy",
               "artifactregistry.repositories.update",
-              "cloudbuild.builds.approve")
+              "cloudbuild.builds.approve",
+              "compute.instances.get")
```
yuhuyoyo, line 110: can you please add it to WRITER? If a user is a writer of the workspace and they creates a notebook, they should be able to see the gpu and CPU number too.

ginaygoogle: Sure. Done. I understand it.  I am just not really sure the reason I add the permission in PROJECT_WRITE, the UI still does not work.

yuhuyoyo: it's equivalent to add it to both PROJECT_WRITER AND PROJECT_OWNER in the console becasue writer is the subset of owner permission 
https://github.com/DataBiosphere/terra-workspace-manager/pull/959/files#diff-4f3b0cebce0b1e5255b6796256d9f390290d1341f1b77be33d217cb48f4a9981R70

zloery: Should this be `WRITER` or `READER`? e.g. we already give readers the `notebooks.instances.get` permission, which seems similar

yuhuyoyo: i was thinking since reader can't create notebook in terra workspace or see other's notebook, this permission doesn't make sense

zloery: they can't view/read the contents of other people's notebooks, but they can see they exist (either from the WSM api or by listing in the cloud console)

yuhuyoyo: fair. let's put it in READER then.

cbookg (approved):

PR #960, opened 2022-12-09
[PF-2075, PF-2280] Use a separate TPS server instead of the library

ddietterich:
This PR moves TPS from an amalgam library to a `policy` service that calls the TPS server.

### Code Changes

1. Rewrote TpsApiDispatch to use the TPS client library. In the process, I realized that the TpsApiDispatch methods were very awkward to use. I changed them up to simplify the calling code. That caused most of the changes, I think
2. Rearranged the OpenApi `tps.yaml` file. For good or bad, WSM API makes use of the TPS structures in its API. Now that they are split, there are two sets of classes: those defined in the WSM API and those defined in the TPS API. I implemented a set of conversions from the TPS to the WSM API form. For now, these are making an exact copy. In the future, they may have skew and then realign. It seems important to keep them independent to allow independent development and release of the components. That removes all of the TPS endpoints from the WSM REST API.
3. Removed the TPS controller
4. Removed the TPS tests; they have moved to TPS and do not need to be run by WSM
5. Removed an (apparently unused) integration test that directly called TPS, and removed the clientUtils. WSM integration tests do not need to directly call TPS.

### Test Environment Changes

I have created a BEE called `wsmtest-eng` that runs all of the components we will need for testing. If you login with your Broad credentials (say a chrome profile), you can look at the BEE [here](https://beehive.dsp-devops.broadinstitute.org/environments/wsmtest-eng). There is a link there to the cloud console for inspecting the logs, if you need to debug. Right now, the only service one we are using is TPS. I made the change in `scripts/write-config.sh` so for local testing (which is on your laptop and github actions) the `wsmtest-eng` BEE TPS server is used and TPS is enabled.

There are more test environment changes to make. When this PR is merged, it will break the nightly test runs. I am thinking of disabling those until we get disposable BEEs from Broad devops that we can use for the nightly runs.

I think BEEs will be very useful for us. More on that topic to come.


ddietterich:
Test failure is `createAiNotebookInstanceDo` :(

service/gradle/dependencies.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
   // Terra deps
   implementation group: "bio.terra", name: "datarepo-client", version: "1.41.0-SNAPSHOT"
   implementation group: "bio.terra", name:"billing-profile-manager-client", version: "0.1.29-SNAPSHOT"
+  implementation group: "bio.terra", name:"terra-policy-client", version:"0.2.20-SNAPSHOT"
```
mikenev, line 18: i see, switching to the client. nice.

integration/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -388,4 +368,16 @@ public static <T> T assertPresent(Optional<T> optional, @Nullable String message
...
+
+  @FunctionalInterface
+  public interface SupplierWithException<T> {
+    T get() throws Exception;
+  }
+
+  /** An interface for an arbitrary workspace operation that throws an {@link ApiException}. */
+  @FunctionalInterface
+  public interface WorkspaceOperation<T> {
+
```
mikenev, line 380: nit: spacing is different in this function vs the one above. I'm curious why spotless is fine with both.

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
```diff
@@ -0,0 +1,211 @@
...
+import javax.ws.rs.client.Client;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.MDC;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Component;
+
+@Component
+public class TpsApiDispatch {
+  // TODO: PF-1714 pass through bearer token in all cases
```
mikenev, line 39: Is this still correct? Seems like we're adding the token in the client on line 64.

mikenev (approved):

service/src/main/java/bio/terra/workspace/app/configuration/external/PolicyServiceConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+/** Configuration for managing connection to Terra Policy Service. * */
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "workspace.policy")
+public class PolicyServiceConfiguration {
+
+  private String basePath;
+  private String clientCredentialFilePath;
+
+  private static final ImmutableList<String> POLICY_SCOPES =
```
yuhuyoyo, line 22: POLICY_SERVICE_ACCOUNT_SCOPE

service/src/main/java/bio/terra/workspace/service/policy/TpsApiConversionUtils.java
```diff
@@ -0,0 +1,141 @@
...
+
+  public static ApiWsmPolicyUpdateResult apiFromTpsUpdateResult(
+      TpsPaoUpdateResult tpsUpdateResult) {
+    return new ApiWsmPolicyUpdateResult()
+        .updateApplied(tpsUpdateResult.isUpdateApplied())
+        .resultingPolicy(apiFromTpsPao(tpsUpdateResult.getResultingPao()))
+        .conflicts(apiFromTpsPaoConflictList(tpsUpdateResult.getConflicts()));
+  }
+
+  public static ApiWsmPolicy apiFromTpsPao(TpsPaoGetResult tpsPao) {
```
yuhuyoyo, line 68: annotate param and return type with @nullable

service/src/main/java/bio/terra/workspace/service/policy/TpsApiConversionUtils.java
```diff
@@ -0,0 +1,141 @@
...
+      additionalData.add(new TpsPolicyPair().key(apiPair.getKey()).value(apiPair.getValue()));
+    }
+
+    return new TpsPolicyInput()
+        .namespace(apiInput.getNamespace())
+        .name(apiInput.getName())
+        .additionalData(additionalData);
+  }
+
+  public static TpsPolicyInputs tpsFromApiTpsPolicyInputs(ApiWsmPolicyInputs apiInputs) {
```
yuhuyoyo, line 123: ditto

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
```diff
@@ -0,0 +1,211 @@
...
+
+  private final FeatureConfiguration features;
+  private final PolicyServiceConfiguration policyServiceConfiguration;
+  private final Client commonHttpClient;
+
+  @Autowired
+  public TpsApiDispatch(
```
yuhuyoyo, line 46: just for my understanding, what's the difference of a service and a dispatch?

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
```diff
@@ -0,0 +1,211 @@
...
+    } catch (IOException e) {
+      throw new PolicyServiceAuthorizationException(
+          String.format(
+              "Error reading or parsing credentials file at %s",
+              policyServiceConfiguration.getClientCredentialFilePath()),
+          e.getCause());
+    }
+  }
+
+  private RuntimeException convertApiException(ApiException ex) {
```
yuhuyoyo, line 82: super nit: put private method at the bottom

service/src/main/java/bio/terra/workspace/service/policy/exception/PolicyServiceAPIException.java
```diff
@@ -0,0 +1,47 @@
+package bio.terra.workspace.service.policy.exception;
+
+import bio.terra.buffer.client.ApiException;
+import bio.terra.common.exception.ErrorReportException;
+import java.util.Collections;
+import java.util.List;
+import org.springframework.http.HttpStatus;
+
+/** Wrapper exception for non-200 responses from calls to Buffer Service. */
```
yuhuyoyo, line 9: do you mean calls to policy service?

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspacePoliciesStep.java
```diff
@@ -38,19 +34,16 @@ public CreateWorkspacePoliciesStep(
...
-            .attributes(
-                new ApiTpsPolicyInputs()
-                    .inputs(policyInputs == null ? null : policyInputs.getInputs()));
     try {
-      tpsApiDispatch.createPao(new BearerToken(userRequest.getRequiredToken()), request);
-    } catch (DuplicateObjectException e) {
+      if (policyInputs == null) {
+        tpsApiDispatch.createEmptyPao(workspace.getWorkspaceId());
+      } else {
+        tpsApiDispatch.createPao(workspace.getWorkspaceId(), policyInputs);
```
yuhuyoyo, line 41: random comment: why don't createPao just accept null policyInput?

yuhuyoyo (approved):

integration/src/main/java/scripts/utils/ClientTestUtils.java
ddietterich: I think the standard is two space tabs. I had previously set mine, but switched IntelliJ versions and didn't reset. It seems we don't have spotless insisting on two space tabs.

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
ddietterich: Removed the comment.

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
ddietterich: No difference. I plan to do a global rename as a separate PR so it would be trivial to review.

service/src/main/java/bio/terra/workspace/service/policy/exception/PolicyServiceAPIException.java
ddietterich: Cut'n'paste strikes again! :)

service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspacePoliciesStep.java
ddietterich: Good idea! Doh!

service/src/main/java/bio/terra/workspace/service/policy/TpsApiConversionUtils.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * The WSM interface uses an identical, but different set of classes for TPS data. This utility
+ * class holds static methods to do extremely simple, if laborious conversions between the two
+ * classes. It is our buffer when TPS API changes and WSM is not in sync.
+ */
+public class TpsApiConversionUtils {
```
zloery, line 33: nit: add an empty private constructor since this is only utils

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
```diff
@@ -0,0 +1,208 @@
...
+        tpsApi.deletePao(workspaceUuid);
+      } catch (ApiException e) {
+        throw convertApiException(e);
+      }
+    } catch (PolicyServiceNotFoundException e) {
+      // Not found is not an error as far as WSM is concerned.
+    }
+  }
+
+  @Traced
```
zloery, line 89: out of curiosity, why is only this endpoint traced?

zloery (approved):
LGTM!

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
ddietterich: Good point. All of the external calls should be traced.

PR #963, opened 2022-12-13
[PF-2341] Recode proxyUrl retry loop

ddietterich:
Moved the retry loop into the NotebookUtil code. This simplifies things and eliminates a bug, since non-exception cases are also retried.

### DIFF
--- a/integration/src/main/java/scripts/utils/NotebookUtils.java
+++ b/integration/src/main/java/scripts/utils/NotebookUtils.java
@@ -1,7 +1,6 @@
 package scripts.utils;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static scripts.utils.CommonResourceFieldsUtil.makeControlledResourceCommonFields;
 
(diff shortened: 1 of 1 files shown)


cbookg (approved):

integration/src/main/java/scripts/utils/NotebookUtils.java
```diff
@@ -186,36 +186,43 @@ public static boolean userHasProxyAccess(
...
-                  throw e;
-                }
-                logger.info("Fails to fetch notebook proxy url, do not retry", e);
-              } catch (Exception ignored) {
-                logger.info("Fails to fetch notebook proxy url, do not retry", ignored);
-              }
-              return null;
-            });
-    assertNotNull(proxyUrl);
+    final int retryMax = 40;
```
yuhuyoyo, line 189: make them constant please

yuhuyoyo (approved):

integration/src/main/java/scripts/utils/NotebookUtils.java
ddietterich: Haha, yes. I'll do that

PR #967, opened 2022-12-14
PF-2347: Log when ControlledResourcesDeleteFlight succeeded too

yuhuyoyo:
When controlledResourceDeleteFilght succeeded, currently we throw because the RESOURCE key in the input map is not set. 

Also fixed a couple test that is not test the right thing.

Clean up the logic a bit:
0. handle flight that modify multiple workspaces
1. handle flight with OperationType == DELETE (SUCCESS OR NOT)
2. handle success flight
3. ignore the rest.

The previous logic misses the scenario when ControlledResourcesDeleteFlight succeeded.


zloery (approved):

service/src/main/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHook.java
```diff
@@ -247,37 +247,37 @@ private void maybeLogFolderDeletion(
...
+                subjectId,
+                OperationType.DELETE,
+                resourceId.toString(),
+                ActivityLogChangedTarget.RESOURCE));
+      }
     }
   }
 
-  private UUID getControlledResourceToDeleteFromFlight(FlightContext context) {
+  private List<UUID> getControlledResourceToDeleteFromFlight(FlightContext context) {
```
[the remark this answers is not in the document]
yuhuyoyo: all the controlled resource deletion takes in a list now https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java#L627.

I made the change because i wanted to reuse the same flight for multi-resource deletion (e.g. a flight to delete a folder).

service/src/main/java/bio/terra/workspace/common/exception/UnhandledActivityLogException.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.workspace.common.exception;
```
ddietterich, line 1: This also showed up in the region PR. Are they stacked?

ddietterich (approved):

service/src/main/java/bio/terra/workspace/common/exception/UnhandledActivityLogException.java
yuhuyoyo: maybe accidentally. It's definitely related to this one tho.

PR #971, opened 2022-12-15
DDO-2528 Add shim to generate fake terra-helmfile versions file

choover-broad:
The information that used to be tracked in terra-helmfile's `versions/` files is now tracked in an API service called Sherlock.

We recently [deleted](https://github.com/broadinstitute/terra-helmfile/commit/07c49af3562e0c3355f2f1185ade4609e655e10f) those files from terra-helmfile, but that seems have broken TestRunner.

This PR adds a new step to WSM's alpha promotion tests that will generate fake versions files based on the information in Sherlock, for backwards compatibility. Ideally TestRunner should be updated to pull information directly from Sherlock.


jack-r-warren (approved):

mflinn-broad (approved):

ichengchang (approved):
Looks good!

PR #972, opened 2022-12-15
PF-2261: check for resource metadata createdBy and createdDate in the test

yuhuyoyo:
Also add test when clone is completed by the second user, destResourceName, destCloudInstanceName and location are *NOT* set to check that those fields are successfully populated.

For example, when bucket is cloned and we don't specify a destResourceName, we generate a name and append cloned- in front to avoid duplication.


ddietterich (approved):
LGTM

zloery (approved):

PR #973, opened 2022-12-15
Revert "[PF-2168] Move from terra-kernel-k8s to broad-dsp-gcr-public"

ddietterich:
Reverts DataBiosphere/terra-workspace-manager#965

### DIFF
--- a/.github/workflows/tag-publish.yml
+++ b/.github/workflows/tag-publish.yml
@@ -44,6 +44,8 @@ on:
         default: 'main'
 env:
   SERVICE_NAME: ${{ github.event.repository.name }}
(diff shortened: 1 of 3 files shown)


zloery (approved):

yuhuyoyo:
do you need a stamp?

ddietterich:
> do you need a stamp?

That would be great. I may close it if Zach gets done with the change over this afternoon. But it would be handy to have two approvals, so if I can merge it if need be.

yuhuyoyo (approved):

zloery:
Got everything merged and dev is now reading from the new GCR

PR #975, opened 2022-12-16
PF-2261: Set lastUpdatedBy and lastUpdatedDate on all wsm resources.

yuhuyoyo:


service/src/main/java/bio/terra/workspace/app/controller/WsmResourceControllerBase.java
```diff
@@ -0,0 +1,26 @@
...
+  public WsmResourceControllerBase(
+      AuthenticatedUserRequestFactory authenticatedUserRequestFactory,
+      HttpServletRequest request,
+      SamService samService,
+      WorkspaceActivityLogService workspaceActivityLogService) {
+    super(authenticatedUserRequestFactory, request, samService);
+    this.workspaceActivityLogService = workspaceActivityLogService;
+  }
+
+  public WsmResourceApiFields getWsmResourceApiFields(UUID workspaceUuid, UUID resourceId) {
```
ddietterich, line 23: Is there a reason this cannot go into `ControllerBase`?

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceApiFields.java
```diff
@@ -0,0 +1,23 @@
...
+import bio.terra.workspace.service.logging.WorkspaceActivityLogService;
+import java.time.OffsetDateTime;
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * Api fields in WsmResource that is not stored as common fields in the WsmResource and needs to be
+ * fetched elsewhere.
+ */
+public record WsmResourceApiFields(String lastUpdatedBy, OffsetDateTime lastUpdatedDate) {
```
ddietterich, line 13: This is awkward. I am thinking we should integrate this into the resource hierarchy rather than have to do this extra action all over the code.

We have a general problem with information that is not stored in the resource row, but is found elsewhere (activity log, cloud platform). I'd like to think of a more general solution than this.

ddietterich (changes requested):
I want to find a more general solution than this PR implements. I don't think this is the best approach.

As I say in a comment, we have an information flow problem in WSM. It happens in two directions:
1) inputs to create and update that are not in the resource object. These are very awkward in the clone code.
2) outputs when retrieving resources that are not in the resource object.

I would like something that integrates into the resource hierarchy to solve these problems.

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceApiFields.java
yuhuyoyo: i just learned that I *can* have a final setter to autowire the ActivityLogservice to `WsmResource`. wdyt?

toApiResource seems OK to me because it's available for all the resource and it's the gateway to convert wsm resource to the api wsm resources. Any alternative that involve computing the non-resource common fields separately is not reliable.

service/src/main/java/bio/terra/workspace/app/controller/WsmResourceControllerBase.java
yuhuyoyo: it's not applicable to workspace controller, application controller, etc.

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceApiFields.java
yuhuyoyo: nvm. I can't inject it into `WsmResource`. Neither can I do it in the `WsmResourceHandler` i dont' think

ddietterich (approved):
A minor comment. Otherwise, LGTM
Thanks for your work on this!

integration/src/main/java/scripts/testscripts/ControlledBigQueryDatasetLifecycle.java
```diff
@@ -277,14 +277,29 @@ private void createControlledDatasetWithBothResourceNameAndDatasetIdSpecified(
...
     ownerResourceApi.deleteBigQueryDataset(
         getWorkspaceId(), createdDatasetWithDifferentDatasetId.getMetadata().getResourceId());
   }
 
+  /**
+   * The dataset at creation can have an earlier LastUpdatedDate than the current dataset. So do not
+   * compare the lastUpdatedDate.
+   */
+  private void assertDatasetsAreEqualIgnoringLastUpdatedDate(
+      GcpBigQueryDatasetResource createdDataset, GcpBigQueryDatasetResource fetchedDataset) {
```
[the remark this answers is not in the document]
yuhuyoyo: good catch

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -1269,7 +1282,8 @@ void updateBqDatasetWithUndefinedExpirationTimes() throws Exception {
         new ApiGcpBigQueryDatasetUpdateParameters()
             .defaultTableLifetime(0L)
             .defaultPartitionLifetime(0L);
-    controlledResourceService.updateBqDataset(resource, updateParameters, null, null);
+    controlledResourceService.updateBqDataset(
+        resource, updateParameters, null, null, userAccessUtils.defaultUserAuthRequest());
```
ginaygoogle, line 1286: nit: maybe could have /* */ for the 2 null parameters.  but not really necessary for the test, maybe

ginaygoogle (approved):
LGTM: thanks for the big work!

PR #979, opened 2022-12-19
PF-2261: Write to log in the flight for cloning in the dest workspace

yuhuyoyo:
I realized that we have not been consistently logging the cloned resource creation in the destination workspace. When cloning the resource as a controlled resource, we call `controlledResourceService#createResourceAsync` so the cloning is logged. but when it's a referenced resource, we directly call into the db to store the referenced resource so the resource is not logged. 

Originally I didn't want to log in the flight as doStep and undoStep will leave garbage traces in the log. But we can't prevent user from calling into service and start a flight in a step that way.  i think "garbage" is better than inconsistency.


yuhuyoyo:
@ddietterich @ginaygoogle I noticed from https://github.com/DataBiosphere/terra-workspace-manager/pull/975 that the log for cloning is inconsistent, thus made this change as a separate PR.

service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
```diff
@@ -1807,6 +1809,46 @@ private ActivityLogChangeDetails getLastChangeDetails(UUID workspaceId, String c
...
+      AuthenticatedUserRequest userRequest)
+      throws InterruptedException {
+    // log in source Workspace
+    ActivityLogChangeDetails sourceChangeDetails =
+        getLastChangeDetails(sourceWorkspaceId, sourceChangeSubjectId.toString());
+    var actorEmail = userRequest.getEmail();
+    var actorSubjectId = samService.getUserStatusInfo(userRequest).getUserSubjectId();
+    assertEquals(
+        new ActivityLogChangeDetails(
+            null,
```
ginaygoogle, line 1828: `/* changeDate */ null`

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneReferencedResourceStep.java
```diff
@@ -24,20 +24,23 @@
 
 public class CloneReferencedResourceStep implements Step {
 
+  private final AuthenticatedUserRequest userRequest;
   private final SamService samService;
-  private final ResourceDao resourceDao;
+  private final ReferencedResourceService referencedResourceService;
```
ginaygoogle, line 29: I am not quite understand why we change the resourceDao to referencedresourceService. how about the controlled  resource?

ddietterich (approved):
LGTM

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneReferencedResourceStep.java
yuhuyoyo: yes this is the meat of the change. the `referencedResourceService` log an activity when creating a referenced resource. Whereas a direct write to the DAO does not log.

This step is called CloneReferencedResourceStep. So it is irrelevant to controlled resource. FWIW, the controlled one is already calling the service.

ginaygoogle (approved):

PR #984, opened 2022-12-20
[no ticket] throw missingRequiredFieldException for getRequired

yuhuyoyo:
to be consistent with validateRequiredEntries and validateCommonEntries.


mikenev (approved):

ddietterich (approved):
Good fix, thanks

PR #992, opened 2023-01-04
[WOR-690] Remove storage account ID arg when creating destination storage container

aherbst-broad:
This regressed as part of this [PR](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStep.java#L106) (my fault, I should have seen this during review).

We do not want to pass a storage account ID during cloning as we do not support non-landing-zone backed workspaces (for cloning). The storage account ID we are passing here is not a workspace manager resource ID but rather a landing zone ID that is tagged in azure. Omitting the storage account ID shifts to using the landing zone's storage account.

Related:
We should fail before dispatching the clone job if the destination does not have a landing zone. This is out of scope for this PR, but I've ticketed [here] for further consideration. (https://broadworkbench.atlassian.net/browse/WOR-695).


service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/container/CopyAzureStorageContainerDefinitionStepTest.java
```diff
@@ -130,9 +133,39 @@ void doStep_clonesDefinition() throws InterruptedException {
...
+    var step =
+        new CopyAzureStorageContainerDefinitionStep(
+            mockSamService(),
+            testUser,
+            sourceContainer,
+            controlledResourceService,
+            CloningInstructions.COPY_DEFINITION);
+
+    var ex = assertThrows(MissingRequiredFieldsException.class, () -> step.doStep(flightContext));
+    assertThat(ex.getMessage(), containsString(ControlledResourceKeys.SHARED_STORAGE_ACCOUNT));
```
aherbst-broad, line 166: Not a big fan of the string matching here, we may want to enhance the missing fields exception to consume an optional field name but I think that's for another PR.

cahrens (approved):

PR #996, opened 2023-01-06
[TOAZ-287] Return correct error message and exception details when storage account validation fails while creating a container

zykonda:
Extended error message for the case when Sam does not have proper setup for the caller's access to Landing Zone resource.

The initial issue was related to an inconsistency in caller's Landing Zone resource setup (Sam) and billing profile access set up.
While the caller had access to a billing profile, same caller did not have proper LIST_PROPERTIES action set up in Sam for a Landing Zone.
Rob has taken action to fix the gap, but in the unlikely event of the reoccurrence, the situation will continue to be difficult to diagnose due to the API contracts design in Terra services. Here is why:

landingZoneService.getLandingZonesByBillingProfile(token, profileId)
is a public API that returns a list of all LZs associated with profile Id. As a public API, it should not expose information about resources that user has no access to. And therefore, the result is filtered by Sam's calls to check resource permitted action for a user (LIST_RESOURCES action must be present).

If user has no required access to a resource, Sam API does not throw exception, nor should Landing Zone service (it is still GET [anything that is found] API).

There is a preceding in the code path, that is checking access to billing account. 
For the issue that raised this bug, the billing profile access check has succeeded by this point, so the situation is not normal.

That is why I chose to extend the error message only to remind the investigation on checking the user's Sam set up.
The chances of surfacing this error should be next to zero when all cases are considered and the gaps addressed. Otherwise, the API contracts need to be changed significantly.


rtitle (approved):

ddietterich:
Why does the caller need to list LZ properties? What is the use case here?
I'm concerned about the direct user access to LZ that is not administrative (CRUD on the LZ itself).

rtitle:
Hm yeah - for CRUD operations and listing LZ resources, the caller needs access to the landing zone. But for internal WSM -> LZ calls, a system identity is used and the end user does _not_ need access to the LZ (at least, after https://github.com/DataBiosphere/terra-workspace-manager/pull/997 is in). On further inspection it looks like this method (`getLandingZoneId`) is just used for the latter case, and thus is not quite accurate?

zykonda:
@ddietterich, @rtitle The error message I'm correcting, should not even be surfaced to end user - as I mentioned, this would be an anomaly that needs to be investigated. Method getLandingZoneId is used in service to service calls (like WSM internal logic). Normal user would not need to access this information, and if user tries to call REST API directly - it will get landing zone Id only in case when user has list_resources access to a specific Landing zone.

aherbst-broad (approved):

PR #997, opened 2023-01-07
[TOAZ-294] Have WSM query LZ service as the WSM SA, not the calling user

rtitle:
Update all _internal_ calls from WSM to `LandingZoneApiDispatch` to use the WSM SA token, instead of the calling user token. The WSM SA is granted `list-resources` permission on all landing zones. The motivation here is that users in various contexts need to be able to create WSM resources without needing full access to the landing zone/billing profile. 

End user tokens are still used for top-level LZ API calls; the scope of this PR is just internal WSM-to-LZ system calls during controlled resource operations.

Tested on a BEE -- I was able to create a VM and Leo-app from the UI using this WSM branch. To migrate I shared pre-existing landing zones with `workspace-dev@broad-dsde-dev.iam.gserviceaccount.com`. I can do similar for dev and persistent BEE environments to not break existing deployments.

Related helmfile PR: https://github.com/broadinstitute/terra-helmfile/pull/3706


service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/CreateAzureDiskStepTest.java
```diff
@@ -32,7 +32,7 @@
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 
-public class CreateAzureDiskStepTest extends BaseAzureConnectedTest {
+public class CreateAzureDiskStepTest extends BaseAzureUnitTest {
```
rtitle, line 35: Made a bunch of Azure step tests unit instead of connected, I think that was the intent and it is just a misconfiguration.

giventocode (approved):
LGTM!

zykonda (approved):
LGTM

aherbst-broad (approved):

ddietterich:
We have good confidence that this code is well tested. No reason to wait for the test runs.

PR #998, opened 2023-01-09
PF-2388: expose getRegionInfo

yuhuyoyo:


yuhuyoyo:
@mikenev

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
```diff
@@ -193,6 +201,26 @@ public List<String> listValidDataCenter(UUID workspaceId, String platform) {
     return new ArrayList<>();
   }
 
+  @Traced
+  @Cacheable(cacheNames = "regionInfo", sync = true)
```
mikenev, line 205: Curious if the tests would also get cached data. Seems like they would. It shouldn't really matter since the data doesn't change that much.

mikenev (approved):

service/src/main/java/bio/terra/workspace/app/configuration/external/PolicyServiceConfiguration.java
```diff
@@ -8,11 +8,13 @@
 import java.io.IOException;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.context.annotation.Configuration;
 
 /** Configuration for managing connection to Terra Policy Service. * */
 @Configuration
 @EnableConfigurationProperties
+@EnableCaching
```
ddietterich, line 17: I am mistrustful of Spring caching. We tried to use it in TDR and it did not work well. We hand-coded instead. It looks like you have not configured a cache provider.

See [here](https://docs.spring.io/spring-boot/docs/2.1.6.RELEASE/reference/html/boot-features-caching.html#boot-features-caching-provider) for choices and how to configure. Also [here](https://docs.spring.io/spring-framework/docs/5.1.8.RELEASE/spring-framework-reference/integration.html#cache) for more info.

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
ddietterich: I wonder if this is a premature optimization. Do we know that we have a performance or load issue? Is WSM the right place to put caching or should it be in TPS? A TPS cache would have the advantage of knowing when to invalidate.

ddietterich (commented):
The code is fine. Two things:
1. I'm not so sure we want caching as implemented and in WSM
2. I wonder if we should get the naming resolved before merging in this new API

service/src/main/java/bio/terra/workspace/app/configuration/external/PolicyServiceConfiguration.java
yuhuyoyo: I'm using spring-boot-starter-cache so I don't need to register a cache provider.

"When using Spring Boot, the mere presence of the starter package on the classpath alongside the EnableCaching annotation would register the same ConcurrentMapCacheManager. So there is no need for a separate bean declaration."

[spring-boot-starter-cache](https://search.maven.org/search?q=g:org.springframework.boot%20a:spring-boot-starter-cache)

You suggested caching in another ticket hence I implemented it. Fine to not cache :)

openapi/src/common/schemas.yaml
```diff
@@ -313,3 +313,10 @@ components:
           $ref: '#/components/schemas/JobReport'
         errorReport:
           $ref: '#/components/schemas/ErrorReport'
+
+    DataCenterList:
```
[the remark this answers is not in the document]
yuhuyoyo: sure!

can't merge it anyway. Would you like me to change it?

ddietterich (approved):

PR #1000, opened 2023-01-11
PF-2172: Integrate regional policy check to controlled resource create

mikenev:
When creating a controlled resource, call TPS to see if the selected region is allowed per policy. If not, fail validation.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledResourceControllerBase.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledResourceControllerBase.java
@@ -80,7 +80,7 @@ public abstract class ControlledResourceControllerBase extends ControllerBase {
     }
 
     resourceValidationUtils.validateControlledResourceRegionAgainstPolicy(
-       workspaceUuid, region, getCloudPlatform());
+        workspaceUuid, region, getCloudPlatform());
 
     return ControlledResourceFields.builder()
         .workspaceUuid(workspaceUuid)
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledResourceControllerBase.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledResourceControllerBase.java
@@ -80,7 +80,7 @@ public abstract class ControlledResourceControllerBase extends ControllerBase {
     }
 
     resourceValidationUtils.validateControlledResourceRegionAgainstPolicy(
-        workspaceUuid, region, getCloudPlatform());
+       workspaceUuid, region, getCloudPlatform());
 
     return ControlledResourceFields.builder()
         .workspaceUuid(workspaceUuid)
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledResourceControllerBase.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledResourceControllerBase.java
@@ -23,7 +23,7 @@ import javax.servlet.http.HttpServletRequest;
  * Super class for controllers containing common code. The code in here requires the @Autowired
  * beans from the @Controller classes, so it is better as a superclass rather than static methods.
  */
-public abstract class ControlledResourceControllerBase extends ControllerBase {
+public class ControlledResourceControllerBase extends ControllerBase {
   private final ControlledResourceService controlledResourceService;
 
   /**
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledResourceControllerBase.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledResourceControllerBase.java
@@ -7,6 +7,7 @@ import bio.terra.workspace.generated.model.ApiControlledResourceCommonFields;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequestFactory;
 import bio.terra.workspace.service.iam.SamService;
+import bio.terra.workspace.service.resource.ResourceValidationUtils;
 import bio.terra.workspace.service.resource.controlled.ControlledResourceService;
 import bio.terra.workspace.service.resource.controlled.model.AccessScopeType;
 import bio.terra.workspace.service.resource.controlled.model.ControlledResourceFields;
@@ -23,8 +24,9 @@ import javax.servlet.http.HttpServletRequest;
  * Super class for controllers containing common code. The code in here requires the @Autowired
  * beans from the @Controller classes, so it is better as a superclass rather than static methods.
  */
-public class ControlledResourceControllerBase extends ControllerBase {
+public abstract class ControlledResourceControllerBase extends ControllerBase {
   private final ControlledResourceService controlledResourceService;
+  private final ResourceValidationUtils resourceValidationUtils;
 
   /**
    * The region field of these wsm resource type are filled during the creation flight because the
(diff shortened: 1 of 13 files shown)


service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
```diff
@@ -148,4 +150,22 @@ public static GoogleCredentials getGoogleCredentialsFromUserRequest(
...
+  /**
+   * Extract the region part from the given location string. If the string is a region, return that.
+   * If the string looks like a zone, return just the region part.
+   */
+  public static String parseRegion(String location) {
+    if (location.length() <= 2) {
+      return location;
+    }
+
+    Pattern pattern = Pattern.compile("-[a-z]$", Pattern.CASE_INSENSITIVE);
```
yuhuyoyo, line 163: nit: zonePattern

i think you can merge l164-166 into one using java's replaceAll with regex? 
location.replaceAll("(?!^)-[a-z]$", "");
https://www.baeldung.com/java-remove-trailing-characters

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
mikenev: that looks nicer, thanks.

service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
```diff
@@ -421,17 +437,30 @@ public static void validateAzureVmSize(String vmSize) {
...
   }
 
+  public void validateGcpRegion(UUID workspaceId, String region) {
+    region = GcpUtils.parseRegion(region);
+
+    // Get the list of valid locations for this workspace from TPS. If there are no regional
+    // constraints applied to the workspace, TPS should return all available regions.
+    List<String> validLocations = tpsApiDispatch.listValidRegions(workspaceId, "gcp");
+
+    if (validLocations.stream().noneMatch(region::equalsIgnoreCase)) {
```
ddietterich, line 458: Should we normalize regions to lowercase when they are assigned to the resource?
Otherwise, we have to always be remembering to ignore case and will return improperly cased region names, right?

We could have this method return the canonical region name rather than whatever the input was. wdyt?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/relayNamespace/ControlledAzureRelayNamespaceResource.java
```diff
@@ -197,7 +197,7 @@ public void validate() {
       throw new MissingRequiredFieldException(
           "Missing required region field for ControlledAzureRelayNamespace.");
     }
-    ResourceValidationUtils.validateRegion(getRegion());
+    ResourceValidationUtils.validateAzureRegion(getRegion());
```
ddietterich, line 200: So if this is here, why is it that we need it in the controller?

ddietterich (changes requested):
Two main concerns:
1. Is there really a need for resourceValidation in the base controller class
2. Use of literals instead of the `CloudPlatform` enum

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/relayNamespace/ControlledAzureRelayNamespaceResource.java
yuhuyoyo: it's not backward compatible for gcp resources which doesn't do region validation before this PR (e.g. if we stored an invalid region, we won't be able to deserialize it after this PR) not a problem for azure resources.

ddietterich: How would we store an invalid region? We couldn't have successfully created the cloud resource if the region was invalid, right?

Good to think about the compatibility aspect. Is this a case where we need to do some data cleaning before we start enforcing the policy?

If you buy my logic that all of the regions have to be valid - if not properly down-cased and de-zoned - then we could do the cleaning on retrieval. Parse the region string and put it in canonical form. If the form changed, then update the database. Maybe??

service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
mikenev: I've been thinking of where to do this and this probably isn't the best spot for it. When we call this, we've already put together the creation params into an object which is now read-only.
In the end, GCP doesn't seem to care what the case is on the resource. There's also some regions that are canonically upper-case in GCP (IE: multi-region = US or EU).
I think we could normalize the region against the generic list of all available regions earlier on, like when we put the common params together, but that feels like it belongs in a separate work item.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/relayNamespace/ControlledAzureRelayNamespaceResource.java
mikenev: Interesting. I think your logic is right in that we shouldn't have an invalid region saved, though maybe not de-zoned. The call to GCP should have failed. So its flight should have too.
I'm not sure that will affect existing resources in the context of this MR now that we're validating during Controlled Resource Create. Existing resources have already been created. 
The cleaning existing resources to use canonical names. I think we need a follow-up task for that.

ddietterich: Putting the resource cleanup into a separate ticket is fine with me. Are we likely to run into the same issue @yuhuyoyo is working on with buckets? That we can't enforce the validation when coming out of the database?

yuhuyoyo: yeah i think that sounds right. there might be zonal string in the db but not invalid ones.

mikenev: I don't think we'll hit an error like that here since we're only calling this validation on create. It should only apply to newly created items.

The next step is to add validation like this to clone. In that case we would be looking at the existing item's region, so we might need some zone => region normalizing there.

service/src/main/java/bio/terra/workspace/app/controller/ControlledResourceControllerBase.java
```diff
@@ -23,7 +23,7 @@
  * Super class for controllers containing common code. The code in here requires the @Autowired
  * beans from the @Controller classes, so it is better as a superclass rather than static methods.
  */
-public class ControlledResourceControllerBase extends ControllerBase {
+public abstract class ControlledResourceControllerBase extends ControllerBase {
```
ddietterich, line 26: Does this need to be abstract?

ddietterich (approved):
LGTM - one minor fix suggested

service/src/main/java/bio/terra/workspace/app/controller/ControlledResourceControllerBase.java
mikenev: Not anymore, will fix.

PR #1010, opened 2023-01-18
[WOR-677] Minimize usage of Azure workspaces with BPM disabled

cahrens:
We are moving towards removing the `bpm_azure_enabled` feature flag because we always run with BPM. This PR does the following:

1. Run Azure connected tests with `bpm_azure_enabled` enabled. To support this, the practice of passing an AzureContext to the CloudContext creation method has been removed from all connected tests. Instead, the SpendProfileService call (BPM) is mocked out to return a spend profile with the MRG coordinates.

Note that all existing code would ignore the AzureContext if BPM was enabled, instead pulling the MRG coordinates from the SpendProfileService via the spend profile ID in the parent workspace.

2. Remove the optional AzureContext argument from cloning. This is another case where the AzureContext was ignored if BPM was enabled (instead the spend profile ID passed in the clone request parameters is used), and we have no test code that relies on the ability to pass an AzureContext.

In a future ticket, we will sort out the handful of integration tests (VM tests) that rely on an AzureContext being provided in CloudContext creation, and then we can completely remove `bpm_azure_enabled` and related code.


service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -631,9 +631,6 @@ public ResponseEntity<ApiCloneWorkspaceResult> cloneWorkspace(
     String generatedDisplayName =
         sourceWorkspace.getDisplayName().orElse(sourceWorkspace.getUserFacingId()) + " (Copy)";
 
-    AzureCloudContext azureCloudContext =
```
cahrens, line 634: I am removing the optional AzureContext argument from cloning, as we intend to remove the `bpm_azure_enabled` entirely flag once the integration tests that rely on Azure managed app coordinates (VM tests) can be transitioned to using a spend profile from BPM.

I debated throwing an assertion in this method if `bpm_azure_enabled` is false, but given that this code path is also used for GCP workspace cloning (for which Azure-specific features and variables don't make any sense anyway), it seemed like somewhat of an odd thing to be checking. Hopefully `bpm_azure_enabled` will soon be gone.

ddietterich: Does that mean you cannot express cloning across LZs through the REST API?

cahrens: You can-- when you clone the workspace, you provide the destination spend profile ID, and that's tied to a landing zone.

cahrens:
The 2 GCP clone connected tests also failed on the last commit to main. https://github.com/DataBiosphere/terra-workspace-manager/runs/10762559732

Integration test failures are unrelated-- the integration tests actually passed on my previous commit, and I only changed a connected test after that.

```
bio.terra.workspace.client.ApiException: {"message":"bio.terra.policy.client.ApiException: {\"message\":\"Failed to obtain JDBC Connection; nested exception is org.postgresql.util.PSQLException: The connection attempt failed.\",\"statusCode\":500}","statusCode":500,"causes":[]}

File: openapi/src/parts/workspace.yaml [hunk 2]
```diff
@@ -408,8 +408,6 @@ components:
             GCP Location to use for cloud-based resources. If omitted, the location of the source
             resource will be used.
           type: string
-        azureContext:
```
ddietterich, line 411: I think this change shows up a mismatch between the expected model for BPM and the actual model.

WSM API is written assuming that the billing profile is _independent_ of the cloud platform. That one profile would hold the billing information for multiple clouds. Therefore, putting billing profile on the workspace object made sense.

BPM does not implement multi-cloud billing profiles. Since that is the case, the billing profile should not be attached to the workspace. It should be attached to the cloud context. If it were, then we would need the `azureContext` here to supply the billing profile.

Essentially for Azure the billing profile has become the context, so you are short-cutting it here. I am uneasy with that choice.

@aherbst-broad @rtitle

cahrens: It seems like this is an existing limitation of the clone workspace API though-- it didn't support providing a CloudContext, and instead an optional `AzureContext` was added on while BPM was in the process of being integrated.

If I am tracing through the [code](https://github.com/DataBiosphere/terra-workspace-manager/blob/a5e171f815d14b7a074d41b23c13e4792fafe218/service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateGcpContextFlightV2.java#L51) properly, GCP workspace cloning always uses the spend profile obtained from the parent workspace and creates a GCP Cloud Context with it.

Passing a CloudContext to clone seems like it would be reasonable if we support cross-cloud cloning in the future. But that will require refactoring in both the GCP and Azure implementations, so I don't see that keeping AzureContext as an optional argument to clone (which currently is ignored if BPM is enabled) is helpful.

(Note also that I filled out the PR description)

ddietterich: Ah, thanks for that explanation. I misunderstood.

I think there is still a mismatch that we should address, but this PR doesn't make it any worse.

cahrens: I agree there is a mismatch in expectations that we need to address-- we've been pretty hand-wavy about whether or not BPM profiles will support multiple clouds. It would be good to nail of this down before we start the migration of Rawls GCP workspaces to WSM/BPM.

marctalbott (approved):

aherbst-broad (approved):

PR #1011, opened 2023-01-18
PF-2289: Fix BigQuery dataset name generation

aaron-wang:


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/bqdataset/ControlledBigQueryDatasetHandler.java
```diff
@@ -67,7 +67,7 @@ public WsmResource makeResourceFromDb(DbResource dbResource) {
    * https://cloud.google.com/bigquery/docs/datasets#dataset-naming.
    */
   public String generateCloudName(@Nullable UUID workspaceUuid, String bqDatasetName) {
-    String generatedName = bqDatasetName.replace("-", "_");
+    String generatedName = bqDatasetName.replace("-", "_").toLowerCase();
```
zloery, line 70: per the link in the comment, datasets can have uppercase letters in their name. is lowercase-only a Terra UI restriction?

PR #1015, opened 2023-01-19
[PF-2422] Add getResource endpoint

rogerwangcs:
Expose the getResource method in ResourceDao as a new resource endpoint

### DIFF
--- a/openapi/src/parts/resource.yaml
+++ b/openapi/src/parts/resource.yaml
@@ -30,6 +30,31 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/workspaces/v1/{workspaceId}/resources/{resourceId}:
+    parameters:
+    - $ref: '#/components/parameters/WorkspaceId'
+    - $ref: '#/components/parameters/ResourceId'
+    get:
+      summary: |
+        Retrieve a resource in a workspace.
+      operationId: getResource
+      tags: [Resource]
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ResourceDescription'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '403':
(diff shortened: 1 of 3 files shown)


openapi/src/parts/resource.yaml
```diff
@@ -30,6 +30,31 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/workspaces/v1/{workspaceId}/resources/{resourceId}:
```
yuhuyoyo, line 33: why is the existing get method not good enough for your usage?

rogerwangcs: The [axon server getFile endpoint](https://docs.google.com/document/d/1_6ruIYYzTnbfrqfjEquRsWUx1ZsNJSzCALa9iMZ6au0/edit?resourcekey=0-L8V4uz_kouNZoUhvjO1R_A#heading=h.mo0z046y10jz) accepts general resource id. We need this endpoint to determine the resourceType and validate whether it is a file containing resource. Without this endpoint, we would unecessarily get all resources in the workspace with enumerateResources then filter for the desired resource.

ddietterich (approved):
LGTM

openapi/src/parts/resource.yaml
ddietterich: @yuhuyoyo I don't think there is an existing method that does this. Is there and I missed it?

yuhuyoyo: no. I meant we have endpoint per resources. i remember it was an intentional decision to not use the union attributes and implement endpoint per resources.

With this endpoint, why do we need GET per resource type at all?

i get why this is useful for the axon use case.

ddietterich: Good point. When we do the "v2" interface design, maybe we skip the per-resource GET in favor of this style!?

yuhuyoyo (approved):

PR #1018, opened 2023-01-19
Log exception so we can get to the underlying cause [WOR-727]

aherbst-broad:
We are getting an exception creating a workspace in staging, but the underlying IllegalStateException is being swallowed. This PR logs the exception to aid debugging.


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/storageContainer/VerifyAzureStorageContainerCanBeCreatedStep.java
```diff
@@ -169,6 +169,8 @@ private StepResult validateLandingZoneSharedStorageAccountExist(
                   "Shared storage account not found in landing zone. Landing zone ID='%s'.",
                   landingZoneId)));
     } catch (IllegalStateException illegalStateException) { // Thrown by landingZoneApiDispatch
+      logger.error(
+          "Landing zone associated with the billing profile not found.", illegalStateException);
       return new StepResult(
```
cahrens, line 174: Can we not include the `illegalStateException` stack trace in the `LandingZoneNotFoundException` itself?

ddietterich: Best practice is not to expose backtraces through the REST API. They can contain data that should not be visible to callers.

Not sure that applies in this case. I would expect LZ to be logging the backtrace before converting the exception.

PR #1024, opened 2023-01-23
PF-2368: Backfill the existing azure controlled resources region column

yuhuyoyo:


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/flight/RetrieveAzureCloudContexts.java
```diff
@@ -0,0 +1,58 @@
...
+  public RetrieveAzureCloudContexts(AzureCloudContextService azureCloudContextService) {
+    this.azureCloudContextService = azureCloudContextService;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    validateRequiredEntries(context.getWorkingMap(), CONTROLLED_RESOURCES_WITHOUT_REGION);
+    List<ControlledResource> controlledResources =
+        context.getWorkingMap().get(CONTROLLED_RESOURCES_WITHOUT_REGION, new TypeReference<>() {});
+    Set<UUID> workspaceIds =
```
ddietterich, line 40: This could be done in a single database transaction in the DAO that would:

1. Make one database query that would return a tuple: (azureCloudContext, controlledResourceWithoutRegion).
2. Stream that to build a map of cloudContext -> region
3. Loop through that to update the region field of the resource

You are basically implementing a hash join over workspaceId in memory that the database has already built.
This would not need to be a flight at all.

The only nasty thing about that approach is calling CRL from inside the DAO. Not a good plan.
However, you could do it in a flight in two steps:
1. Step 1: Query the database as above
2. Step 2: Build the map of context -> region using CRL. Then call DAO with the map and the list. Have it perform the updates in one transaction.

Since this is single-use-code that you have debugged, it is not worth redoing. However, when you find yourself building and probing maps, maybe check if the database can do it for you.

ddietterich (approved):
This is an inefficient implementation with more code than is needed. See comments.
No worries. Since it is a single use and then remove, I won't request changes.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/flight/RetrieveAzureResourcesRegionStep.java
```diff
@@ -0,0 +1,242 @@
...
+                new TypeReference<>() {});
+    Map<UUID, String> resourceIdToRegionMap = new HashMap<>();
+    Map<UUID, String> resourceIdToWorkspaceIdMap = new HashMap<>();
+    for (var resource : controlledResources) {
+      WsmResourceType resourceType = resource.getResourceType();
+      logger.info(
+          "Getting cloud region for resource {} in workspace {}",
+          resource.getResourceId(),
+          resource.getWorkspaceId());
+      AzureCloudContext azureCloudContext =
```
aherbst-broad, line 86: Is it possible there could be a disconnect between the `workspaceIdToAzureCloudContextMap` and the resources in `controlledResources` that would result in an NPE here?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/flight/RetrieveAzureResourcesRegionStep.java
```diff
@@ -0,0 +1,242 @@
...
+            resource,
+            getAzureStorageContainerRegion(resource.castByEnum(resourceType), azureCloudContext));
+        case CONTROLLED_AZURE_VM -> populateMapsWithResourceIdKey(
+            resourceIdToRegionMap,
+            resourceIdToWorkspaceIdMap,
+            resource,
+            getAzureVmRegion(resource.castByEnum(resourceType), azureCloudContext));
+        default -> throw new UnsupportedOperationException(
+            String.format(
+                "resource of type %s is not a GCP resource or is a referenced resource",
```
aherbst-broad, line 127: "is not an Azure resource" right?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/flight/RetrieveAzureResourcesRegionStep.java
```diff
@@ -0,0 +1,242 @@
...
+        landingZoneApiDispatch.getSharedStorageAccount(
+            new BearerToken(userRequest.getRequiredToken()), landingZoneId);
+    return existingSharedStorageAccount
+        .map(
+            apiAzureLandingZoneDeployedResource ->
+                storageManager
+                    .storageAccounts()
+                    .getById(apiAzureLandingZoneDeployedResource.getResourceId())
+                    .regionName())
+        .orElse(null);
```
aherbst-broad, line 211: Should this throw an exception? If we have a container with no storage account ID and no existing shared storage account it seems like we're in uncharted territory and should bail out, or at least log a warning.

service/src/test/java/bio/terra/workspace/connected/WorkspaceConnectedTestUtils.java
```diff
@@ -51,7 +51,8 @@ public Workspace createWorkspace(AuthenticatedUserRequest userRequest) {
     return workspaceService.getWorkspace(workspaceUuid);
   }
 
-  public void deleteWorkspaceAndGcpContext(AuthenticatedUserRequest userRequest, UUID workspaceId) {
+  public void deleteWorkspaceAndCloudContext(
```
aherbst-broad, line 54: 👍 I like this rename.

aherbst-broad (commented):
Overall this seems like a reasonable approach. Are there any pre-requisite PRs that need to go out before this backfill merges?

Another thought: I tend to be paranoid for this sort of thing, esp. now that we have Azure in production. What do you think about adding a dry run mode to the `UpdateControlledResourcesRegionStep` where it'd just log a string saying what it would do instead of doing the actual update so we can inspect the potential changes? Once we're satisfied we can turn the dry-run mode off and let it run for real.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/flight/RetrieveAzureResourcesRegionStep.java
yuhuyoyo: i don't want to abandon the whole operation because one workspace is "corrupted". logging warning sg.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/flight/RetrieveAzureResourcesRegionStep.java
yuhuyoyo: i don't think so?
cuz we iterate all the workspaces of the controlled resources and fetch the azure context (and throw if not found) in the previous step. 

however I can add a check here for best practice.

yuhuyoyo:
> Overall this seems like a reasonable approach. Are there any pre-requisite PRs that need to go out before this backfill merges?
> 
> Another thought: I tend to be paranoid for this sort of thing, esp. now that we have Azure in production. What do you think about adding a dry run mode to the `UpdateControlledResourcesRegionStep` where it'd just log a string saying what it would do instead of doing the actual update so we can inspect the potential changes? Once we're satisfied we can turn the dry-run mode off and let it run for real.

agree. have moved this to an admin api endpoint and added a wet-run mode. swagger code-gen doesn't allow default true so I have to use wet-run.

aherbst-broad (approved):
👍 wet run / dry run mode looks, thanks for doing that.

PR #1025, opened 2023-01-23
[PF-2333] Remove cloud-platform requirement for creating context

zloery:
Currently, creating a GCP cloud context involves polling GCP using the user's credentials to determine when cloud permissions are fully synchronized. However, this means that the user provided credentials must have the `cloud-platform` OAuth scope. This breaks the CLI, which intentionally does not include this scope for user auth tokens (as it gives broader permissions than we want for Terra).

I've switched this polling to use the calling user's pet SA instead, as we always generate pet tokens with cloud-platform scope. There was already a step in the flight to create the pet SA for the calling user, so the account is guaranteed to exist.


ddietterich (approved):
LGTM

yuhuyoyo (approved):

PR #1027, opened 2023-01-23
PF-2427: enable the TPS connected test in presubmit run, convert ApiCloudPlatform enum to lower case

yuhuyoyo:


service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
```diff
@@ -194,16 +198,20 @@ public List<String> listValidRegions(UUID workspaceId, String platform) {
...
     TpsApi tpsApi = policyApi();
     try {
-      return tpsApi.getLocationInfo(platform, location);
+      return tpsApi.getLocationInfo(convertApiCloudPlatformToTpsCloudPlatform(platform), location);
     } catch (ApiException e) {
       throw convertApiException(e);
     }
   }
 
+  private static String convertApiCloudPlatformToTpsCloudPlatform(ApiCloudPlatform platform) {
```
ddietterich, line 211: No, not like this. Now you are assuming that the strings and enums will always line up.
Please do this by putting another field into the WSM `CloudPlatform` enum. We convert from API -> CloudPlatform enum -> TPS platform string.

ddietterich (changes requested):

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
yuhuyoyo: done

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
```diff
@@ -194,16 +197,20 @@ public List<String> listValidRegions(UUID workspaceId, String platform) {
...
     try {
-      return tpsApi.getLocationInfo(platform, location);
+      return tpsApi.getLocationInfo(convertCloudPlatformToTpsCloudPlatform(platform), location);
     } catch (ApiException e) {
       throw convertApiException(e);
     }
   }
 
+  private static String convertCloudPlatformToTpsCloudPlatform(CloudPlatform platform) {
+    return platform.toSql().toLowerCase(Locale.ROOT);
```
ddietterich, line 211: You assume here that the SQL string is identical to the TPS string.
I think those should be independent.

Suggest in CloudPlatform add
```
private final tpsString;
```
Fill the new field in on each enum. (BTW, looks like AZURE needs to have the ApiCloudPlatform set; it shouldn't be `null`!)

Add a getter, maybe like:
```
public String getTpsString() {
 if (tpsString == null) { throw InternalLogicError(); }
 return tpsString;
}
```

ddietterich (changes requested):
I would like the TPS name captured in the CloudPlatform enum and not done by lowercasing the database name.

ddietterich (approved):
Thank you!

PR #1028, opened 2023-01-23
Use the right validation for existing bucket name

yuhuyoyo:
Our current database has bucket with name that has underscore in it. It is *allowed* in gcp so they likely has a gcs bucket cloud instance associated with. However, we imposed a rule to not allow underscore for some nextflow reason. that change is *not* backward compatible and makes all those previous workspace with underscore bucket unusable.

Now when those workspaces are not able to be get, delete, update because of this validation error. 

I don't want to just clean up the row in the db because the cloud bucket is still there.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gcsbucket/ControlledGcsBucketResource.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gcsbucket/ControlledGcsBucketResource.java
@@ -7,6 +7,7 @@ import bio.terra.stairway.RetryRule;
 import bio.terra.workspace.common.utils.FlightBeanBag;
 import bio.terra.workspace.common.utils.RetryRules;
 import bio.terra.workspace.db.DbSerDes;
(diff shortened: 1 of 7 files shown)


yuhuyoyo:
@ddietterich I ran into this as I'm backfilling the gcp resource region in the dev env. 

this is the minimum change I can think of to get around the validation. wdyt?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gcsbucket/ControlledGcsBucketResource.java
```diff
@@ -91,6 +92,11 @@ private ControlledGcsBucketResource(ControlledResourceFields common, String buck
     validate();
   }
 
+  public ControlledGcsBucketResource(DbResource dbResource, String bucketName) {
```
ddietterich, line 95: Maybe comment that this constructor is only used for building the resource from the database and not validating the bucket name.

ddietterich (approved):
lgtm

zloery (approved):

PR #1030, opened 2023-01-24
[PF-2432] Fix PrivateControlledGcsBucketLifecycle

zloery:
Another integration test which needed retries.

Also interesting: I've seen a number of failures where the primary test user could delete one blob from a bucket but later could not delete a different one. This might have been an issue of deleting an object immediately after it was created, but still odd.

### DIFF
--- a/integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
+++ b/integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
@@ -74,11 +74,13 @@ public class GcsBucketAccessTester implements AutoCloseable {
   }
 
   @Override
-  public void close() throws IOException {
+  public void close() throws Exception {
     // Clean up any created blob so that delete happens quickly, lest the test timeout waiting.
     Storage creatorClient = ClientTestUtils.getGcpStorageClient(creatorTestUser, projectId);
+    // Blob deletion permission can be different for individual blobs, so use retries here.
+    needToWait = true;
     for (BlobId blobId : createdBlobs) {
-      boolean found = creatorClient.delete(blobId);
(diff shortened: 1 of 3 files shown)


integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
```diff
@@ -138,6 +146,8 @@ public void checkAccess(TestUserSpecification testUser, @Nullable ControlledReso
         assertTrue(doWithOptionalWait(this::blobUpdate), "editor can update");
         assertTrue(doWithOptionalWait(this::blobDelete), "editor can delete");
         assertTrue(doWithOptionalWait(this::bucketGet), "editor can bucket get");
+        // Do not wait for permissions which won't appear
+        needToWait = false;
```
yuhuyoyo, line 150: why do we need to call doWithOptionalWait if we are not waiting? can't we just do

yuhuyoyo (approved):

ddietterich (changes requested):
I don't think your fix works right. I think there are some structural problems with this code (that I wrote, so blame me!).
1. The needToWait should be final - it is a property of the class, not of a run. So changing it breaks a second use of the object.
2. The designed usage is that the "creator" user has all of the owner permissions, so it can configure the bucket and resources and then test. We don't use it that way.
3. The code assumed that if you had one permission, the others would be sync'd. That appears to be not true for buckets.
4. Does the negative case - waiting for a user to lose a permission - work in this code? I haven't thought that through. It does seem like we have cases - as you have here - where you don't want to wait on other permissions.

I have a piece of fix for this class. You have another. Let's make a separate PR to fix up this class and verify correct usage in all of its users. I'll make a ticket for it.

PR #1032, opened 2023-01-25
[PF-2442] fix access tester

ddietterich:
This PR takes a stab at fixing up the access tester.

I changed the check code so that it would just check and not assert. Split out the wait and nowait cases, so instead of setting the boolean, we just call `doNoWait` for the "cannot XXX" cases. (h/t Yu).

I renamed the check methods to assertXXX and put the assert at the top level. A little hacky using the String to determine error or not, but it works.

I added a method that I plan to use in remove-user testing. Kicking off the PR now so you can review it and find out if it breaks any existing PR tests. In the meantime, I am locally running broken tests that use the tester to see if they behave better.

### DIFF
--- a/.github/workflows/run-integration-suite.yml
+++ b/.github/workflows/run-integration-suite.yml
@@ -1,0 +1,99 @@
+# Action for running a user-defined suite of integration tests against PRs. This test uses local
+# server changes but runs a published version of the client, so it will not pick
+# up local client changes.
+#
+# The idea of this workflow is to help debug integration tests. On your branch:
+# - modify service and integration code
+# - edit suites/TempTestSuite.json with the tests you want to run
+# - push your branch to github - NO PR NEEDED!
+# - use the workflow dispatch to run this workflow.
+# This does not report results anywhere, so you have to check the github action page
+# to see when it finished.
+
+name: Run Integration Temp Test Suite
+on:
+  workflow_dispatch: {}
+
+jobs:
+  any-integration-job:
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:13.1
+        env:
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+        - 5432:5432
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Initialize Postgres DB for local server test run
+      env:
+        PGPASSWORD: postgres
+      run: psql -h 127.0.0.1 -U postgres -f ./service/local-dev/local-postgres-init.sql
+
+    - name: Set up JDK
+      uses: actions/setup-java@v3
+      with:
+        distribution: 'temurin'
+        java-version: 17
+
+    - name: Cache Gradle packages
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+        key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+        restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+
+    - name: Get Vault token
+      id: vault-token-step
+      env:
+        VAULT_ADDR: https://clotho.broadinstitute.org:8200
+      run: |
+        VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
+          -e "VAULT_ADDR=${VAULT_ADDR}" \
+          vault:1.1.0 \
+          vault write -field token \
+            auth/approle/login role_id=${{ secrets.VAULT_APPROLE_ROLE_ID }} \
+            secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+        echo ::add-mask::$VAULT_TOKEN    
+        echo vault-token=$VAULT_TOKEN >> $GITHUB_OUTPUT
+
+    - name: Write config
+      id: config
+      uses: ./.github/actions/write-config
+      with:
+        vault-token: ${{ steps.vault-token-step.outputs.vault-token }}
+        target: local
+
+    - name: Launch local server
+      uses: ./.github/actions/start-local-server
+
+    - name: Run the integration test suite
+      id: integration-test
+      uses: ./.github/actions/integration-test
+      with:
+        test-server: workspace-local.json
+        test: suites/TempTestSuite.json
+
+    - name: Archive WSM and TestRunner logs
+      id: archive_logs
+      if: always()
+      uses: actions/upload-artifact@v3
+      with:
+        name: wsm-and-testrunner-logs
+        path: |
+          wsm.log
+          ${{ steps.integration-test.outputs.results-dir }}
--- a/integration/src/main/java/scripts/testscripts/ControlledApplicationPrivateGcsBucketLifecycle.java
+++ b/integration/src/main/java/scripts/testscripts/ControlledApplicationPrivateGcsBucketLifecycle.java
@@ -84,7 +84,8 @@ public class ControlledApplicationPrivateGcsBucketLifecycle
   protected void doCleanup(List<TestUserSpecification> testUsers, WorkspaceApi workspaceApi)
       throws Exception {
     try {
-      // Clean any buckets on the list. There might be some in a failure case
+      // Clean any buckets on the list. There might be some if one of the test cases
+      // fails and we do not execute the deletes as part of the test.
       deleteBucketList();
     } finally {
       super.doCleanup(testUsers, workspaceApi);
@@ -133,8 +134,11 @@ public class ControlledApplicationPrivateGcsBucketLifecycle
     assertEquals(3, bucketList.getResources().size());
     MultiResourcesUtils.assertResourceType(ResourceType.GCS_BUCKET, bucketList);
 
+    // TODO: PF-2446 This is flaky due to IAM latency. Solution is to make delete
+    //  which is async, retry the delete if it fails for permission reasons.
+    //  This still gets cleaned up (or not) in doCleanup
     // Try the delete as part of the successful test
-    deleteBucketList();
+    // deleteBucketList();
   }
 
   private void testNoAssignedUser(ControlledGcpResourceApi resourceApi, String projectId)
@@ -152,7 +156,8 @@ public class ControlledApplicationPrivateGcsBucketLifecycle
     assertNotNull(bucketName);
     logger.info("Created no-assigned-user bucket {}", bucketName);
 
-    // Creating the tester will wait for wsmapp to have EDITOR permission
+    // Constructing the tester will wait for wsmapp to have EDITOR permission; no additional
+    // test is needed here.
     try (GcsBucketAccessTester tester = new GcsBucketAccessTester(wsmapp, bucketName, projectId)) {
       tester.assertAccess(owner, null);
       // Don't bother testing reader and writer here.
@@ -178,7 +183,8 @@ public class ControlledApplicationPrivateGcsBucketLifecycle
     assertNotNull(bucketName);
     logger.info("Created assigned-reader bucket {}", bucketName);
 
-    // Creating tester waits for wsmapp to have EDITOR permissions
+    // Constructing the tester will wait for wsmapp to have EDITOR permission; no additional
+    // test is needed here.
     try (GcsBucketAccessTester tester = new GcsBucketAccessTester(wsmapp, bucketName, projectId)) {
       tester.assertAccess(reader, null);
       tester.assertAccessWait(writer, ControlledResourceIamRole.READER);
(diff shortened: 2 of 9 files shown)


integration/src/main/java/scripts/testscripts/ControlledApplicationPrivateGcsBucketLifecycle.java
```diff
@@ -181,33 +199,25 @@ private void testAssignedWriter(ControlledGcpResourceApi resourceApi, String pro
...
-    if (bucketName != null) {
-      logger.warn("Test failed to cleanup bucket " + bucketName);
+  private void deleteBucketList() throws Exception {
+    for (CreatedControlledGcpGcsBucket bucket : bucketList) {
+      GcsBucketUtils.deleteControlledGcsBucket(
+          bucket.getResourceId(), getWorkspaceId(), wsmappResourceApi);
+      logger.info(
+          "Application deleted bucket {}", bucket.getGcpBucket().getAttributes().getBucketName());
     }
+    bucketList.clear();
```
yuhuyoyo, line 221: if you clear this in doUserJourney, there's nothing to delete in cleanUp()

integration/src/main/java/scripts/testscripts/ControlledApplicationPrivateGcsBucketLifecycle.java
```diff
@@ -155,16 +173,16 @@ private void testAssignedReader(ControlledGcpResourceApi resourceApi, String pro
             bucketResourceName,
             CloningInstructionsEnum.NOTHING,
             privateUser);
-    bucketName = createdBucket.getGcpBucket().getAttributes().getBucketName();
+    bucketList.add(createdBucket);
+    String bucketName = createdBucket.getGcpBucket().getAttributes().getBucketName();
     assertNotNull(bucketName);
     logger.info("Created assigned-reader bucket {}", bucketName);
 
+    // Creating tester waits for wsmapp to have EDITOR permissions
```
yuhuyoyo, line 181: ... so we don't need to check wsmapp's EDITOR permission.

integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
```diff
@@ -141,7 +143,14 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
...
 
-    // Create the bucket - should work this time
+    // -- Test Case 3 --
+    // Clone bucket test
+    // - create a bucket
+    // - fail to create a duplicate named bucket (duplicate test of above, needed here?)
+    // - test that reader has access to the bucket
+    // - testUser puts files into the bucket
+    // - reader clones the bucket to destination workspace
+    // -
```
yuhuyoyo, line 153: nit. rm this line

integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
```diff
@@ -83,19 +87,76 @@ public void close() throws IOException {
...
+   * expected to be denied, so this method cannot be used to wait for a user to lose access.
+   *
+   * @param testUser user to check
+   * @param role the IamRole to test for; null to test the "user has no role" case
+   * @throws Exception from asserts if the user has unexpected access
+   */
+  public void assertAccessWait(
+      TestUserSpecification testUser, @Nullable ControlledResourceIamRole role) throws Exception {
+    assertNotEquals(creatorTestUser, testUser);
+    String result = checkAccessWorker(testUser, role, true);
```
yuhuyoyo, line 101: /*wait=*/

same for below

integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
```diff
@@ -154,17 +265,34 @@ public interface TestFunction {
   }
 
   private boolean doWithOptionalWait(TestFunction function) throws Exception {
```
yuhuyoyo, line 267: can we get rid of the field needToWait and pass this in through the param? Similar to `checkAccessWorker`

integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
```diff
@@ -104,45 +165,95 @@ public void checkAccess(TestUserSpecification testUser, @Nullable ControlledReso
...
     testClient = ClientTestUtils.getGcpStorageClient(testUser, projectId);
     if (role == null) {
-      assertFalse(doWithOptionalWait(() -> blobCreate(testClient)), "no role cannot create");
-      assertFalse(doWithOptionalWait(this::blobRead), "no role cannot read");
-      assertFalse(doWithOptionalWait(this::blobUpdate), "no role cannot update");
-      assertFalse(doWithOptionalWait(this::blobDelete), "no role cannot delete");
-      assertFalse(doWithOptionalWait(this::bucketGet), "no role cannot bucket get");
-      assertFalse(doWithOptionalWait(this::bucketDelete), "no role cannot bucket delete");
-      return;
+      if (doNoWait(() -> blobCreate(testClient))) {
```
yuhuyoyo, line 168: can you add some comment to explain why these use `doNoWait` regardless of what the `wait` param is?

integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
```diff
@@ -104,45 +165,95 @@ public void checkAccess(TestUserSpecification testUser, @Nullable ControlledReso
...
-        return;
-
-      default:
-        throw new IllegalArgumentException("Invalid IAM role: " + role);
+      case READER -> {
+        if (!doWithOptionalWait(this::blobRead)) {
+          return "reader can read";
+        }
+        if (doNoWait(() -> blobCreate(testClient))) {
+          return "reader cannot create";
```
yuhuyoyo, line 195: why does it return "cannot create", when doNoWait return true, I thought that means the action succeeded?

integration/src/main/java/scripts/testscripts/ControlledApplicationPrivateGcsBucketLifecycle.java
ddietterich: The idea is that on a successful test run, the deletes happen as part of the test. So we notice a failure if they fail.
On some other test failure, we will still delete any buckets we have made so far. Do you think that works?

integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
ddietterich: I don't understand the suggestion.

integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
ddietterich: I added a comment at the top of the method, since it applies to all of the cases.

integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
yuhuyoyo: doWithOptionalWait(TestFunction function, boolean wait)

ddietterich: I don't think I want to do that.

yuhuyoyo: why not? it's only used by `checkAccessWorker` and `needToWait` is only set in `checkAccessWorker`.

ddietterich: Because it is a style change for a blocking PR

integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
```diff
@@ -154,17 +272,34 @@ public interface TestFunction {
...
-      if (needToWait) {
-        ClientTestUtils.getWithRetryOnException(function::apply);
-        needToWait = false;
-      } else {
-        function.apply();
+      ClientTestUtils.getWithRetryOnException(function::apply);
+      return true;
+    } catch (StorageException e) {
+      if (e.getCode() == HttpStatusCodes.STATUS_CODE_FORBIDDEN) {
+        assertEquals(HttpStatusCodes.STATUS_CODE_FORBIDDEN, e.getCode());
```
zloery, line 286: this assertion will always be true given the branch above

integration/src/main/resources/suites/TempTestSuite.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "Temp Test Suite",
```
zloery, line 2: This is temp, should it be committed?

integration/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -53,8 +53,10 @@
 
 public class ClientTestUtils {
   // Retry default parameters
-  public static final Duration DEFAULT_RETRY_TOTAL_DURATION = Duration.ofMinutes(30);
-  public static final Duration DEFAULT_SLEEP_DURATION = Duration.ofSeconds(15);
+  // The total allowed duration is a guess. It may be too long, but we have no guidance.
+  // The sleep duration is based on guidance for the fastest we should consider polling.
+  public static final Duration DEFAULT_RETRY_TOTAL_DURATION = Duration.ofMinutes(20);
+  public static final Duration DEFAULT_SLEEP_DURATION = Duration.ofSeconds(60);
```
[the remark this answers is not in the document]
ddietterich: I am understanding that 30s is the fastest they want us to poll. Thought I'd try slower.
It is another arbitrary guess.

integration/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -329,6 +331,19 @@ public static <T> T getWithRetryOnException(
     return result;
   }
 
+  public static void getWithRetryOnFalse(SupplierWithException<Boolean> function) throws Exception {
+    getWithRetryOnException(function);
```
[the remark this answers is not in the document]
ddietterich: Good catch - I was supposed to be calling `tryBooleanFunction` which converts a false into an exception.
But rereading the code, it really doesn't work right. If the boolean function throws, then we retry instead of
allowing the exception. It is unused code, so I'll delete it and code a different way when I get back to RemoveUser.

integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
ddietterich: Removed the assert.

integration/src/main/java/scripts/utils/GcsBucketAccessTester.java
ddietterich: I fixed this and the annotation of /wait=/

integration/src/main/resources/suites/TempTestSuite.json
ddietterich: Yes, but I could pick a different name. Any suggestions?

It is a kind of scratch suite for running some subset of integration tests in the github runner. I made run-integration-test.yml to do that.

zloery: sgtm! I missed the new action and wasn't sure if this was for local testing or unmerged branches

yuhuyoyo (approved):

.github/workflows/run-integration-suite.yml
```diff
@@ -0,0 +1,99 @@
+# Action for running a user-defined suite of integration tests against PRs. This test uses local
+# server changes but runs a published version of the client, so it will not pick
+# up local client changes.
+#
+# The idea of this workflow is to help debug integration tests. On your branch:
```
mikenev, line 5: nice!

integration/src/main/java/scripts/testscripts/ControlledApplicationPrivateGcsBucketLifecycle.java
```diff
@@ -155,16 +173,16 @@ private void testAssignedReader(ControlledGcpResourceApi resourceApi, String pro
...
     try (GcsBucketAccessTester tester = new GcsBucketAccessTester(wsmapp, bucketName, projectId)) {
-      tester.checkAccess(wsmapp, ControlledResourceIamRole.EDITOR);
-      tester.checkAccess(reader, null);
-      tester.checkAccess(writer, ControlledResourceIamRole.READER);
+      tester.assertAccess(reader, null);
+      tester.assertAccessWait(writer, ControlledResourceIamRole.READER);
```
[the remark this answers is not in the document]
mikenev: +1 to that, reading through it does make you stop and think.

integration/src/main/java/scripts/testscripts/ControlledGcsBucketLifecycle.java
```diff
@@ -258,7 +274,10 @@ public void doUserJourney(TestUserSpecification testUser, WorkspaceApi workspace
     assertEquals(1, bucketList.getResources().size());
     MultiResourcesUtils.assertResourceType(ResourceType.GCS_BUCKET, bucketList);
 
-    // Owner can delete the bucket through WSM
+    // -- Test Case 6 --
```
mikenev, line 277: Not specific to this review - It is troubling that we have 6 cases in this one test. I can see why for these integration tests where they're testing the whole lifecycle, but I can also see why this would take a long time. Especially if we have to wait between cases.

mikenev (approved):

zloery (approved):

PR #1034, opened 2023-01-26
PF-2269: Copy source defaultTableLifetime and defaultPartitionLifetime to destination, when cloning controlled BQ datasets

aaron-wang:


openapi/src/parts/controlled_gcp_big_query_dataset.yaml
```diff
@@ -187,6 +187,14 @@ components:
             Name for destination dataset. Must not be set if cloningInstructions
             is COPY_REFERENCE. Letters, numbers and underscores only.
           type: string
+        defaultTableLifetime:
+          description: Optional. Default lifetime (in seconds) for newly created tables.
+          type: integer
```
yuhuyoyo, line 192: nit: can you extract this as a schema as share it with createRequest

yuhuyoyo (approved):

zloery (changes requested):
I don't think this is the right approach for copying parameters during BQ clones. 

`ControlledBigQueryDatasetAttributes` and other `Controlled____Attrbutes` classes are meant to be the minimum set of attributes to uniquely identify a cloud resource, which we use for serialization. WSM stores these in its database, but it doesn't track all the other cloud parameters that might apply to controlled resources, we'd rather query the cloud for those instead of keeping multiple copies of state. `defaultTableLifetime` and `defaultPartitionLifetime` aren't required to identify a dataset, they're just parameters on it.

I think a better approach would be more like what we do for GCS buckets, where we clone things like storage class/lifecycle rules:

- Early in the flight, read the source dataset's values for `defaultTableLifetime` and `defaultPartitionLifetime` from the cloud, and store them in the flight's working map. We already have a `RetrieveBigQueryDatasetCloudAttributesStep` in the flight, which could be a good spot for that
  - e.g. for GCS, we do this in [RetrieveGcsBucketCloudAttributesStep](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gcsbucket/RetrieveGcsBucketCloudAttributesStep.java#L80)
- Later in the flight when you're creating a new controlled BQ dataset, you can read those values from the flight working map [for the new object](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CopyBigQueryDatasetDefinitionStep.java#L103)

zloery:
> i discussed with Dan and we decided to store it in the attributes to avoid around of cloud calls.
> Aaron is going to do a round of backfill for the existing resource using what you described.

We're already making that cloud call in `RetrieveBigQueryDatasetCloudAttributesStep` since we're reading source dataset location from the cloud... though looking at this again I'm not sure why we do that instead of just reading it from the source object now that we store region inside WSM. We should get rid of that step if we're taking this approach

ddietterich:
> We should get rid of that step if we're taking this approach

Yes, let's do that!

aaron-wang:
> We should get rid of that step if we're taking this approach

Ok, I'll make a separate PR for that!

zloery (approved):

PR #1037, opened 2023-01-27
PF-2269: Remove extraneous BQ dataset retrieve step

aaron-wang:
Per discussion in #1034: 

Remove the `RetrieveBigQueryDatasetCloudAttributesStep` which originally queried the location.

- We're already storing the location in the resource, so this step isn't needed


yuhuyoyo (approved):
thanks.

zloery (approved):

PR #1039, opened 2023-01-27
PF-2449: add sonarqube to gradle build task and update gha

yuhuyoyo:


yuhuyoyo:
@ddietterich i think we also need to remove codacy from the required test?

ddietterich (approved):
LGTM - I'll remove Codacy as a requirement

mikenev (approved):
Seems ok. I wonder if it should have been split into 2 items - one for GHA and one for SonarQube.

PR #1041, opened 2023-01-30
PF-2269: Backfill the lifetime for existing controlled BQ datasets

aaron-wang:


service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -537,6 +574,64 @@ public boolean updateControlledResourceRegion(UUID resourceId, @Nullable String
     return updated;
   }
 
+  /**
+   * Update a BigQuery dataset's default table lifetime
+   *
+   * @return whether the resource's region is successfully updated.
+   */
+  @WriteTransaction
+  public boolean updateBigQueryDatasetDefaultTableLifetime(
```
yuhuyoyo, line 583: can we add a test for this one and the one below in ResourceDao please

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -646,6 +647,31 @@ public String updateGcpControlledResourcesRegionAsync(
...
+    }
+    AuthenticatedUserRequest wsmSaRequest =
+        new AuthenticatedUserRequest().token(Optional.of(wsmSaToken));
+    return jobService
+        .newJob()
+        .description(
+            "A flight to update controlled BigQuery datasets' missing "
+                + "default table lifetime and default partition lifetime "
+                + "in all the existing terra managed gcp projects")
+        .jobId(UUID.randomUUID().toString())
```
yuhuyoyo, line 669: no need to set this actually. a job id will be generated when none is provided.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/flight/RetrieveGcpControlledBigQueryDatasetLifetimeStep.java
```diff
@@ -0,0 +1,145 @@
...
+
+  public RetrieveGcpControlledBigQueryDatasetLifetimeStep(CrlService crlService) {
+    this.crlService = crlService;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    validateRequiredEntries(
+        context.getWorkingMap(), CONTROLLED_BIG_QUERY_DATASETS_WITHOUT_LIFETIME);
+    List<ControlledResource> controlledBigQueryDatasets =
```
yuhuyoyo, line 43: we can just deserialize to List<ControlledGcpBqDatasetResource> here.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/RetrieveControlledBigQueryDatasetWithoutLifetimeStep.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public RetrieveControlledBigQueryDatasetWithoutLifetimeStep(
+      CloudPlatform cloudPlatform, ResourceDao resourceDao) {
+    this.cloudPlatform = cloudPlatform;
+    this.resourceDao = resourceDao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    List<ControlledResource> controlledBigQueryDatasets =
```
yuhuyoyo, line 26: it's guaranteed to return List<ControlledGcpBqDataset> here right?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledBigQueryDatasetsLifetimeStep.java
```diff
@@ -0,0 +1,112 @@
...
+        context
+            .getWorkingMap()
+            .get(CONTROLLED_RESOURCE_ID_TO_WORKSPACE_ID_MAP, new TypeReference<>() {});
+
+    List<ControlledResource> controlledBigQueryDatasets =
+        context
+            .getWorkingMap()
+            .get(CONTROLLED_BIG_QUERY_DATASETS_WITHOUT_LIFETIME, new TypeReference<>() {});
+
+    if (controlledBigQueryDatasets.isEmpty()) {
```
yuhuyoyo, line 101: is null or empty

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledBigQueryDatasetsLifetimeStep.java
```diff
@@ -0,0 +1,112 @@
...
+
+    System.out.println("id to table lifetime " + resourceIdToDefaultTableLifetimeMap);
+
+    for (var resource : controlledBigQueryDatasets) {
+      var id = resource.getResourceId();
+      System.out.println("BQ dataset without lifetime " + id);
+      boolean updated =
+          resourceDao.updateBigQueryDatasetDefaultTableLifetime(
+                  resource.castByEnum(CONTROLLED_GCP_BIG_QUERY_DATASET),
+                  resourceIdToDefaultTableLifetimeMap.get(id))
```
yuhuyoyo, line 73: should you only call updateBqDatasetDefaultTableLifetime only when the map has the key?

service/src/test/java/bio/terra/workspace/db/ResourceDaoTest.java
```diff
@@ -587,6 +595,28 @@ public void listControlledResourceWithoutRegion() {
...
+    resourceDao.deleteAllControlledResources(workspaceUuid2, CloudPlatform.GCP);
+    resourceDao.deleteAllControlledResources(workspaceUuid3, CloudPlatform.GCP);
+  }
+
+  @Test
+  public void listControlledBigQueryDatasetsWithoutLifetime() {
+    UUID workspaceUuid = createWorkspaceWithGcpContext(workspaceDao);
+    for (int i = 1; i <= 6; i++) {
+      ControlledBigQueryDatasetResource dataset =
+          ControlledResourceFixtures.makeDefaultControlledBqDatasetBuilder(workspaceUuid)
```
yuhuyoyo, line 608: what about you just do
for (int i = 0; i < 5; i++) {
   ControlledBigQueryDatasetResource dataset =
          ControlledResourceFixtures.makeDefaultControlledBqDatasetBuilder(workspaceUuid)
                .defaultTableLifetime(DEFAULT_CREATED_BIG_QUERY_TABLE_LIFETIME)
                .defaultPartitionLifetime(DEFAULT_CREATED_BIG_QUERY_PARTITION_LIFETIME)
                .build();
       resourceDao.createControlledResource(dataset);
}

ControlledBigQueryDatasetResource datasetWithoutLifetime =
          ControlledResourceFixtures.makeDefaultControlledBqDatasetBuilder(workspaceUuid)
              .defaultTableLifetime(null)
              .defaultPartitionLifetime(null)
              .build();
              
     resourceDao.createControlledResource(datasetWithoutLifeTime);

assertEquals(5, ans);

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/flight/RetrieveGcpControlledBigQueryDatasetLifetimeStep.java
aaron-wang: Can you clarify what you mean by that?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/RetrieveControlledBigQueryDatasetWithoutLifetimeStep.java
aaron-wang: Yes, now it does.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/flight/RetrieveGcpControlledBigQueryDatasetLifetimeStep.java
yuhuyoyo: i'm sorry. i was thinking one thing and writing another.

List<ControlledBqDatasetResource> datasets
    = context
            .getWorkingMap()
            .get(CONTROLLED_BIG_QUERY_DATASETS_WITHOUT_LIFETIME, new TypeReference<>() {});

so we don't need to do the castbyEnum

service/src/test/java/bio/terra/workspace/db/ResourceDaoTest.java
```diff
@@ -587,6 +602,71 @@ public void listControlledResourceWithoutRegion() {
...
+
+      ControlledBigQueryDatasetResource datasetWithLifetime =
+          ControlledResourceFixtures.makeDefaultControlledBqDatasetBuilder(workspaceUuid)
+              .defaultTableLifetime(DEFAULT_CREATED_BIG_QUERY_TABLE_LIFETIME)
+              .defaultPartitionLifetime(DEFAULT_CREATED_BIG_QUERY_PARTITION_LIFETIME)
+              .build();
+
+      resourceDao.createControlledResource(datasetWithLifetime);
+
+      var ans = resourceDao.listControlledBigQueryDatasetsWithoutLifetime().size();
```
yuhuyoyo, line 633: can you do this assert after line 614 and assert that it's 0.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledBigQueryDatasetsLifetimeStep.java
```diff
@@ -0,0 +1,102 @@
...
+    List<ControlledBigQueryDatasetResource> controlledBigQueryDatasets =
+        context
+            .getWorkingMap()
+            .get(CONTROLLED_BIG_QUERY_DATASETS_WITHOUT_LIFETIME, new TypeReference<>() {});
+
+    if (controlledBigQueryDatasets == null || controlledBigQueryDatasets.isEmpty()) {
+      return StepResult.getStepResultSuccess();
+    }
+    for (var resource : controlledBigQueryDatasets) {
+      resourceDao.updateBigQueryDatasetDefaultTableAndPartitionLifetime(resource, null, null);
```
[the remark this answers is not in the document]
aaron-wang: Oh, good catch! So, since we iterate through the BQ datasets from the initial `listControlledBigQueryDatasetsWithoutLifetime` query, we can use those lifetime values for our undo.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/flight/RetrieveControlledBigQueryDatasetLifetimeStep.java
```diff
@@ -0,0 +1,134 @@
...
+  private Long getBqDatasetDefaultPartitionLifetime(ControlledBigQueryDatasetResource resource) {
+    Dataset dataset = getBqDataset(resource);
+    if (dataset != null) {
+      return dataset.getDefaultPartitionExpirationMs() / 1000;
+    }
+    return null;
+  }
+
+  @Nullable
+  private Dataset getBqDataset(ControlledBigQueryDatasetResource resource) {
```
ddietterich, line 115: I'd like to understand the failure/retry model here. If there is a cloud glitch and we fail to get the dataset, then we will not back fill it. I guess we will rerun this every time we restart a pod, so in theory it will converge.

But how will we know it is safe to remove this code? Can we have this code run the "what needs backfill?" query at the end and report when it is done?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/flight/RetrieveControlledBigQueryDatasetLifetimeStep.java
```diff
@@ -0,0 +1,134 @@
...
+      logger.info(
+          "Getting default table lifetime and partition life for resource (BigQuery dataset) {} in workspace {}",
+          resource.getResourceId(),
+          resource.getWorkspaceId());
+      populateMapsWithResourceIdKey(
+          resourceIdToDefaultTableLifetimeMap,
+          resourceIdToDefaultPartitionLifetimeMap,
+          resourceIdToWorkspaceIdMap,
+          resource,
+          getBqDatasetDefaultTableLifetime(resource),
```
ddietterich, line 60: This causes two cloud requests to lookup the same information. I think it would be better to make one request to get that, and then pull the two pieces of info from it.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -646,6 +647,30 @@ public String updateGcpControlledResourcesRegionAsync(
         .addParameter(IS_WET_RUN, wetRun)
         .submit();
   }
+  // TODO (PF-2269): Clean this up once the back-fill is done in all Terra environments.
+  @Traced
+  @Nullable
+  public String updateGcpControlledBigQueryDatasetsLifetimeAsync() {
```
ddietterich, line 653: Why does this return anything? No one is waiting.
If there is a reason, make a comment.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/flight/UpdateControlledBigQueryDatasetsLifetimeFlight.java
```diff
@@ -0,0 +1,32 @@
...
+   * @param inputParameters FlightMap of the inputs for the flight
+   * @param beanBag Anonymous context meaningful to the application using Stairway
+   */
+  public UpdateControlledBigQueryDatasetsLifetimeFlight(FlightMap inputParameters, Object beanBag) {
+    super(inputParameters, beanBag);
+    FlightBeanBag flightBeanBag = FlightBeanBag.getFromObject(beanBag);
+
+    addStep(
+        new RetrieveControlledBigQueryDatasetWithoutLifetimeStep(flightBeanBag.getResourceDao()),
+        RetryRules.shortDatabase());
```
ddietterich, line 24: I don't think anything generates a retry in this step. The database retries are done in the database layer. If you want to retry, there needs to be some code to ask Stairway to retry. I don't see any.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/RetrieveControlledBigQueryDatasetWithoutLifetimeStep.java
```diff
@@ -0,0 +1,39 @@
...
+  private final ResourceDao resourceDao;
+
+  public RetrieveControlledBigQueryDatasetWithoutLifetimeStep(ResourceDao resourceDao) {
+    this.resourceDao = resourceDao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    List<ControlledBigQueryDatasetResource> controlledBigQueryDatasets =
+        Optional.ofNullable(resourceDao.listControlledBigQueryDatasetsWithoutLifetime())
```
ddietterich, line 25: This is where you would know that there is nothing left to do.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/flight/RetrieveControlledBigQueryDatasetLifetimeStep.java
```diff
@@ -0,0 +1,134 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RetrieveControlledBigQueryDatasetLifetimeStep implements Step {
```
ddietterich, line 28: I'm confused by the module organization. You have a fight and one step right under "flight" and the other steps under "flight/update".

Suggest you make a directory of "flight/backfill" and co-locate everything. That will make removing it easier also.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledBigQueryDatasetsLifetimeStep.java
```diff
@@ -0,0 +1,101 @@
...
+          updatedResources.add(
+              resourceDao
+                  .getResource(UUID.fromString(resourceIdsToWorkspaceIdMap.get(id)), id)
+                  .castToControlledResource()
+                  .castByEnum(CONTROLLED_GCP_BIG_QUERY_DATASET));
+        }
+      }
+    }
+
+    setResponse(context, updatedResources, HttpStatus.OK);
```
ddietterich, line 81: Why do you need this? What is the result used for? Seems like extra work for no benefit.
If it is for testing, I suggest moving the resource retrieval into the test, rather than have it in this code and unused.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledBigQueryDatasetsLifetimeStep.java
```diff
@@ -0,0 +1,101 @@
...
+    List<ControlledBigQueryDatasetResource> controlledBigQueryDatasets =
+        context
+            .getWorkingMap()
+            .get(CONTROLLED_BIG_QUERY_DATASETS_WITHOUT_LIFETIME, new TypeReference<>() {});
+
+    if (controlledBigQueryDatasets == null || controlledBigQueryDatasets.isEmpty()) {
+      return StepResult.getStepResultSuccess();
+    }
+    for (var resource : controlledBigQueryDatasets) {
+      resourceDao.updateBigQueryDatasetDefaultTableAndPartitionLifetime(
```
ddietterich, line 96: I'm very confused by this code. In what way does this undo actually undo anything?
Won't it be filling in things in the attributes that were not there before? Are the values correct? Are they null? Where did they come from?

ddietterich (changes requested):
At a minimum, the modules need better organization. There is a lot I do not understand that I think I should be able to understand by reading the code and comments.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/flight/RetrieveControlledBigQueryDatasetLifetimeStep.java
yuhuyoyo: so I realize as I'm doing the region backfill, some bucket or dataset have a lifetime so they are deleted in the gcp project but we still have them in the database. 

i manually checked them for the preprod env and manually deleted the db row when i don't see it in the cloud project. Prod env is very clean.

there are so many in devel that I didn't bother to check manually (they can be from test glitches or just admin users manually deleting them from the gcp project too).

that's being said, we can just check manuallly the log of the preprod and prod env once this code is in.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
yuhuyoyo: +1. my update region methods are fed into api endpoints. but if we are doing this at startup, we don't need to return jobid.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/update/UpdateControlledBigQueryDatasetsLifetimeStep.java
yuhuyoyo: the resourceDao logic to retrieve bq dataset without lifetime uses OR so the resources may have one of the attributes set. 

I think it's probably enough to use AND because the old resource definitely don't have either but this is fine with me too..

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
yuhuyoyo: it is easier for testing though..

as this is async, the test will need a way to pull the job till it's done.

service/src/main/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHook.java
```diff
@@ -353,30 +355,28 @@ private void maybeLogForSyncGcpIamRolesFlight(
...
-        context
-            .getWorkingMap()
-            .get(CONTROLLED_RESOURCE_ID_TO_WORKSPACE_ID_MAP, new TypeReference<>() {});
+        Preconditions.checkNotNull(
+            context
+                .getWorkingMap()
+                .get(CONTROLLED_RESOURCE_ID_TO_WORKSPACE_ID_MAP, new TypeReference<>() {}));
 
-    for (var resourceId : resourceIdToRegionMap.keySet()) {
+    for (var pair : resourceIdToWorkspaceIdMap.entrySet()) {
```
ddietterich, line 372: Nit: by convention, we do not use `var` unless it is obvious what the type is; e.g., a builder or a `new`.

service/src/main/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHook.java
```diff
@@ -353,30 +355,28 @@ private void maybeLogForSyncGcpIamRolesFlight(
...
       return;
     }
     FlightUtils.validateRequiredEntries(
-        context.getWorkingMap(),
-        CONTROLLED_RESOURCE_ID_TO_REGION_MAP,
-        CONTROLLED_RESOURCE_ID_TO_WORKSPACE_ID_MAP);
-    Map<UUID, String> resourceIdToRegionMap =
-        context.getWorkingMap().get(CONTROLLED_RESOURCE_ID_TO_REGION_MAP, new TypeReference<>() {});
+        context.getWorkingMap(), CONTROLLED_RESOURCE_ID_TO_WORKSPACE_ID_MAP);
+
```
ddietterich, line 365: Why is the RESOURCE_ID_TO_REGION_MAP no longer needed?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/backfill/RetrieveControlledBigQueryDatasetsLifetimeStep.java
```diff
@@ -0,0 +1,141 @@
...
+    List<ControlledBigQueryDatasetResource> controlledBigQueryDatasets =
+        Preconditions.checkNotNull(
+            context
+                .getWorkingMap()
+                .get(CONTROLLED_BIG_QUERY_DATASETS_WITHOUT_LIFETIME, new TypeReference<>() {}));
+    Map<UUID, String> resourceIdToDefaultTableLifetimeMap = new HashMap<>();
+    Map<UUID, String> resourceIdToDefaultPartitionLifetimeMap = new HashMap<>();
+    Map<UUID, String> resourceIdToWorkspaceIdMap = new HashMap<>();
+
+    for (var resource : controlledBigQueryDatasets) {
```
ddietterich, line 53: Nit: non-obvious `var`

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/backfill/RetrieveControlledBigQueryDatasetsLifetimeStep.java
```diff
@@ -0,0 +1,141 @@
...
+          "Getting default table lifetime and partition life for resource (BigQuery dataset) {} in workspace {}",
+          resource.getResourceId(),
+          resource.getWorkspaceId());
+
+      populateMapsWithResourceIdKey(
+          resourceIdToDefaultTableLifetimeMap,
+          resourceIdToDefaultPartitionLifetimeMap,
+          resourceIdToWorkspaceIdMap,
+          resource,
+          getBqDatasetDefaultTableLifetimeAndPartitionLifetime(resource));
```
ddietterich, line 64: Nit: Why pass this as a parameter? I think it would be clearer if you called it inside the populate code.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/backfill/UpdateControlledBigQueryDatasetsLifetimeFlight.java
```diff
@@ -0,0 +1,29 @@
...
+  public UpdateControlledBigQueryDatasetsLifetimeFlight(FlightMap inputParameters, Object beanBag) {
+    super(inputParameters, beanBag);
+    FlightBeanBag flightBeanBag = FlightBeanBag.getFromObject(beanBag);
+
+    addStep(
+        new RetrieveControlledBigQueryDatasetsWithoutLifetimeStep(flightBeanBag.getResourceDao()));
+
+    addStep(
+        new RetrieveControlledBigQueryDatasetsLifetimeStep(flightBeanBag.getCrlService()),
+        RetryRules.shortExponential());
```
ddietterich, line 25: I don't see anything retry-able in the step. Are you expecting this retry will do something?

ddietterich (approved):
A bunch of nits to fix up, but otherwise looks good.

service/src/main/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHook.java
aaron-wang: The `maybeLogUpdateControlledResourceFieldsFlight` logs the GCP region, Azure region, and the BQ lifetime backfill update flights.

Before, the code was specific to the two region flights. Now, it's generalized to include the BQ lifetime flight (see Yu's comment from earlier: https://github.com/DataBiosphere/terra-workspace-manager/pull/1041#discussion_r1104640810).

yuhuyoyo (approved):

PR #1043, opened 2023-01-30
PF-2172: Integrate regional policy check to controlled resource create (re-add)

mikenev:
The previous commit broke 2 things and was reverted.
1) Creating Azure workspaces - since they don't pass a region to validate
2) Working with GCP workspaces created without policies - caused policy not found exceptions.

This PR re-introduces the changes to support regional policy checks, but also fixes those two pieces.

This is like the previous PR with a couple small changes I'll call out.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
@@ -200,7 +200,11 @@ public class ResourceValidationUtils {
   public static void validateControlledResourceRegionAgainstPolicy(
       TpsApiDispatch tpsApiDispatch, UUID workspaceUuid, String location, CloudPlatform platform) {
     switch (platform) {
-      case AZURE -> validateAzureRegion(location);
+      case AZURE -> {
+        // TODO: enable policy check in Azure when we support Azure regions in the TPS ontology.
+        // validateAzureRegion(location);
+        return;
+      }
       case GCP -> validateGcpRegion(tpsApiDispatch, workspaceUuid, location);
       default -> throw new InvalidControlledResourceException("Unrecognized platform");
     }
--- a/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
@@ -484,6 +484,9 @@ public class ResourceValidationUtils {
         throw new InvalidControlledResourceException(
             String.format("Specified location %s is not allowed by effective policy.", region));
       }
+    } else {
+      logger.warn(
+          "Workspace {} has no policy attached, cannot validate region constraints.", workspaceId);
     }
   }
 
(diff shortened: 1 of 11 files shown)


service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
```diff
@@ -441,14 +455,35 @@ public static void validateAzureVmSize(String vmSize) {
     }
   }
 
-  public static void validateRegion(String region) {
+  public static void validateAzureRegion(String region) {
+    if (StringUtils.isEmpty(region)) {
+      // Azure resources like workspaces may not have a region.
```
mikenev, line 460: This is one change to account for Azure workspace creation.

service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
```diff
@@ -441,14 +455,35 @@ public static void validateAzureVmSize(String vmSize) {
...
+      throw new InvalidControlledResourceException("Invalid Azure Region specified.");
+    }
+  }
+
+  public static void validateGcpRegion(
+      TpsApiDispatch tpsApiDispatch, UUID workspaceId, String region) {
+    region = GcpUtils.parseRegion(region);
+
+    // Old workspaces may not have a policy. We can only validate against policy if one exists.
+    if (tpsApiDispatch.getPaoIfExists(workspaceId).isPresent()) {
```
mikenev, line 478: Another change to verify we have a policy before checking regions against it.

service/src/test/java/bio/terra/workspace/service/resource/ValidationUtilsTest.java
```diff
@@ -452,4 +460,80 @@ public void validateProperties_folderIdNotUuid_throwsBadRequestException() {
...
+        InvalidControlledResourceException.class,
+        () ->
+            validationUtils.validateControlledResourceRegionAgainstPolicy(
+                mockTpsApiDispatch(), workspaceId, "badregion", CloudPlatform.AZURE));
+  }
+
+  @Test
+  public void validateControlledResourceRegion_workspaceWithoutPolicy() {
+    UUID workspaceId = UUID.randomUUID();
+    when(mockTpsApiDispatch().getPao(workspaceId))
```
mikenev, line 505: Another test for when the workspace doesn't have a policy attached.

service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
ddietterich: Should you have a `logger.warn` here also?

ddietterich (approved):
LGTM - one suggestion

service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
mikenev: Good idea

yuhuyoyo (approved):

PR #1045, opened 2023-01-31
PF-2368: More on gcs bucket with underscore name.

yuhuyoyo:
Still getting Json deserilization exception from getting the resources via typeReference and castByEnum.

Make this an endpoint so we don't print out a blob of deserilization failure everytime we start up wsm.


ddietterich (approved):
lgtm

zloery (approved):

PR #1047, opened 2023-02-01
PF-2448: Add mergeCheck endpoint to check policy conflicts

aaron-wang:
**Edit:** Add a `mergeCheck` endpoint that calls `mergePao` strictly with the `DRY_RUN` update mode.


openapi/src/parts/tps.yaml
```diff
@@ -258,4 +258,11 @@ components:
             $ref: '#/components/schemas/WsmPolicyLocation'
         regions:
           $ref: '#/components/schemas/Regions'
- 
+
+    WsmPolicySourceRequestBody:
+      description: Request structure shared by merge
```
yuhuyoyo, line 263: Request with a WSM Policy object id (e.g. a workspace id)

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
```diff
@@ -350,6 +360,78 @@ public void explainPolicies() throws Exception {
...
+        mergeCheck(userAccessUtils.defaultUserAuthRequest(), workspace.getId(), workspace.getId());
+
+    assertEquals(0, result.getConflicts().size());
+  }
+
+  @Test
+  @EnabledIf(expression = "${feature.tps-enabled}", loadContext = true)
+  public void mergeCheck_differentRegion() throws Exception {
+    // Create workspace with US region constraint.
+    ApiCreateWorkspaceRequestBody targetRequest =
```
yuhuyoyo, line 376: ah you can extract l376 - l396 to a helper method createWorkspaceWithRegionConstraint(String regionName)

yuhuyoyo (approved):

cbookg:
So in discussion with @mikenev I think there's two things this API needs to check to know if Workspace A can take a policy from Workspace B:
1) Does workspace A's policy conflict with Workspace B's policy.
2) Does workspace A have any controlled resources that would conflict with Workspace B's policy.

I believe this API is solving #1, but not #2.  Mike has said he's doing this check as part of this PR to modify the clone resource API, so perhaps that logic can be added here.

openapi/src/parts/tps.yaml
```diff
@@ -258,4 +258,11 @@ components:
             $ref: '#/components/schemas/WsmPolicyLocation'
         regions:
           $ref: '#/components/schemas/Regions'
- 
+
+    WsmPolicySourceRequestBody:
```
ddietterich, line 262: Since we are in the WSM API, I would expect this to always be the workspace id, not a generic `sourceObjectId`.

Since this body is only used by the merge check, I would expect it to live in the `workspace.yaml` and be called `MergeCheckRequest` in line with other schema objects.

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
```diff
@@ -350,6 +355,34 @@ public void explainPolicies() throws Exception {
     assertEquals(0, result.getExplanation().size());
   }
 
+  @Test
+  @EnabledIf(expression = "${feature.tps-enabled}", loadContext = true)
```
ddietterich, line 359: Oh, nice one! Didn't know about EnabledIf.

ddietterich (commented):
I would like some name changes. Otherwise, the code looks fine.

ddietterich:
> should we change the response and logic of the mergeCheck endpoint?

As Chris suggests, this endpoint should check both the TPS state and whether the workspace has conflicting resources.
Seems like the error response should be clear about whether the conflict is due to existing resources (those can be fixed) or due to direct policy conflict (that cannot easily be addressed).

aaron-wang:
@ddietterich Ok, the endpoint has been updated to additionally check resource conflicts. Ready for a review :)

Is there a better way to indicate the type of conflict error (rather than having different text in the message body)?

openapi/src/parts/workspace.yaml
```diff
@@ -356,6 +356,34 @@ paths:
...
+      operationId: mergeCheck
+      tags: [ Workspace ]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MergeCheckRequest'
+      responses:
+        '200':
```
ddietterich, line 379: Since there is no response body, this would be a 204 - NO_CONTENT

I think you need to specify 409 - CONFLICT, since that is what you are throwing when there is a conflict.

I'll leave it to @cbookg to say whether getting an error message is sufficient, or if you would like a response body with more explanation for the conflict.

The TPS response for merge is a 200 even when there is a conflict so that we can return the conflict information on the response.

ddietterich (commented):
There is a little cleanup to do in the YAML. Please discuss with Chris whether the error return is sufficient for the UI or more info is needed.

openapi/src/parts/workspace.yaml
aaron-wang: After talking with Chris, we decided to always return a successful response with a response body (even when there are conflicts)

So, the endpoint now returns:
1. a list of conflicts (between the workspace policies themselves)
2. a list of existing resources with conflicts (if the other workspace was to be merged into the target)

Then, the UI can infer what to do next.

openapi/src/parts/tps.yaml
```diff
@@ -133,6 +133,24 @@ components:
...
+      properties:
+        conflicts:
+          description: |
+            List of workspace conflicts.
+          type: array
+          items:
+            $ref: '#/components/schemas/WsmPolicyConflict'
+        resourcesWithConflict:
+          description: |
+            List of resources (by ID) in the target workspace with conflicting policy.
```
yuhuyoyo, line 148: curious about the schema here.

would it be helpful when the resourcesWithConflict contains an array of  resource uuid + the conflicts so we know exactly why resource A is a conflicted resources?

ddietterich: I note that the caller has to test for two empty lists to decide if the check succeeded. Seems OK to me if it is OK with the UI folks.

ddietterich (approved):
LGTM

PR #1049, opened 2023-02-01
PF-1687 - Validate region constraints when adding data collections

mikenev:
The operation for adding data collection resources into your workspace show as cloning referenced resources to your workspace. 
When this happens, we need to check the policy on the data collection (workspace) and verify any region policies that exist there are compatible with region policies (if any) on the target workspace. Those need to be merged with the least privilege. Calculating that should be handled by the policy service.
When adding resources into the current workspace, we need to also verify that no resources already exist in the workspace that would violate the policy after merging.


service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
```diff
@@ -68,7 +78,17 @@ public ReferencedResource createReferenceResourceForClone(
       ReferencedResource resourceToClone,
       ReferencedResource sourceReferencedResource,
       AuthenticatedUserRequest userRequest) {
+
```
yuhuyoyo, line 81: should we do this check for all cloning endpoint? I know the UI don't let you clone a workspace partially right now but the wsm service should handle that flow correctly regarding policy.

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
```diff
@@ -248,4 +268,32 @@ public ReferencedResource cloneReferencedResource(
...
         destinationResource, sourceReferencedResource, userRequest);
   }
+
+  private void validateDestinationWorkspacePolicies(
+      UUID sourceWorkspaceId, UUID destinationWorkspaceId, CloudPlatform platform) {
+    TpsPaoUpdateResult dryRunResults =
+        tpsApiDispatch.mergePao(sourceWorkspaceId, destinationWorkspaceId, TpsUpdateMode.DRY_RUN);
+
+    if (!dryRunResults.getConflicts().isEmpty()) {
+      throw new ConflictException("Policy merge has conflicts");
```
yuhuyoyo, line 278: should we include the content of the conflict in the exception?

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
ddietterich: We do need the check for all clone paths. Any clone needs to validate and merge, not only the reference case; the copy controlled resource definition and copy controlled resource data cases, too.

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
```diff
@@ -68,7 +78,17 @@ public ReferencedResource createReferenceResourceForClone(
       ReferencedResource resourceToClone,
       ReferencedResource sourceReferencedResource,
       AuthenticatedUserRequest userRequest) {
+
+    validateDestinationWorkspacePolicies(
+        sourceReferencedResource.getWorkspaceId(),
+        resourceToClone.getWorkspaceId(),
+        sourceReferencedResource.getResourceType().getCloudPlatform());
     resourceDao.createReferencedResource(resourceToClone);
+    var result =
```
ddietterich, line 87: I think this has to happen in a flight. If the `mergePao` fails, we need to delete the created referenced resource. We need it to behave transactionally.

There is no guarantee that the dry run test will be valid for the actual merge.

service/src/main/java/bio/terra/workspace/service/resource/referenced/ReferencedResourceService.java
```diff
@@ -248,4 +268,32 @@ public ReferencedResource cloneReferencedResource(
...
+      UUID sourceWorkspaceId, UUID destinationWorkspaceId, CloudPlatform platform) {
+    TpsPaoUpdateResult dryRunResults =
+        tpsApiDispatch.mergePao(sourceWorkspaceId, destinationWorkspaceId, TpsUpdateMode.DRY_RUN);
+
+    if (!dryRunResults.getConflicts().isEmpty()) {
+      throw new ConflictException("Policy merge has conflicts");
+    }
+
+    if (platform == CloudPlatform.ANY) {
+      logger.warn("Cannot validate regions for CloudPlatform.ANY");
```
ddietterich, line 282: I think this is a normal case for things like datarepo snapshots and github repos. So maybe not worth logging.

ddietterich (changes requested):
I think we need to move the reference creation into a flight.

integration/src/main/java/scripts/testscripts/ImportDataCollection.java
```diff
@@ -0,0 +1,454 @@
...
+            .getMessage()
+            .contains("The specified destination location violates region policies"));
+
+    workspaceApi.deleteWorkspace(scenario8Workspace.getId());
+
+    // Clean up the data collection used in most of the scenarios.
+    workspaceApi.deleteWorkspace(centralDataCollection.getId());
+  }
+
+  private CreatedWorkspace CreateWorkspaceWithRegionPolicy(
```
ddietterich, line 398: nit - methods should start with lowercase letter

integration/src/main/resources/suites/PRIntegration.json
```diff
@@ -8,6 +8,7 @@
     "integration/DeleteWorkspaceWithControlledResource.json",
     "integration/EnumerateResources.json",
     "integration/GetRoles.json",
+    "integration/ImportDataCollection.json",
```
ddietterich, line 11: How long does this test take to run?  If it is long-running, it should probably be in the Full suite

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CloneControlledGcpBigQueryDatasetResourceFlight.java
```diff
@@ -70,9 +73,25 @@ public CloneControlledGcpBigQueryDatasetResourceFlight(
         RetryRules.shortExponential());
     if (mergePolicies) {
       addStep(
-          new ClonePolicyAttributesStep(
+          new MergePolicyAttributesDryRunStep(
```
ddietterich, line 76: Could this step generation be factored out and reused across controlled resource clones?
Might be more trouble than its worth, but it will be a burden to visit duplicate places (if they are in fact duplicate...)

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CloneControlledGcpBigQueryDatasetResourceFlight.java
```diff
@@ -70,9 +73,25 @@ public CloneControlledGcpBigQueryDatasetResourceFlight(
...
+              sourceResource.getResourceType().getCloudPlatform(),
+              destLocation,
+              userRequest,
+              flightBeanBag.getResourceDao(),
+              flightBeanBag.getTpsApiDispatch()));
+
+      addStep(
+          new MergePolicyAttributesStep(
+              destinationWorkspaceId,
+              sourceResource.getWorkspaceId(),
```
ddietterich, line 94: Does this flight need the `ValidateWorkspaceAgainstPolicyStep` like buckets?

service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/clone/CloneReferenceResourceStep.java
```diff
@@ -0,0 +1,106 @@
...
+          referencedResourceService.createReferenceResourceForClone(
+              destinationResource, sourceResource, userRequest);
+      FlightUtils.setResponse(context, createdResource, HttpStatus.OK);
+      cloneDetails.setResult(WsmCloneResourceResult.SUCCEEDED);
+    } catch (Exception e) {
+      cloneDetails.setResult(WsmCloneResourceResult.FAILED).setErrorMessage(e.getMessage());
+    }
+
+    // add to the map
+    addCloneDetailsToWorkingMap(context, cloneDetails);
```
ddietterich, line 67: The error case appears to fall through. Is that what you intend?
Maybe some comment about what is going on?

service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/clone/CloneReferenceResourceStep.java
```diff
@@ -0,0 +1,106 @@
...
+      var createdResource =
+          referencedResourceService.createReferenceResourceForClone(
+              destinationResource, sourceResource, userRequest);
+      FlightUtils.setResponse(context, createdResource, HttpStatus.OK);
+      cloneDetails.setResult(WsmCloneResourceResult.SUCCEEDED);
+    } catch (Exception e) {
+      cloneDetails.setResult(WsmCloneResourceResult.FAILED).setErrorMessage(e.getMessage());
+    }
+
+    // add to the map
```
ddietterich, line 66: nit - useless comment

service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/clone/SetCloneReferencedResourceResponseStep.java
```diff
@@ -0,0 +1,32 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import bio.terra.workspace.common.utils.FlightUtils;
+import bio.terra.workspace.service.resource.referenced.model.ReferencedResource;
+import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys;
+import org.springframework.http.HttpStatus;
+
+public class SetCloneReferencedResourceResponseStep implements Step {
```
ddietterich, line 12: I didn't see where this step was used. Maybe I missed it. Is it used?

ddietterich (commented):
Looking good. I have a few comments and questions.

integration/src/main/resources/suites/PRIntegration.json
mikenev: It does take a long time ~20 min. Moved to the Full suite.

service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/clone/SetCloneReferencedResourceResponseStep.java
mikenev: It was, but it's not used anymore. Removed.

service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/clone/CloneReferenceResourceStep.java
mikenev: it should fail here. Fixed.

service/src/main/java/bio/terra/workspace/service/policy/flight/MergePolicyAttributesUtils.java
```diff
@@ -0,0 +1,20 @@
...
+import bio.terra.policy.model.TpsPaoGetResult;
+import bio.terra.workspace.service.policy.TpsApiDispatch;
+import java.util.Optional;
+import java.util.UUID;
+
+public class MergePolicyAttributesUtils {
+  // Since policy attributes were added later in development, not all existing
+  // workspaces have an associated policy attribute object. This method creates
+  // an empty one if it does not exist.
+  protected static void createPaoIfNotExist(TpsApiDispatch tpsApiDispatch, UUID workspaceId) {
```
yuhuyoyo, line 12: this seems very useful. 

should we just put that into TpsApiDispatch and do that for all the TpsApiDispatch method that's modifying policy on a workspace?

service/src/main/java/bio/terra/workspace/service/policy/flight/ValidateWorkspaceAgainstPolicyStep.java
```diff
@@ -0,0 +1,88 @@
...
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    final TpsPaoGetResult effectivePolicies =
+        flightContext
+            .getWorkingMap()
+            .get(WorkspaceFlightMapKeys.EFFECTIVE_POLICIES, TpsPaoGetResult.class);
+
+    if (cloudPlatform != CloudPlatform.GCP) {
+      // We can only validate GCP right now. We'll need to add other platforms
```
yuhuyoyo, line 51: i thought tps knows azure regions

service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/clone/CloneReferenceResourceStep.java
```diff
@@ -0,0 +1,106 @@
...
+      var createdResource =
+          referencedResourceService.createReferenceResourceForClone(
+              destinationResource, sourceResource, userRequest);
+      FlightUtils.setResponse(context, createdResource, HttpStatus.OK);
+      cloneDetails.setResult(WsmCloneResourceResult.SUCCEEDED);
+    } catch (Exception e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+    }
+
+    addCloneDetailsToWorkingMap(context, cloneDetails);
```
yuhuyoyo, line 67: i don't think we need this. we are not waiting for the clone of referenced resources.

service/src/main/java/bio/terra/workspace/service/policy/flight/ValidateWorkspaceAgainstPolicyStep.java
```diff
@@ -0,0 +1,88 @@
...
+      if (existingResource.getRegion() == null) {
+        // Some resources don't have regions. IE: Git repos.
+        continue;
+      }
+      if (!validRegions.contains(existingResource.getRegion().toLowerCase())) {
+        throw new PolicyConflictException("Workspace contains resources in violation of policy.");
+      }
+    }
+
+    if (destinationLocation != null && !validRegions.contains(destinationLocation.toLowerCase())) {
```
yuhuyoyo, line 75: nit: you can use equalsIgnoreCase()

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/policy/flight/MergePolicyAttributesUtils.java
mikenev: That is a really good idea. I think I'll do that as part of PF-2502 so I don't keep expanding the scope of this review.

service/src/main/java/bio/terra/workspace/service/policy/flight/ValidateWorkspaceAgainstPolicyStep.java
mikenev: It's built to support it, but it doesn't do much with it right now. We took Azure out of the Ontology and aren't validating regions on resource create right now.

service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/clone/CloneReferenceResourceStep.java
mikenev: You're right, we aren't using that in this case. Removed.

service/src/main/java/bio/terra/workspace/service/policy/flight/ValidateWorkspaceAgainstPolicyStep.java
mikenev: I didn't see a way to do that with contains(). There are ways it can be done, but I'm not sure it makes it more succinct.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CloneControlledGcpBigQueryDatasetResourceFlight.java
mikenev: I wondered about that. I started going down that path but it felt like you mentioned - that it was more trouble than it was worth. If you feel strongly about it, I can look into factoring it out again.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/dataset/CloneControlledGcpBigQueryDatasetResourceFlight.java
ddietterich: I agree, it seems like overkill.

ddietterich (approved):

PR #1050, opened 2023-02-02
PF-2468: Correct clone log

yuhuyoyo:
So the log is written on the destination workspace id. But the change_subject_id should be the source because that is what is cloned.


yuhuyoyo:
i will add a ticket and attach later

ddietterich (approved):

zloery (approved):

PR #1051, opened 2023-02-03
[PF-2469] Add jersey-hk2 inject

rogerwangcs:
Add wsm openapi client depedency needed by terra-cli and terra-axon-server to properly build and and use the wsm client.

https://github.com/DataBiosphere/terra-cli/blob/main/build.gradle#L57


zloery (approved):

yuhuyoyo (approved):

client/gradle/dependencies.gradle
```diff
@@ -10,6 +10,7 @@ dependencies {
   implementation group: "org.glassfish.jersey.media", name: "jersey-media-json-jackson", version: jerseyVersion
   implementation group: "org.glassfish.jersey.media", name: "jersey-media-multipart", version: jerseyVersion
   implementation group: "org.glassfish.jersey.connectors", name: "jersey-jdk-connector", version:jerseyVersion
+  implementation group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: jerseyVersion
```
ddietterich, line 13: I _think_ for this to be consumed by users of the jar, it should be `application` instead of `implementation`.

rogerwangcs: Ah, this helped point me to https://docs.gradle.org/current/userguide/library_vs_application.html). Looks like the keyword is `api`.

PR #1053, opened 2023-02-07
Fix OpenAPI enums

harrismendell:
Enums in this file have type: object, which is breaking Enterprise Terra's Typescript codegen, as I believe the enums are meant to be strings.


ddietterich (approved):

harrismendell:
Hi @sergiygetlin mind taking a look? Let me know if you have any questions / concerns. Thanks!

zloery (approved):

PR #1055, opened 2023-02-08
TOAZ-314 Adjust uami payload for azure batch pool creation endpoint

sergiygetlin:
Adjust payload for user assigned identities for batch pool:
-make resourceGroupName parameter optional
-add clientId parameter. Batch pool API required UAMI name, but when this parameter is set UAMI and it's name will be found using clientId.
-make name and clientId parameters mutually exclusive

### DIFF
--- a/openapi/src/parts/controlled_azure_batch_pool.yaml
+++ b/openapi/src/parts/controlled_azure_batch_pool.yaml
@@ -154,14 +154,21 @@ components:
 
     AzureBatchPoolUserAssignedIdentity:
       type: object
-      description: User assigned identity associated with a batch pool
-      required: [ name, resourceGroupName ]
+      description: User assigned identity associated with a batch pool.
       properties:
         name:
-          description: The name of user assigned identity.
+          description: |
+            The name of the user assigned identity. This parameter is mutually exclusive with clientId.
           type: string
+        clientId:
+          description: |
+            The client id of the UAMI. This parameter is mutually exclusive with name. If this parameter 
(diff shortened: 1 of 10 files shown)


giventocode (approved):
LGTM

openapi/src/parts/controlled_azure_batch_pool.yaml
```diff
@@ -154,14 +154,21 @@ components:
...
     AzureBatchPoolUserAssignedIdentity:
       type: object
-      description: User assigned identity associated with a batch pool
-      required: [ name, resourceGroupName ]
+      description: User assigned identity associated with a batch pool.
       properties:
         name:
-          description: The name of user assigned identity.
+          description: |
+            The name of the user assigned identity. This parameter is mutually exclusive with clientId.
```
rtitle, line 161: Given `name` and `clientId` are mutually exclusive, how would a caller decide to pass one or the other? Are there circumstances where `name` is required and others where `clientId` is required?

rtitle (approved):

openapi/src/parts/controlled_azure_batch_pool.yaml
sergiygetlin: It depends on what a caller has. Both options will give the same result. But Azure Batch SDK accept name of managed identity. So, in case a caller provide name it is straightforward to call corresponding API. Otherwise, we need to look up for it using clientId. So, it is an additional call to Azure.

zykonda (approved):
lgtm

PR #1057, opened 2023-02-09
[WOR-757] Fail Azure workspace/cloud context deletion if there are controlled resources that user cannot delete

marctalbott:
Ticket: [WOR-757](https://broadworkbench.atlassian.net/browse/WOR-757)
- Users cannot delete controlled application resources in their workspace. We need to check that a user can actually delete all controlled resources in their workspace _before_ we start deleting the cloud objects, not after.

[WOR-757]: https://broadworkbench.atlassian.net/browse/WOR-757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteControlledAzureResourcesStepTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteControlledAzureResourcesStepTest.java
@@ -1,5 +1,12 @@
 package bio.terra.workspace.service.workspace.flight;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.StepResult;
 import bio.terra.stairway.StepStatus;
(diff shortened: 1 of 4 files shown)


blakery (approved):

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteControlledAzureResourcesStepTest.java
```diff
@@ -0,0 +1,109 @@
...
+import bio.terra.workspace.service.resource.controlled.model.ManagedByType;
+import bio.terra.workspace.service.resource.model.CloningInstructions;
+import bio.terra.workspace.service.workspace.model.CloudPlatform;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+
+public class DeleteControlledAzureResourcesStepTest extends BaseAzureUnitTest {
```
cahrens, line 33: Should we also add a test of the happy path? There is some coverage in `deleteMcWorkspaceWithAzureContextAndResource`, but it seems odd that this file didn't already have a unit test.

cahrens (approved):
Consider adding a happy path test as well.

service/src/test/java/bio/terra/workspace/service/workspace/flight/DeleteControlledAzureResourcesStepTest.java
marctalbott: I was on the fence about this, and this is enough to push me over. I'll add it

cahrens: Thanks!

PR #1061, opened 2023-02-09
[PF-2478] Only run checks when PR targets 'main' branch

dexamundsen:
Only run PR checks when the PR targets main/master branch

currently, all PRs (irrespective of the branch they target) automatically trigger PR checks. This is unnecessary and may cause the checks to be queued up on GitHub runner.

### DIFF
--- a/.github/workflows/linter-tests.yml
+++ b/.github/workflows/linter-tests.yml
@@ -6,7 +6,7 @@ on:
     - main
   pull_request:
     branches:
-    - '**'
+    - main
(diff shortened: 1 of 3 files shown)


ddietterich (approved):
This seems like a reasonable change to me. AFAIK, we got the '**' via cut'n'paste from other repos, not from any decision about where it was needed.

I'll be interested to hear what others think.

zloery (approved):

PR #1063, opened 2023-02-10
TOAZ-316 Enrich payload for Batch pool creation with metadata

sergiygetlin:
Add 'metadata' to the batch pool create payload which is required by TES


giventocode (approved):
LGTM!

zykonda (approved):
lgtm

PR #1068, opened 2023-02-15
fix on the client side?

yuhuyoyo:
I found this PR https://github.com/DataBiosphere/terra-workspace-manager/pull/637/files

my attempt in tps: https://github.com/DataBiosphere/terra-policy-service/pull/43/files but it doesn't work.


service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
```diff
@@ -248,6 +251,9 @@ private ApiClient getApiClient(String accessToken) {
             .addDefaultHeader(
                 RequestIdFilter.REQUEST_ID_HEADER, MDC.get(RequestIdFilter.REQUEST_ID_MDC_KEY));
     client.setAccessToken(accessToken);
+    ClientConfig clientConfig = new ClientConfig();
```
ddietterich, line 254: This fix should be made on the TPS side in the TPS client.
If you look at the way the WSM client is set up, it provides mustache template overrides that configure the JdkConnector. That is, when WSM does a `new ApiClient()` the generated client code would do that bit of configuration instead of making every caller do it.

We need to copy the whole source tree from WSM [here](https://github.com/DataBiosphere/terra-workspace-manager/tree/main/client/src/main/resources/swaggercodegen) and put it in TPS.

And tell the TPS codegen to use it, like WSM does [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/ee97989509e24f3efbe2a50067caab3aaf9353d1/client/gradle/openapi.gradle#L12)

yuhuyoyo: oh templateDir.

i will add you to the tps pr.

PR #1069, opened 2023-02-15
PF-2458: Add Dataproc permissions

rogerwangcs:
Also add dataproc cluster/job read permissions for READER. With this change,
workspace owners can:
1. Create Dataproc clusters, submit jobs to them, and apply autoscaling policies via cloud console or `gcloud`/`terra gcloud`
2. Open jupyterlab and other web interfaces via cloud console
and workspace readers can:
1. List clusters and jobs and view their details

The tasks/agents/compute/monitoring permissions are required for users to successfully create clusters.


ddietterich (commented):
I don't think this will work as written. The feature flag configuration relies on Spring setup. If you `new` it, you won't get that configuration.

The problem is that in order to `@Autowire` the flags into the module, you would have to make it a `@Component`. That would mean that we would have to pass the thing into all of the flights and steps that need it. That might be the best idea, since I expect we will need to be flexible about which project permission we use. Not only as a feature flag, but also as a parameter on the workspace.

Another (somewhat hacky) approach would be to make the class an `@Component` and give it a `@PostConstruct` to store the address of the class in a static. Then steps that use it could access the static. We use that trick in a few places in the code.

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -110,23 +113,91 @@ public class CloudSyncRoleMapping {
...
+              "compute.networks.get",
+              "compute.networks.list",
+              "compute.projects.get",
+              "compute.regions.get",
+              "compute.regions.list",
+              "compute.zones.get",
+              "compute.zones.list")
+          .build();
+
+  public CloudSyncRoleMapping() {
```
zloery, line 174: I don't think this will work as-is. FeatureConfiguration is a Spring component (because it's annotated with `@Configuration` so the FeatureConfiguration singleton will have values from `application.yml`, but a newly-created FeatureConfiguration object (like here) won't have anything.
There are two approaches you could take here:

1. Change this constructor to take a FeatureConfiguration object as a parameter, and have all callers pass in the FeatureConfiguration bean. Inside a flight you can get this from the `FlightBeanBag`, and you can then pass it to steps as needed

2. Make `CloudSyncRoleMapping` a Spring component as well. Then, you can add a FeatureConfiguration parameter to the constructor and use `@Autowired` to ensure this is wired up to the FeatureConfiguration bean. To call this component's methods inside flights, you would add `CloudSyncRoleMapping` to FlightBeanBag and then pass the object from that to all steps as needed

Either approach should work, but I'd probably lean towards 2 a bit

rogerwangcs: Ah thanks! Flight bean bag was the missing piece.

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -110,23 +113,91 @@ public class CloudSyncRoleMapping {
...
+    this.featureConfiguration = new FeatureConfiguration();
```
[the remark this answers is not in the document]
rogerwangcs: I had tested all the dataproc permissions before I moved it behind a flag. Just added CloudSyncRoleMappings as spring component to flight bean bag. The flag works correctly now.

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -110,23 +113,91 @@ public class CloudSyncRoleMapping {
...
+  }
+
+  public ImmutableMap<WsmIamRole, CustomGcpIamRole> getCustomGcpProjectIamRoles() {
+    CustomGcpIamRole PROJECT_READER =
```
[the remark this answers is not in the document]
rogerwangcs: No longer using static methods/fields.

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -110,23 +116,90 @@ public class CloudSyncRoleMapping {
...
+              "dataproc.jobs.list")
+          .build();
+
+  private final List<String> projectOwnerDataprocPermissions =
+      new ImmutableList.Builder<String>()
+          .addAll(projectReaderDataprocPermissions)
+          .addAll(projectOwnerPermissions)
+          .add(
+              // Dataproc CRUD permissions
+              // TODO: Move read permissions to projectReader and revoke modify permissions once
```
yuhuyoyo, line 135: please attach ticket to TODO.
what does revoke modify permissions once mean?

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
```diff
@@ -110,23 +116,90 @@ public class CloudSyncRoleMapping {
...
+
+  public ImmutableMap<WsmIamRole, CustomGcpIamRole> getCustomGcpProjectIamRoles() {
+    CustomGcpIamRole projectReader =
+        CustomGcpIamRole.of("PROJECT_READER", projectReaderPermissions);
+    CustomGcpIamRole projectWriter =
+        CustomGcpIamRole.of("PROJECT_WRITER", projectWriterPermissions);
+    CustomGcpIamRole projectOwner = CustomGcpIamRole.of("PROJECT_OWNER", projectOwnerPermissions);
+    if (featureConfiguration.isDataprocEnabled()) {
+      projectReader = CustomGcpIamRole.of("PROJECT_READER", projectReaderDataprocPermissions);
+      projectOwner = CustomGcpIamRole.of("PROJECT_OWNER", projectOwnerDataprocPermissions);
```
yuhuyoyo, line 189: what about writer? i think they should also have data proc permission

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
yuhuyoyo: +1 for 2.

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
rogerwangcs: Because the permissions allow for direct resource creation on gcp, it made sense to limit them to the most priveleged workspace users. Also this isn't the intended long term way of allowing sharing and collaboration for dataproc.

yuhuyoyo: but why does reader can access data proc while writer can't.

rogerwangcs: Oh right for read permissions. Let me add read permissions for writers as well.

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
rogerwangcs: Attached ticket and updated comment. Remove all permissions in projectOwnerDataprocPermissions role and then run cloudIamSync endpoint.

service/src/main/java/bio/terra/workspace/service/workspace/CloudSyncRoleMapping.java
yuhuyoyo: personally for consistency i still think writer should be able to create data proc because writer are able to create other resources in terra workspace.

service/src/test/java/bio/terra/workspace/service/workspace/CloudSyncRoleMappingTest.java
```diff
@@ -7,47 +7,57 @@
...
 import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
 
 /**
  * These tests assert architectural design invariants about the workspace role hierarchy. These are
  * not expected to change.
  */
 public class CloudSyncRoleMappingTest extends BaseUnitTest {
+
+  @Autowired CloudSyncRoleMapping cloudSyncRoleMapping;
```
yuhuyoyo, line 18: can you add test for flag ON?

service/src/test/java/bio/terra/workspace/service/admin/AdminServiceTest.java
```diff
@@ -52,6 +52,7 @@ public class AdminServiceTest extends BaseConnectedTest {
   @Autowired WorkspaceDao workspaceDao;
   @Autowired WorkspaceConnectedTestUtils connectedTestUtils;
   @Autowired UserAccessUtils userAccessUtils;
+  @Autowired CloudSyncRoleMapping cloudSyncRoleMapping;
```
yuhuyoyo, line 55: can you add test for flag on?

service/src/test/java/bio/terra/workspace/service/workspace/CloudSyncRoleMappingTest.java
rogerwangcs: Added tests for flag on and flag off

yuhuyoyo (approved):

ddietterich (approved):
LGTM

PR #1070, opened 2023-02-16
PF-2407: Update tps version and also add connected test

yuhuyoyo:


service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
```diff
@@ -443,6 +446,79 @@ public void listValidRegions() throws Exception {
...
+
+    assertTrue(result.isUpdateApplied());
+    ApiWorkspaceDescription updatedWorkspace =
+        mockMvcUtils.getWorkspace(userAccessUtils.defaultUserAuthRequest(), workspace.getId());
+    assertEquals(1, updatedWorkspace.getPolicies().size());
+
+    mockMvcUtils.updatePoliciesExpect(
+        userAccessUtils.defaultUserAuthRequest(),
+        workspace.getId(),
+        HttpStatus.SC_INTERNAL_SERVER_ERROR,
```
ddietterich, line 500: I'm surprised this is a 500 error. Where is that coming from?
If this is a conflict, we should be returning success with conflict information.

So see, the test already has found a bug! :)

Not sure if we should apply the test until we get that straighted out. wdyt?

yuhuyoyo: oh ok. I wasn't sure what's the right behavior is.

the fix should be in TPS right? we are getting 409 from TPS and then wrap it in a 500 in TpsApidispatch. are you saying updatePao should return 200 for ENFORCE_CONFLICT and put conflict information in the response?

yuhuyoyo: 409 are wrapped now.

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
```diff
@@ -443,6 +446,77 @@ public void listValidRegions() throws Exception {
...
 
+  @Test
+  @EnabledIf(expression = "${feature.tps-enabled}", loadContext = true)
+  public void updatePolicies_tpsEnabledAndPolicyUpdated() throws Exception {
+    ApiWorkspaceDescription workspaceWithoutPolicy =
+        mockMvcUtils.getWorkspace(userAccessUtils.defaultUserAuthRequest(), workspace.getId());
+    assertTrue(workspaceWithoutPolicy.getPolicies().isEmpty());
+
+    // add attributes
+    String usRegion = "US";
```
mikenev, line 457: This looks ok, but I'm not sure why we have capital "US" here. That's not one of the regular regions. And we use lower case "us" on 487. Maybe we should use the "usa" region?

yuhuyoyo: it does allow me to update policy attributes to US though. i thought it's tps being smart. is it a bug?

mikenev (approved):

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
mikenev: We don't currently validate the value of the constraint. So you can set it to US if you want. If you attempt to merge that policy with something later or retrieve valid locations for it, then you won't be able to merge any policies or find any valid locations.
Maybe we should be validating that input.

ddietterich (approved):

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerConnectedTest.java
yuhuyoyo: agree with validating the input.

PR #1075, opened 2023-02-16
PF-2368: Catch api error 404.

yuhuyoyo:
manage group is sometimes not found.


ddietterich (approved):
IIUC this lets us tolerate no region for azure resources.
LGTM

yuhuyoyo:
> IIUC this lets us tolerate no region for azure resources. LGTM

yeah.. or just let me pass the dry_run at least so I know how many are cases like this.

my guess is that azure folks removed some resource group in devel but the resource info are still stored in our db?

zloery (approved):
LGTM, but I'm not super familiar with Azure resources so it might be worth someone who has more context taking a look

PR #1076, opened 2023-02-16
[TOAZ-319] Removes vm size validation for batch pool creation

giventocode:
- PR removes the VM size value validation that relies on the [deprecated](https://learn.microsoft.com/en-us/java/api/com.azure.resourcemanager.compute.models.VirtualMachineSizeTypes?view=azure-java-stable) enum `VirtualMachineSizeTypes`.
  -  This validation was preventing TES from creating a batch pool using a valid VM size not included in the enum.

- This validation is redundant as TES performs pre-validation of the VMsize considering, pricing, quota and region availability.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/batchpool/ControlledAzureBatchPoolResource.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/batchpool/ControlledAzureBatchPoolResource.java
@@ -266,7 +266,6 @@ public class ControlledAzureBatchPoolResource extends ControlledResource {
       throw new InconsistentFieldsException("Expected controlled AZURE_BATCH_POOL");
     }
     ResourceValidationUtils.validateAzureBatchPoolId(getId());
-    ResourceValidationUtils.validateAzureVmSize(getVmSize());
     ResourceValidationUtils.validateBatchPoolDisplayName(getDisplayName());
     if (deploymentConfiguration != null) {
       if (deploymentConfiguration.virtualMachineConfiguration() != null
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/batchpool/ControlledAzureBatchPoolResource.java
```diff
@@ -266,7 +266,6 @@ public void validate() {
       throw new InconsistentFieldsException("Expected controlled AZURE_BATCH_POOL");
     }
     ResourceValidationUtils.validateAzureBatchPoolId(getId());
-    ResourceValidationUtils.validateAzureVmSize(getVmSize());
```
jgainerdewar, line 269: Should we now remove `ResourceValidationUtils.validateAzureVmSize`?

jgainerdewar (approved):
Left a comment about some additional cleanup, not a blocker.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/batchpool/ControlledAzureBatchPoolResource.java
giventocode: The method is also used when creating a VM. I didn’t remove it altogether because for VMs we need to thinking through a replacement using ARM APIs - unlike the batch pool, I don’t think we can assume the VM size is already validated. This should be follow up work.

jgainerdewar:
@sergiygetlin are you able to take a look at this?

aherbst-broad (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/batchpool/ControlledAzureBatchPoolResource.java
ddietterich: Can you call Azure to do a validation instead of having to maintain a redundant list?

PR #1079, opened 2023-02-21
[PF-2484] cleanup common fields

ddietterich:
This is the first phase of a two step process to refactor common fields. With this change, it should be possible to add new common fields to all resources or controlled resources without visiting every resource.

In the first phase, the existing constructors are left as is so that any in-flight serialized resources will properly deserialize. When we serialize these revised resources, we serialize both the existing individual fields as well as the two common structures: WsmResourceFields and WsmControlledResourceFields. In phase two, we will only serialize and deserialize the new structures.

### Azure Region Handling Changes

There are two Api paths to setting region in the Azure resources: an explicit region as part of the resource attributes and the newly added region in the common fields. I changed the construction to duplicate the explicit region into the region field and removed the field from the Azure resource classes.

I **did not** change the API. It accepts the region through the resource attributes. When we complete the PF-2252 epic and move to the V2 interface, we can drop the API field and only use the common field.

I added detection of a mismatch between the attribute region and the common region settings. That revealed a number of issues with region handling - especially in the tests that had be happily ignoring the common region setting. That also revealed that the region returned from Azure for relay namespace is in a different format ("West Central US") instead of "westcentralus". Looking at the Azure Region.java code, I see that they mutate that name into the lower-case-no-spaces version. So I added that in the UpdateAzureControlledResourceRegionFlight. It may be that we want that normalization on the incoming API regions. I did not do that. Let me know what you think.

### Stairway Mapper Change

Jackson Json does not "see" `@JsonIgnore` annotations on inherited classes; we have been serializing all kinds of random things that (thankfully) are ignored on deserialization. I updated the Stairway setup in JobManager to ignore all inherited setters by using an annotation introspector. Directly fiddling with the Stairway mapper is not really a good plan. I made PF-2505 to add a Stairway method allowing a client-specified mapper.

### Resource Changes

There are repetitive changes to all of the resources. They include:

- Modifying the constructor to build the common fields to pass to `super`
- Commented out constructor to be used in phase 2. (Wishing for good old C #ifdef's...)
- Consolidated list of getters for what is to be serialized. Since the inherited classes are ignored, this controls exactly what gets written out. You'll notice some of these labelled to be removed in phase 2.

### Test Changes

- One additional test, to be removed in Phase 2, that confirms an old serialized Ai Notebook will deserialize with the new code.
- Cleaned up repetitive code in ControlledResourceFixtures and ReferenceResourceFixtures.
- A bunch of test changes to accommodate the changes above.


service/src/test/java/bio/terra/workspace/common/fixtures/ReferenceResourceFixtures.java
```diff
@@ -41,99 +28,44 @@ public static WsmResourceFields.Builder<?> makeDefaultWsmResourceFieldBuilder(UU
...
       UUID workspaceId, String projectId, String bqDataset) {
     UUID resourceId = UUID.randomUUID();
     String resourceName = "testbq-" + resourceId.toString();
     return new ReferencedBigQueryDatasetResource(
-        workspaceId,
-        resourceId,
-        resourceName,
-        "a description",
-        CloningInstructions.COPY_NOTHING,
+        WsmResourceFields.builder()
```
zloery, line 54: can this just use `makeDefaultWsmResourceFields`?

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/StoreControlledResourceMetadataStepTest.java
```diff
@@ -67,6 +67,9 @@ public void testEntersInfo() throws InterruptedException, RetryException {
     assertThat(daoResource.getResourceType(), equalTo(WsmResourceType.CONTROLLED_GCP_GCS_BUCKET));
 
     ControlledGcsBucketResource daoBucket = (ControlledGcsBucketResource) daoResource;
+    if (bucketResource.equals(daoBucket)) {
+      System.out.println("equals");
```
zloery, line 71: leftover from debugging?

zloery (approved):

service/src/test/java/bio/terra/workspace/unit/TestTransitionalSerdes.java
```diff
@@ -0,0 +1,292 @@
...
+      """;
+
+  @Test
+  public void testOldNotebookSerdes() throws Exception {
+    JobService.configureMapper();
+
+    FlightMap testMap = new FlightMap();
+    testMap.putRaw("foo", RESOURCE_NOTEBOOK_JSON);
+    ControlledAiNotebookInstanceResource resource =
+        testMap.get("foo", ControlledAiNotebookInstanceResource.class);
```
yuhuyoyo, line 214: should we test the deserialization for typeReference

service/src/main/java/bio/terra/workspace/service/resource/controlled/model/ControlledResource.java
```diff
@@ -187,18 +159,31 @@ public ApiResourceMetadata toApiMetadata() {
...
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (!(o instanceof ControlledResource that)) return false;
+    if (!super.equals(o)) return false;
+    return Objects.equal(wsmControlledResourceFields, that.wsmControlledResourceFields);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(super.hashCode(), wsmControlledResourceFields);
```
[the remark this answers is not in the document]
ddietterich: That is the auto-generated form. `Objects.hashcode(Object objects...)` is the signature, so It is hashing the hashcode (which seems redundant) and the wsmControlledResourceFields. Seems OK.

service/src/test/java/bio/terra/workspace/unit/TestTransitionalSerdes.java
ddietterich: I don't think we need an additional test. The deserialization of something like `List<WsmResource>` uses the individual object deserialization inside.

yuhuyoyo (approved):
can you take a look of the code smells from the scan?

ddietterich:
> can you take a look of the code smells from the scan?

I will do a pass with PF-2512 next week. Many of them will go away with that branch.

PR #1082, opened 2023-02-21
[PF-2514] Handle deleted projects in revokeActAs

zloery:
There are a bunch of revokeGrant flights churning in devel + autopush caused by test workspaces which have been deleted before the grants can be revoked, this should handle them. The `revokeProject` and `revokeResource` methods already check whether the GCP context has been deleted so they don't need any changes.


yuhuyoyo (approved):

ddietterich (approved):
Thanks for sorting this out, Zach!

PR #1084, opened 2023-02-22
[PF-2512] remove transition code from common fields refactoring

ddietterich:


ddietterich:
I'll post this back up closer to when we want to merge it. Has to wait until next week.

PR #1085, opened 2023-02-22
PF-2502: mergeCheck can fail if workspace has no PAO

mikenev:
Couldn't think of a good way to add automation for this since all of our create workspace paths also add a PAO. We need an old workspace that doesn't have one. But this code would create a PAO for that workspace. But the concept is pretty well proven already in the merge PAO path, this is just adding it to the mergeCheck API.


zloery (approved):

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
```diff
@@ -87,6 +87,19 @@ public void createPao(UUID workspaceUuid, @Nullable TpsPolicyInputs policyInputs
     }
   }
 
+  // Since policy attributes were added later in development, not all existing
+  // workspaces have an associated policy attribute object. This function creates
+  // an empty one if it does not exist.
+  @Traced
+  public void createPaoIfNotExist(UUID workspaceId) {
```
yuhuyoyo, line 94: is the idea that we don't backfill but we fill as we encounter them?

if so, should we make `getPaoIfExists` private and always use `createPaoIfNotExist`? is there any case we want to leave the old workspace with empty pao?

mikenev: Yeah, we aren't backfilling but filling in as we encounter them. That's a good idea, let me see what it will take. I don't think there's a case where we'd need to leave the old workspace without a PAO.

yuhuyoyo (approved):
thanks

PR #1086, opened 2023-02-22
Remove `terra cromwell generate-config` command from post-startup script

wnojopra:
I recently updated this command to require a bucket path parameter. The expectation is that this config file includes a workspace bucket. A user needs to specify the bucket when calling this command. 

I don't think the startup script should have called this command in the first place. Before my change, it produced a config file that wouldn't work because there is no workspace bucket.


harrismendell (approved):

yuhuyoyo (approved):
the original idea is that we write a config template in the notebook and the user will modify it by passing the bucket gs:// path before using it. 

i don't feel strongly either way.

PR #1089, opened 2023-02-23
PF-2468: Remove redundant logging

yuhuyoyo:
Clone referenced resources is now a flight so log will be written in the WorkspaceActivityLogHook instead. `createReferenceResourceForClone` doesn't need to log the activity anymore.

All the clone test in the *controllerTest now also assert that a clone activity is logged in the destination workspace with the source resource id as the changeSubjectId.
i.e. foo is cloned from workspace 1 to workspace 2 by yu.
The log should be: in workspace 2, foo is cloned by yu at <timestamp>.


mikenev (approved):
lgtm, thanks for cleaning these up.

zloery (approved):

PR #1094, opened 2023-02-24
PF-2497 - Add groups from WS policy as auth domains during WS create

mikenev:
The approach works when the policy group name == auth domain name, that makes sense to me. Two things I'm not as happy with:
* Passing the list of auth domains. It seems appropriate though most of the time we aren't using this so it's usually null. Just like we're passing the applicationIds which are usually empty.
* How to test this. Manual testing seems to work fine. Other than that there's not a lot of functionality like asking SAM - which auth domains are set on this workspace?


service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -22,13 +22,7 @@
...
 import io.opencensus.trace.Tracing;
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
zloery, line 25: nit: avoid * imports. there's a setting somewhere in intelliJ to stop it from replacing import lists with these

service/src/main/java/bio/terra/workspace/service/policy/TpsUtilities.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class TpsUtilities {
+  public static final String TERRA_NAMESPACE = "terra";
+  public static final String GROUP_CONSTRAINT = "group-constraint";
+
+  public static List<String> getGroupConstraintsFromInputs(TpsPolicyInputs inputs) {
+    List<String> result = new ArrayList<>();
+
+    for (var input : inputs.getInputs()) {
+      if (input.getNamespace() == TERRA_NAMESPACE && input.getName() == GROUP_CONSTRAINT) {
```
zloery, line 15: these string comparisons need to be `.equals()` instead of `==`

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -447,8 +451,12 @@ public String cloneWorkspace(
     List<String> applicationIds =
         applicationDao.listWorkspaceApplicationsForClone(sourceWorkspace.getWorkspaceId());
 
+    TpsPaoGetResult policy = tpsApiDispatch.getPao(sourceWorkspace.getWorkspaceId());
```
zloery, line 454: We still have a TPS feature flag - does this call need to be behind that check? I think this is causing a unit test to fail, but I'm not sure how long we want to keep using the flag for

mikenev: I'll add it behind that check too. I'm not sure how long we want to keep using it either, but it's probably a bigger task to take it out.

service/src/main/java/bio/terra/workspace/service/policy/TpsUtilities.java
mikenev: Thanks, fixed. Sonar caught that too.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
mikenev: Thanks, fixed. I updated the intelliJ settings for this but I guess made this change before that update.

zloery (approved):

service/src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
```diff
@@ -46,11 +46,7 @@
 import bio.terra.workspace.service.workspace.model.Workspace;
 import bio.terra.workspace.service.workspace.model.WorkspaceStage;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
```
ddietterich, line 49: No `*` imports. You can make sure this doesn't happen with IntelliJ settings. I can show you.

service/src/test/java/bio/terra/workspace/service/iam/SamServiceTest.java
```diff
@@ -225,11 +221,11 @@ void nonOwnerCannotAddReader() {
   void permissionsApiFailsInRawlsWorkspace() throws Exception {
     UUID workspaceUuid = UUID.randomUUID();
     // RAWLS_WORKSPACEs do not own their own Sam resources, so we need to manage them separately.
-    samService.createWorkspaceWithDefaults(defaultUserRequest(), workspaceUuid);
+    samService.createWorkspaceWithDefaults(defaultUserRequest(), workspaceUuid, new ArrayList<>());
 
     Workspace rawlsWorkspace =
         WorkspaceFixtures.buildWorkspace(workspaceUuid, WorkspaceStage.RAWLS_WORKSPACE);
-    workspaceService.createWorkspace(rawlsWorkspace, null, null, defaultUserRequest());
+    workspaceService.createWorkspace(rawlsWorkspace, null, null, null, defaultUserRequest());
```
ddietterich, line 228: It would have been easier to make a `createWorkspaceWithPolicy` method and avoid adding all of the nulls to all of the tests.

ddietterich (approved):
A couple of minor changes I'd like. Otherwise, LGTM

ddietterich (approved):
Suggest pulling the auth domains out inside of the flight.
Otherwise, LGTM

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -125,6 +127,9 @@ public UUID createWorkspace(
             .addParameter(
                 WorkspaceFlightMapKeys.WORKSPACE_STAGE, workspace.getWorkspaceStage().name())
             .addParameter(WorkspaceFlightMapKeys.POLICIES, policies)
+            .addParameter(
```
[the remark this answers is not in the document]
mikenev: You're right, will do that. The previous commit made this possible but could have simplified even further.

PR #1095, opened 2023-02-27
TOAZ-313 Delete Batch Pools through WSM

sergiygetlin:
Synchronous endpoint for Azure Batch Pool deletion. The related deletion step was implemented earlier.


giventocode (approved):
LGTM! Thanks.

PR #1098, opened 2023-02-28
Add OpenAPI specification for flexible resources.

aaron-wang:
Part one of a few PRs which will add "Flexible resources" to WSM. Previously, these were known as "Blob resources".

The next PR adds the bulk of the resource code (class, attributes, handler), and includes the create, get, and delete endpoints.

### DIFF
--- a/openapi/src/resources/flexible_resource.yaml
+++ b/openapi/src/resources/flexible_resource.yaml
@@ -12,7 +12,7 @@ components:
           description: Type of the flexible resource.
           type: string
         data:
-          description: JSON encoded in base-64.
+          description: JSON encoded in base64.
           type: string
           format: byte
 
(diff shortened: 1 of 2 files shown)


ddietterich (commented):
This is fine as is, but I'm curious why you need it when you did not also add the paths.

If you did it so you could program with the API classes, I would ask that you not do that. You will see other code that makes direct use of the API classes in the flight code. This is not good practice and it is on my list to eliminate. We should always convert API classes into internal classes in the controller. Then operate on the internal classes for all service and flight processing.

aaron-wang:
> This is fine as is, but I'm curious why you need it when you did not also add the paths.

So, I wanted to see if there were any major issues with the OpenAPI specification (or naming).

I can publish a larger change which has functioning endpoints and close this PR.

PR #1101, opened 2023-03-01
[PF-2538] Add flexible resources and basic endpoints.

aaron-wang:
Flexible resources are controlled resources; however, they are not stored on the cloud.

### Changes:
- Add bulk of resource code (class, attributes, and handler)
- Add create, get, and delete endpoints
- Classify flexible resources to be ANY `CloudPlatform`. 
  - Currently, this is the only controlled resource that is ANY `CloudPlatform`
  - `GetCloudContextStep` no longer throws if the controlled resource is ANY `CloudPlatform`
  - Flexible resources do not have regions (so, regional policies do not apply)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/ControlledFlexibleResource.java
```diff
@@ -0,0 +1,364 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+public class ControlledFlexibleResource extends ControlledResource {
+  private final String typeNamespace;
+  private final String type;
+  private final String data;
```
ddietterich, line 43: Mark this as `@Nullable`

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/ControlledFlexibleResource.java
```diff
@@ -0,0 +1,364 @@
...
+     this.typeNamespace = typeNamespace;
+     this.type = type;
+     this.data = data;
+     validate();
+   }
+    */
+
+  // Constructor for the builder
+  private ControlledFlexibleResource(
+      ControlledResourceFields common, String typeNamespace, String type, String data) {
```
ddietterich, line 114: `@Nullable`  here also, I think

I think it is worth marking it throughout the code. It is needed in a couple of other places. Helps us catch spots where we are not handling null.

ddietterich (approved):
Looks great! And very quickly done!
My only nitpick is to annotate `@Nullable` as you pass the data field through the various classes.

service/src/main/java/bio/terra/workspace/app/controller/ControlledFlexibleResourceApiController.java
```diff
@@ -0,0 +1,130 @@
...
+        controlledResourceMetadataManager
+            .validateControlledResourceAndAction(
+                userRequest,
+                workspaceUuid,
+                resourceId,
+                SamConstants.SamControlledResourceActions.READ_ACTION)
+            .castByEnum(WsmResourceType.CONTROLLED_FLEXIBLE_RESOURCE);
+    return new ResponseEntity<>(resource.toApiResource(), HttpStatus.OK);
+  }
+}
```
yuhuyoyo, line 130: just curious, are you planning to add update/clone endpoint in the future?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/ControlledFlexibleResource.java
```diff
@@ -0,0 +1,366 @@
...
+            .build());
+    this.typeNamespace = typeNamespace;
+    this.type = type;
+    this.data = data;
+    validate();
+  }
+
+  /*
+     // TODO: PF-2512 remove constructor above and enable this constructor
+   @JsonCreator
```
yuhuyoyo, line 96: @ddietterich correct me if I'm wrong, can we enable this now that the blocking tickets are done?

service/src/main/java/bio/terra/workspace/app/controller/ControlledFlexibleResourceApiController.java
```diff
@@ -0,0 +1,130 @@
...
+        resourceId.toString(),
+        workspaceUuid.toString());
+    getControlledResourceService()
+        .deleteControlledResourceSync(workspaceUuid, resourceId, userRequest);
+    return new ResponseEntity<>(HttpStatus.NO_CONTENT);
+  }
+
+  @Traced
+  @Override
+  public ResponseEntity<ApiFlexibleResource> getFlexibleResource(
```
yuhuyoyo, line 117: now that we have this endpoint /api/workspaces/v1/{workspaceId}/resources/{resourceId}

do we still need this? @ddietterich

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -648,9 +648,10 @@ public boolean updateResource(
   public void createControlledResource(ControlledResource controlledResource)
       throws DuplicateResourceException {
 
-    if (!cloudContextExists(
```
yuhuyoyo, line 651: nit: can you add a resourcedao test that create a controlled flexible resource?

yuhuyoyo (approved):
have some question for dan. lgtm!

service/src/main/java/bio/terra/workspace/app/controller/ControlledFlexibleResourceApiController.java
aaron-wang: Yes. The update endpoint is next, and then the clone will come after 🙂

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
aaron-wang: added.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/ControlledFlexibleResource.java
ddietterich: You are right, Yu - good catch! The only reason to have the old constructor is to migrate old flights to the new model. This will never have old flights, so it can exclusively use the new model.

If you look at my branch dd-PF-2512-remove-transition you will see what the end state looks like. You can go directly there.

service/src/main/java/bio/terra/workspace/app/controller/ControlledFlexibleResourceApiController.java
ddietterich: It keeps things consistent, so I'm happy to have it.

PR #1104, opened 2023-03-01
[DDO-2694] Automatically update TPS in wsmtest-eng BEE

jack-r-warren:
This PR works in two parts:

- This actual workflow here will "refresh" the wsmtest-eng BEE and sync the entire environment in ArgoCD
- That "refresh" will actually do something because over in Beehive, tps-wsmtest-eng is configured to get it's app version from tps-dev. The refresh will get the latest tps-dev version and the sync will deploy it.

This action also tries to curl TPS's version endpoint and print it to the GitHub Actions summary so it should be visible without going through logs.


ddietterich (approved):

ddietterich:
This is a github action - independent of all of the WSM code, so not worth clearing the test bars.

PR #1105, opened 2023-03-02
[PF-2544] Add custom ser / deser to CloudContextHolder

dexamundsen:
Problem - basic object mapper used for ser / deser included via stairway does not support custom serialization / deserialization of included fields. 
Serialize() / Deserialize() methods of gcpCloudContext & azureCloudContext need to be explicitly called. 

Proposed solution - add custom serializer & deserializer for CloudContextHolder. 
This is preferred over alternates - add config to objectMapper / change gcpCloudContext & azureCloudContext to make them serializable directly and override  relevant functions.

--
encountered this error while adding additional platform cloud context in CCH


dexamundsen:
No longer needed, changed the implementation of cloudContext

PR #1109, opened 2023-03-02
[PF-2531] Revert to non-apache connector to unblock tests

zloery:
This reverts https://github.com/DataBiosphere/terra-workspace-manager/pull/1102 until I can get the client working properly


ddietterich (approved):
OK - we haven't added any new endpoints, so I think running the old client should be fine.

ddietterich:
I expect to see 10 failures in connected tests due to the new TPS.

ddietterich:
> I expect to see 10 failures in connected tests due to the new TPS.

And that is what we got.

yuhuyoyo (approved):

PR #1111, opened 2023-03-03
PF-2533: Case insensitive check for existing resources when performing the merge check

mikenev:
Issue came up where we were marking a conflict if a workspace had an existing resource with a region that didn't match exactly one of the regions from TPS. Checking existing resources should be a case-insensitive comparison.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -758,7 +758,7 @@ public class WorkspaceApiController extends ControllerBase implements WorkspaceA
               targetWorkspaceId, CloudPlatform.fromApiCloudPlatform(platform));
 
       for (var existingResource : existingResources) {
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -758,7 +758,7 @@ public ResponseEntity<ApiWsmPolicyMergeCheckResult> mergeCheck(
               targetWorkspaceId, CloudPlatform.fromApiCloudPlatform(platform));
 
       for (var existingResource : existingResources) {
-        if (!validRegions.contains(existingResource.getRegion())) {
+        if (!validRegions.stream().anyMatch(existingResource.getRegion()::equalsIgnoreCase)) {
```
ddietterich, line 761: Seems a waste of a hashset to iterate through it, but I guess we have mixed case valid regions?

ddietterich (approved):

yuhuyoyo (approved):

PR #1114, opened 2023-03-06
Fix for ImportDataCollection test

mikenev:
Problem was that the test was creating a workspace with a region policy and then updating the policy to have a group policy. Our update method used to do a replace so the group policy was taken out, but that was a bug that was fixed so now that update leaves the group policy in place and adds a region policy. The rest of the verification was only expecting the group policy to exist.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/ImportDataCollection.java
+++ b/integration/src/main/java/scripts/testscripts/ImportDataCollection.java
@@ -397,7 +397,7 @@ public class ImportDataCollection extends WorkspaceAllocateTestScriptBase {
      Scenario 9: Group policy merging. WS(groupA) can merge DC(nogroup).
     */
     CreatedWorkspace groupTestWorkspace =
-        createWorkspaceWithRegionPolicy(workspaceApi, usaLocation);
+        createWorkspace(UUID.randomUUID(), getSpendProfileId(), workspaceApi);
     var request =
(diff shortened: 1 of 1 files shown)


zloery (approved):

integration/src/main/java/scripts/testscripts/ImportDataCollection.java
```diff
@@ -397,7 +397,7 @@ gain region from data collection. Workspace (policy=empty) + Data Collection (po
      Scenario 9: Group policy merging. WS(groupA) can merge DC(nogroup).
     */
     CreatedWorkspace groupTestWorkspace =
-        createWorkspaceWithRegionPolicy(workspaceApi, usaLocation);
```
yuhuyoyo, line 400: i don't follow why we need to remove the region policy? it shouldn't affect the group-constraint policy.

mikenev: It doesn't necessarily need to be removed. But all these scenarios were written with the idea that the region policy wasn't there. It seemed more straightforward to remove it rather than go back to adjust the comments and validation steps to accommodate it.

yuhuyoyo: sure

yuhuyoyo (approved):

ddietterich (approved):

PR #1115, opened 2023-03-06
[PF-2539] Add update endpoint for flexible resources.

aaron-wang:
### Changes:
- Add an update endpoint
- Allow editing the `data` field of flexible resources
- Assume that `type` and `typeNamespace` are immutable for now


ddietterich (approved):
A question about consistent handling of null data inputs.
Not necessary to fix in this PR, but something to think about.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flight/update/UpdateControlledFlexibleResourceAttributesStep.java
```diff
@@ -0,0 +1,72 @@
...
+
+    String newAttributes =
+        DbSerDes.toJson(
+            new FlexibleResourceAttributes(
+                flexResource.getTypeNamespace(), flexResource.getType(), updateData));
+    boolean updated =
+        resourceDao.updateResource(
+            flexResource.getWorkspaceId(),
+            flexResource.getResourceId(),
+            null,
```
yuhuyoyo, line 46: nit: add comment for the null field so it's clear what they are. e.g. /*param=*/null

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flight/update/UpdateControlledFlexibleResourceAttributesStep.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    String previousAttributes =
+        flightContext
+            .getWorkingMap()
+            .get(WorkspaceFlightMapKeys.ResourceKeys.PREVIOUS_ATTRIBUTES, String.class);
```
yuhuyoyo, line 62: if the key doesn't exist, return early?

service/src/test/java/bio/terra/workspace/common/utils/MockMvcUtils.java
```diff
@@ -2398,6 +2466,22 @@ public void postExpect(
         .andExpect(status().is(httpStatus));
   }
 
+  /** Patch http request and expect error thrown. */
+  public ResultActions patchExpect(
```
yuhuyoyo, line 2470: there is an updateresource method already in MockMvcUtils. can you use that?

yuhuyoyo (approved):

PR #1116, opened 2023-03-06
[PF-2541] Add LINK_REFERENCE cloning instruction

ddietterich:
We need explicit control over when we are cloning for a copy and when we a cloning to build a link between a workspace and a data collection. This cloning instruction is used for that.

On a resource clone operation, if the instruction is LINK_REFERENCE, then a persistent policy connection is made from the source to the destination workspace. If the instruction is something else, then the source policy is merged with the destination, but no persistent connection is made between the workspaces.

Similarly, on a workspace clone operation, if any of the source workspace's resources has the LINK_REFERENCE cloning instruction, then the workspace policies are linked. Otherwise, the workspace policies are merged.

I changed the logic for clone workspace to allow policy inputs to be provided on the clone endpoint. The sequence is now:
1. Create the workspace policy with any policy inputs
2. Based on cloning instructions: do nothing (normal create), merge (clone), or link (clone with LINKing resources). For the clone cases, the source workspace policy is integrated with the destination workspace policy (merged or linked)
3. Create the Sam resource with auth domains based on step 2


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/SetReferencedDestinationGcsBucketResponseStep.java
```diff
@@ -67,7 +67,7 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
 
     final ApiClonedControlledGcpGcsBucket apiClonedBucket =
         new ApiClonedControlledGcpGcsBucket()
-            .effectiveCloningInstructions(ApiCloningInstructionsEnum.REFERENCE)
```
yuhuyoyo, line 70: why is it always copy_reference here?

yuhuyoyo (approved):

openapi/src/common/schemas.yaml
```diff
@@ -36,21 +36,30 @@ components:
...
             from source bucket to new bucket.
-          * COPY_REFERENCE: Used for controlled and referenced resources. Create new referenced resource
+          * COPY_REFERENCE and REFERENCE: Used for controlled and referenced resources. Create new referenced resource
             that points to same cloud resource as source resource.
           * LINK_REFERENCE: Used for controlled and referenced resources. Create a new referenced resource
             that points to the same cloud resource as the source resource, AND link the source workspace
             policy to the destination workspace policy; changes in the source will propagate to the destination.
-      enum: ['COPY_NOTHING', 'COPY_DEFINITION', 'COPY_RESOURCE', 'COPY_REFERENCE', 'LINK_REFERENCE']
+      enum:
+        - COPY_NOTHING
```
yuhuyoyo, line 54: Does it show NOTHING, COPY_NOTHING as one enum in the client API?

I wonder what's more confusing, having two duplicate enum for the same function or having LINK_REFERENCE called COPY_LINK_REFERENCE..

ddietterich: I want to disable the prefix trim so this doesn't happen again. If I do that, then I need to specify the extra enum values.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/SetReferencedDestinationGcsBucketResponseStep.java
ddietterich: Good catch! That needs the resolved cloning instructions

mikenev (approved):
Enum update means lots of files to touch. Update makes sense though.

PR #1119, opened 2023-03-07
PF-2543: list workspaces will include missing authdomains 

yuhuyoyo:


openapi/src/parts/workspace.yaml
```diff
@@ -659,6 +659,11 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/WsmPolicyInput'
+        missingAuthDomains:
+          type: array
+          description: The auth domains that the user is missing for this workspace
```
ddietterich, line 664: Maybe check with @cbookg about what he'd like to call it? In Sam-land it is an auth domain, but in policy-land it is a failed group constraint. I don't feel strongly about what we call it here.

ddietterich (approved):
Nice!

mikenev (approved):
lgtm

openapi/src/parts/workspace.yaml
yuhuyoyo: chris approved the field in the ticket comment

PR #1123, opened 2023-03-08
WOR-812 - update landing zone service version to save BP id in delete result

blakery:
update landing zone service version


cahrens:
@blakery it would be good to always have a descriptive name in the PR title, along with the ticket number. When glancing through merged PRs to try to figure out what has changed (which I've been doing recently!), having just the ticket number isn't helpful.

cahrens (approved):
Code looks good, but please update the title of the PR to explain what the PR is about (for instance, "Update landing zone service to pass billing profile ID in delete result").

PR #1124, opened 2023-03-08
Add new guest attribute writes to post startup script for success/error

harrismendell:
https://broadworkbench.atlassian.net/browse/PF-2550

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -14,7 +14,6 @@
 # https://github.com/DataBiosphere/terra-workspace-manager/tree/main/integration#Run-nightly-only-test-suite-locally
 # for instruction on how to run the test.
 
-set -o errexit
 set -o nounset
 set -o pipefail
 set -o xtrace
@@ -23,6 +22,9 @@ set -o xtrace
 # user space, such as setting Terra CLI settings which are persisted in the user's $HOME.
 # This post startup script is not run by the same user.
 readonly JUPYTER_USER="jupyter"
+readonly STATUS_ATTRIBUTE="startup_script/status"
+readonly MESSAGE_ATTRIBUTE="startup_script/message"
+readonly OUTPUT_FILE=".terra/post-startup-output.txt"
 
 # Move to the /tmp directory to let any artifacts left behind by this script can be removed.
 cd /tmp || exit
@@ -30,7 +32,7 @@ cd /tmp || exit
 # Send stdout and stderr from this script to a file for debugging.
 # Make the .terra directory as the user so that they own it and have correct linux permissions.
 sudo -u "${JUPYTER_USER}" sh -c "mkdir -p /home/${JUPYTER_USER}/.terra"
-exec >> /home/"${JUPYTER_USER}"/.terra/post-startup-output.txt
+exec >> /home/"${JUPYTER_USER}"/"${OUTPUT_FILE}"
 exec 2>&1
 
 #######################################
@@ -46,6 +48,30 @@ function get_metadata_value() {
     -H "Metadata-Flavor: Google" \
     "http://metadata/computeMetadata/v1/$1"
 }
+
+#######################################
+# Set guest attributes on GCE. Used here to log completion status of the script.
+# See https://cloud.google.com/compute/docs/metadata/manage-guest-attributes
+# Arguments:
+#   $1: The guest attribute domain and key IE startup_script/status
+#   $2  The data to write to the guest attribute
+#######################################
+function set_guest_attributes() {
+  curl -s -X PUT --data "$2" \
+    -H "Metadata-Flavor: Google" \
+    "http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/$1"
+}
+
+# Write error status and message to guest attributes should an error occur.
+function error_handler {
+  set_guest_attributes "$STATUS_ATTRIBUTE" "ERROR"
+  set_guest_attributes "$MESSAGE_ATTRIBUTE" "Error on line $1, command \"$2\". See $OUTPUT_FILE for more information"
+  exit 1
+}
+trap 'error_handler $LINENO $BASH_COMMAND' ERR
+
+# Let the UI know the script has started
+set_guest_attributes STATUS_ATTRIBUTE "STARTED"
 
 # Install common packages. Use pip instead of conda because conda is slow.
 /opt/conda/bin/pip install pre-commit nbdime nbstripout pylint pytest dsub pandas_gbq
@@ -228,6 +254,8 @@ readonly TERRA_GCP_NOTEBOOK_RESOURCE_NAME="$(get_metadata_value "instance/attrib
 if [[ -n "${TERRA_TEST_VALUE}" ]]; then
   sudo -u "${JUPYTER_USER}" sh -c "terra resource update gcp-notebook --name=${TERRA_GCP_NOTEBOOK_RESOURCE_NAME} --new-metadata=terra-test-result=${TERRA_TEST_VALUE}"
 fi
+
+set_guest_attributes "$STATUS_ATTRIBUTE" "COMPLETE"
 
 ####################################
 # Restart kernel so environment variables work in notebook. See PF-2178.


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -14,7 +14,6 @@
 # https://github.com/DataBiosphere/terra-workspace-manager/tree/main/integration#Run-nightly-only-test-suite-locally
 # for instruction on how to run the test.
 
-set -o errexit
```
wnojopra, line 17: Why are we removing this?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -47,6 +49,30 @@ function get_metadata_value() {
...
+# See https://cloud.google.com/compute/docs/metadata/manage-guest-attributes
+# Arguments:
+#   $1: The guest attribute domain and key IE startup_script/status
+#   $2  The data to write to the guest attribute
+#######################################
+function set_guest_attributes() {
+  curl -s -X PUT --data "$2" \
+    -H "Metadata-Flavor: Google" \
+    "http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/$1"
+}
```
wnojopra, line 63: This is optional, but you can clarify the arguments by doing something like

```suggestion
function set_guest_attributes() {
  local attr_key="${1}"
  local attr_value="${2}"

  curl -s -X PUT --data "${attr_value}" \
    -H "Metadata-Flavor: Google" \
    "http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/${attr_key}"
}
```

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
mbookman: This should not be removed.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -47,6 +49,30 @@ function get_metadata_value() {
...
+
+# Write error status and message to guest attributes should an error occur.
+function error_handler {
+  set_guest_attributes "$STATUS_ATTRIBUTE" "ERROR"
```
mbookman, line 67: https://google.github.io/styleguide/shellguide.html#variable-expansion

> quote your variables; prefer "${var}" over "$var".

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
mbookman: Good example, Willy. I consider it non-optional. It's an important precedent to set in this file.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
harrismendell: If we leave it the error trap won't work because the script will exit immediately. We also don't need it anymore, because the trap itself exits on completion.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -23,14 +22,17 @@ set -o xtrace
 # user space, such as setting Terra CLI settings which are persisted in the user's $HOME.
 # This post startup script is not run by the same user.
 readonly JUPYTER_USER="jupyter"
+readonly STATUS_ATTRIBUTE="startup_script/status"
```
mbookman, line 25: I'd like to suggest:

```
readonly USER_TERRA_CONFIG_DIR=/home/${JUPYTER_USER}/.terra
readonly OUTPUT_FILE="${USER_TERRA_CONFIG_DIR}/post-startup-output.txt"
```

such that the `USER_TERRA_CONFIG_DIR` can be used as needed as new commands are added to this script.

mbookman (changes requested):
Thanks @harrismendell! This is looking great.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
mbookman: That doesn't sound right. errexit should cause faulty commands to start the exit process, including going through the ERR trap.

https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

> -e
> Exit immediately... <snip>
>
> A trap on ERR, if set, is executed before the shell exits.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -47,6 +49,30 @@ function get_metadata_value() {
...
+    "http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/$1"
+}
+
+# Write error status and message to guest attributes should an error occur.
+function error_handler {
+  set_guest_attributes "$STATUS_ATTRIBUTE" "ERROR"
+  set_guest_attributes "$MESSAGE_ATTRIBUTE" "Error on line $1, command \"$2\". See $OUTPUT_FILE for more information"
+  exit 1
+}
+trap 'error_handler $LINENO $BASH_COMMAND' ERR
```
mbookman, line 71: Reflecting a bit further, I'm unclear what trips the ERR trap specifically.
I have generally preferred trapping EXIT and then just having an if/then/else flow based on the exit code.

```
function exit_handler {
  local exit_code="$1"
  local line_no="$2"
  local command="$3"

  # Success! Set the guest attributes and exit cleanly
  if [[ "${exit_code}" -eq 0 ]]; then
    set_guest_attributes "$STATUS_ATTRIBUTE" "COMPLETE"
    exit 0
  fi

  # Write error status and message to guest attributes
  set_guest_attributes "$STATUS_ATTRIBUTE" "ERROR"
  set_guest_attributes "$MESSAGE_ATTRIBUTE" "Error on line $1, command \"$2\". See $OUTPUT_FILE for more information"
  exit "${exit_code}"
}
trap 'exit_handler $? $LINENO $BASH_COMMAND' EXIT
```

mbookman (changes requested):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
mbookman: This article captures (traps?) the essence and intention of the above approach:

https://medium.com/@dirk.avery/the-bash-trap-trap-ce6083f36700

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
harrismendell: 👍

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -42,11 +46,48 @@ exec 2>&1
...
+  local attr_path="${1}"
+  local attr_value="${2}"
+  curl -s -X PUT --data "${attr_value}" \
+    -H "Metadata-Flavor: Google" \
+    "http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/${attr_path}"
+}
+
+# If the script exits without error let the UI know it completed successfully
+# Otherwise if an error occurred write the line and command that failed to guest attributes.
+function exit_handler {
```
harrismendell, line 72: @mbookman switched to your suggestion, thanks for the input / learnings.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
harrismendell: Thanks for the learning! Made the changes

wnojopra (approved):
LGTM. Let's get Matt's approval as well.

mbookman (approved):
LGTM. Thanks Harris!

PR #1132, opened 2023-03-09
[PF-2512] remove common fields transition

ddietterich:
Now that the transition code has made it to prod on both Broad and Verily, we can remove it.
This PR removes the verbose constructors and accessors from individual resources.


zloery (approved):
nice!

service/src/main/java/bio/terra/workspace/service/resource/controlled/model/ControlledResource.java
```diff
@@ -51,7 +65,7 @@ public ControlledResource(DbResource dbResource) {
    *
    * @param fields container for building WsmResource and Controlled resource
    */
-  public ControlledResource(ControlledResourceFields fields) {
```
yuhuyoyo, line 54: why is this visibility changed to protected?

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/model/ControlledResource.java
ddietterich: I expect this constructor only to be used from subclasses.
Do you think it needs to be public?

PR #1133, opened 2023-03-09
[PF-2540] Add clone endpoint for flexible resources.

aaron-wang:
Only allow changing name and description upon cloning (attributes are immutable).

Also, now we can specify the cloning instructions in the update endpoint.


service/src/test/java/bio/terra/workspace/app/controller/ControlledFlexibleResourceApiControllerConnectedTest.java
```diff
@@ -0,0 +1,309 @@
...
+    mockMvcUtils.deleteWorkspace(userAccessUtils.defaultUserAuthRequest(), workspaceId2);
+  }
+
+  // Destination workspace policy is the merge of source workspace policy and pre-clone destination
+  // workspace policy
+  @Test
+  @Disabled(
+      "PF-2563 seems to cause failure upon setup() for controlled api connected tests due to "
+          + "failure with createWorkspaceWithPolicy(...")
+  void clone_policiesMerged() throws Exception {
```
aaron-wang, line 116: I'm having issues running connected controller tests (including GCS bucket and BQ dataset) locally. 

`createWorkspaceWithPolicy` causes an `InitializationError` for me (in all three of these connected controller tests).

I note that the other two connected controller tests are entirely disabled, so maybe this slipped through undetected?

ddietterich (approved):
Minor comments, otherwise LGTM.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/flexibleresource/CloneFlexibleResourceStep.java
```diff
@@ -0,0 +1,129 @@
...
+    UUID destinationWorkspaceId =
+        inputParameters.get(
+            WorkspaceFlightMapKeys.ControlledResourceKeys.DESTINATION_WORKSPACE_ID, UUID.class);
+    UUID destinationResourceId =
+        inputParameters.get(
+            WorkspaceFlightMapKeys.ControlledResourceKeys.DESTINATION_RESOURCE_ID, UUID.class);
+
+    controlledResourceService.deleteControlledResourceSync(
+        destinationWorkspaceId, destinationResourceId, userRequest);
+    return null;
```
yuhuyoyo, line 127: return stepresult.getstepresultsuccess?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/flexibleresource/CloneFlexibleResourceStep.java
```diff
@@ -0,0 +1,129 @@
...
+            resourceName,
+            description,
+            samService.getUserEmailFromSamAndRethrowOnInterrupt(userRequest));
+
+    ControlledResourceIamRole iamRole =
+        IamRoleUtils.getIamRoleForAccessScope(sourceFlexResource.getAccessScope());
+
+    ControlledFlexibleResource clonedFlexResource =
+        controlledResourceService
+            .createControlledFlexResourceSyncWithoutCreationParameters(
```
yuhuyoyo, line 93: this is a very specific method that i don't suspect other call path will be using for creating flex resources. Can you move it here?

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/flexibleresource/CloneFlexibleResourceStep.java
aaron-wang: Yeah, I agree this method is quite specific. However, Dan said it's bad practice to have API objects in the flight.

So, I want to avoid using these here, and handle them in the controller.

ddietterich: > However, Dan said it's bad practice to have API objects in the flight.

I want to keep all API objects in the controller. I never want to use them as parameters to a flight; they should never be serialized.

What that means is, we need an internal class for the "creation parameters"; let's call it `FlexResourceCreationParameters`. So you would not pass `body` [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/0121a575a4734c586d845fed394cf7d514488751/service/src/main/java/bio/terra/workspace/app/controller/ControlledFlexibleResourceApiController.java#L85). You would convert the relevant fields of `body` into `FlexResourceCreationParameters` and use that.

I am asking you to do it differently than what we have done so far. I think all of the other controlled resources are passing around the API object as creation parameters. And as you see, one of the issues that creates is it makes implementing clone awkward, because you end up creating the API object.

I _should_ have caught this on the review. Sorry about that.

ddietterich: I think you could have a single endpoint in ControlledResourceService that would work for both cases, instead of needing the `createControlledFlexResourceSyncWithoutCreationParameters`

PR #1135, opened 2023-03-10
[PF-2566] Handle null case in BigQuery lifetime backfill.

aaron-wang:
Previously, if the BigQuery datasets had one or two null lifetimes on the cloud, this would cause the flight to crash.

In retrieval:
- Now the `Pair<Long,Long>` can contain nulls in the coordinates,

In update:
- If the coordinate is null, then update with null (instead of throwing a null error upon conversion).

Tests:
- Added two tests to ensure the flight won't crash in both cases.

Re: #1041 (PF-2269).

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
@@ -131,6 +131,8 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 
+import javax.annotation.Nullable;
+
 // Per-class lifecycle on this test to allow a shared workspace object across tests, which saves
 // time creating and deleting GCP contexts.
 @Tag("connected")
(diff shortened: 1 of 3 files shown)


service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -1957,6 +1958,92 @@ public void updateControlledBigQueryDatasetLifetime_onlyUpdateWhenLifetimesAreEm
...
+    assertControlledBigQueryDatasetLifetimeIsUpdatedAndActivityIsLogged(
+        updatedResourceList,
+        createdDataset.getResourceId(),
+        /*expectedTableLifetime=*/ null,
+        DEFAULT_CREATED_BIG_QUERY_PARTITION_LIFETIME);
+  }
+  // Ensure the flight doesn't crash when there are BigQuery datasets with two null lifetimes on the
+  // cloud.
+  @Test
+  public void updateControlledBigQueryDataset_HandlesDatasets_WithTwoNullLifetimes()
```
yuhuyoyo, line 1995: nit, can we just keep this test and not the one above? this seems to cover the one null case too.

yuhuyoyo (approved):
thanks for fixing this

PR #1136, opened 2023-03-10
[PF-2253] resource states

ddietterich:
This is the first part of putting in resource states. It adds states to the database and enforces the state transitions. It provides a feature configuration to control how we handle failed create operations.

Missing pieces: update states, exposing the state information through the REST API.


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/bqdataset/ControlledBigQueryDatasetResource.java
```diff
@@ -301,11 +293,8 @@ public boolean partialEqual(Object o) {
...
   @Override
   public int hashCode() {
-    return new HashCodeBuilder(17, 37)
-        .appendSuper(super.hashCode())
-        .append(datasetName)
-        .append(projectId)
-        .toHashCode();
+    return Objects.hashCode(
+        super.hashCode(), datasetName, projectId, defaultTableLifetime, defaultPartitionLifetime);
   }
```
aaron-wang, line 298: Should the `equals` and `hashCode` methods for `ControlledFlexibleResource` be updated as well?

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -312,6 +312,7 @@ public <T> ControlledResource createControlledResourceSync(
     JobBuilder jobBuilder =
         commonCreationJobBuilder(resource, privateResourceIamRole, userRequest)
             .addParameter(ControlledResourceKeys.CREATION_PARAMETERS, creationParameters);
+    logger.info(">>> state rule: {}", features.getStateRule());
```
ddietterich, line 315: Stray debug log. Will remove

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/bqdataset/ControlledBigQueryDatasetResource.java
ddietterich: We should standardize on one library for equals and hash, but it doesn't affect correct operation.
One thing I note is that with the conversion of `equals` to `partialEquals` we do not have `equals` methods on many of the classes. So maybe we can pick a consistent equals/hash implementation and make a ticket to apply it everywhere.

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceState.java
```diff
@@ -0,0 +1,110 @@
...
+  NOT_EXISTS("invalid not exists", null);
+
+  // -- Valid state transitions --
+  private static final List<WsmResourceState> BROKEN_TRANSITIONS = List.of(DELETING);
+  // BROKEN_DELETING requires manual intervention to clean up
+  private static final List<WsmResourceState> BROKEN_DELETING_TRANSITIONS = List.of();
+  // We allow the NOT_EXISTS transition for the V1 apis where we move failed resources from
+  // creating to non-existent.
+  private static final List<WsmResourceState> CREATING_TRANSITIONS =
+      List.of(BROKEN, READY, NOT_EXISTS);
```
yuhuyoyo, line 30: to check my understanding, for v1, they will all be NOT_EXIST when creation fail but for v2, we will keep the resource in the db with a broken state?

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
```diff
@@ -86,17 +98,66 @@ public ControlledResource validateControlledResourceAndAction(
...
+        // and cannot be deleted by a user. It has to be cleaned up manually by Terra personnel.
+        throw new ResourceIsCorruptException(
+            "Resource is in an internally corrupted state. Contact support.");
+
+      case BROKEN:
+        // If the resource is in the BROKEN state, then there may be no Sam resource. We do our
+        // best with what we know. We handle two cases:
+        //  1. If the resource is an application resource, and the application is making the
+        // request,
+        //     we go ahead.
```
yuhuyoyo, line 132: nit. comment lint.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
```diff
@@ -86,17 +98,66 @@ public ControlledResource validateControlledResourceAndAction(
...
+                    + application.getDisplayName()
+                    + "can delete a broken application resource");
+          }
+        } else {
+          // broken user resource
+          SamRethrow.onInterrupted(
+              () ->
+                  samService.checkAuthz(
+                      userRequest,
+                      SamConstants.SamResource.WORKSPACE,
```
yuhuyoyo, line 151: why is the type WORKSPACE?

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceState.java
```diff
@@ -0,0 +1,110 @@
...
+    if (startState == endState) {
+      return true;
+    }
+
+    for (WsmResourceState validState : validTransitionList) {
+      if (validState == endState) {
+        return true;
+      }
+    }
+    return false;
```
zloery, line 108: nit: `return validTransitionList.contains(endState);`

service/src/main/resources/db/changesets/20230222_states.yaml
```diff
@@ -0,0 +1,43 @@
...
+    id: Add state to resources and cloud contexts
+    author: dd
+    changes:
+    - addColumn:
+        tableName: resource
+        columns:
+            - column:
+                  name: state
+                  type: text
+                  remarks: |
```
zloery, line 12: Does this need a default/backfill to `CREATED` for existing resources?

service/src/main/resources/db/changesets/20230222_states.yaml
```diff
@@ -0,0 +1,43 @@
...
+                      For transition states (e.g., updating), the flight doing the transition
+            - column:
+                  name: error
+                  type: text
+                  remarks: |
+                      Serialized exception; must be a subclass of ErrorReportException
+    -addColumn:
+        tableName: cloud_context
```
zloery, line 25: I don't see where this is used in this PR (outside of tests). if you're adding that in another PR can this change be part of that instead?

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -663,32 +661,245 @@ public boolean updateResource(
...
+    switch (resourceStateRule) {
+      case DELETE_ON_FAILURE -> {
+        deleteResourceWorker(
+            resource.getWorkspaceId(), resource.getResourceId(), /* resourceType */ null);
+      }
+
+      case BROKEN_ON_FAILURE -> {
+        // If the resource is already in the broken state with an empty flightId
+        // we assume we already completed this operation and this is a retry.
+        // Nothing to do here.
```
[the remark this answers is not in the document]
ddietterich: Ah, yes. I had more code here before, but it all moved to `updateState`. Will clarify.

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceState.java
ddietterich: Right, exactly. I want to move away from the "v1" and "v2", so I updated the comment to refer to the state rule.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
ddietterich: I updated the comment. It is because the Sam resource creation might have failed or we will have executed the "UNDO" on the create; since the Sam resource doesn't exist, we cannot test authz against it. So we test against the workspace.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
```diff
@@ -86,17 +98,66 @@ public ControlledResource validateControlledResourceAndAction(
...
+            "Another operation is running on the resource; wait and try again");
+
+      case BROKEN_DELETE:
+        // If the resource is in the BROKEN_DELETING state, then it is in an unknown broken state
+        // and cannot be deleted by a user. It has to be cleaned up manually by Terra personnel.
+        throw new ResourceIsCorruptException(
+            "Resource is in an internally corrupted state. Contact support.");
+
+      case BROKEN:
+        // If the resource is in the BROKEN state, then there may be no Sam resource. We do our
```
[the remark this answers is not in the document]
ddietterich: Delete is the only operation allowed on BROKEN. No operations are allowed on BROKEN_DELETE. It requires manual intervention. It is the state for dismal failures on the delete path. I don't think any current code can arrive in BROKEN_DELETE because the flight will stop with a dismal failure. My (unimplemented) idea is that we can alter the undo steps to mark a failed undo (in the working map) and successfully continue. The final step would leave the metadata in the database and mark it BROKEN_DELETE.

That said, I am uncertain it is the right design. On the one hand, it makes it clear to the UI user that there is a problem, and provides some explanation. Otherwise, the object is stuck in DELETING state with nothing happening. On the other hand, suppose the failure was some outage that is now restored. We could use `stairctl` to for the flight state to READY and allow the undo to complete normally. Which is better than manually going in to Sam or the cloud to delete resources.

I'll give this issue more visibility. Thanks for the prompt.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteMetadataStartStep.java
```diff
@@ -0,0 +1,60 @@
...
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    // If we successfully changed state, then we assume that something bad happened
+    // during delete processing and call this a broken delete.
+    // Otherwise, we assume it failed because the resource is busy and simply return.
+    var resourceStateChanged =
+        flightContext
+            .getWorkingMap()
+            .get(WorkspaceFlightMapKeys.ResourceKeys.RESOURCE_STATE_CHANGED, Boolean.class);
+    if (TRUE.equals(resourceStateChanged)) {
+      resourceDao.deleteResourceFailure(
```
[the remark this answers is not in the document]
ddietterich: Right. I think there are problems with a lot of the undo. And in the current code, I don't know of any cases that will execute that code. The "surfacing the failure" code is a mistake and will immediately cause dismal failure even when one is not merited. Stairway already holds on to the original failure. Undo code is suppose to return success if it succeeds in performing the undo.

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceState.java
ddietterich: TIL

service/src/main/resources/db/changesets/20230222_states.yaml
ddietterich: OK

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteMetadataStartStep.java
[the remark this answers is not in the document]
zloery: I think we chose the "surface the failure" approach to avoid hiding the real failure when we have steps we know we can't undo, and a dismal failure is the right approach.
e.g. for `DeleteMetadataStep` we can't undo the delete, so returning `success` isn't right, but if the step throws a new exception then the flight's exception is overwritten with something unhelpful like "can't undo DeleteMetadataStep" instead of the real failure cause.  Does Stairway support separate do and undo exceptions?

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/delete/DeleteMetadataStartStep.java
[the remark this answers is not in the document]
ddietterich: I believe it is not the case that returning a new FATAL err overwrites the real failure. If that is the behavior, it is a bug in Stairway. The intent is to preserve the original failure and return that as the flight status failure.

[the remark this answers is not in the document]
ddietterich: Damn! I write a Stairway test and it does what you said it would. I see the Stairway bug and will fix with a Stairway ticket.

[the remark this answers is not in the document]
ddietterich: I won't change this code, though.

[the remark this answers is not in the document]
zloery: sgtm! I thought the overwriting was intended behavior in stairway, not using the "surface failure" pattern sounds right given that's actually a stairway bug

zloery (approved):

PR #1138, opened 2023-03-10
[PF-1650] Modify WSM client default connection timeout

zloery:
Currently, the default WSM client believes idle connections will stay open for 1000s. However, the WSM server in all live environments (or at least the proxy in front of it) closes idle connections much faster, leading to `Connection Reset` errors on the client side. These don't tend to cause test failures as the JdkConnector automatically retries the requests (which might be an underlying cause for the unexpected retries we've seen before?), but they do flood test logs.

Tested with Verily's nightly integration test workflow, which went from dozens of `Connection Reset` issues to 0.


client/src/main/resources/swaggercodegen/libraries/jersey2/ApiClient.mustache
```diff
@@ -829,6 +829,9 @@ public class ApiClient {
     // Set idle timeout to 2m rather than the default of 30s, as clients may poll more
     // slowly for long-running operations.
     clientConfig.property(JdkConnectorProperties.CONTAINER_IDLE_TIMEOUT, /*120s*/ 120_000);
+    // Set client-side connection idle timeout to 30s rather than the default of 1000s to match
+    // server-side connection timeouts more accurately.
+    clientConfig.property(JdkConnectorProperties.CONNECTION_IDLE_TIMEOUT, /*30s*/ 30_000);
```
cbookg, line 834: maybe the client should be a somewhat lower value so we don't 'race', where exactly around 30 seconds the client thinks it is okay but the server times out?

zloery: Yeah - I'm still experimenting with this value, but generally our backend services will drop connections after ~a little over a minute, so 30s was a safe place to start

cbookg (approved):

PR #1139, opened 2023-03-10
[PF-2566] Fix BigQuery backfill lifetime test flakiness.

aaron-wang:
The BQ lifetime backfill picks up more datasets now, which leads to flakiness.

### Changes:
- Cleaned up a few more stray BQ datasets with undefined lifetime
- Removed deep equals comparison (when checking if the created dataset matches the original)

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
@@ -1364,9 +1364,8 @@ public class ControlledResourceServiceTest extends BaseConnectedTest {
                   null,
                   userAccessUtils.defaultUserAuthRequest()));
 
-      // check the expiration times stored on the cloud are still undefined, because the update
-      // above
-      // failed
+      // check the expiration times stored on the cloud are still undefined,
+      // because the update above failed
       validateBigQueryDatasetCloudMetadata(
           projectId, createdDataset.getDatasetName(), location, null, null);
 
@@ -1385,9 +1384,8 @@ public class ControlledResourceServiceTest extends BaseConnectedTest {
                   null,
                   userAccessUtils.defaultUserAuthRequest()));
 
-      // check the expiration times stored on the cloud are still undefined, because the update
-      // above
-      // failed
+      // check the expiration times stored on the cloud are still undefined,
+      // because the update above failed
       validateBigQueryDatasetCloudMetadata(
           projectId, createdDataset.getDatasetName(), location, null, null);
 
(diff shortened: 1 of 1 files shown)


service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -1348,41 +1348,54 @@ void updateBqDatasetWithInvalidExpirationTimes() throws Exception {
...
 
-    // Remove dataset to not conflict with other test that checks for empty lifetime
-    controlledResourceService.deleteControlledResourceSync(
-        workspaceId, resource.getResourceId(), userAccessUtils.defaultUserAuthRequest());
+    } finally {
+      // Remove dataset to not conflict with other test that checks for empty lifetime
+      controlledResourceService.deleteControlledResourceSync(
+          workspaceId, resource.getResourceId(), userAccessUtils.defaultUserAuthRequest());
+    }
   }
```
aaron-wang, line 1399: Surrounding this test with a `try... finally` clause so the BQ dataset is always removed.

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -1968,28 +1981,33 @@ public void updateControlledBigQueryDataset_HandlesDatasets_WithOneNullLifetime(
...
+          updatedResourceList,
+          createdDataset.getResourceId(),
+          /*expectedTableLifetime=*/ null,
+          DEFAULT_CREATED_BIG_QUERY_PARTITION_LIFETIME);
+    } finally {
+      // Remove dataset to not conflict with other test that checks for empty lifetime
+      controlledResourceService.deleteControlledResourceSync(
+          workspaceId, createdDataset.getResourceId(), userAccessUtils.defaultUserAuthRequest());
+    }
   }
```
aaron-wang, line 2011: Surrounding this test with a `try... finally` clause so the BQ dataset is always removed.

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceTest.java
```diff
@@ -2042,7 +2060,12 @@ private ControlledBigQueryDatasetResource makeBigQueryDatasetWithLifetime(
                 resource, null, user.getAuthenticatedRequest(), creationParameters)
             .castByEnum(WsmResourceType.CONTROLLED_GCP_BIG_QUERY_DATASET);
 
-    assertEquals(resource, createdDataset);
+    assertEquals(resource.getResourceId(), createdDataset.getResourceId());
+    assertEquals(datasetId, createdDataset.getDatasetName());
+    assertEquals(projectId, createdDataset.getProjectId());
+    assertEquals(defaultTableLifetime, createdDataset.getDefaultTableLifetime());
+    assertEquals(defaultPartitionLifetime, createdDataset.getDefaultPartitionLifetime());
+
```
aaron-wang, line 2068: No longer using deep equals comparison. We should only care if the resource ID, and the attributes match up (most importantly, the lifetime)

aaron-wang:
The connected test for the first commit had a failed policy test (unrelated to this change).

zloery (approved):

PR #1142, opened 2023-03-13
PF-2571: use partialEqual instead

yuhuyoyo:


rogerwangcs (approved):
Would there be any other instances where this needs to be added or is it just bqdataset.

ddietterich (approved):

PR #1143, opened 2023-03-13
[PF-2572] Use internal class for flex resource creation parameters.

aaron-wang:
Adds internal class `FlexResourceCreationParameters` which will be used in creation / cloning flights.

#1133 will be merged on top of this change.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/FlexResourceCreationParameters.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/FlexResourceCreationParameters.java
@@ -3,8 +3,6 @@ package bio.terra.workspace.service.resource.controlled.cloud.any.flexibleresour
 import bio.terra.workspace.generated.model.ApiControlledFlexibleResourceCreationParameters;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import org.jetbrains.annotations.NotNull;
-
 import javax.annotation.Nullable;
 
 public class FlexResourceCreationParameters {
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/FlexResourceCreationParameters.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/FlexResourceCreationParameters.java
@@ -1,5 +1,6 @@
 package bio.terra.workspace.service.resource.controlled.cloud.any.flexibleresource;
 
+import bio.terra.workspace.generated.model.ApiControlledFlexibleResourceCreationParameters;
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/FlexResourceCreationParameters.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  public static FlexResourceCreationParameters fromApiCreationParameters(
+      ApiControlledFlexibleResourceCreationParameters apiCreationParamters) {
+    return new FlexResourceCreationParameters()
+        .typeNamespace(apiCreationParamters.getTypeNamespace())
+        .type(apiCreationParamters.getType())
+        .data(apiCreationParamters.getData());
+  }
+}
```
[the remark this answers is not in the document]
ddietterich: Did you consider making this a record instead of a class?
You can add static methods to a record.

ddietterich (approved):
This is fine. Consider using a record.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/FlexResourceCreationParameters.java
```diff
@@ -0,0 +1,28 @@
...
+import bio.terra.workspace.generated.model.ApiControlledFlexibleResourceCreationParameters;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import javax.annotation.Nullable;
+
+public record FlexResourceCreationParameters(
+    String typeNamespace, String type, @Nullable byte[] data) {
+
+  @JsonCreator
+  public FlexResourceCreationParameters(
```
yuhuyoyo, line 12: Do we need json serilization of this class? in another word, can we not have this constructor? (kinda defeat the purpose of making it into a record)

we can add @JsonSerialize if we must. https://dev.to/brunooliveira/practical-java-16-using-jackson-to-serialize-records-4og4

aaron-wang: I don't think so. I can remove this constructor. We'll see later if we need the JSON serialization (probably not).

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/FlexResourceCreationParameters.java
ddietterich: The structure will be serialized when used as a Stairway parameter. Part of my goal here is to use internally defined classes for everything we serialize in Stairway. It can be hard to correlate an API change to a deserialization error in a flight.

Jackson is very good at serdes of Java records. You shouldn't need a special serializer

PR #1145, opened 2023-03-14
[WOR-799] Bump landing zone lib version

aherbst-broad:
Pull in the latest landing zone lib that has changes to facilitate Azure E2E testing, add optional landing zone ID parameter (the LZ service API already accepts this, this is just exposing it in the WSM API).


openapi/src/parts/azure_landing_zone.yaml
```diff
@@ -189,6 +189,10 @@ components:
       type: object
       required: [definition, billingProfileId]
       properties:
+        landingZoneId:
+          description: The ID of the landing zone (optional)
```
cahrens, line 193: It would be good to describe more how this would be used. And if the `landingZoneId` is specified, then the `definition` is _not_ used, correct? Is `definition` still a required property in that case?

aherbst-broad: Definition is required and always used as it specifies what type of landing zone to create (i.e, CromwellBaseResources in almost all cases); if the ID is omitted (as it is today) one is auto generated.

cahrens: I was assuming here that if the ID is specified, then attachment is done to an existing landing zone. That's not what is going on?

aherbst-broad: No. We need to be able to specify the ID for the landing zone from test code as the pre deployed landing zone resources will already be tagged with that ID.

cahrens (approved):

blakery (approved):

PR #1146, opened 2023-03-14
[PF-2323] Allow updating the CPU and GPU config of AI notebooks.

aaron-wang:
Updating the CPU or GPU configuration requires the notebook instance to be stopped.
### Changes:
- Add `machineType` and `acceleratorConfig` to AI notebook attributes.
- Add steps for updating the CPU/GPU config in the cloud.
  - `CheckAiNotebookStoppedForGpuAndCpuUpdateStep` -> `UpdateNotebookCpuAndGpuAttributesStep` -> `UpdateAiNotebookCpuAndGpuStep`. 
  - See `ControlledAiNotebookInstanceResource` for new update steps (we recently added an `addUpdateSteps` method for controlled resources).
 - Set the `machineType` and `acceleratorConfig` during notebook creation (write to the database)


aaron-wang:
The GitHub actions of the last few commits were failing due to policy tests.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -61,6 +62,7 @@
 import bio.terra.workspace.service.workspace.model.GcpCloudContext;
 import bio.terra.workspace.service.workspace.model.OperationType;
 import bio.terra.workspace.service.workspace.model.WsmApplication;
+import com.google.api.services.notebooks.v1.model.AcceleratorConfig;
```
ddietterich, line 65: We should not serialize the Google datatype for AcceleratorConfig. Suggest you specify a record for that using types that properly serialize. You can make to/fromGcp methods in there.

We do not want a serialized data dependency on an externally defined structure. If that structure changes we will have a compatibility problem outside of our control and a custom serdes that doesn't work.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/AcceleratorConfigDeserializer.java
```diff
@@ -0,0 +1,38 @@
+package bio.terra.workspace.service.resource.controlled.cloud.gcp.ainotebook;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.google.api.services.notebooks.v1.model.AcceleratorConfig;
+import java.io.IOException;
+
+public class AcceleratorConfigDeserializer extends StdDeserializer<AcceleratorConfig> {
```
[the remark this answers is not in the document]
ddietterich: As above - please don't do it this way.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/AiNotebookApiConversions.java
```diff
@@ -0,0 +1,32 @@
+package bio.terra.workspace.service.resource.controlled.cloud.gcp.ainotebook;
+
+import bio.terra.workspace.generated.model.ApiGcpAiNotebookInstanceAcceleratorConfig;
+import com.google.api.services.notebooks.v1.model.AcceleratorConfig;
+import javax.annotation.Nullable;
+
+/** Utility method for converting AI notebook objects between WSM formats and GCP formats. */
+public class AiNotebookApiConversions {
```
ddietterich, line 8: If you make a record or class for the internal and serializable format of the accelerator config, you can put these methods into that class.

ddietterich (changes requested):
Do not serialize the Google Accelerator config. More detail in the comments.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -505,6 +507,13 @@ public String createAiNotebookInstance(
             "enablePet");
     jobBuilder.addParameter(ControlledResourceKeys.CREATE_NOTEBOOK_PARAMETERS, creationParameters);
     jobBuilder.addParameter(ControlledResourceKeys.NOTEBOOK_PET_SERVICE_ACCOUNT, petSaEmail);
+    jobBuilder.addParameter(
+        ControlledResourceKeys.CREATE_NOTEBOOK_MACHINE_TYPE, creationParameters.getMachineType());
+    jobBuilder.addParameter(
+        ControlledResourceKeys.CREATE_NOTEBOOK_ACCELERATOR_CONFIG,
+        AiNotebookApiConversions.fromApiAcceleratorConfig(
+            creationParameters.getAcceleratorConfig()));
+
```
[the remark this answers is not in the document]
aaron-wang: So, when updating these attributes in the database `UpdateNotebookResourceAttributesStep` (previously named `UpdateNotebookResourceLocationAttributesStep`) I wanted to avoid deserializing the API object. However, seeing as the other create steps use the API object regardless, I can remove these parameters.

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -584,9 +594,20 @@ public ControlledAiNotebookInstanceResource updateAiNotebookInstance(
             .workspaceId(resource.getWorkspaceId().toString())
             .resourceType(resource.getResourceType())
             .stewardshipType(resource.getStewardshipType())
-            .addParameter(ControlledResourceKeys.UPDATE_PARAMETERS, updateParameters)
             .addParameter(ResourceKeys.RESOURCE_NAME, newName)
             .addParameter(ResourceKeys.RESOURCE_DESCRIPTION, newDescription);
+    if (updateParameters != null) {
```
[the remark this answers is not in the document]
aaron-wang: It's on! ;)

openapi/src/resources/gcp_ai_notebook_instance.yaml
```diff
@@ -5,7 +5,7 @@ components:
         AI Notebook instance properties included in post-creation, get, and update. Others must be
         retrieved from GCP using the specification.
       type: object
-      required: [projectId, location, instanceId]
+      required: [projectId, location, instanceId, machineType]
```
ddietterich, line 8: Making this required will not be upward compatible. You need to be able to handle defaulting the machine type.

service/gradle/dependencies.gradle
```diff
@@ -8,6 +8,7 @@ dependencies {
 
   implementation 'com.google.cloud:google-cloud-bigquery'
   implementation 'com.google.cloud:google-cloud-bigquerydatatransfer'
+  implementation 'com.google.cloud:google-cloud-notebooks:1.10.0'
```
ddietterich, line 11: I don't think we need the version here, unless the right version is not supplied by the google BOM.
The idea of the BOM is that all of the different libraries play nice with each other.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/AcceleratorConfig.java
```diff
@@ -0,0 +1,35 @@
+package bio.terra.workspace.service.resource.controlled.cloud.gcp.ainotebook;
+
+import bio.terra.workspace.generated.model.ApiGcpAiNotebookInstanceAcceleratorConfig;
+import javax.annotation.Nullable;
+
+/**
+ * Interal record for the AcceleratorConfig of AI notebooks {@link
```
ddietterich, line 7: nit - spelling

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/UpdateNotebookCpuAndGpuAttributesStep.java
```diff
@@ -0,0 +1,69 @@
...
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import bio.terra.workspace.db.DbSerDes;
+import bio.terra.workspace.db.ResourceDao;
+import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys;
+import java.util.Optional;
+
+// Update the CPU and GPU attributes in the database.
+public class UpdateNotebookCpuAndGpuAttributesStep implements Step {
```
ddietterich, line 17: I think the code would be simpler if you combined the UpdateAiNotebookMetadataAttributesStep and this step into a single update of the attribute data. I don't think there is a reason the metadata update has to come before updating the CPU/GPU.

ddietterich (changes requested):
I'd like a few changes

yuhuyoyo (approved):
thanks!

openapi/src/resources/gcp_ai_notebook_instance.yaml
ddietterich: I took a look at the code. IIUC, the existing API requires a machine type to be specified on create. There is no defaulting. And the machine type is currently not update-able. So my initial comment about making the field required is nonsense.

I think that we will want to backfill from adding the machine type and accelerator to the notebook attributes. That can be done as you suggest, by reading the machineType from the cloud platform and recording it in the attributes.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
```diff
@@ -50,16 +63,53 @@ public WsmResource makeResourceFromDb(DbResource dbResource) {
...
-            .build();
-    return resource;
+    // Notebook attributes created prior to PF-2323 do not have machineType or acceleratorConfig
+    // stored. Note: machineType is required.
+    // Thus, retrieve these two values from the cloud.
+    String retrievedMachineType = attributes.getMachineType();
+    AcceleratorConfig retrievedAcceleratorConfig = attributes.getAcceleratorConfig();
+
+    if (retrievedMachineType == null) {
+      try (NotebookServiceClient notebookServiceClient = NotebookServiceClient.create()) {
```
aaron-wang, line 73: @ddietterich Do we backfill attributes as we read the notebooks? Or do we want a separate process akin to the BQ backfill?

ddietterich: I vote for the BQ-backfill type of process

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookInstanceResource.java
```diff
@@ -73,21 +75,32 @@ public ControlledAiNotebookInstanceResource(
...
   }
 
   // Constructor for the builder
   private ControlledAiNotebookInstanceResource(
-      ControlledResourceFields common, String instanceId, String location, String projectId) {
+      ControlledResourceFields common,
+      String instanceId,
+      String location,
+      String projectId,
+      String machineType,
```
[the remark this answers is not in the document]
ddietterich: Yes, I think not.

service/src/test/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiControllerAiNotebookTest.java
```diff
@@ -116,6 +120,48 @@ public void createAiNotebookInstance_correctZone() throws Exception {
...
+                notebook.getAttributes().getInstanceId());
+    // Stop the notebook so the CPU and GPU can be updated.
+    crlService.getAIPlatformNotebooksCow().instances().stop(instanceName).execute();
+
+    var updatedNotebook =
+        mockMvcUtils.updateAiNotebookInstance(
+            userAccessUtils.defaultUserAuthRequest(),
+            workspaceId,
+            notebook.getMetadata().getResourceId(),
+            "n1-highmem-2",
```
ddietterich, line 154: Suggest putting the test values in static finals, so you can do compares using those and not repeating the literals.

service/src/test/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiControllerAiNotebookTest.java
```diff
@@ -116,6 +120,48 @@ public void createAiNotebookInstance_correctZone() throws Exception {
         List.of(WorkspaceConstants.Properties.DEFAULT_RESOURCE_LOCATION));
   }
 
+  @Test
```
ddietterich, line 123: I don't see a test that validates the notebook state checking. Is there one?

Could be added in here or in the integration tests, but seems like it would be good to cover the error case of the notebook not stopped.

ddietterich (commented):
Some comments. Biggest suggestion is to add a test for the notebook-not-stopped error case. Not sure we could contrive a test that would hit the "stopping-retry" case. Maybe semi-randomly hit it by requesting the notebook to stop and immediately trying to update the machine type.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
ddietterich: I think that covers it well

service/gradle/dependencies.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
 
   implementation 'com.google.cloud:google-cloud-bigquery'
   implementation 'com.google.cloud:google-cloud-bigquerydatatransfer'
+  implementation 'com.google.cloud:google-cloud-notebooks'
```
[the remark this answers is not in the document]
ddietterich: Hmm... yeah. Looks like we should update that BOM version - there are known vulnerabilities :(
I made PF-2655 to track this.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/UpdateAiNotebookCpuAndGpuStep.java
```diff
@@ -0,0 +1,137 @@
...
+        RetryUtils.getWithRetryOnException(
+            () ->
+                notebookServiceClient
+                    .setInstanceMachineTypeAsync(
+                        SetInstanceMachineTypeRequest.newBuilder()
+                            .setName(instanceName.formatName())
+                            .setMachineType(effectiveMachineType)
+                            .build())
+                    .get(),
+            Duration.ofMinutes(7),
```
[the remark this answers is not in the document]
ddietterich: You could make a method that did both calls. On an error, it would redo both calls, but maybe that is OK.

ddietterich (approved):
LGTM - let me know if I can help with the conflicts.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/UpdateAiNotebookCpuAndGpuStep.java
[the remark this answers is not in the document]
aaron-wang: Ah, makes sense!

I've run into queuing issues before, so I'll keep this separate for now.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/ControlledAiNotebookHandler.java
```diff
@@ -50,11 +63,51 @@ public WsmResource makeResourceFromDb(DbResource dbResource) {
...
+                Duration.ofMinutes(5),
+                DEFAULT_RETRY_SLEEP_DURATION,
+                DEFAULT_RETRY_FACTOR_INCREASE,
+                DEFAULT_RETRY_SLEEP_DURATION_MAX,
+                retryableErrors);
+
+        retrievedMachineType = parseMachineTypeFromUrl(cloudInstance.getMachineType());
+        retrievedAcceleratorConfig =
+            new AcceleratorConfig(
+                cloudInstance.getAcceleratorConfig().getType().toString(),
```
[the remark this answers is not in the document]
aaron-wang: So, the `getAcceleratorConfig` method of an `Instance` (`google.cloud.notebooks.v1.Instance`) is annotated with "not null". 

If "it is null", it instead returns a *default instance*, which is an empty `AcceleratorConfig` object.

Interestingly, the `getType` method is also marked not null.

By default, it returns "UNSPECIFIED", and will fallback to "UNRECOGNIZED" in other cases.

I've added a brief comment in the code documenting this.

```java
// Note: getAcceleratorConfig (of google.cloud.notebooks.v1.Instance) never returns null.
// Instead, an empty AcceleratorConfig object is returned.
```

PR #1149, opened 2023-03-16
PF-2596: doesn't allow same instance in a workspace because UI doesn't pass in location

yuhuyoyo:
When we use default location for resources, we use us-central1 but later we upddate it to us-central1-a. So if i rapidly create two notebooks, the second one will fail to be created because the first one hasn't got the location updated yet. However, if the first one finished creation, the location will get changed to us-central1-a and then there's seemingly no conflict.


ddietterich (approved):
Seems like a reasonable restriction to me.

yuhuyoyo:
> Seems like a reasonable restriction to me.

so I think the more right thing to do is checkuniquness on update too but that's harder to do.

zloery (approved):

PR #1151, opened 2023-03-17
[PF-2269] Remove BigQuery lifetime backfill.

aaron-wang:
Removes the code supporting the BQ lifetime backfill.

Also updated a few stray tests (in `ControlledGcpResourceApiControllerBqDatasetTest`) related to the completion of PF-2269.


ddietterich:
I'm looking at the BigQuery update path and I am not seeing any place where the updated BQ attributes are stored the database. It looks like they are only updated on the cloud platform. I feel like I must be missing something. Can you double check and see if the metadata attributes are updated in the current code path? If we are not doing the update, then we shouldn't remove the backfill yet.

Maybe I broke this somehow as part of my update changes. I am going to make a fix in my branch.

aaron-wang:
> I'm looking at the BigQuery update path and I am not seeing any place where the updated BQ attributes are stored the database.

Correct. I checked and currently we're only updating the attributes on the cloud platform. The only metadata being updated are the common controlled resource metadata (name, description, etc.).

I can put up a PR to add an "update BQ attributes in the database" step, if the update steps are further away from being published.

aaron-wang:
@ddietterich I opened a quick PR #1155 to resolve the BQ update bug :)

aaron-wang:
#1231 will resolve this.

PR #1155, opened 2023-03-20
[PF-2604] Update BigQuery lifetime attributes in database.

aaron-wang:
We added `defaultTableLifetime` and `defaultPartitionLifetime` to the controlled BigQuery attributes in #1034.

However, we only update the BQ-specific attributes on the cloud.

This PR adds a step to update the lifetime attributes in the database (before updating the cloud).

I used the API object for the update step, but I figured this would be replaced / fixed by Dan's update refactoring.


aaron-wang:
Sounds good! We can keep the backfill until then.

PR #1157, opened 2023-03-20
[Pf-2452] Handle exceeded quota limit in create notebook step.

aaron-wang:
Previously, `GcpUtils.pollAndRetry` would continually retry creating the notebook, despite the error:

- `Quota 'CPUS' exceeded.  Limit: 72.0 in region us-east1.`

Previously, we only threw for a [subset](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java#L103) of client errors (4XX codes) 

Now, there's a clause that throws an exception if permission is denied. I added this separately so the error message can be cleaner in the UI.


aaron-wang:
Update: will use unit test instead and mock the quota limit error instead.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStepTest.java
```diff
@@ -31,6 +54,79 @@ public class CreateAiNotebookInstanceStepTest extends BaseUnitTest {
...
+  @Mock private AIPlatformNotebooksCow mockAIPlatformNotebookCow;
+  @Mock private AIPlatformNotebooksCow.Instances mockInstances;
+  @Mock private AIPlatformNotebooksCow.Instances.Create mockInstancesCreate;
+
+  @Mock private AIPlatformNotebooksCow.Operations mockOperations;
+  @Mock private OperationCow<Operation> mockOperationCow;
+  @Mock private OperationCow.OperationAdapter<Operation> mockOperationAdapter;
+  @Mock private OperationCow.OperationAdapter.StatusAdapter mockOperationStatusAdapter;
+  @Mock private Operation mockOperation;
+
```
aaron-wang, line 71: @yuhuyoyo Should I split the mock notebook COW stuff into a separate test file or leave it here?

I would prefer having a `setup()` function, but the other tests don't use the COW.

service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
```diff
@@ -82,6 +83,9 @@ public static void pollAndRetry(
         // do not waste time retrying on client error.
         throw new BadRequestException(
             String.format("Gcp calls failed with client error code %s. Do not retry", code));
+      } else if (code == Code.PERMISSION_DENIED.value()) {
```
yuhuyoyo, line 86: Is permission_denied 403? how do we even get to here and skip the is4xxClientError check? 

also surprising that it throws permission_denied for quota limit. could you add a comment here explaining that this also handles quota limit execeed error?

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStepTest.java
```diff
@@ -31,6 +54,79 @@ public class CreateAiNotebookInstanceStepTest extends BaseUnitTest {
...
+    GcpCloudContext fakeCloudContext = new GcpCloudContext(FAKE_PROJECT_ID);
+
+    workingFlightMap.put(
+        WorkspaceFlightMapKeys.ControlledResourceKeys.GCP_CLOUD_CONTEXT, fakeCloudContext);
+    workingFlightMap.put(CREATE_NOTEBOOK_LOCATION, "fake-location");
+    doReturn(workingFlightMap).when(mockFlightContext).getWorkingMap();
+
+    // Mock the permission error thrown when polling the error (e.g., quota exceeded limit)
+    when(mockOperationAdapter.getError()).thenReturn(mockOperationStatusAdapter);
+    when(mockOperationStatusAdapter.getCode()).thenReturn(Code.PERMISSION_DENIED_VALUE);
```
yuhuyoyo, line 126: all the set up before this part is not really necessary. can we just test this part in GcpUtilsTest? sorry i don't think GcputilsTest doesn't exists yet -.-

aaron-wang: Sure! Will do.

Goodbye cow 🐄 . \o/

service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
aaron-wang: Yes, it's 403. So, the `is4xxClientError` check has the following [subset](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java#L103) of codes:

```java
private static boolean is4xxClientError(int code) {
    return Code.INVALID_ARGUMENT.value() == code // 400
        || Code.OUT_OF_RANGE.value() == code // 400
        || Code.FAILED_PRECONDITION.value() == code // 400
        || Code.ALREADY_EXISTS.value() == code // 409
        || Code.ABORTED.value() == code // 409
        || Code.UNAUTHENTICATED.value() == code // 401
        || Code.RESOURCE_EXHAUSTED.value() == code; // 429
  }
```

I think the `PERMISSION_DENIED` refers to *not being allowed* to create *because* of the quota limit.

I'll add a comment about the quota limit!

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/CreateAiNotebookInstanceStepTest.java
yuhuyoyo: pollAndRetry still takes in OperationCow so i don't  think you can completely say goodbye to it

aaron-wang: :(

Welcome back cow!

service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
yuhuyoyo: i'm personally ok with putting it in the is4xxClientError and just call it a badRequest too... since we are not throwing conflict for 409 or forbidden for 401.

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
```diff
@@ -77,11 +80,26 @@ public static void pollAndRetry(
...
+              errorMessage);
+            // 401
+          case UNAUTHENTICATED -> throw new UnauthorizedException(errorMessage);
+            // 403
+          case PERMISSION_DENIED -> throw new ForbiddenException(errorMessage);
+            // 409
+          case ALREADY_EXISTS, ABORTED -> throw new ConflictException(errorMessage);
+            // 429
+          case RESOURCE_EXHAUSTED -> throw new BadRequestException(errorMessage);
+          default -> throw new BadRequestException(
```
zloery, line 100: You've already checked for every value of `code` that satisfies `is4xxClientError`, so I don't think you'll ever hit this default case. I think you can just remove `is4xxClientError` and throw the RetryException as the default case

zloery (approved):

PR #1165, opened 2023-03-22
Fix bug with missing blob name in URL (WOR-830).

cahrens:
See https://broadworkbench.atlassian.net/browse/WOR-830 for details.

As far as I can tell (searching in repos and posting on [Slack](https://broadinstitute.slack.com/archives/C02GJ782BAA/p1679513399512369)), nobody is currently using the current URL format and injecting the blob name into it. I thought terra-ui was because it generates URLs with blob names in them, but the SAS token it uses (and URL modifies) is actually returned by a call to get the SAS token for the entire container.


aherbst-broad (approved):
lgtm

marctalbott (approved):

PR #1174, opened 2023-03-24
[TERRA-314] Move cloud-specific schema / parameter definitions to separate file

dexamundsen:
Move cloud-specific (Azure) schema / parameter definitions to separate file
This is in preparation for AWS related changes coming in with similar separate files

### DIFF
--- a/openapi/src/common/schemas.yaml
+++ b/openapi/src/common/schemas.yaml
@@ -11,7 +11,7 @@ components:
     CloudPlatform:
       type: string
       description: Enum representing a cloud platform type.
-      enum: ['AZURE', 'GCP']
+      enum: ['GCP', 'AZURE']
 
     CloneReferencedResourceRequestBody:
       description: >-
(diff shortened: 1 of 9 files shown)


openapi/src/common/schemas.yaml
```diff
@@ -11,7 +11,7 @@ components:
     CloudPlatform:
       type: string
       description: Enum representing a cloud platform type.
-      enum: ['AZURE', 'GCP']
+      enum: ['GCP', 'AZURE']
```
zloery, line 14: out of curiosity, why this swap?

zloery (approved):
LGTM
Can you also update `openapi/README` with this new pattern?

openapi/src/common/schemas.yaml
dexamundsen: Every other place - GCP was listed before Azure. keeping it consistent.

jmczerk (approved):

PR #1176, opened 2023-03-28
WOR-827 enforce region policy for azure

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-827

Check region policy when an azure cloud context is created. When cloning a workspace, change order of stairway steps to populate policy before creating cloud context.

Azure connected tests no longer create resources that are expected to be created by landing zones. The resources that the tests require have been preallocated in MRG `mrg-terra-integration-test-20211118`. The related test code to create these resource types has been removed. This saves time and removes cross test interference.


service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
```diff
@@ -198,18 +198,15 @@ public static void validateBucketName(
...
-        // validateAzureRegion(location);
-      }
-      case GCP -> validateGcpRegion(tpsApiDispatch, workspaceUuid, location);
+      case AZURE -> validateRegion(tpsApiDispatch, workspaceUuid, region, platform);
+      case GCP -> validateRegion(
+          tpsApiDispatch, workspaceUuid, GcpUtils.parseRegion(region), platform);
       case ANY -> {
         // Flexible resources are not stored on the cloud. Thus, they have no region policies.
       }
-      default -> throw new InvalidControlledResourceException("Unrecognized platform");
```
dvoet, line 212: I removed this because a) it is not required and b) I changed `InvalidControlledResourceException` into a more specific exception: `RegionNotAllowedException`

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneWorkspaceFlight.java
```diff
@@ -22,15 +21,16 @@ public class CloneWorkspaceFlight extends Flight {
...
-    // 0. Clone all folders in the workspace
-    // 1. Build a list of resources to clone and attach the updated cloned folder id
-    // 2. Create job IDs for future sub-flights and a couple other things
-    // 3. Launch a flight to create a cloud context if necessary
-    // 3a. Await the context flight
-    // TODO: [PF-1972] 4. Merge Policy Attributes
-    // 5. Launch a flight to clone all resources on the list
-    // 5a. Await the clone all resources flight and build a response
+    // Flight Plan
+    // * Merge Policy Attributes
```
dvoet, line 25: Merge Policy Attributes was moved up from step 4

service/src/test/java/bio/terra/workspace/common/StairwayTestUtils.java
```diff
@@ -53,7 +53,7 @@ public static FlightState blockUntilFlightCompletes(
 
     stairway.submitWithDebugInfo(
         flightId, flightClass, inputParameters, /* shouldQueue= */ false, debugInfo);
-    return pollUntilComplete(flightId, stairway, timeout.dividedBy(20), timeout);
+    return pollUntilComplete(flightId, stairway, Duration.ofSeconds(1), timeout);
```
dvoet, line 56: This should reduce overall test time. The timeout is currently hardcoded to 75m so this polling interval was set to 3.75m meaning any test that calls this waits a minimum of that time. This is a simple db lookup, doing it once per second should not be a problem.

service/src/test/java/bio/terra/workspace/service/workspace/flight/azure/CreateAzureContextFlightTest.java
```diff
@@ -52,7 +77,7 @@ void successCreatesContext() throws Exception {
     // Wait for the job to complete
     FlightState flightState =
         StairwayTestUtils.pollUntilComplete(
-            jobId, jobService.getStairway(), Duration.ofSeconds(30), STAIRWAY_FLIGHT_TIMEOUT);
+            jobId, jobService.getStairway(), Duration.ofSeconds(1), STAIRWAY_FLIGHT_TIMEOUT);
```
dvoet, line 80: I changed all the polling intervals in this file to 1s. Each tests now takes only a few seconds instead of 30s.

service/src/main/java/bio/terra/workspace/common/utils/GcpUtils.java
```diff
@@ -181,7 +181,7 @@ public static GoogleCredentials getGoogleCredentialsFromUserRequest(
    * "-[a-z]".
    */
   public static String parseRegion(String location) {
-    return location.replaceAll("(?!^)-[a-z]$", "");
+    return location.replaceAll("(?!^)-[a-zA-Z]$", "");
```
dvoet, line 184: case insensitive

cahrens (approved):

PR #1179, opened 2023-03-29
[PF-2626] Include flex resources in workspace clone flight.

aaron-wang:
All the cloning logic is done from #1133 . We just need to put it in the clone workspace flight.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/flexibleresource/CloneControlledFlexibleResourceFlight.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/flexibleresource/CloneControlledFlexibleResourceFlight.java
@@ -42,7 +42,6 @@ public class CloneControlledFlexibleResourceFlight extends Flight {
         inputParameters,
         WorkspaceFlightMapKeys.ResourceKeys.RESOURCE,
         JobMapKeys.AUTH_USER_INFO.getKeyName(),
-        WorkspaceFlightMapKeys.ResourceKeys.CLONING_INSTRUCTIONS,
         WorkspaceFlightMapKeys.ControlledResourceKeys.DESTINATION_RESOURCE_ID,
         WorkspaceFlightMapKeys.ControlledResourceKeys.DESTINATION_WORKSPACE_ID);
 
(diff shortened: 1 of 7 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/flexibleresource/CloneControlledFlexibleResourceFlight.java
```diff
@@ -42,7 +42,6 @@ public CloneControlledFlexibleResourceFlight(FlightMap inputParameters, Object b
         inputParameters,
         WorkspaceFlightMapKeys.ResourceKeys.RESOURCE,
         JobMapKeys.AUTH_USER_INFO.getKeyName(),
-        WorkspaceFlightMapKeys.ResourceKeys.CLONING_INSTRUCTIONS,
```
aaron-wang, line 45: Removed this unnecessary restriction: The resolved cloning instructions are calculated later regardless.

If the cloning instructions are not provided in the input parameters, then we use the source flex resource cloning instructions.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCloneControlledFlexibleResourceFlightStep.java
```diff
@@ -0,0 +1,97 @@
...
+import bio.terra.stairway.exception.StairwayExecutionException;
+import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
+import bio.terra.workspace.service.job.JobMapKeys;
+import bio.terra.workspace.service.resource.controlled.cloud.any.flexibleresource.ControlledFlexibleResource;
+import bio.terra.workspace.service.resource.controlled.flight.clone.flexibleresource.CloneControlledFlexibleResourceFlight;
+import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys;
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+public class LaunchCloneControlledFlexibleResourceFlightStep implements Step {
```
aaron-wang, line 22: So, I tested this locally with a few cases via Swagger. 

Most importantly: I had flex resource A with `COPY_RESOURCE`, and flex resource B with `COPY_NOTHING`. 

Resource A copied over, and resource B did not copy over (thus, the correct behavior).

Do we need to add tests confirming the flex resources are cloned over during the workspace clone flight?

ddietterich: You already validate the resource clone operation. So the additional piece to cover is making sure that the resources are included in the clone. Perhaps add a flex resource to the integration `CloneWorkspace` test. I think the test structure is pretty obvious.

ddietterich (approved):
LGTM - I would suggest extending the CloneWorkspace integration test to add a flex resource.

zloery (approved):

PR #1166, opened 2023-03-23
remove unused azure resources and region parameters

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-825
https://broadworkbench.atlassian.net/browse/WOR-838

This is a big PR mostly because it deletes a bunch of code: every Azure controlled resource that is handled by the landing zone. This also means some connected tests needed changing to simulate required landing zone resources.

Since landing zone dictates region for Azure workspaces, remaining Azure controlled resources are populated up front with that region. This means stairway steps to populate or backfill region are not required and also deleted.

Looks like this cuts the time for azureConnectedTest in half!

PR #1184, opened 2023-04-03
Always run with BPM enabled for Azure (WOR-743).

cahrens:
Ticket:  https://broadworkbench.atlassian.net/browse/WOR-743

We previously minimized the usage of Azure workspaces with BPM disabled in https://github.com/DataBiosphere/terra-workspace-manager/pull/1010. The original hope was to remove the feature flag with that PR, but Azure VM integration tests that relied on creating a CloudContext without BPM prevented complete removal of the feature flag.

However, https://github.com/DataBiosphere/terra-workspace-manager/pull/1166 deleted the problematic integration test code, so we can now change Azure CloudContext creation to unconditionally pull MRG coordinates from the BPM spend profile.

### DIFF
--- a/integration/src/main/java/scripts/utils/CloudContextMaker.java
+++ b/integration/src/main/java/scripts/utils/CloudContextMaker.java
@@ -1,7 +1,6 @@
 package scripts.utils;
 
 import bio.terra.workspace.api.WorkspaceApi;
-import bio.terra.workspace.model.AzureContext;
 import bio.terra.workspace.model.CloudPlatform;
 import bio.terra.workspace.model.CreateCloudContextRequest;
 import bio.terra.workspace.model.CreateCloudContextResult;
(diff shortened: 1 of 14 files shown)


integration/src/main/java/scripts/utils/CloudContextMaker.java
```diff
@@ -36,26 +35,6 @@ public static String createGcpCloudContext(UUID workspaceUuid, WorkspaceApi work
     return contextResult.getGcpContext().getProjectId();
   }
 
-  /**
-   * Creates an Azure cloud context for a given workspace. Returns the Azure resource group ID as a
-   * string
-   */
-  public static String createAzureCloudContext(
```
cahrens, line 43: Method is now unused after https://github.com/DataBiosphere/terra-workspace-manager/pull/1166.

aherbst-broad (approved):
Very nice, thanks for paying down this debt.

blakery (approved):

PR #1187, opened 2023-04-05
Fix notebook integration test error and connected test flakiness.

aaron-wang:
Overall, the biggest test change is waiting for the stop request to be properly executed (and not just queued), before continuing.

#### Integration
In the integration test, we never queued the stop request before. Pretty sure we needed to call `.execute()` on that request.

Changes pass manually triggered full integration tests run: https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/4622277782

#### Connected

To resolve flakiness in the connected test, wait until the notebook is stopping/stopped before continuing with the update.

- Before, the stop operation was still queued when we tried to update, as the notebook was still provisioning.

### DIFF
--- a/integration/src/main/java/scripts/utils/ClientTestUtils.java
+++ b/integration/src/main/java/scripts/utils/ClientTestUtils.java
@@ -48,6 +48,7 @@ import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


integration/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -338,6 +341,76 @@ public static <T> T getWithRetryOnException(
     return result;
   }
 
+  /**
+   * Get a result from a call that might throw an exception. If the supplier finishes, the result is
```
zloery, line 345: update javadoc - this method will only return if the supplier finishes and the supplied value satisfies the predicate

integration/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -338,6 +341,76 @@ public static <T> T getWithRetryOnException(
...
+    Duration sleepDuration = initialSleepDuration;
+
+    while (true) {
+      result = supplier.get();
+      if (predicate.test(result)) {
+        break;
+      } else {
+        // If we are out of time
+        if (Instant.now().isAfter(endTime)) {
+          throw new Exception();
```
zloery, line 381: this should be a more specific exception, and contain a message indicating what happened (e.g. "timed out during getWithRetry")

PR #1188, opened 2023-04-05
Support limiting which blobs to clone (WOR-544).

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-544

Allows passing an array of prefixes that limit which blobs to clone. If the array is not present or is of zero length, all blobs will be cloned.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
@@ -500,7 +500,7 @@ public class ControlledAzureResourceApiController extends ControlledResourceCont
             body.getDescription(),
             body.getName(),
             body.getCloningInstructions(),
-            body.getPrefixesToClone().toArray(new String[0]));
+            body.getPrefixesToClone());
 
     final ApiCloneControlledAzureStorageContainerResult result =
         fetchCloneAzureContainerResult(jobId);
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
@@ -499,7 +499,8 @@ public class ControlledAzureResourceApiController extends ControlledResourceCont
             body.getName(),
             body.getDescription(),
             body.getName(),
-            body.getCloningInstructions());
+            body.getCloningInstructions(),
+            body.getPrefixesToClone().toArray(new String[0]));
 
     final ApiCloneControlledAzureStorageContainerResult result =
         fetchCloneAzureContainerResult(jobId);
--- a/service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopierConnectedTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopierConnectedTest.java
@@ -105,31 +105,7 @@ public class BlobCopierConnectedTest extends BaseAzureConnectedTest {
   @ParameterizedTest
   @MethodSource("getPrefixesToCopyAllFiles")
   void copyAllBlobs(@Nullable String[] prefixesToCopy, String[] blobNames) {
-    // upload blob to source container
-    var sourceContainerClient =
-        azureStorageAccessService.buildBlobContainerClient(sourceContainer, storageAcct);
-
-    BlobCopier bc = new BlobCopier(azureStorageAccessService, userRequest);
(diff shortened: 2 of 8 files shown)


service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopierConnectedTest.java
```diff
@@ -32,11 +31,11 @@
...
-import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.TestInstance;
+import javax.annotation.Nullable;
+import org.junit.jupiter.api.*;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.springframework.beans.factory.annotation.Autowired;
 
 @Tag("azureConnectedPlus")
```
cahrens, line 41: This test doesn't run on PR commits, but I ran it locally and [here](https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/4628382886) (failure was a sporadic unrelated test failure… rerunning just to be sure).

aherbst-broad (commented):
Had a question about the prefixes serialization.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopierConnectedTest.java
blakery: I know it can be a lot of obnoxious setup at first, with how some of this code is structured.... but might it be a good time to break some of this out into a proper unit test?

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -499,7 +499,8 @@ public ResponseEntity<ApiCloneControlledAzureStorageContainerResult> cloneAzureS
             body.getName(),
             body.getDescription(),
             body.getName(),
-            body.getCloningInstructions());
+            body.getCloningInstructions(),
+            body.getPrefixesToClone().toArray(new String[0]));
```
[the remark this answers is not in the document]
cahrens: TBH, I just thought of it as an array of strings instead of a list (and the matching implementation using a List was just the easiest way to write it). But when I tried to convert it to a List of strings throughout just now I did hit a serialization issue (even using TypeReference which is supposed to be the way around this). I spent about 45 minutes trying to figure it out and looking at other examples, but no luck. Does it really bother you if it's an array instead of a List?

```
Failed to deserialize value '["analyses/"]' from JSON to type java.util.List<java.lang.String>
bio.terra.stairway.exception.JsonConversionException: Failed to deserialize value '["analyses/"]' from JSON to type java.util.List<java.lang.String>
	at app//bio.terra.stairway.FlightMap.get(FlightMap.java:117)

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopierConnectedTest.java
cahrens: I don't think so because this is the only coverage that we have of actually using the Azure APIs to copy blobs. I found it really useful to be able to see what Azure would return when we had "directory" structures, and if it would prefix blob names with "/", etc.

aherbst-broad: Yes, this test is intended to be a real (limited) exercise of our ability to copy blobs.

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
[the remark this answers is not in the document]
aherbst-broad: Nope, makes sense, thanks for the clarification.

aherbst-broad (approved):

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopierConnectedTest.java
blakery: I wasn't necessarily saying that what's here has to go away (particularly since it doesn't run on PRs).

That said, I realize there isn't a ton of logic in the BlobCopier class in the first place, so the answer might be that it's not worth it atm.

blakery (approved):

PR #1192, opened 2023-04-06
Get the resource from the correct map (WOR-544).

cahrens:
Follow-up PR from https://github.com/DataBiosphere/terra-workspace-manager/pull/1188.

I confess that my manual testing of this (since we have no full integration test) is pretty sketchy-- I ended up commenting out a ton of code along the path to where I made this fix so that I could verify that the expected value is in the inputParameters map (without properly setting up a billing profile attached to an actual landing zone and creating a storage container in it). I feel like the risk here is low though because in the absence of the prefixes cloning continues to copy everything over.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
@@ -240,7 +240,7 @@ public class ControlledResourceService {
                 Optional.ofNullable(cloningInstructionsOverride)
                     .map(CloningInstructions::fromApiModel)
                     .orElse(sourceContainer.getCloningInstructions()))
-            .addParameter(ResourceKeys.PREFIXES_TO_CLONE, prefixesToClone);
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -240,7 +240,7 @@ public String cloneAzureContainer(
                 Optional.ofNullable(cloningInstructionsOverride)
                     .map(CloningInstructions::fromApiModel)
                     .orElse(sourceContainer.getCloningInstructions()))
-            .addParameter(ResourceKeys.PREFIXES_TO_CLONE, prefixesToClone);
+            .addParameter(ControlledResourceKeys.PREFIXES_TO_CLONE, prefixesToClone);
```
cahrens, line 243: I think I should have defined this in `ControlledResourceKeys`.

aherbst-broad (approved):

blakery (approved):

PR #1194, opened 2023-04-07
[PF-2668] test fixes

ddietterich:
PR fixing the two tests I broke.

Integration test: ImportDataCollection depended on a specific error message. I changed the error. This is a compensating change to the test.

ConnectedPlus test: ControlledGcpResourceApiControllerBqDatasetTest needed to use the proper default lifetimes, since that TODO got implemented.


mikenev (approved):
lgtm.
I'm not too happy with checking the message contents against a hard-coded string (I added this one originally), but it is a pattern we do a lot.

PR #1195, opened 2023-04-07
[PF-2662] Break up ControlledResourceServiceTest

ddietterich:
The ControlledResourceServiceTest takes 20 minutes. This PR breaks it up by controlled resource type. I'm hoping with parallelism, it will shorted the PR validation check time.

I also moved all of the API tests into the connectedPlus category


mikenev (approved):
lgtm.
breaking up the monotest

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceNotebookTest.java
```diff
@@ -0,0 +1,768 @@
...
+            petCredentials);
+    jobService.waitForJob(petJobId);
+    // Confirm this flight status is ERROR, not FATAL.
+    assertEquals(
+        FlightStatus.ERROR, stairwayComponent.get().getFlightState(petJobId).getFlightStatus());
+  }
+
+  @Test
+  @DisabledIfEnvironmentVariable(named = "TEST_ENV", matches = BUFFER_SERVICE_DISABLED_ENVS_REG_EX)
+  void updateAiNotebookResourceDo() throws InterruptedException, IOException {
```
zloery, line 354: out of scope for this PR, but I wonder if we could shrink test time more by combining some tests to use the same instance - e.g. create the notebook and test do steps, test do/undo updates on that instance, etc.

zloery (approved):
Thanks for this!

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceNotebookTest.java
ddietterich: Yes. In general, we should do a pass over our tests to find those cases and ones where multiple tests are exercising the same code paths.

PR #1196, opened 2023-04-10
[PF-2581] Handle a user removing themselves from a workspace better

zloery:
This changes WSM behavior in 3 ways:

- Later steps in `RemoveUserFromFlight` now use the WSM SA credentials when talking to Sam instead of the requesting user's credentials. This stops the flight from throwing dismal failures when a user removes themself as an owner.
- Additionally, WSM will no longer allow a user to remove themselves as the sole owner of a workspace to prevent abandoning workspaces. They can remove themselves after adding another owner, and can always delete the workspace. It's still possible to abandon a workspace via groups (e.g. if the only owner is an empty/deleted group), but this should prevent accidental abandonment.
- This also changes existing behavior when calling the "removeRole" endpoint for a user who is not in the workspace. Previously, this would return a 204 while silently doing nothing. Now, it returns a 400 with an error message. The old behavior matched Sam, but given that we've already validated that the user has permission to list workspace users, it does not leak information to give this response. Users without OWNER permission in the workspace will continue to receive a 403 as before.


service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -625,6 +630,14 @@ public void removeWorkspaceRoleFromUser(
     if (!roleMembers.contains(targetUserEmail)) {
       return;
     }
+    // Additionally, validate that the user is not removing themselves as the sole owner. WSM does
+    // not allow users to abandon resources this way.
+    if (role.equals(WsmIamRole.OWNER)
+        && roleMembers.size() == 1
+        && roleMembers.get(0).equals(targetUserEmail)) {
+      throw new BadRequestException(
```
ddietterich, line 638: What do you think about making a specific exception for this?

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -625,6 +630,14 @@ public void removeWorkspaceRoleFromUser(
     if (!roleMembers.contains(targetUserEmail)) {
       return;
     }
+    // Additionally, validate that the user is not removing themselves as the sole owner. WSM does
+    // not allow users to abandon resources this way.
+    if (role.equals(WsmIamRole.OWNER)
+        && roleMembers.size() == 1
+        && roleMembers.get(0).equals(targetUserEmail)) {
```
ddietterich, line 637: This clause isn't strictly needed, due to the contains above.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -625,6 +630,14 @@ public void removeWorkspaceRoleFromUser(
     if (!roleMembers.contains(targetUserEmail)) {
       return;
     }
+    // Additionally, validate that the user is not removing themselves as the sole owner. WSM does
+    // not allow users to abandon resources this way.
+    if (role.equals(WsmIamRole.OWNER)
```
ddietterich, line 635: Putting this check into the `RemoveUserFromSamStep` would shrink the window where the result of this check becomes obsolete due to concurrent action.

ddietterich (commented):
I think I would prefer to have the "last owner" check in the flight to limit the window between the check and the Sam action.

yuhuyoyo (approved):

PR #1014, opened 2023-01-18
[TERRA-314] Merge AWS OpenApi spec in WSM

dexamundsen:
- Merge AWS OpenApi spec in WSM
- Add stubs (with exceptions thrown) for AWS related endpoints

Note on PR check failure - 

- Integration test: PrivateControlledGcsBucketLifecycle failed due to Google bucket access failure (https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/3953978458/)
- AzureConnected test: createAndDeleteAzureVmControlledResource failed due to landing zone with billing profile is not found (https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/3953978461)

PR #1198, opened 2023-04-12
[TERRA-304] Add AWS S3 Storage support

dexamundsen:
Add support to - create, get and delete a S3 storage folder

Resource named awsStorageFolder instead of awsS3StorageFolder based on https://github.com/DataBiosphere/terra-workspace-manager/pull/1014#discussion_r1082698611 and inline with AzureStorageContainer

Tested locally
```
# Create a new s3 storage bucket
http://localhost:8080/swagger-ui.html#/ControlledAwsResource/createAwsS3StorageFolder
{
  "common": {
    "name": "dex-s3-414-1",
    "description": "dex-s3-414-1-description",
    "cloningInstructions": "COPY_NOTHING",
    "accessScope": "SHARED_ACCESS",
    "managedBy": "USER",

    "resourceId": "e7d5ff64-6364-47ad-a6ae-2335019534ff"
  },
  "AwsS3StorageFolder": {
    "region": "us-east-1"
  }
}

{
  "resourceId": "e7d5ff64-6364-47ad-a6ae-2335019534ff",
  "AwsS3StorageFolder": {
    "metadata": {
      "workspaceId": "4c30fe7a-52fb-4528-9e06-86600ccc2a5c",
      "resourceId": "e7d5ff64-6364-47ad-a6ae-2335019534ff",
      "name": "dex-s3-414-1",
      "description": "dex-s3-414-1-description",
      "resourceType": "AWS_S3_STORAGE_FOLDER",
      "stewardshipType": "CONTROLLED",
      "cloudPlatform": "AWS",
      "cloningInstructions": "COPY_NOTHING",
      "controlledResourceMetadata": {
        "accessScope": "SHARED_ACCESS",
        "managedBy": "USER",
        "privateResourceUser": {},
        "privateResourceState": "NOT_APPLICABLE",
        "region": "us-east-1"
      },
      "resourceLineage": [],
      "properties": [],
      "createdBy": "dexamundsen@google.com",
      "createdDate": "2023-04-14T18:28:28.137172Z",
      "lastUpdatedBy": "dexamundsen@google.com",
      "lastUpdatedDate": "2023-04-14T18:28:28.137172Z",
      "state": "READY"
    },
    "attributes": {
      "s3BucketName": "v0-saas-devel-us-east-1-terra",
      "prefix": "dex-s3-414-1"
    }
  }
}
.... verified on AWS web console
```

```
# get an existing storage folder
http://localhost:8080/swagger-ui.html#/ControlledAwsResource/getAwsS3StorageFolder
.... verified
```

```
# list resources in the workspace
http://localhost:8080/swagger-ui.html#/Resource/enumerateResources
.... verified
```

```
# delete an existing storage folder
http://localhost:8080/swagger-ui.html#/ControlledAwsResource/deleteAwsS3StorageFolder
.... verified
.... verified on AWS web console, get resource and list resources
```

Tested that DB entries are deleted on resource deletion

Tests added in https://verily.atlassian.net/browse/TERRA-306
Credential support added in https://verily.atlassian.net/browse/TERRA-318


service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
```diff
@@ -140,19 +162,35 @@ public AwsCloudContext getRequiredAwsCloudContext(UUID workspaceUuid) {
...
+   * @param region AWS region
+   * @return AWS landing zone, if supported for the Cloud context region
+   * @throws StaleConfigurationException StaleConfigurationException
+   */
+  public Optional<LandingZone> getLandingZone(
+      @NotNull AwsCloudContext awsCloudContext, @NotNull Region region) {
+    Environment environment = discoverEnvironment();
+
+    AwsCloudContext awsCloudContextFromEnv = getCloudContext(environment);
+    if (!awsCloudContext.equals(awsCloudContextFromEnv)) {
```
jmczerk, line 188: I think that this is a good thing to check, but object equality feels a little coarse.

Maybe we should just check against some of the more "invariant" fields, specifically the major version number and the AWS account number feel like things that would be a problem if they changed, whereas the "alias" fields and the org are not really meaningful to the rest of the WSM now.

service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
```diff
@@ -0,0 +1,185 @@
...
+    return new ResponseEntity<>(HttpStatus.NO_CONTENT);
+  }
+
+  @Override
+  public ResponseEntity<ApiAwsCredential> getAwsStorageFolderCredential(
+      UUID workspaceUuid,
+      UUID resourceId,
+      ApiAwsCredentialAccessScope accessScope,
+      Integer duration) {
+    throw new NotImplementedException("not implemented");
```
jmczerk, line 183: This is really important functionality, given the user can't access the folder without it.  I don't think we want to defer this.

service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
```diff
@@ -0,0 +1,185 @@
...
+        userRequest,
+        workspaceUuid,
+        resourceId,
+        SamConstants.SamControlledResourceActions.DELETE_ACTION);
+    logger.info(
+        "deleteAwsStorageFolder workspace: {}, resourceId: {}",
+        workspaceUuid.toString(),
+        resourceId.toString());
+
+    controlledResourceService.deleteControlledResourceSync(workspaceUuid, resourceId, userRequest);
```
jmczerk, line 173: I expected folder deletion to be async, as object deletion could be a long running operation.

service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java
```diff
@@ -205,8 +204,9 @@ public static void validateRegionAgainstPolicy(
       case GCP -> validateRegion(
           tpsApiDispatch, workspaceUuid, GcpUtils.parseRegion(region), platform);
       case AZURE -> validateRegion(tpsApiDispatch, workspaceUuid, region, platform);
-      case AWS -> throw new FeatureNotSupportedException(
-          "validateRegionAgainstPolicy.AWS not supported");
+      case AWS -> {
+        // TODO(TERRA-496) check if user has access to landing zone / region
```
jmczerk, line 208: FYI, AWS regions have been added to TPS, no specific action here.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/CreateAwsStorageFolderStep.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    AwsCredentialsProvider credentialsProvider =
+        AwsUtils.createWsmCredentialProvider(
+            awsCloudContextService.getRequiredAuthentication(),
+            awsCloudContextService.discoverEnvironment());
+
+    if (AwsUtils.checkFolderExists(
```
jmczerk, line 44: Doesn't this need to be in a separate step if it's going to be fatal?  If the subsequent folder creation succeeds, but the flight does not finish, this will fail the flight on retry.

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
```diff
@@ -38,33 +40,46 @@ public WorkspaceDeleteFlight(FlightMap inputParameters, Object applicationContex
...
+    // which will also delete all resources
     addStep(
         new DeleteControlledAzureResourcesStep(
             appContext.getResourceDao(),
             appContext.getControlledResourceService(),
             appContext.getSamService(),
             workspaceUuid,
             userRequest));
+    addStep(
+        new DeleteControlledAwsResourcesStep(
```
jmczerk, line 54: Is Cloud Context deletion async?  These deletions could take a long time (thus previous comment that Storage Folder delete must be async).  I suspect this has already been encountered in Azure...

service/src/main/java/bio/terra/workspace/service/workspace/flight/aws/DeleteControlledAwsResourcesStep.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+
+    List<ControlledResource> controlledResourceList =
+        resourceDao.listControlledResources(workspaceUuid, CloudPlatform.AWS);
+    // TODO(TERRA-279): check permissions to delete
+
+    // TODO(TERRA-320) delete notebooks first, since they may be using underlying S3 folders
```
jmczerk, line 49: There's no dependency of notebooks on folders, so I wouldn't worry too much about this.  The "default folder" thing we did in the prototype notebook was demo-enabling and should probably not be a WSM construct.

jmczerk (commented):
A few inline comments and a couple general concerns:
* The guidance from UX was to use the term "S3 Storage Folder", and I thought they made a pretty good case for this.  Not sure it matters too much at this level, but it will matter in the CLI.
* Seems like a lot of code here that will be replaced with CRL code, do we really want to create this debt?
* Also on debt, what is the plan for automated testing?  If not here, I'd like to see testing addressed before we move onto any other AWS resource types.

service/src/main/java/bio/terra/workspace/service/workspace/flight/aws/DeleteControlledAwsResourcesStep.java
dexamundsen: Noted, removing comment

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/CreateAwsStorageFolderStep.java
dexamundsen: Makes sense, changed

service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
dexamundsen: Agree, I plan to work on it today in a subsequent PR.

dexamundsen:
@jmczerk 

> * The guidance from UX was to use the term "S3 Storage Folder", and I thought they made a pretty good case for this.  Not sure it matters too much at this level, but it will matter in the CLI. --> The (only) reason I used storageFolder was to keep it consistent with Azure resources. However it adds more value to keep it consistent with UI, rather than Azure, Happy to change it . 

> * Seems like a lot of code here that will be replaced with CRL code, do we really want to create this debt? --> I spoke to Zach on this last week, to update CRL code to handle errors. Hence the tech-debt. I will work with Zach to resolve this this week and go ahead with functions in this PR, if it is not a blocker. 

> * Also on debt, what is the plan for automated testing?  If not here, I'd like to see testing addressed before we move onto any other AWS resource types. -> this is a targeted for second part of the week (testing in WSM via TERRA-306). In the meantime as a backstop, connected tests are part of CLI PR for workspace / s3 operations

jmczerk:
> @jmczerk
> 
> > * The guidance from UX was to use the term "S3 Storage Folder", and I thought they made a pretty good case for this.  Not sure it matters too much at this level, but it will matter in the CLI. --> The (only) reason I used storageFolder was to keep it consistent with Azure resources. However it adds more value to keep it consistent with UI, rather than Azure, Happy to change it .
> 
> > * Seems like a lot of code here that will be replaced with CRL code, do we really want to create this debt? --> I spoke to Zach on this last week, to update CRL code to handle errors. Hence the tech-debt. I will work with Zach to resolve this this week and go ahead with functions in this PR, if it is not a blocker.
> 
> > * Also on debt, what is the plan for automated testing?  If not here, I'd like to see testing addressed before we move onto any other AWS resource types. -> this is a targeted for second part of the week (testing in WSM via TERRA-306). In the meantime as a backstop, connected tests are part of CLI PR for workspace / s3 operations

Thanks.  Regarding the first item, lets see what others say before changing anything.

Regarding 2nd and 3rd, I'm OK with treating them as non-blockers, but would like to see resolved ahead of notebook work.

service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
```diff
@@ -0,0 +1,185 @@
...
+
+    AwsCloudContext awsCloudContext =
+        awsCloudContextService.getRequiredAwsCloudContext(workspaceUuid);
+    ApiAwsStorageFolderCreationParameters creationParameters = body.getAwsStorageFolder();
+
+    Region requestedRegion;
+    LandingZone landingZone;
+    try {
+      String prefixName = commonFields.getName();
+      if (prefixName.isEmpty() || prefixName.length() > 1024) {
```
yuhuyoyo, line 98: move this to validationutils

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
```diff
@@ -480,6 +483,32 @@ public ControlledResource getControlledResource(UUID workspaceUuid, UUID resourc
...
+   * Synchronously delete all controlled resources instances of the specified type returning all
+   * remaining resources
+   *
+   * @param workspaceUuid workspace UUID
+   * @param allResources list of all controlled resources in the workspace
+   * @param type resource type to delete
+   * @param userRequest the user request
+   * @return list of controlled resources, after deleting those of specified type
+   */
+  public @NotNull List<ControlledResource> deleteControlledResourceSyncOfType(
```
yuhuyoyo, line 496: nit: rm @notnull. generally we mark nullable and not the other way around.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/ControlledAwsStorageFolderResource.java
```diff
@@ -0,0 +1,213 @@
...
+
+  @Override
+  public void validate() {
+    super.validate();
+    if (getResourceType() != WsmResourceType.CONTROLLED_AWS_STORAGE_FOLDER
+        || getResourceFamily() != WsmResourceFamily.AWS_STORAGE_FOLDER
+        || getStewardshipType() != StewardshipType.CONTROLLED) {
+      throw new InconsistentFieldsException("Expected CONTROLLED_AWS_STORAGE_FOLDER");
+    }
+    if ((getBucketName() == null) || (getPrefix() == null) || (getRegion() == null)) {
```
yuhuyoyo, line 183: move the folder name validation from controller to here

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/ControlledAwsStorageFolderResource.java
```diff
@@ -0,0 +1,213 @@
...
+      this.common = common;
+      return this;
+    }
+
+    public Builder bucketName(String bucketName) {
+      this.bucketName = bucketName;
+      return this;
+    }
+
+    public Builder prefix(String prefix) {
```
yuhuyoyo, line 204: what about just call this folderPath or objectPath

prefix is a little confusing as i don't think the s3 official document uses prefix to refer to s3 object either.

service/src/test/java/bio/terra/workspace/common/utils/AwsUtilsTest.java
```diff
@@ -52,7 +52,7 @@ void hello_bucket() throws IOException {
     // This lifetime should track that of the discovered environment where possible (stale
     // credentials will get refreshed under the hood).
     AwsCredentialsProvider awsCredentialsProvider =
-        AwsUtils.createWsmCredentialProvider(awsConfiguration, environment);
```
yuhuyoyo, line 55: please add CRUD tests
- controller test, examples https://github.com/DataBiosphere/terra-workspace-manager/tree/main/service/src/test/java/bio/terra/workspace/app/controller

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/DeleteAwsStorageFolderStep.java
```diff
@@ -0,0 +1,53 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    AwsCredentialsProvider credentialsProvider =
+        AwsUtils.createWsmCredentialProvider(
+            awsCloudContextService.getRequiredAuthentication(),
+            awsCloudContextService.discoverEnvironment());
+
+    // TODO(TERRA-279): check permissions to delete
```
yuhuyoyo, line 33: We already checked this permission in the controller

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/ControlledAwsStorageFolderResource.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public void addCreateSteps(
+      CreateControlledResourceFlight flight,
+      String petSaEmail,
+      AuthenticatedUserRequest userRequest,
+      FlightBeanBag flightBeanBag) {
+    RetryRule cloudRetry = RetryRules.cloud();
```
yuhuyoyo, line 128: nit: inline

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/ControlledAwsStorageFolderResource.java
yuhuyoyo: also nit: rm the unnecessary ()

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -173,6 +179,11 @@ public String getUserEmailFromSam(AuthenticatedUserRequest userRequest)
     return getUserStatusInfo(userRequest).getUserEmail();
   }
 
+  /** Fetch the Sam user info associated with the request. */
+  public SamUser getSamUser(HttpServletRequest request) {
```
yuhuyoyo, line 183: is this used?

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
```diff
@@ -140,19 +162,35 @@ public AwsCloudContext getRequiredAwsCloudContext(UUID workspaceUuid) {
...
+  /**
+   * Return the landing zone to for the given cloud context's region
+   *
+   * @param awsCloudContext AWS cloud context
+   * @param region AWS region
+   * @return AWS landing zone, if supported for the Cloud context region
+   * @throws StaleConfigurationException StaleConfigurationException
+   */
+  public Optional<LandingZone> getLandingZone(
+      @NotNull AwsCloudContext awsCloudContext, @NotNull Region region) {
```
yuhuyoyo, line 184: nit: rm notNull

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
```diff
@@ -121,8 +134,17 @@ public AwsCloudContext getRequiredAwsCloudContext(UUID workspaceUuid) {
    *
    * @return AWS cloud context
    */
-  public @NotNull AwsCloudContext getCloudContextFromConfiguration() {
-    Environment environment = discoverEnvironment();
+  public @NotNull AwsCloudContext getCloudContext() {
```
yuhuyoyo, line 137: nit: rm notnull

service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
dexamundsen: Done, changed

openapi/src/common/schemas_aws.yaml
```diff
@@ -0,0 +1,28 @@
+# Common AWS schemas
+# These schemas should be referenced from the source like this:
+#  $ref: '#/components/schemas/<item-you-want>
+components:
+  schemas:
+    AwsCredentialDuration:
```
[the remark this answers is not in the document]
dexamundsen: Done, changed to AwsCredentialDurationSeconds

openapi/src/parts/controlled_aws_storage_folder.yaml
```diff
@@ -0,0 +1,130 @@
+paths:
+  /api/workspaces/v1/{workspaceId}/resources/controlled/aws/storageFolder:
```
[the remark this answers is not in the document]
dexamundsen: Re: https://github.com/DataBiosphere/terra-workspace-manager/pull/1198#issuecomment-1511871242

I would be open to renaming this. Considerations -
1) Recommendation from Dan in a previous PR (+ potential other considerations) to avoid using copyrighted or marketing names like 'S3' or 'sagemaker'
2) kept in line with Azure Storage container (more recent than GCP)

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/DeleteAwsStorageFolderStep.java
dexamundsen: Retaining this since AWS account / landing zone handling 'may' be different

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
dexamundsen: Reason for this - 
specifying NotNull has 2 uses here - 1) compiler can enforce that null is never returned. 2) caller can be assured that this will never be null and hence avoid checks

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/ControlledAwsStorageFolderResource.java
dexamundsen: Folder is only a visual concept in AWS. It uses prefixes
re: https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-prefixes.html
Furthermore internal operations with AWS SDK work based prefixes, hence prefer to retain prefix

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/ControlledAwsStorageFolderResource.java
dexamundsen: NA, used in 2 steps (new changes)

service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
```diff
@@ -0,0 +1,185 @@
...
+    this.workspaceService = workspaceService;
+    this.awsCloudContextService = awsCloudContextService;
+  }
+
+  @Override
+  public ResponseEntity<ApiCreatedControlledAwsStorageFolder> createAwsStorageFolder(
+      UUID workspaceUuid, @Valid ApiCreateControlledAwsStorageFolderRequestBody body) {
+    features.awsEnabledCheck();
+
+    final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
```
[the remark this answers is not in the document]
dexamundsen: done, hand copied some code from other controllers

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
yuhuyoyo: https://engdoc.corp.google.com/eng/doc/devguide/java/practices/null.md#legacy

service/src/test/java/bio/terra/workspace/common/utils/AwsUtilsTest.java
dexamundsen: handled in https://verily.atlassian.net/browse/TERRA-306

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceService.java
dexamundsen: Done, all places

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/ControlledAwsStorageFolderResource.java
dexamundsen: Retaining () to aid readability

zloery (commented):
Generally LGTM, some comments

For the earlier thread:
- For resource type names I'd lean towards consistency with the UI + CLI over other cloud's resource types
- Ack on the error-handling CRL work, hoping to get to that tomorrow and pull it into WSM soon after
- +1 to at least unit + connected testing before going too much further, we should have some established patterns for other resource types

service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
dexamundsen: done

openapi/src/common/schemas_aws.yaml
```diff
@@ -0,0 +1,28 @@
...
+      type: integer
+      minimum: 900
+      maximum: 3600
+      default: 900
```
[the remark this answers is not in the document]
dexamundsen: makes sense, removed

openapi/src/common/schemas_aws.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+    AwsCredentialAccessScope:
+      type: string
+      enum: [ 'READ_ONLY', 'WRITE_READ' ]
+
+    AwsCredential:
+      description: AWS CLI/SDK-compatible credential to use to access a bucket
```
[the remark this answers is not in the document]
dexamundsen: done

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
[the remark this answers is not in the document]
dexamundsen: Removed in new changeset, it is needed for credential though

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
dexamundsen: Looks like workspace and cloud context deletion are sync
WS deletion flow - 

- https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java#L429
- https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java#L370

delete job submit-and-wait till it completes. 

and within the deleteWorkspaceFlight 
- https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java#L44
Azure deletion functions internally call the deleteSync functions 

1) Just deleting the cloud context deletes the DB entries (not the resources). Is there any use case for this? 
2) ideally the entire workspace deletion must be async. I can create a new task for this, it touches GCP & Azure and will be a larger changeset.

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
dexamundsen: Checking the tenant and env definitely are an overkill atm. 
since there was already a equals method, leveraged it. Plus, all the fields ought to be the same anyway, so decided to not write code to check individual fields

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
jmczerk: We should definitely not take this on in this PR.  Let's keep workspace deletion as you've written it, and medium term think about making this async (a ticket might already exist for this in Broad Jira).  If AWS workspace deletion gives us timeout trouble in the meantime we can maybe do some soft enforcement in Axon UI/CLI.

dexamundsen: created https://verily.atlassian.net/browse/TERRA-502

dexamundsen:
Tested async resource deletion locally, success

service/src/main/java/bio/terra/workspace/service/workspace/flight/aws/DeleteControlledAwsResourcesStep.java
```diff
@@ -0,0 +1,65 @@
...
+    this.userRequest = userRequest;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+
+    List<ControlledResource> controlledResourceList =
+        resourceDao.listControlledResources(workspaceUuid, CloudPlatform.AWS);
+    // TODO(TERRA-279): check permissions to delete
```
ddietterich, line 46: I would expect a permission check on the delete workspace and the delete cloud context. I would not expect it to be done in this flight. By the time you are here, we should have checked permissions.

service/src/main/java/bio/terra/workspace/service/workspace/flight/aws/DeleteControlledAwsResourcesStep.java
```diff
@@ -0,0 +1,65 @@
...
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    // Resource deletion can't be undone, so this just surfaces the error from the DO direction
+    // instead.
+    logger.error(
+        "Unable to undo deletion of controlled AWS resources for workspace {}", workspaceUuid);
+    return flightContext.getResult();
```
ddietterich, line 63: The better way to do this is:
```
new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, new InternalLogicException("Unable to..."));
```
With the Stairway fix in (may not be updated in WSM. I'll do that), the flight properly reports the initial error and not the dismal failure.

ddietterich (changes requested):
Out of 45 files in the PR, I count 17 that are not related to adding storage support. About half of those are adding a `final` that is contrary to the coding norms in WSM.

I do not want to continue with the practice of leaving tests to a separate PR. I would much rather have you implement, say, create and have create tests. Then delete and add in delete tests, etc. Given the scale of this PR, I won't request for that now. But for future planning, please plan to include tests with the feature.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/storageFolder/ControlledAwsStorageFolderHandler.java
```diff
@@ -0,0 +1,36 @@
...
+  public WsmResource makeResourceFromDb(DbResource dbResource) {
+    ControlledAwsStorageFolderAttributes attributes =
+        DbSerDes.fromJson(dbResource.getAttributes(), ControlledAwsStorageFolderAttributes.class);
+
+    return new ControlledAwsStorageFolderResource(
+        dbResource, attributes.getBucketName(), attributes.getPrefix());
+  }
+
+  @Override
+  public String generateCloudName(@Nullable UUID workspaceUuid, String resourceName) {
```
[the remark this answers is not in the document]
dexamundsen: done

service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
```diff
@@ -0,0 +1,239 @@
...
+    try {
+      AwsResourceValidationUtils.validateAwsStorageFolderName(commonFields.getName());
+
+      requestedRegion = Region.of(creationParameters.getRegion());
+      landingZone =
+          awsCloudContextService
+              .getLandingZone(awsCloudContext, requestedRegion)
+              .orElseThrow(
+                  () -> {
+                    throw new BadRequestException(
```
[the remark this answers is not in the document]
dexamundsen: looked up more on the AWS documentation & implementation. It creates a new region with any string, even if it doesnt exist in the Regions enum. Updated validation based on this (never throws IllegalArgumentException)

dexamundsen:
@jmczerk @ddietterich @zloery  changes since your last reviews - 
- All pending comments addressed
- Renamed AwsStorageFolder to AwsS3StorageFolder: resource and internal implementation only, endpoints do not expose 's3'
- Updated and re tested - create, delete, getcredentials

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3storageFolder/ValidateAwsS3StorageFolderCreateStep.java
```diff
@@ -0,0 +1,52 @@
...
+
+    if (AwsUtils.checkFolderExists(
+        credentialsProvider,
+        Region.of(resource.getRegion()),
+        resource.getBucketName(),
+        resource.getPrefix())) {
+      return new StepResult(
+          StepStatus.STEP_RESULT_FAILURE_FATAL,
+          new ConflictException(
+              String.format(
```
jmczerk, line 41: FOLLOW-ON, NOT THIS PR!

We should have a test that makes sure that folder `foo` can be created if `foo_bar` exists to make sure we're doing the right things in regards to delimiters.  Worst case this is overly restrictive now, we can revisit if the test shows this to be the case.  **_No action for this PR._**

jmczerk (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3storageFolder/ControlledAwsS3StorageFolderResource.java
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  @JsonIgnore
+  public Optional<UniquenessCheckAttributes> getUniquenessCheckAttributes() {
+    return Optional.of(
+        new UniquenessCheckAttributes()
+            .uniquenessScope(UniquenessScope.GLOBAL)
+            .addParameter("prefix", prefix));
```
zloery, line 118: I think this also needs a parameter for region if I understand the AWS infra properly. Since buckets are regional, it's valid in AWS to have both `my-us-bucket/foo/bar.txt` and `my-eu-bucket/foo/bar.txt` even though they have the same prefix

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3storageFolder/ControlledAwsS3StorageFolderResource.java
```diff
@@ -0,0 +1,218 @@
...
+    super.validate();
+    if (getResourceType() != WsmResourceType.CONTROLLED_AWS_S3_STORAGE_FOLDER
+        || getResourceFamily() != WsmResourceFamily.AWS_S3_STORAGE_FOLDER
+        || getStewardshipType() != StewardshipType.CONTROLLED) {
+      throw new InconsistentFieldsException("Expected CONTROLLED_AWS_S3_STORAGE_FOLDER");
+    }
+    if ((getBucketName() == null) || (getPrefix() == null) || (getRegion() == null)) {
+      throw new MissingRequiredFieldException(
+          "Missing required field for ControlledAwsS3StorageFolderResource.");
+    }
```
zloery, line 191: I think we need to validate that the prefix is a top-level folder in the bucket (i.e. does not contain `/`). 
I think you could get into a weird state if you had nested controlled resources (i.e. one at `bucket/folder/` and one at `bucket/folder/subfolder`) where deleting one resource would affect other ones.

dexamundsen: Agree. already added as item in https://verily.atlassian.net/browse/TERRA-470

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3storageFolder/ControlledAwsS3StorageFolderResource.java
jmczerk: It would be valid to have different folders with the same name in different buckets, but this would get more complicated if/when we have multiple Environments in a WSM instance that might have region overlaps.

I think the safe way to do this would be to include the bucket name rather than the region, but I still think that maybe we should side-step the issue via the global uniqueness check, and accept that we lose the ability to create `foo` in more than on workspace.  WDYT?  

_This assumes that I'm reading line 117 correctly (we can only have one folder named 'foo' in any Workspace managed by this WSM instance)._

(one thing that informs my desire to be conservative is the fact that, unlike GCS, [S3 has no "create if do not exist" option](https://repost.aws/questions/QUPOqd552USXuyHJVKC9yAHg/questions/QUPOqd552USXuyHJVKC9yAHg/can-we-do-write-object-if-not-exist-in-s3-similar-to-conditional-put-in-dynamo-db?) for atomic object creation so we are dependent on the database to avoid collisions)

jmczerk: Unresolved, as there's more to discuss here.

dexamundsen: created https://verily.atlassian.net/browse/TERRA-519 so that it is not missed in other tickets. 
had marked it resolved only since it was tracked elsewhere and doesnt block this PR.

dexamundsen: this might be what we need 

prefix + account for the check. along with disallowing / in prefix
reason - the scope 'workspace' isnt going to work for storage since all folders across a wsm instance 

Bucket names are expected to be globally unique and not needed (thanks John)

zloery: Leaving this as a global uniqueness check on prefix sgtm if we're not quite sure how we want to handle multiple environments, it's easier to start with the more restrictive validation and loosen than to go the other way

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3storageFolder/ValidateAwsS3StorageFolderCreateStep.java
```diff
@@ -0,0 +1,52 @@
...
+                  "Prefix '%s/' already exists in bucket '%s'.",
```
ddietterich, line 42: Do we expect our users know what we are talking about when we use the terms prefix and bucket here?

ddietterich (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3storageFolder/ValidateAwsS3StorageFolderCreateStep.java
dexamundsen: Noted, changing the text to folder in all error messages which is expected UX nomenclature.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3storageFolder/ControlledAwsS3StorageFolderResource.java
dexamundsen: handled in https://github.com/DataBiosphere/terra-workspace-manager/pull/1214

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3storageFolder/ValidateAwsS3StorageFolderCreateStep.java
dexamundsen: handled in https://github.com/DataBiosphere/terra-workspace-manager/pull/1214

PR #1201, opened 2023-04-12
[PF-2581] Make RemoveUserFromSamStep idempotent

zloery:
In #1196 I moved some validation inside `RemoveUserFromSamStep`. However, removing the user's role in the later part of the step means that validation no longer succeeds on rerun, and the step is no longer idempotent. This pulls the validation into a different step first.
This does introduce slightly more time between the validation and role removal, but I think that's worthwhile given the alternative of failing on step retry.


ddietterich (approved):
LGTM

yuhuyoyo (approved):

PR #1203, opened 2023-04-18
[TERRA-318] Implement getCredentials

dexamundsen:
```
# get s3 credentials from endpoint via swagger (local deployment)
http://localhost:8080/swagger-ui.html#/ControlledAwsResource/getAwsStorageFolderCredential

{
  "AccessKeyId": "xxx",
  "SecretAccessKey": "xxx",
  "SessionToken": "xxx",
  "Expiration": "2023-04-18T04:50:28Z"
}
```
```
# verify credentials
~ >> aws configure --profile testcred
AWS Access Key ID [***]: xxx
AWS Secret Access Key [***]: xxx
Default region name [us-east-1]: 
Default output format [None]: 
~ >> aws configure --profile testcred set aws_session_token <session-token>
~ >> aws sts --profile testcred get-caller-identity
{
    "UserId": "<access id>:<user email>,
    "Account": "<account>",
    "Arn": "arn:aws:sts::<account>:assumed-role/<role>/<email>"
}
```


dexamundsen:
FYI 

@jmczerk @yuhuyoyo @zloery @ddietterich 

Implemented getCredentials based on the same flow as in mc-feature. 
this was tested locally (see description). 

Merging this to the main S3 PR, tagging you since you may have already reviewed the rest of the changes in the S3 PR 
https://github.com/DataBiosphere/terra-workspace-manager/pull/1198

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
```diff
@@ -207,6 +222,67 @@ public static AwsCredentialsProvider createWsmCredentialProvider(
         authentication.getGoogleJwtAudience());
   }
 
+  public static Credentials getAssumeServiceRoleCredentials(
```
jmczerk, line 225: Isn't this method equivalent to calling `createWsmCredentialProvider().resolveCredentials()`?  The duration passed to the "get resource credential" WSM API should not apply to the service credential, just the user credential, so does not need to be passed here.

That method makes use of the SDK's `StsAssumeRoleWithWebIdentityCredentialsProvider` so we don't have to make the `assumeRoleWithWebIdentity` call directly.

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
```diff
@@ -207,6 +222,67 @@ public static AwsCredentialsProvider createWsmCredentialProvider(
...
+    logger.info(
+        "Assuming Service role ('{}') with session name `{}`, duration {} seconds.",
+        request.roleArn(),
+        request.roleSessionName(),
+        request.durationSeconds());
+
+    return getStsClient().assumeRoleWithWebIdentity(request).credentials();
+  }
+
+  public static Credentials getAssumeUserRoleCredentials(
```
jmczerk, line 241: This need to pass a container of tags that get consolidated with the tags that the `assumeUserRoleFromServiceCredentials()` method adds.  As this is written, the tags necessary to access the folder cannot be passed, so the resulting credential will not have access to the storage folder.

service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
```diff
@@ -206,6 +215,25 @@ public ResponseEntity<ApiAwsCredential> getAwsStorageFolderCredential(
...
       ApiAwsCredentialAccessScope accessScope,
       Integer duration) {
-    throw new NotImplementedException("not implemented");
+    final AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    controlledResourceMetadataManager
+        .validateControlledResourceAndAction(
+            userRequest, workspaceUuid, resourceId, getSamAction(accessScope))
+        .castByEnum(WsmResourceType.CONTROLLED_AWS_STORAGE_FOLDER);
+
+    Credentials awsCredentials =
```
jmczerk, line 224: Discussed below, but this needs generate and pass the tags that allow access to the bucket as described in [the Principal Tags section of the design doc](https://docs.google.com/document/d/1qWFVAUwidqzx_mqda8vmkjjYnO3MvWwKi_w9z2n3mTA/edit#heading=h.ya9vyen240zq)...  these are echoed [here](https://github.com/verily-src/mc-terra-aws-wsm-landingzone/blob/3b03e739e4e01e5339c539909635fcbbbde59517/modules/environment/common.tf#L10)

(note the tag key names are a little stale, but let's solve one problem at a time and just go with them)

jmczerk (commented):
Adding comments here even though this already merged to the dev branch.

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
dexamundsen: createWsmCredentialProvider().resolveCredentials() returns 
[software.amazon.awssdk.auth.credentials.AwsCredentials](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/auth/credentials/AwsCredentials.html) , which only includes keyId & accessKey. 

fwiu, we need [software.amazon.awssdk.services.sts.model.Credentials](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/sts/model/Credentials.html) for API authentication, it includes sessionToken, expiration time along with keyId & accessKey

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
dexamundsen: Agree, relates to https://github.com/DataBiosphere/terra-workspace-manager/pull/1203#discussion_r1170295343 (missed adding [principal tags).

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
jmczerk: Let's leave it as is for now and I'll take an action to create a CRL COW for STS to streamline this later.

PR #1205, opened 2023-04-18
PF-2708: switch aws_enabled to use flagsmith

yuhuyoyo:


service/src/main/resources/application.yml
```diff
@@ -205,8 +208,6 @@ landingzone:
 feature:
   # azure-enabled - Controls inclusion of Azure support in WSM
   azure-enabled: true
-  # aws-enabled - Controls inclusion of AWS support in WSM
-  aws-enabled: false
```
dexamundsen, line 209: Can we keen aws-enabled in place till feature service is set up and tested in all env?
i.e. keep the plumbing for aws-enabled as is, but replace the call to isAwsEnabled with the call to feature service

yuhuyoyo: i set the default value to false in the FeatureService.

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
```diff
@@ -4,41 +4,46 @@
...
-      AwsConfiguration awsConfiguration) {
+      WorkspaceDao workspaceDao, FeatureService featureService, AwsConfiguration awsConfiguration) {
+    this.awsConfiguration = awsConfiguration;
     this.workspaceDao = workspaceDao;
+    this.featureService = featureService;
     this.environmentDiscovery =
-        featureConfiguration.isAwsEnabled()
-            ? AwsUtils.createEnvironmentDiscovery(awsConfiguration)
-            : null;
+        featureService.awsEnabled() ? AwsUtils.createEnvironmentDiscovery(awsConfiguration) : null;
```
yuhuyoyo, line 41: i don't like how we are realizing on flag values at app start up. 

there might be a delay of flag fetching and app start up so we might be getting the default value here. also we could switch flag on independent of service redeployment now.

service/src/main/java/bio/terra/workspace/service/features/FeatureService.java
```diff
@@ -0,0 +1,26 @@
...
+@Component
+public class FeatureService {
+  private final FlagsmithService flagsmithService;
+
+  @Autowired
+  FeatureService(FlagsmithService flagsmithService) {
+    this.flagsmithService = flagsmithService;
+  }
+
+  public boolean awsEnabled() {
```
yuhuyoyo, line 17: @ddietterich we can just inject featureconfiguration in here for features that broad would care about. but i don't think they care about aws atm so I don't want to do that. 

if it's a broad only flag, they can just keep using featureconfigurations. as long as application.yaml keep it as false by default, we don't need to worry about it.

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
jmczerk: We do want the `EnvironmentDiscovery` class to be long-lived, as it does some caching  to reduce cloud reads of infrequently-updated data in an S3 bucket.

So  perhaps the `maybeInitializeEnvironmentDiscovery()` method below should also be only call `AwsUtils. createEnvironmentDiscovery()` if `this.environmentDiscovery` is `null` AND AWS is enabled?

service/src/main/resources/application.yml
dexamundsen: Sure, no concern if tested

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
```diff
@@ -4,41 +4,45 @@
...
+  }
+
+  private void maybeInitializeEnvironmentDiscovery() {
+    this.environmentDiscovery =
+        featureService.awsEnabled() ? AwsUtils.createEnvironmentDiscovery(awsConfiguration) : null;
```
dexamundsen, line 45:
```suggestion
        (this.environmentDiscovery == null && featureService.awsEnabled()) ? AwsUtils.createEnvironmentDiscovery(awsConfiguration) : null;
```

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
```diff
@@ -4,41 +4,45 @@
...
     this.workspaceDao = workspaceDao;
+    this.featureService = featureService;
     this.environmentDiscovery =
-        featureConfiguration.isAwsEnabled()
-            ? AwsUtils.createEnvironmentDiscovery(awsConfiguration)
-            : null;
+        featureService.awsEnabled() ? AwsUtils.createEnvironmentDiscovery(awsConfiguration) : null;
+  }
+
+  private void maybeInitializeEnvironmentDiscovery() {
```
dexamundsen, line 43:
```suggestion
  private void initializeEnvironmentDiscovery() {
```

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
```diff
@@ -137,15 +141,16 @@ public AwsCloudContext getRequiredAwsCloudContext(UUID workspaceUuid) {
    *
    * @return AWS environment
    */
-  public @NotNull Environment discoverEnvironment()
-      throws IllegalArgumentException, InternalLogicException {
+  public Environment discoverEnvironment() throws IllegalArgumentException, InternalLogicException {
     try {
-      Assert.notNull(this.environmentDiscovery, "environmentDiscovery not configured");
+      maybeInitializeEnvironmentDiscovery();
```
dexamundsen, line 146:
```suggestion
      initializeEnvironmentDiscovery();
```

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
```diff
@@ -4,41 +4,45 @@
...
-      AwsConfiguration awsConfiguration) {
+      WorkspaceDao workspaceDao, FeatureService featureService, AwsConfiguration awsConfiguration) {
+    this.awsConfiguration = awsConfiguration;
     this.workspaceDao = workspaceDao;
+    this.featureService = featureService;
     this.environmentDiscovery =
-        featureConfiguration.isAwsEnabled()
-            ? AwsUtils.createEnvironmentDiscovery(awsConfiguration)
-            : null;
+        featureService.awsEnabled() ? AwsUtils.createEnvironmentDiscovery(awsConfiguration) : null;
```
dexamundsen, line 40:
```suggestion
        initializeEnvironmentDiscovery()
```

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
yuhuyoyo: done, dex suggested the same

btw i assume this is a known thing, i tried turn on aws and wsm doesn't build if it's on, complaning about discoveryConfig.

```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'workspaceApiController' defined in file [/Users/yuhuyoyo/databiosphere/terra-workspace-manager/service/build/classes/java/main/bio/terra/workspace/app/controller/WorkspaceApiController.class]: Unsatisfied dependency expressed through constructor parameter 10; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'awsCloudContextService' defined in file [/Users/yuhuyoyo/databiosphere/terra-workspace-manager/service/build/classes/java/main/bio/terra/workspace/service/workspace/AwsCloudContextService.class]: Bean instantiation via constructor failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [bio.terra.workspace.service.workspace.AwsCloudContextService]: Constructor threw exception; nested exception is java.lang.NullPointerException: Cannot invoke "bio.terra.workspace.app.configuration.external.AwsConfiguration$Discovery.getBucket()" because "discoveryConfig" is null
```

jmczerk: We can only turn AWS on connected/integration tests when running in Verily `devel`, as that's the only environment with AWS backing.  Currently we only turn AWS on in `awsConnectedTests`, which do get run in downstream (`verily-src` repo) `pr-test.yaml` Github workflow.  

This workflow has a `workflow_dispatch` trigger, so you could trigger a run against a branch of the `verily-src` repo with the submodule pointed at your development branch.

service/src/main/resources/application.yml
```diff
@@ -154,6 +155,8 @@ terra.common:
   tracing:
     stackdriverExportEnabled: ${env.tracing.exportEnabled}
     samplingRate: ${env.tracing.samplingRate}
+  flagsmith:
```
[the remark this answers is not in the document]
yuhuyoyo: it's to enable the TCL flagsmithService. but I'm going to just do that through env var. I will just remove this from application.yaml

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
yuhuyoyo: ah ok, i see you triggered it. I didn't turn the flag on in devel yet. i will flip it and re-run that test

ddietterich (approved):

dexamundsen (approved):

PR #1209, opened 2023-04-19
Credential tags update

dexamundsen:


service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
```diff
@@ -71,6 +73,33 @@ private static String getRoleSessionName(String value) {
...
+    tags.add(Tag.builder().key("version").value(awsCloudContext.getMajorVersion()).build());
+    tags.add(Tag.builder().key("tenant").value(awsCloudContext.getTenantAlias()).build());
+    tags.add(Tag.builder().key("environment").value(awsCloudContext.getEnvironmentAlias()).build());
+  }
+
+  public static void appendPrincipalTags(
+      Collection<Tag> tags,
+      AwsCloudContext awsCloudContext,
+      ControlledAwsStorageFolderResource awsStorageFolderResource) {
+    tags.add(Tag.builder().key("version").value(awsCloudContext.getMajorVersion()).build());
```
jmczerk, line 90: The newer environment deployments (including `devel`) use the more conventional camel casing for the tag keys (defined as the values [here](https://github.com/verily-src/mc-terra-aws-wsm-landingzone/blob/main/modules/environment/common.tf#L10)):

* `version -> Version`
* `bucket_id -> S3BucketID`
* `terra_bucket_id -> TerraBucketID`

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
```diff
@@ -71,6 +73,33 @@ private static String getRoleSessionName(String value) {
...
+    tags.add(
+        Tag.builder().key("bucket_id").value(awsStorageFolderResource.getBucketName()).build());
+    tags.add(
+        Tag.builder().key("terra_bucket_id").value(awsStorageFolderResource.getPrefix()).build());
+  }
+
+  public static void appendRoleTags(Collection<Tag> tags, ApiAwsCredentialAccessScope accessScope) {
+    if (accessScope == ApiAwsCredentialAccessScope.WRITE_READ) {
+      tags.add(Tag.builder().key("ws_role").value("writer").build());
```
jmczerk, line 99: `ws_role -> WorkspaceRole` per above comment.

Also, `WorkspaceRole` tag must be set to `reader` if the role is `READ_ONLY` (i.e. if not `WRITE_READ`).

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
```diff
@@ -71,6 +73,33 @@ private static String getRoleSessionName(String value) {
         : value;
   }
 
+  public static void appendUserTags(Collection<Tag> tags, AuthenticatedUserRequest userRequest) {
+    tags.add(Tag.builder().key("user_id").value(userRequest.getSubjectId()).build());
```
jmczerk, line 77: `user_id -> UserID`

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
```diff
@@ -71,6 +73,33 @@ private static String getRoleSessionName(String value) {
...
+  }
+
+  public static void appendResourceTags(Collection<Tag> tags, AwsCloudContext awsCloudContext) {
+    tags.add(Tag.builder().key("version").value(awsCloudContext.getMajorVersion()).build());
```
jmczerk, line 81: * `version -> Version`
* `tenant -> Tenant`
* `environment -> Environment`

jmczerk (commented):
Casing needs to match the values rather than the keys defined in the Terraform, but otherwise LGTM.

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
jmczerk: I read this as "will only be mentioned _**in the document**_ when it is writer" (and this is the case in the subsequent sections describing the IAM roles).  @jwtracy can confirm.

The sentence before that states "For the TerraUser role, this tag must always exist.", and reading the IAM policies aligns to this

dexamundsen: noted

jmczerk (approved):

PR #1211, opened 2023-04-19
Correct quoting of terra tab completion command in post-startup.sh

mbookman:
Fixes the error noted in https://broadworkbench.atlassian.net/browse/PF-2703.


wnojopra (approved):
LGTM

yuhuyoyo (approved):

PR #1213, opened 2023-04-20
WOR-903 don't leave workspace in a corrupted state when deletion fails

blakery:
Ticket: [WOR-903](https://broadworkbench.atlassian.net/browse/WOR-903)

Adds a step to the workspace delete flight to check that no child resources remain before finalizing the delete.

I have unit test coverage on the new step, and I ran the existing 'connectedPlus' tests for workspace deletion (`WorkspaceDeleteFlightConnectedTest`), which passed. 
The logic is pretty much the same as when rawls checks for GCP child resources that would block deletion, so it's not doing anything novel.

I also have the urge to go through and write unit tests for all of the individual flight steps, but that's a larger task, for another day(s).

[WOR-903]: https://broadworkbench.atlassian.net/browse/WOR-903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
```diff
@@ -73,35 +71,53 @@ public WorkspaceDeleteFlight(FlightMap inputParameters, Object applicationContex
...
+    addStep(
+        new DeleteWorkspaceStateStep(appContext.getWorkspaceDao(), workspaceUuid), terraRetryRule);
+  }
+
+  /**
+   * Adds steps to delete sam resources owned by WSM Workspace authz is handled differently
+   * depending on whether WSM owns the underlying Sam resource or not, as indicated by the workspace
+   * stage enum.
+   */
+  protected void addOauthZSteps(
```
blakery, line 94: The main function was getting cumbersome to read, so I moved this part out. 
The only functional change I made to this file was adding the `EnsureNoWorkspaceChildrenStep`.

service/src/main/java/bio/terra/workspace/service/workspace/exceptions/ChildrenBlockingDeletionException.java
```diff
@@ -0,0 +1,13 @@
...
+
+import bio.terra.common.exception.ErrorReportException;
+import java.util.List;
+import org.springframework.http.HttpStatus;
+
+/** Exception thrown when a user attempts to delete a workspace that still has children */
+public class ChildrenBlockingDeletionException extends ErrorReportException {
+
+  public ChildrenBlockingDeletionException(String message, List<String> causes) {
+    super(message, causes, HttpStatus.CONFLICT);
```
blakery, line 11: IETF status codes:

> [6.5.8](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.8).  409 Conflict
> 
>    The 409 (Conflict) status code indicates that the request could not
>    be completed due to a conflict with the current state of the target
>    resource.  This code is used in situations where the user might be
>    able to resolve the conflict and resubmit the request.  The server
>    SHOULD generate a payload that includes enough information for a user
>    to recognize the source of the conflict.
> 
>    Conflicts are most likely to occur in response to a PUT request.  For
>    example, if versioning were being used and the representation being
>    PUT included changes to a resource that conflict with those made by
>    an earlier (third-party) request, the origin server might use a 409
>    response to indicate that it can't complete the request.  In this
>    case, the response representation would likely contain information
>    useful for merging the differences based on the revision history.

I think that lines up, particularly when combined with the error message enumerating the resources that are blocking deletion.

service/src/test/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlightTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.mockito.Mockito;
+import org.springframework.test.context.ActiveProfiles;
+
+// This doesn't extend BaseUnitTest because we don't need BaseUnitTestMocks
+@Tag("unit")
+@ActiveProfiles({"unit-test"})
+public class WorkspaceDeleteFlightTest {
+
+  @Test
+  void stepsForRawlsWorkspace() {
```
blakery, line 29: I realize these aren't the most meaningful tests, but it seems useful to have spelled out something to the effect of 'this flight contains these steps in this order'.

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
marctalbott: I'm a bit confused about the naming since this is workspace authz not oauth?

blakery: Force of habit

service/src/main/java/bio/terra/workspace/service/workspace/exceptions/ChildrenBlockingDeletionException.java
blakery: rawls returns 400.
I'm not sure consistency with rawls matters much here since:

1. It's still 4xx, just more semantic
2. this is after all the checks in rawls, so it's additive anyway
3. More specific error codes tend to make things easier to track down

marctalbott (approved):

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -292,6 +292,18 @@ private void registerWsmServiceAccount(UsersApi usersApi) throws InterruptedExce
     }
   }
 
+  public List<FullyQualifiedResourceId> getWorkspaceChildResources(
```
cahrens, line 295: Should we be adding JavaDoc for public API methods?

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
```diff
@@ -73,35 +71,53 @@ public WorkspaceDeleteFlight(FlightMap inputParameters, Object applicationContex
         new DeleteGcpProjectStep(
             appContext.getCrlService(), appContext.getGcpCloudContextService()),
         cloudRetryRule);
+
+    addStep(
+        new EnsureNoWorkspaceChildrenStep(appContext.getSamService(), userRequest, workspaceUuid));
```
cahrens, line 76: I was originally thinking that we would check if any Sam child resources existed that we would not be able to delete _before_ we called `DeleteControlledAzureResourcesStep` and `DeleteControlledSamResourcesStep` (so that the workspace would stay in its original state). But is there no way for us to know what is deletable without actually first attempting to delete things?

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
```diff
@@ -73,35 +71,53 @@ public WorkspaceDeleteFlight(FlightMap inputParameters, Object applicationContex
...
     addStep(
         new DeleteAwsContextStep(appContext.getAwsCloudContextService(), workspaceUuid),
         cloudRetryRule);
+    addAuthZSteps(appContext, inputParameters, userRequest, workspaceUuid, terraRetryRule);
+    addStep(
+        new DeleteWorkspaceStateStep(appContext.getWorkspaceDao(), workspaceUuid), terraRetryRule);
+  }
+
+  /**
+   * Adds steps to delete sam resources owned by WSM Workspace authz is handled differently
```
cahrens, line 90: Nit: I think you are missing a period/semi-colon after "WSM Workspace".

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
```diff
@@ -73,35 +71,53 @@ public WorkspaceDeleteFlight(FlightMap inputParameters, Object applicationContex
...
+   * depending on whether WSM owns the underlying Sam resource or not, as indicated by the workspace
+   * stage enum.
+   */
+  protected void addAuthZSteps(
```
cahrens, line 94: Why is this method protected?

service/src/main/java/bio/terra/workspace/service/workspace/flight/EnsureNoWorkspaceChildrenStep.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    var children = samService.getWorkspaceChildResources(userRequest, workspaceUuid);
+    if (!children.isEmpty()) {
+      var blocking =
+          children.stream()
+              .map(r -> String.format("%s resource %s", r.getResourceTypeName(), r.getResourceId()))
+              .toList();
+      throw new ChildrenBlockingDeletionException(
```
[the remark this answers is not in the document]
blakery: >     2023-04-24 23:46:04.443   INFO 39469 --- [pool-8-thread-5] bio.terra.stairway.impl.FlightRunner     : Caught exception: (bio.terra.workspace.service.workspace.exceptions.ChildrenBlockingDeletionException@3cfa59bd[causes=[controlled-user-shared-workspace-resource resource 202a9fb2-92fc-4112-9206-d6061fc97ef5],statusCode=409 CONFLICT]) flight id: k0MINFm6SUOobaxKe3s8gg step: 0 direction: DO

The message and list of resources is basically the same as rawls gives for child resources blocking deletion.

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
blakery: I'm missing a line break
I moved the inline comment from the block of code this was factoring out to the top of the method, and I think some formatting got weird.

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
blakery: I was thinking of unit tests, but there's really no point: form a testing standpoint, it's all the same 'unit'.

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
blakery: I don't see any way to know, especially as we might be depending upon external services to actually delete resources, and some of those might have child resources, etc.

aherbst-broad (approved):

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceDeleteFlight.java
blakery: It was actually that without a period, running spotlessApply tries to reformat it.

PR #1218, opened 2023-04-24
[PF-2296] Propagate policy created/lastUpdated time through WSM

ddietterich:
This is copying data from TPS objects out through the WSM interface.


openapi/src/parts/tps.yaml
```diff
@@ -68,6 +68,10 @@ components:
           uniqueItems: true
           items:
             $ref: '#/components/schemas/WsmPolicyObjectId'
+        createdDate:
+          type: string
```
mikenev, line 72: Do we want to give these the `format: date-time` like we did in WSM?

mikenev (approved):
Minor comment otherwise lgtm.

openapi/src/parts/tps.yaml
ddietterich: No, I don't want to do that. The data is already in string form coming out of TPS, so it seems useless to parse and convert it just to convert it back again.

yuhuyoyo: what does the client-side output looks like? my only concern is that ui need to convert it?

yuhuyoyo (approved):

openapi/src/parts/tps.yaml
ddietterich: It is the standard ISO 8601 format. I'll add that to the documentation.

PR #1222, opened 2023-04-26
[PF-2694] Reorganize workspace folder into packages

ddietterich:
As a first step to revising the cloud context and workspace flights, put the workspace folder into a better structure.
This makes **no code changes**. It only rearranges the files into packages.


zloery (approved):
One thought (and it's hard to comment on unchanged files in git!), otherwise LGTM:

do Application flights/steps need to be under /cloud? I think of these as Terra concepts rather than cloud ones, so I'd expect them to be somewhere like service/workspace/flight/application rather than .../cloud/any

ddietterich:
> do Application flights/steps need to be under /cloud? I think of these as Terra concepts rather than cloud ones, so I'd expect them to be somewhere like service/workspace/flight/application rather than .../cloud/any

Good question. I was thinking of it as following the same pattern as resources, where we have cloud specific ones and not. But it doesn't really fit the pattern. I'll pull them out to the outer flight level.

yuhuyoyo (approved):

PR #1225, opened 2023-04-26
[PF-2694] Factor out StateDao

ddietterich:
This change creates a generic StateDao that processes state transitions. It relies on a new interface `DbStateful` that I implemented in `DbResource`.

I plan to reuse this when handling the states in the workspace and cloud_context tables.


service/src/main/java/bio/terra/workspace/db/StateDao.java
```diff
@@ -0,0 +1,158 @@
...
+      @Nullable DbStateful dbStateful, WsmResourceState state, String flightId) {
+    if (dbStateful == null) {
+      return (state == WsmResourceState.NOT_EXISTS && flightId == null);
+    }
+
+    return (dbStateful.getState() == state
+        && StringUtils.equals(dbStateful.getFlightId(), flightId));
+  }
+
+  public void updateState(
```
zloery, line 48: I think this needs `@WriteTransaction`

zloery (commented):
LGTM, but is the stuff under `service/workspace/` intended for the next PR?

service/src/main/java/bio/terra/workspace/db/model/DbStateful.java
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Get the flightId of the DB object
+   *
+   * @return nullable flight id
+   */
+  @Nullable
+  String getFlightId();
+
+  /** Get the text string of the type of object for logging */
+  String getObjectTypeString();
```
yuhuyoyo, line 25: what about reuse the enum?

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/cloudcontext/CreateDbCloudContextStartStep.java
```diff
@@ -0,0 +1,46 @@
...
+import bio.terra.workspace.db.WorkspaceDao;
+import bio.terra.workspace.service.workspace.exceptions.DuplicateCloudContextException;
+import bio.terra.workspace.service.workspace.model.CloudPlatform;
+import java.util.UUID;
+
+/**
+ * Stores the previously generated Google Project Id in the {@link WorkspaceDao} as the Google cloud
+ * context for the workspace.
+ */
+public class CreateDbCloudContextStartStep implements Step {
```
yuhuyoyo, line 15: same with this one

service/src/main/java/bio/terra/workspace/db/StateDao.java
ddietterich: I renamed the module `StateDbUtils`. It is always called from other DAOs that should have the proper transaction markings. I also added a comment about that.

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/cloudcontext/CreateCloudContextFlight.java
```diff
@@ -0,0 +1,3 @@
+package bio.terra.workspace.service.workspace.flight.create.cloudcontext;
+
+public class CreateCloudContextFlight {}
```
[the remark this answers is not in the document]
ddietterich: Yes, these two are not used. I will remove them.

service/src/main/java/bio/terra/workspace/db/model/DbStateful.java
ddietterich: Which enum? I'm not sure what you mean

yuhuyoyo: https://github.com/DataBiosphere/terra-workspace-manager/blob/34b6d06e2d5af4cac1c8c9fde69b84e5301320c6/service/src/main/java/bio/terra/workspace/common/logging/model/ActivityLogChangedTarget.java#L4

yuhuyoyo (approved):

PR #1226, opened 2023-04-27
[PF-2656] Create a vm startup script and associated systemd service in post startup script

rogerwangcs:
To mount workspace resources automatically on every vm bootup, we create a systemd service `terra-setup-environment` to run the mount command along with future commands or scripts that we need to run on startup/restart.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -313,14 +313,16 @@ sudo -u "$JUPYTER_USER" sh -c 'terra git clone --all'
 ################################
 # Setup instance startup service
 ################################
-# The following steps are performed
+# Create a script to perform the following steps on instance startup:
 # 1. Mount terra workspace resources. This command requires system user home
 #    directories to be mounted. We run the startup service after
 #    jupyter.service to meet this requirement.
 
-# Create a script to be executed by the startup service.
-TERRA_SETUP_SCRIPT="${USER_TERRA_CONFIG_DIR}/setup-environment.sh"
-cat <<EOF >${TERRA_SETUP_SCRIPT}
+# Create the startup script
+readonly TERRA_SETUP_SCRIPT="${USER_TERRA_CONFIG_DIR}/setup-environment.sh"
+readonly TERRA_SETUP_SERVICE="/etc/systemd/system/terra-setup-environment.service"
+
+cat <<EOF >"${TERRA_SETUP_SCRIPT}"
 #!/bin/bash
 
 # Mount terra workspace resources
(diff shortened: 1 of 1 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -310,6 +310,44 @@ fi
...
+chmod +x "${TERRA_SETUP_SCRIPT}"
+
+# Create a systemd service to run the script on instance startup.
+cat <<EOF >"/etc/systemd/system/terra-setup-environment.service"
+[Unit]
+Description=Configure environment for terra
+After=jupyter.service
+
+[Service]
+ExecStart=/home/jupyter/.terra/setup-environment.sh
```
mbookman, line 338:
```
ExecStart="${TERRA_SETUP_SCRIPT}"
User="${JUPYTER_USER}"
```

Shell variables will expand at the time this is written (unless you put the HereDoc delimiter "EOF" in quotes.

mbookman (changes requested):
Great to be getting this capability @rogerwangcs !

Just a few small things...

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
rogerwangcs: Thanks Matt, appreciate the tips! Using variables and substitution instead of hardcoding now.

mbookman (approved):
LGTM

yuhuyoyo (commented):
i'm not familiar with what's going here so I can't review...

but why are we creating a startup script in the post-startup script?

rogerwangcs:
> i'm not familiar with what's going here so I can't review...
> 
> but why are we creating a startup script in the post-startup script?

I could use help on giving it a better name. The post startup script runs once on instance **creation**, while this startup script runs on _every_ instance **startup/boot**.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -310,6 +310,46 @@ fi
 # integration test will ensure that everything in script worked.
 sudo -u "$JUPYTER_USER" sh -c 'terra git clone --all'
 
+################################
+# Setup instance startup service
+################################
+# Create a script to perform the following steps on instance startup:
```
cbookg, line 316: Maybe to make this more clear, "on each instance reboot"?  Or just in general make it clear in the comment that we have this script to run on every reboot, not just one-time?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -310,6 +310,46 @@ fi
...
+# 1. Mount terra workspace resources. This command requires system user home
+#    directories to be mounted. We run the startup service after
+#    jupyter.service to meet this requirement.
+
+# Create the startup script
+readonly TERRA_SETUP_SCRIPT="${USER_TERRA_CONFIG_DIR}/setup-environment.sh"
```
cbookg, line 322: i feel like this naming doesn't really explain the purpose.

not sure i have a better suggestion though.  maybe something like "instance-boot.sh", "terra-instance-boot.service"?

cbookg (approved):
code looks good.  naming/comments are not really explaining it.

PR #1232, opened 2023-04-28
WOR-907 add contract tests between WSM and BPM

blakery:
Ticket: [WOR-907](https://broadworkbench.atlassian.net/browse/WOR-907)

Add contract tests between WSM and BPM using [pact](https://docs.pact.io/)

* add and configure pact
* add consumer tests between WSM and BPM

There are 3 endpoints in BPM called from WSM, Creating, Deleting, and Retrieving a billing profile.
Creation and deletion are both used for setup in integration tests for retrieving a billing profile.
Since the pact tests serve the same purpose as those integration tests (not the same purpose as _all_ integration tests - just those ones), we should be able to remove them, once we have more confidence in how pact functions.
So I decided to only include the endpoint we actually care about, for our pact between services.

[WOR-907]: https://broadworkbench.atlassian.net/browse/WOR-907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/build.gradle
```diff
@@ -59,6 +60,17 @@ jacoco {
     applyTo jacocoBootRun
 }
 
+test {
+    systemProperties['pact.rootDir'] = "$buildDir/pacts"
+    systemProperties['pact.provider.version'] = "$project.version"
+}
+
+tasks.register("pactTests", Test) {
```
blakery, line 68: This is all the same configuration used in [BPM](https://github.com/DataBiosphere/terra-billing-profile-manager/pull/168)

service/src/test/java/bio/terra/workspace/pact/ProfileApiTest.java
```diff
@@ -0,0 +1,108 @@
...
+            .uuid("tenantId")
+            .uuid("subscriptionId")
+            .stringType("managedResourceGroupId")
+            .stringType("billingAccountId")
+            .uuid("id");
+    return builder
+        .given("an existing billing profile")
+        .uponReceiving("A request to retrieve a billing profile")
+        .method("GET")
+        .pathFromProviderState(
```
blakery, line 54: From what I can tell, this will allow us to use a value in the consumer-side tests that will get dynamically replaced when run on the provider side.
This prevents having to agree on magic values that don't have any real meaning, like randomly generated ids.
More details in [the pact-jvm docs](https://github.com/pact-foundation/pact-jvm/tree/master/consumer#having-values-injected-from-provider-state-callbacks)

service/src/test/java/bio/terra/workspace/pact/ProfileApiTest.java
```diff
@@ -0,0 +1,108 @@
...
+@ExtendWith(PactConsumerTestExt.class)
+public class ProfileApiTest {
+
+  // This will need to be set up with BPM
+  static final String billingProfileId = "4a5afeaa-b3b2-fa51-8e4e-9dbf294b7837";
+
+  @Pact(consumer = "wsm-consumer", provider = "bpm-provider")
+  public RequestResponsePact billingProfileAvailable(PactDslWithProvider builder) {
+    // Using the WSM version of this, instead of the BPM version,
+    // because what we care about is not what BPM sends, but that WSM can handle the value
```
marctalbott, line 37: Wouldn't we care that WSM can actually handle what BPM sends? Or is that enforced by provider-side tests? In other words, this defines what WSM _can_ handle, and the provider tests enforce that BPM sends something that fits the parameters defined by WSM?

service/src/test/java/bio/terra/workspace/pact/ProfileApiTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mockito;
+
+@Tag("pact-test")
+@ExtendWith(PactConsumerTestExt.class)
+public class ProfileApiTest {
+
+  // This will need to be set up with BPM
```
cahrens, line 31: I am not sure what you mean about this.

service/src/test/java/bio/terra/workspace/pact/ProfileApiTest.java
```diff
@@ -0,0 +1,108 @@
...
+        .toPact();
+  }
+
+  @Pact(consumer = "wsm-consumer", provider = "bpm-provider")
+  public RequestResponsePact billingProfileUnAvailable(PactDslWithProvider builder) {
+    return builder
+        .uponReceiving("A request to retrieve a billing profile")
+        .method("GET")
+        .pathFromProviderState(
+            "/api/profiles/v1/${profileId}", String.format("/api/profiles/v1/%s", billingProfileId))
```
cahrens, line 68: I don't follow-- for the previous test, the billing profile ID returns a 200. How does it get replaced here with something that doesn't exist?

service/src/test/java/bio/terra/workspace/pact/ProfileApiTest.java
cahrens: Hmmm, I read that but I'm still not getting it.

service/src/test/java/bio/terra/workspace/pact/ProfileApiTest.java
blakery: outdated comment from before discovering the pathFromProviderState functionality

blakery: outdated comment from before discovering the pathFromProviderState functionality

service/src/test/java/bio/terra/workspace/pact/ProfileApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mockito;
+
+@Tag("pact-test")
+@ExtendWith(PactConsumerTestExt.class)
+public class ProfileApiTest {
+
+  static final String billingProfileId = "4a5afeaa-b3b2-fa51-8e4e-9dbf294b7837";
```
cahrens, line 29: Is there any reason to just not make this a random UUID? I feel like having a constant in the file makes it seem like the value is somehow important.

blakery: In general, using random (changing) is bad practice with pact - it tends to give incorrect indications that the pact has changed.
It's easiest to have it as a static value so it can be used in both the pact mock setup, and the test itself.

I'll see if I can make it more clear what's going on, though.

service/src/test/java/bio/terra/workspace/pact/ProfileApiTest.java
blakery: Basically, yes.
This defines the pact, which is then run against the provider. 
So we're saying basically "we can handle 'GCP' and 'Azure'". I defined it like this to ensure the string values match up. 

Note that I split this into two tests, one for GCP and one for Azure, after we determined last week that the old test setup is demanding mutually contradictory values.

service/gradle/testing.gradle
```diff
@@ -158,4 +158,17 @@ task awsConnectedTest(type: Test) {
   outputs.upToDateWhen { false }
   finalizedBy tasks.combinedJaCoCoReport
   maxParallelForks = runnerThreads
-}
\ No newline at end of file
+}
+
+// contract tests
+test {
```
marctalbott, line 164: I haven't tried to run these myself, but I assume it's fine to have multiple `test` blocks in the same file? Would it be possible to move all of this into the `test` block up above?

cahrens (approved):

service/gradle/testing.gradle
blakery: yes, you can have an arbitrary number - you just use as many blocks as you need for your organization.
I was just trying to keep the contract testing separate, under the assumption it would be easier to understand.

I changed them to the same pattern as the other tests in the file.
The definition styles are equivalent, but stand out a bit less

marctalbott (approved):

PR #1233, opened 2023-04-28
PF-2724: Fix bigquery data transfer test

yuhuyoyo:
The pet SA is missing the permissions of the service account token creator role. I manually tested by adding these permissions into the PROJECT_OWNER role. It took a few min for the permissions to be synced to the actual pet SA but it works.

it's passing for me locally 👯

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/CustomGcpIamRoleMapping.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/CustomGcpIamRoleMapping.java
@@ -46,7 +46,13 @@ public class CustomGcpIamRoleMapping {
           "bigquery.tables.export",
           "bigquery.tables.getData",
           "bigquery.tables.list",
-          "bigquery.tables.get");
+          "bigquery.tables.get",
+          // Permissions of Service Account Token creator, required for bq data transfer service.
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/CustomGcpIamRoleMapping.java
```diff
@@ -46,7 +46,13 @@ public class CustomGcpIamRoleMapping {
           "bigquery.tables.export",
           "bigquery.tables.getData",
           "bigquery.tables.list",
-          "bigquery.tables.get");
+          "bigquery.tables.get",
+          // Permissions of Service Account Token creator, required for bq data transfer service.
+          "iam.serviceAccounts.getAccessToken",
```
zloery, line 51: This block of permissions is granted on a BQ dataset, I don't think the SA related ones will have any effect here

zloery (changes requested):
The given error message is that a service agent (GCP managed SA) is missing permission on the pet SA, but this change (and your manual testing one) are giving additional permissions to the user and pet, so I'm not sure how that helped.

Running locally I also see the tests pass with just the retry changes from this PR, so maybe this was actually a permissions delay issue with a misleading error message all along 🤷

zloery (approved):

ddietterich (approved):

PR #1234, opened 2023-04-28
Set permissions of instance-boot script so the user can edit the file without sudo

rogerwangcs:
We want to allow users to modify the instance boot script to turn off behaviors they don't want.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -324,11 +324,13 @@ readonly TERRA_BOOT_SERVICE="/etc/systemd/system/terra-instance-boot.service"
 
 cat <<EOF >"${TERRA_BOOT_SCRIPT}"
 #!/bin/bash
+# This script is run on instance boot to configure the instance for terra.
 
 # Mount terra workspace resources
 /usr/bin/terra resource mount
 EOF
 chmod +x "${TERRA_BOOT_SCRIPT}"
+chown ${JUPYTER_USER}:${JUPYTER_USER} "${TERRA_BOOT_SCRIPT}"
 
 # Create a systemd service to run the boot script on system boot
 cat <<EOF >"${TERRA_BOOT_SERVICE}"


wnojopra (approved):
I think this change is fine - but I'm confused on how the user is expected to edit and re-run the file. At the top of this script, there are instructions for copying this script to a GCS location and then creating a new instance with that script. Is there some other path for starting an instance with a modified script that I'm unaware of?

rogerwangcs:
> I think this change is fine - but I'm confused on how the user is expected to edit and re-run the file. At the top of this script, there are instructions for copying this script to a GCS location and then creating a new instance with that script. Is there some other path for starting an instance with a modified script that I'm unaware of?

So the post startup script now creates a new script file called `instance-boot.sh` in the terra config folder. `instance-boot.sh` is configured to run on every instance startup, and that's the file I'm setting ownership on. User's can edit that file and stop/start their instance. Open to ideas on how to make this less confusing.

mbookman:
I think that the approach I would prefer would be something _like_ adding to the TERRA_BOOT_SCRIPT:

```
readonly TERRA_BOOT_SCRIPT_USER="${JUPYTER_USER}/.terra/user/instance_boot.sh"
...
# Check if the user has a custom boot script and if so, run it as well
if [[ -x "${TERRA_BOOT_SCRIPT_USER}" ]]; then
  "${TERRA_BOOT_SCRIPT_USER}"
fi
```

So rather than having the user edit the boot script, they can just drop down their own. If it doesn't exist, that's fine.

mbookman:
How about a simple YAML, JSON, or CSV file with a few options. Since we're looking for something simple to be implemented in Bash, I'd lean towards a CSV file.

`/home/${JUPYTER_USER}/.terra/mount_cfg.csv`

```
resource_a,ro
resource_b,rw
*,rw
```

This specific example is a little off from a real-world example, but what it indicates is:

- mount resource_a read only
- mount resource_b read write
- mount all other resources read write

If the final line there is omitted, then resources not specifically included in the list are omitted from mounting.

If you are looking for some tips on parsing CSVs in Bash, see:

https://www.baeldung.com/linux/csv-parsing

That said, how about building this into the Terra CLI, such that one can pass a config file. That gives you all the power of Java to properly handle the file.

rogerwangcs:
So this feature isn't something that is captured in the prd, but I think this change is be an easy fix to allow users the edit the boot script (without sudo) as a bonus. I can link it in the doc as something to consider for milestone 2.

mbookman (approved):
LGTM (for now).

PR #1240, opened 2023-05-02
WOR-900 link policy api

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-900


service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -54,15 +54,11 @@ public class LandingZoneApiDispatch {
   private static final String AZURE_BATCH_ACCOUNT_RESOURCE_TYPE = "Microsoft.Batch/batchAccounts";
 
   private final LandingZoneService landingZoneService;
-  private final WorkspaceService workspaceService;
```
dvoet, line 57: `LandingZoneApiDispatch` should not depend on `WorkspaceService`. It appears to have been added as a convenience but causes an unresolved circular dependency in this new code. Removed.

dvoet:
@ddietterich I added you as an FYI and also in the know for policy stuff

openapi/src/parts/workspace.yaml
```diff
@@ -202,6 +202,37 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/workspaces/v1/{workspaceId}/policies/link:
```
ddietterich, line 205: Could you make a one pager explaining what this API is for? It is not obvious to me.
In general, I prefer having new APIs written down and looked over before making their appearance in the YAML...

Having read more of the code, I am don't think this is the right level of abstraction. If a workspace should inherit a policy from a billing profile or from a landing zone, then we should build that in WSM. Not build an open ended interface at the policy level that allows any workspace writer to tinker with the policy database.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -261,8 +257,7 @@ private ApiAzureLandingZoneResult toApiAzureLandingZoneResult(
         .landingZone(azureLandingZone);
   }
 
-  public UUID getLandingZoneId(BearerToken token, UUID workspaceId) {
-    Workspace workspace = workspaceService.getWorkspace(workspaceId);
+  public UUID getLandingZoneId(BearerToken token, Workspace workspace) {
     Optional<UUID> profileId = workspace.getSpendProfileId().map(sp -> UUID.fromString(sp.getId()));
```
ddietterich, line 261: FYI - I am in process of moving spend profiles to the cloud context and out of the workspace. As per, [New Create/Delete Workspace Endpoints](https://docs.google.com/document/d/1zbNppE6JJWbFGbasO5iHZHg3EjFVqHeOuj8MULP11qo/edit?resourcekey=0-Qm_6ORckNLekwOSqQ6f83Q#heading=h.potz6upymld7).

ddietterich:
@mikenev Maybe take a look at the policy validator added here. I had the idea we had done something along those lines before. Or was that just on the TPS side?

ddietterich (changes requested):

openapi/src/parts/workspace.yaml
dvoet: this is for linking a workspace's policy to a snapshot's. It is essentially a wrapper around https://tps.dsde-dev.broadinstitute.org/#/Tps/linkPao that verifies everything complies first.

ddietterich: If we made the connection to the snapshot a referenced resource, then this policy would flow in the same way as we do for data collections and other workspaces. I don't think we want the naked link as the API here.

mikenev:
> 

We do some of that validation in ValidateWorkspaceAgainstPolicyStep.java . Perhaps that could be updated to call this function instead.

openapi/src/parts/workspace.yaml
dvoet: here are some docs https://docs.google.com/document/d/1AtnBFMw5YhxWNMVeXRN5g9Z8JBKxbfp1iNnD8IsKdZY/edit#heading=h.90m9ste9j9qq

I am not 100% sure modeling this as a reference makes sense. We are actually copying data into a workspace, not referencing anything.

How does this work today for copying data from other workspaces?

dvoet:
> We do some of that validation in ValidateWorkspaceAgainstPolicyStep.java . Perhaps that could be updated to call this function instead.

I agree there is work ahead to make this more consistent. My current target is Azure specific. I did not want to change the behavior for GCP yet. A bridge too far for this PR.

openapi/src/parts/workspace.yaml
ddietterich: This works as part of the clone operation. Whether you are making a reference or copying the data, the workspace inherits the policy of the source. Maybe we need a variant of clone that has a snapshot as a source. Presumably, you could make a reference or a controlled resource as the target based on cloning instructions.

Alternately, some policy-aware data movement component that can interact with TPS to configure the policies. But that seems like too big a lift.

dvoet:
@ddietterich I removed the new api and added the policy linking to the create snapshot reference api

openapi/src/parts/tps.yaml
```diff
@@ -169,6 +169,16 @@ components:
         updateMode:
           $ref: '#/components/schemas/WsmPolicyUpdateMode'
 
+    WsmLinkPoliciesRequest:
```
ddietterich, line 172: I think this can go away

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -476,11 +481,29 @@ public ResponseEntity<ApiDataRepoSnapshotResource> createDataRepoSnapshotReferen
...
+                    .component(TpsComponent.TDR)
+                    .objectType(TpsObjectType.SNAPSHOT),
+                TpsUpdateMode.FAIL_ON_CONFLICT,
+                userRequest)
+            : null;
+
+    if (linkPoliciesResults == null || linkPoliciesResults.isUpdateApplied()) {
+      ReferencedDataRepoSnapshotResource referenceResource =
+          referencedResourceService
+              .createReferenceResource(resource, userRequest)
```
ddietterich, line 499: If this create fails, the link will exist, but the resource won't. This would be better as a two step flight. It is a tiny window, so I won't sweat it for this PR, but suggest doing this as cleanup work.

service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java
```diff
@@ -93,13 +97,13 @@ public void createPao(UUID workspaceUuid, @Nullable TpsPolicyInputs policyInputs
   // workspaces have an associated policy attribute object. This function creates
   // an empty one if it does not exist.
   @Traced
-  public void createPaoIfNotExist(UUID workspaceId) {
-    Optional<TpsPaoGetResult> pao = getPaoIfExists(workspaceId);
+  public void createPaoIfNotExist(UUID objectId, TpsComponent component, TpsObjectType objectType) {
```
ddietterich, line 100: Is this change needed?
An alternative would have been to add an overload that populates the component and type.

dvoet: I made this change because I want to populate TDR snapshot PAOs if they don't exist. I agree that this is not really the province of WSM. However, I think it is important enough to do this anyway so that the PAO exists and future policy changes to the snapshot can trickle down appropriately.

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
dvoet: so long are you are not sweating, I will add a comment and let this slide

dvoet: I don't see a way to remove the link via the TPS api in case of failure in any case

ddietterich: TPS itself has the unlink ability. Maybe we don't surface that API in `TpsApiDispatch`. Which means we probably have other undo problems elsewhere...

service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java
```diff
@@ -476,11 +481,31 @@ public ResponseEntity<ApiDataRepoSnapshotResource> createDataRepoSnapshotReferen
             .snapshotId(body.getSnapshot().getSnapshot())
             .build();
 
-    ReferencedDataRepoSnapshotResource referenceResource =
-        referencedResourceService
-            .createReferenceResource(resource, userRequest)
-            .castByEnum(WsmResourceType.REFERENCED_ANY_DATA_REPO_SNAPSHOT);
-    return new ResponseEntity<>(referenceResource.toApiResource(), HttpStatus.OK);
+    var linkPoliciesResults =
```
dvoet, line 484: This is the entry point for the change to link policies to a snapshot. Start here to follow the new feature.

ddietterich:
This change is fine with me. I'll let Blake and Marc do the approving.

marctalbott (approved):

PR #1241, opened 2023-05-03
[TERRA-548] Use Workspace Userfacing id in generateCloudName

dexamundsen:
Use Workspace Userfacing id in generateCloudName
- add a second generateCloudName method in interface with string as first param
- add default definition to generateCloudName in interface to avoid repeated definition of both function in all inheritors

Others:
- Add aws-unit test tag
- Replace tag awsConnected with aws-connected (downstream  workflow files will be updated)
- Move workspaceService to ControlledResourceControllerBase
- minor optimization of gcp.generateCloudName


service/src/test/java/bio/terra/workspace/common/BaseUnitTest.java
```diff
@@ -3,6 +3,7 @@
 import org.junit.jupiter.api.Tag;
 import org.springframework.test.context.ActiveProfiles;
 
+/** Base class for (GCP) unit tests: not connected (to GCP) */
```
zloery, line 6: nit: Some of these are GCP functionality unit tests, others are strictly WSM component unit tests. we didn't split those apart when GCP was the only cloud

service/src/test/java/bio/terra/workspace/common/BaseAwsConnectedTest.java
```diff
@@ -8,8 +8,8 @@
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.test.context.ActiveProfiles;
 
-/** Base class for AWS tests. Treat these as connected tests: connected to AWS */
-@Tag("aws")
+/** Base class for AWS connected tests: connected to AWS */
+@Tag("aws-connected")
```
zloery, line 12: I think it's better not to include this tag in `BaseAwsConnectedTest` and continue annotating each test class directly. If we follow the `connectedPlus` pattern for pre- vs. post-submit tests from other clouds, having this here would tag tests with both `aws-connected` and `aws-connectedPlus`

zloery (approved):

service/src/test/java/bio/terra/workspace/common/BaseAwsConnectedTest.java
dexamundsen: Makes sense. changed

service/src/test/java/bio/terra/workspace/common/BaseUnitTest.java
dexamundsen: Yep, hence added gcp within brackets. Will update comment to be more descriptive
splitting GCP tests - thought of it during this PR. https://verily.atlassian.net/browse/TERRA-549

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3storageFolder/ControlledAwsS3StorageFolderHandler.java
```diff
@@ -39,28 +40,28 @@ public WsmResource makeResourceFromDb(DbResource dbResource) {
...
-    // The regular expression only allow legal character combinations containing
+    // The regular expression only allows legal character combinations containing
     // alphanumeric characters and one or more of "!-_.*'()". It trims any other combinations.
-    generatedName = generatedName.replaceAll("\\[", "").replaceAll("\\]", "");
+    generatedName = CharMatcher.anyOf("{}^%`<>~#|@*+[]'\"\\").replaceFrom(generatedName, "");
     generatedName =
         CharMatcher.inRange('0', '9')
             .or(CharMatcher.inRange('A', 'z'))
-            .or(CharMatcher.anyOf("!-_.*'()"))
+            .or(CharMatcher.anyOf("!-_.()"))
```
yuhuyoyo, line 54: why does this part not trim the character in line 50?

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceHandler.java
```diff
@@ -23,10 +25,24 @@ public interface WsmResourceHandler {
...
+  }
+
+  /**
+   * Generate the resource cloud-native name for resource
+   *
+   * @param workspaceUserFacingId workspace UserFacingId
+   * @param resourceName resource name
+   * @return cloud-native name
+   */
+  default String generateCloudName(@NotNull String workspaceUserFacingId, String resourceName) {
```
yuhuyoyo, line 44: rm notnull.

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/model/WsmResourceHandler.java
dexamundsen: handled in https://github.com/DataBiosphere/terra-workspace-manager/pull/1227

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3storageFolder/ControlledAwsS3StorageFolderHandler.java
yuhuyoyo: i mean why this is not enough?

PR #1242, opened 2023-05-03
Move storage.bucket.get permission from bucket editor to workspace reader

rogerwangcs:
To enable the CLI to properly display bucket resource metadata when listed or described, and allow mounting referenced buckets where the user only has read access.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/CustomGcpIamRoleMapping.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/CustomGcpIamRoleMapping.java
@@ -32,7 +32,6 @@ public class CustomGcpIamRoleMapping {
   static final ImmutableList<String> GCS_BUCKET_EDITOR_PERMISSIONS =
       new ImmutableList.Builder<String>()
           .addAll(GCS_BUCKET_WRITER_PERMISSIONS)
-          .add("storage.buckets.get")
           .build();
   static final ImmutableList<String> BIG_QUERY_DATASET_READER_PERMISSIONS =
       ImmutableList.of(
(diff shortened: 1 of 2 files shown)


ddietterich (commented):
I thought you would move the permission to the READER on the **bucket**, not for the whole workspace. Now bucket get will work for any bucket in the workspace - not just the ones the user has permission on.

Is that also necessary for some reason?

rogerwangcs:
On a related note, this change means that GCS_BUCKET_EDITORs have the same permissions as GCS_BUCKET_WRITER. Looking at the other resource permissions, perhaps we had originally intended to give `storage.buckets.update` to bucket editors?

From the gcs [docs](https://cloud.google.com/storage/docs/access-control/iam-permissions#bucket_permissions):
> Update bucket metadata, excluding IAM policies, and add or remove a Pub/Sub notification configuration on a bucket.

zloery:
> On a related note, this change means that GCS_BUCKET_EDITORs have the same permissions as GCS_BUCKET_WRITER. Looking at the other resource permissions, perhaps we had originally intended to give storage.buckets.update to bucket editors?

We intentionally don't give `storage.buckets.update` permission to any user role, as it lets users change metadata that's generally reserved for WSM. We do give `storage.objects.update` (objects vs. buckets) to the bucket writer, but that feels appropriate given that it's modifying bucket contents

zloery (approved):
LGTM, +1 to this being on the bucket reader role and not the workspace reader.

ddietterich (approved):

rogerwangcs:
> > On a related note, this change means that GCS_BUCKET_EDITORs have the same permissions as GCS_BUCKET_WRITER. Looking at the other resource permissions, perhaps we had originally intended to give storage.buckets.update to bucket editors?
> 
> We intentionally don't give `storage.buckets.update` permission to any user role, as it lets users change metadata that's generally reserved for WSM. We do give `storage.objects.update` (objects vs. buckets) to the bucket writer, but that feels appropriate given that it's modifying bucket contents

Oh gotcha, that makes sense.

PR #1245, opened 2023-05-04
[WOR-990] Compute and log a sha256 for Azure SAS signature

aherbst-broad:
## Why
We will need to be able to correlate container logs stored in Azure with SAS creation requests from the Terra control plane. Azure container logs do not directly capture the SAS for security reasons, instead capturing a sha256 of the SAS's signature. MSFT's [guidance](https://learn.microsoft.com/en-us/azure/storage/blobs/blob-storage-monitoring-scenarios#identifying-the-sas-token-used-to-authorize-a-request) is to compute a sha256 when  minting the token and use that for correlation between sources. 

I tested this locally by minting a token, firing an `azcopy list` request using the token and observing the logged sha256 matches the value from the Azure storage container logs.

## This PR
* Extracts the signature from the minted token, URL decodes it and computes a sha256 of the value and then logs it

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
@@ -25,6 +25,8 @@ import com.azure.storage.blob.sas.BlobServiceSasSignatureValues;
 import com.azure.storage.common.StorageSharedKeyCredential;
 import com.azure.storage.common.sas.SasIpRange;
 import com.azure.storage.common.sas.SasProtocol;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -224,12 +227,19 @@ public AzureSasBundle createAzureStorageContainerSasToken(
       resourceName = storageData.storageContainerResource().getStorageContainerName();
     }
 
+    var sig = token.split("sig=")[1];
```
aherbst-broad, line 230: Really wish the azure libs gave us a better object to work with for SAS's than a raw string.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessServiceUnitTest.java
```diff
@@ -431,6 +433,55 @@ public void createAzureStorageContainerSasToken_basedOnLzSharedStorageAccount_Su
...
+                null,
+                OffsetDateTime.parse(
+                    "2023-05-03T09:15:30-05:00", DateTimeFormatter.ISO_OFFSET_DATE_TIME),
+                OffsetDateTime.parse(
+                    "2023-05-03T10:15:30-05:00", DateTimeFormatter.ISO_OFFSET_DATE_TIME),
+                null,
+                null));
+
+    assertEquals(
+        "9FFE137AEB017B2BBF4E9DD724EEC987A4BE60321C0E54B279F8C558031DDE8B", result.sha256());
```
cahrens, line 482: Wow, I found it confusing that it's calling a function on a String. But I guess that it's automatically adding a getter for the parameter… Java has come a long way in 10+ years!

cahrens (approved):

blakery (approved):

PR #1248, opened 2023-05-04
PF-2739:Add state to reference resource cloning

yuhuyoyo:
Reuse the storeMetadataStep for cloneReferenceResourceFlight.

And set state to READY when the referencedResource is successfully created.


service/src/main/java/bio/terra/workspace/service/resource/referenced/flight/clone/CloneReferenceResourceStep.java
ddietterich, line 1: The name of this step is misleading now. It doesn't do any cloning. It just finishes the create process.
Maybe there is away to rename `StoreMetadataStep` and this step so they are paired in some way.
Maybe `CreateResourceMetadataStartStep` and `CreateResourceMetadataFinishStep`?
You can probably think of something better...

ddietterich (approved):
The code looks fine. My only suggestion is to clean up the step naming.

PR #1249, opened 2023-05-04
post-startup.sh: Don't start multiple ssh-agents, don't restart jupyter service for container launches

mbookman:
Fixes:
- [PF-2705](https://broadworkbench.atlassian.net/browse/PF-2705): Creating cloud environment from Docker image fails in post-startup.sh (restarting Jupyter service)
- [PF-2625](https://broadworkbench.atlassian.net/browse/PF-2625): WSM: Avoid creating many ssh-agent instances on cloud environment

[PF-2705]: https://broadworkbench.atlassian.net/browse/PF-2705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PF-2625]: https://broadworkbench.atlassian.net/browse/PF-2625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -47,10 +47,15 @@ set -o xtrace
 # user space, such as setting Terra CLI settings which are persisted in the user's $HOME.
 # This post startup script is not run by the same user.
 readonly JUPYTER_USER="jupyter"
+
 readonly STATUS_ATTRIBUTE="startup_script/status"
 readonly MESSAGE_ATTRIBUTE="startup_script/message"
-readonly USER_TERRA_CONFIG_DIR="/home/${JUPYTER_USER}/.terra"
-readonly OUTPUT_FILE="${USER_TERRA_CONFIG_DIR}/post-startup-output.txt"
+
+readonly USER_HOME_DIR="/home/${JUPYTER_USER}"
+readonly USER_SSH_DIR="${USER_HOME_DIR}/.ssh"
+readonly USER_BASH_PROFILE="${USER_HOME_DIR}/.bash_profile"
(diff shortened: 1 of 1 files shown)


wnojopra (approved):
LGTM. Hooray for custom containers!

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -337,8 +337,33 @@ echo "Setting up git integration..."
 # user starts a shell, we start the ssh-agent.
 cat << 'EOF' >> "${USER_BASH_PROFILE}"
 
-# Start the ssh-agent; "eval" it to set SSH_AUTH_SOCK and SSH_AGENT_PID
-eval "$(ssh-agent -s)"
+# Start a new ssh-agent if one is not already running.
```
cbookg, line 340: LGTM, but should this be moved to the "instance boot service" part so that it is run after reboot as well?

cbookg (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
mbookman: That is an interesting thought, now that we have that service available.

I'd actually be interested in adding it as a standalone service that gets restarted if it crashes.
My experience with ssh-agent across the years is that I periodically find that it has stopped running, for reasons unclear, and I need to restart it myself. Doing this in the` .bash_profile` feels slightly wrong, but it does provide a level of control and simplicity for users.

I'm in the process of trying to figure out how the custom container on Vertex AI Workbench works. The answer to that might impact our long term decision. So I'm inclined to continue our current approach (just improved) and revisit soon,

PR #1255, opened 2023-05-05
[PF-2742] Stop generating names ending in dashes

zloery:
CLI tests relying on name generation are occasionally failing due to invalid names being returned, e.g. `gcsbucketwithoutspecifyingbucketname-terra-wsm-d-genial-orange-` which does not end in an alphanumeric character.


dexamundsen (approved):
Thanks

yuhuyoyo (approved):
ah yes... i knew there was a reason we didn't do truncate in the end but couldn't remember

PR #1257, opened 2023-05-08
[PF-2743] Fix resource name issue; add logging contain assert

ddietterich:
The `ImportDataCollection` test was re-using a resource name. The sequence of checks changes and the name conflict appeared instead of the policy merge conflict. I made two changes:

1. Generated unique resource names in the test
2. Wrote `TestUtils.assertContains` that does the contains test and logs the contents if there is a mismatch. I replaced current usage with calls to that method in several tests.


yuhuyoyo (approved):

mikenev (approved):
lgtm

PR #1258, opened 2023-05-08
[PF-2694] cloud context state

ddietterich:
This PR makes two contributions:

1. It makes cloud context stateful. That is, cloud context operations are mutexed by state and owning flight. They are able to support either delete-on-failure or broken-on-failure error semantics. The state changes are straightforward. They reuse the `StateDao` previously factored out for this purpose.
2. Refactors cloud context and cloud context service. There was a lot of cut'n'paste code in the cloud context services and cloud context flights. The refactor requires more explanation.

Adding state to cloud context means there is now common data across the three cloud contexts. I made these changes:
- added `CloudContext` interface, to allow common code to operate on the different context types
- added `CloudContextCommonFields` to hold the common data for cloud contexts. This follows the pattern used with resources and allows for additional fields to be added without needing to visit every context implementation.

I made two common flights: `CreateCloudContextFlight` and `DeleteCloudContextFlight`. Each has a common start step and end step that handle the context state transition. Each cloud context service generates the cloud-specific steps needed. To make that work I added `CloudContextService` interface, to allow the common flight code to operate on different context types. So all of the cloud-specific services (e.g., `AzureCloudContextService`) implement that interface and provide methods for adding steps to the create and delete flights.

I added a `DbCloudContext` object that is now the way `WorkspaceDao` handles cloud contexts. Each of the cloud-specific services provides a method to create a cloud-specific context object from a `DbCloudContext`. This is similar to how the ResourceDao creates resource-specific objects from the `DbResource` object.

Lastly, permission checking on the spend profile was not being done in the controllers. It turned out that non-privileged users could spend without permission. I moved the spend profile to the controller. As a consequence, I pass the SpendProfile object into the cloud context flights. That required many test changes, since the tests had previously gotten away with passing in a fake id and mocking the permission call.

 I'm happy to point out that adding the state and removing the redundancy resulted in only about +300 LOC.
Sadly, the refactor touched a lot of files.


service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/workspace/WorkspaceDeleteFlight.java
```diff
@@ -77,12 +77,17 @@ public WorkspaceDeleteFlight(FlightMap inputParameters, Object applicationContex
...
 
-    addStep(
-        new DeleteAzureContextStep(appContext.getAzureCloudContextService(), workspaceUuid),
-        cloudRetryRule);
-    addStep(
-        new DeleteAwsContextStep(appContext.getAwsCloudContextService(), workspaceUuid),
-        cloudRetryRule);
+    // This step just does the DeleteCloudFinishStep.
+    // I do not think need this at all, since we are deleting the workspace object.
+    // but I am also thinking we should run these context deletes as separate flights.
```
yuhuyoyo, line 82: so what's the final plan? :)

service/src/main/resources/db/changesets/20230427_workspace_states.yaml
```diff
@@ -0,0 +1,56 @@
+databaseChangeLog:
+- changeSet:
+    id: Add state to resources and cloud contexts
```
zloery, line 3: nit: to workspaces and cloud contexts

service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/workspace/WorkspaceDeleteFlight.java
yuhuyoyo: for the first option, if we don't delete workspace row and leave it broken, then cloudcontext will stuck in deleting i think ..

service/src/main/resources/db/changesets/20230427_workspace_states.yaml
yuhuyoyo, line 3: to workspaces and cloud contexts.

yuhuyoyo (commented):
first pass through.. i wil look at it again tmr

zloery:
> Lastly, permission checking on the spend profile was not being done in the controllers. It turned out that non-privileged users could spend without permission.

Were you able to create a cloud context without permission on the spend profile? My understanding of the old pattern was that we would check as part of the create-context flight which prevented them from actually spending anything. It was possible to create a workspace with any spend profile (whether or not the user had permission/it actually existed) and we wouldn't check until context creation time, though.

service/src/main/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHook.java
```diff
@@ -399,7 +397,7 @@ private void maybeLogGcpContextBackfillFlight(
               subjectId,
               operationType,
               workspaceIdString,
-              ActivityLogChangedTarget.GCP_CLOUD_CONTEXT));
+              ActivityLogChangedTarget.CLOUD_CONTEXT));
```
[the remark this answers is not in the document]
ddietterich: I could put that logic in the hook, but I don't see how to do that in the `ActivityFlight` enum. Moving to a common flight means I cannot distinguish the type there. Any suggestions?

service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/workspace/WorkspaceDeleteFlight.java
ddietterich: Sorry, that is a note to myself. The follow-on PR will change the implementation of workspace delete and I plan to run the cloud context delete as a separate flight. I'll make that a proper TODO.

service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/cloudcontext/DeleteCloudContextFinishStep.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    logger.error(
+        "Cannot undo delete of WSM {} cloud context in workspace {}.",
+        cloudPlatform,
+        workspaceUuid);
+    // Surface whatever error caused Stairway to begin undoing.
+    return flightContext.getResult();
```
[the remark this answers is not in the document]
ddietterich: Yes, that fix it in. I'll revise this.

service/src/main/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHook.java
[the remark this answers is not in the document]
yuhuyoyo: yeah i did that in the hook as well for resource.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -72,14 +78,47 @@ public class WorkspaceDao {
...
+  private static final String BASE_CLOUD_CONTEXT_SELECT_SQL =
+      """
+        SELECT workspace_id, cloud_platform, spend_profile, context, state, flight_id, error
+        FROM cloud_context
+    """;
+
+  /** SQL query for reading a cloud context */
+  private static final String CLOUD_CONTEXT_SELECT_SQL =
+      BASE_CLOUD_CONTEXT_SELECT_SQL
+          + "WHERE workspace_id = :workspace_id AND cloud_platform = :cloud_platform";
```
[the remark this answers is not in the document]
ddietterich: I'll put one in so others won't worry about it, but there are lots of spaces in the `"""` string form.

zloery (approved):

zloery:
> > It was possible to create a workspace with any spend profile...

> You may be right. What I observed was that we had integration tests where the default user did not have link permission, but the test passed. So I'm not sure that we were checking on all paths.
This will be cleaner when the spend profile is only on the cloud context.

Just checked, all integration test configs modified in this PR were tests which created a workspace without a cloud context, so I think everything was WAI there and users couldn't spend without permission on the spend-profile. Agreed it's better to check this earlier in the controller instead though.

yuhuyoyo (approved):

PR #1263, opened 2023-05-10
[PF-2749] Fix test failures from the context state merge

ddietterich:
Several test failures in the "connectedPlus" tests handling GCP failures.

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/workspace/flight/gcp/CreateGcpContextFlightTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/workspace/flight/gcp/CreateGcpContextFlightTest.java
@@ -3,7 +3,7 @@ package bio.terra.workspace.service.workspace.flight.gcp;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import bio.terra.stairway.FlightDebugInfo;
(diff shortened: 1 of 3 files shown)


service/src/test/java/bio/terra/workspace/service/workspace/flight/gcp/CreateGcpContextFlightTest.java
```diff
@@ -168,14 +175,27 @@ void errorRevertsChanges() throws Exception {
     assertTrue(gcpCloudContextService.getGcpCloudContext(workspaceUuid).isEmpty());
 
     // Submit a flight class that always errors.
-    Map<String, StepStatus> retrySteps = getStepNameToStepStatusMap();
+    Map<String, StepStatus> undoRetrySteps = getStepNameToStepStatusMap();
+    // Cause a failure on the penultimate step. Cannot fail after the final step,
```
yuhuyoyo, line 179: i don't understand this comment about cannot fail after the final step. Isn't waitforprojectPermissionsStep that final step of cloudcontext creation?

service/src/test/java/bio/terra/workspace/service/workspace/flight/gcp/DeleteGcpContextFlightTest.java
```diff
@@ -181,22 +181,4 @@ void deleteContextUndo() throws Exception {
     // Because this flight cannot be undone, the context should still be deleted even after undoing.
     assertTrue(gcpCloudContextService.getGcpCloudContext(workspaceUuid).isEmpty());
   }
-
-  @Test
-  void deleteNonExistentContextIsOk() throws Exception {
```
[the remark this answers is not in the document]
ddietterich: Because I couldn't think of a reason why we should hide the not-found error. It was not hidden for the Azure case.
Do you know a reason we should hide this from callers?

service/src/test/java/bio/terra/workspace/service/workspace/flight/gcp/CreateGcpContextFlightTest.java
ddietterich: The final step is the common CreateContextFinishStep. Once the state change is made to READY, it is not safe to undo it, so the flight will generate a dismal failure

zloery (approved):

PR #1264, opened 2023-05-11
[TERRA-312] Fix case in sageMaker name

dexamundsen:
change the case from Sagemaker to SageMaker, keeping inline with AWS usage
No other changes


service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
```diff
@@ -597,15 +597,15 @@ private static void deleteS3Objects(
...
   }
 
-  // AWS Sagemaker Notebook
+  // AWS SageMaker Notebook
 
   // TODO(TERRA-500) Move notebook functions below to CRL
 
   /**
-   * Create a AWS sagemaker notebook
+   * Create a AWS sageMaker notebook
```
zloery, line 605: As a proper noun In comments/error strings/text I think this should be `SageMaker`, not `sageMaker`.

IMO `sageMaker___` is fine as a var name though

zloery (approved):

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
dexamundsen: Agree, this was missed due to a project wide find and replace

jmczerk (approved):
FYI @rogerwangcs @cbookg in case you're already using `Sagemaker` references against current client

PR #1265, opened 2023-05-11
[PF-2762] Restore activity log targets; backfill cloud context spend profile

ddietterich:


service/src/main/java/bio/terra/workspace/common/logging/model/ActivityLogChangedTarget.java
```diff
@@ -31,7 +31,10 @@ public enum ActivityLogChangedTarget {
   USER,
   WORKSPACE,
   POLICIES,
-  CLOUD_CONTEXT;
+  CLOUD_CONTEXT,
+  GCP_CLOUD_CONTEXT,
```
yuhuyoyo, line 35: can you log them in the activity log hook?

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/common/logging/model/ActivityLogChangedTarget.java
ddietterich: I made https://broadworkbench.atlassian.net/browse/PF-2764 to track restoring the activity log detail.
I wrote two options in there. I'd like your input.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -782,4 +782,19 @@ public List<WorkspaceUserPair> getPrivateResourceUsers() {
 
     return jdbcTemplate.query(sql, params, WORKSPACE_USER_PAIR_ROW_MAPPER);
   }
+
+  /**
+   * Temporary backfill of cloud context spend profile from workspace spend profile TODO: PF-2762
+   * remove this backfill.
+   */
+  @WriteTransaction
+  public void backfillCloudContextSpendProfile() {
```
zloery, line 791: fwiw I think you could also do this in another liquibase changeset, but either way seems fine

zloery (approved):

PR #1267, opened 2023-05-12
[PF-2765] Use UUID for snapshot id in test

ddietterich:
Snapshot ids are UUIDs and we are now using them for policy. (TPS requires UUID for an object id)
The `ReferencedGcpResourceControllerDataRepoSnapshotTest` was using a name for the snapshot and that failed.
This PR is a one line change to use a generated UUID instead of a string for the snapshot id.


mikenev (approved):
makes sense

yuhuyoyo (approved):

PR #1269, opened 2023-05-12
Minor fixes to Sagemaker creation

dexamundsen:
- AWS sagemaker should use the Workspace.userfacingid and not workspace.uuid
- Add Nullable annotation to generateCloudName(String,String) function, similar to generateCloudName(Uuid,String)
- update resource.validate() method to only allow private notebooks
- leverage validationUtils for resource.validate steps


zloery (approved):

zloery:
One other issue I noticed: I think you can currently created shared SageMaker notebooks. My understanding was that these should be like AI Platform notebooks, where we require them to be private resources (due to credentials on the machine) and enforce that in the resource's `validate` method

dexamundsen:
> One other issue I noticed: I think you can currently created shared SageMaker notebooks. My understanding was that these should be like AI Platform notebooks, where we require them to be private resources (due to credentials on the machine) and enforce that in the resource's `validate` method

Done. Updated

zloery (approved):

PR #1271, opened 2023-05-12
[PF-2768] Handle deserializing placeholder gcp contexts

zloery:
Currently, calling `createCloudContextStart` puts a placeholder value in the cloud_context DB. This is important for state management, but it does not have any GCP context information (policy groups, project ID) and so deserialization on this placeholder object will fail. IMO we should hide this placeholder as it will not have the fields most callers expect.

This is currently calling `getWorkspace` calls immediately after `createGcpCloudContext` calls to fail with a 400. After the context creation finishes, the `getWorkspace` calls start working again.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
@@ -4,6 +4,7 @@ import bio.terra.workspace.common.exception.InternalLogicException;
 import bio.terra.workspace.db.DbSerDes;
 import bio.terra.workspace.db.model.DbCloudContext;
 import bio.terra.workspace.generated.model.ApiGcpContext;
+import bio.terra.workspace.service.resource.model.WsmResourceState;
 import bio.terra.workspace.service.workspace.exceptions.InvalidSerializedVersionException;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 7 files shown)


service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
```diff
@@ -80,14 +81,29 @@ public String serialize() {
...
+   */
+  public static Optional<GcpCloudContext> deserialize(DbCloudContext dbCloudContext) {
     // Sanity check the input
     if (dbCloudContext == null
         || dbCloudContext.getSpendProfile() == null
         || dbCloudContext.getCloudPlatform() != CloudPlatform.GCP) {
       throw new InternalLogicException("Invalid GCP cloud context state");
     }
 
+    if (dbCloudContext.getState() == WsmResourceState.CREATING) {
```
yuhuyoyo, line 103: or  broken?

service/src/test/java/bio/terra/workspace/service/workspace/GcpCloudContextUnitTest.java
```diff
@@ -92,6 +93,37 @@ void serdesTest() {
...
+            "a-user-facing-id",
+            "deleteGcpContextDeletesControlledResources",
+            "description",
+            new SpendProfileId("spend-profile"),
+            Collections.emptyMap(),
+            WorkspaceStage.MC_WORKSPACE,
+            DEFAULT_USER_EMAIL,
+            null);
+    workspaceDao.createWorkspace(workspace, /* applicationIds= */ null);
+    workspaceDao.createCloudContextStart(
```
yuhuyoyo, line 111: nit: line break between setup, test, assert

yuhuyoyo (approved):

cbookg (approved):

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
ddietterich: BROKEN also probably doesn't have the fields filled in. Right now, no one is using BROKEN, so it probably doesn't matter.

ddietterich (approved):
Fine for now. I will have to make a bigger fix in order to get the target behavior.

ddietterich:
I made [PF-2771](https://broadworkbench.atlassian.net/browse/PF-2771) to track the proper fix

PR #1272, opened 2023-05-15
[WOR-1013] Check for hardcoded spend profile first

aherbst-broad:
Integration tests expect a spend profile named "wm-default-spend-profile" to be present. Now that we have the BPM flag enabled in all environments, this check fails as BPM does not know about that profile. This PR checks the configured map of hardcoded spend profiles prior to reaching out to BPM.


ddietterich (changes requested):
Need one more Sam check

ddietterich (approved):

marctalbott (approved):

PR #1275, opened 2023-05-17
[PF-2771] Allow cloud context retrieval in intermediate states

ddietterich:
This change rearranges the cloud context classes by separating the "cloud context fields" into a separate class. That class looks a lot like the old cloud context classes; it knows how to serialize into the database format.

All of the cloud context classes look similar. They have a `commonFields` member that holds the `CloudContextCommonFields` (spendProfile, state, flight, and error). That is always present.

They have a `contextFields` member that holds the cloud-specific data; e.g., `AzureCloudContextFields`.

There are other checks to be made for concurrency, but those will come in PF-2770 and PF-2788.

-------------
On Zach's prompting I reviewed all uses of `getRequiredCloudContext`, `getCloudContext`, `getRequiredGcpProject` and `getGcpProject`. I added two new methods: `getRequiredReadyCloudContext` and `getRequiredReadyGcpProject` that enforce READY state.

In updating the code, I found that notebooks were doing many projectId lookups. The projectId is a member of that resource, so there was no reason for those lookups. I removed them in favor of direct reference.

I would appreciate a review of the last two commits. The next-to-last one is the bulk of the changes. The last one is the merge fixup from Yu's load work.


service/src/main/java/bio/terra/workspace/service/workspace/model/AwsCloudContext.java
```diff
@@ -1,72 +1,69 @@
...
-      @JsonProperty("tenantAlias") String tenantAlias,
-      @JsonProperty("environmentAlias") String environmentAlias,
+      @JsonProperty("contextFields") @Nullable AwsCloudContextFields contextFields,
       @JsonProperty("commonFields") CloudContextCommonFields commonFields) {
-    this.majorVersion = majorVersion;
-    this.organizationId = organizationId;
-    this.accountId = accountId;
-    this.tenantAlias = tenantAlias;
-    this.environmentAlias = environmentAlias;
+    this.contextFields = contextFields;
```
zloery, line 23: Just to confirm: the AwsCloudContext objects were previously serialized in flights but never in the DB?
I think the change to fields here will cause issues with any in-progress flights, but given the infrequency and how fast those flights are, I think that should be safe enough

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
```diff
@@ -1,61 +1,57 @@
...
-      @JsonProperty("samPolicyReader") String samPolicyReader,
-      @JsonProperty("samPolicyApplication") String samPolicyApplication,
-      @JsonProperty("commonFields") @Nullable CloudContextCommonFields commonFields) {
-    this.gcpProjectId = gcpProjectId;
-    this.samPolicyOwner = samPolicyOwner;
-    this.samPolicyWriter = samPolicyWriter;
-    this.samPolicyReader = samPolicyReader;
-    this.samPolicyApplication = samPolicyApplication;
+      @JsonProperty("contextFields") @Nullable GcpCloudContextFields contextFields,
+      @JsonProperty("commonFields") CloudContextCommonFields commonFields) {
```
zloery, line 18: IIUC this will break in-progress "createGcpContext" flights. Those can sometimes run longer (10-15m if DRS is grumpy), so we should probably be aware of this when deploying to prod

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
```diff
@@ -160,30 +156,27 @@ public CloudContext makeCloudContextFromDb(DbCloudContext dbCloudContext) {
...
-      UUID workspaceUuid, AuthenticatedUserRequest userRequest) throws InterruptedException {
-    return getGcpCloudContext(workspaceUuid)
-        .orElseThrow(
-            () -> new CloudContextRequiredException("Operation requires GCP cloud context"));
+  public GcpCloudContext getRequiredGcpCloudContext(UUID workspaceUuid) {
+    GcpCloudContext cloudContext =
+        getGcpCloudContext(workspaceUuid)
+            .orElseThrow(
+                () -> new CloudContextRequiredException("Operation requires GCP cloud context"));
+    return cloudContext;
```
zloery, line 173: I think this step should also validate that the context is in the READY state. As-is, I think this won't throw an exception here, but callers will expect the context fields to be populated per the javadoc.

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
```diff
@@ -1,61 +1,57 @@
...
 
+  @Nullable
+  public GcpCloudContextFields getContextFields() {
+    return contextFields;
+  }
+
+  @JsonIgnore
   public String getGcpProjectId() {
-    return gcpProjectId;
+    return contextFields.getGcpProjectId();
```
zloery, line 30: contextFields is nullable, this needs NPE handling

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
```diff
@@ -160,30 +156,27 @@ public CloudContext makeCloudContextFromDb(DbCloudContext dbCloudContext) {
...
 
   /**
    * Helper method for looking up the GCP project ID for a given workspace ID, if one exists. Unlike
-   * {@link #getRequiredGcpProject(UUID)}, this returns an empty Optional instead of throwing if the
-   * given workspace does not have a GCP cloud context. NOTE: no user auth validation
+   * {@link #getRequiredGcpProject}, this returns an empty Optional instead of throwing if the given
+   * workspace does not have a GCP cloud context.
    *
    * @param workspaceUuid workspace identifier of the cloud context
    * @return optional GCP project from the cloud context
```
zloery, line 182: If you call this method on a `CREATING` context, it will return an `Optional<String>` with value `null` instead of an empty Optional, which seems misleading to callers

zloery (commented):
For GCP contexts in particular, I think there's one more place that won't handle a `GcpCloudContext` object without a project ID:
- `PetSaService.getWorkspacePetCredentials`

I think those need to be updated to check that both 1. the cloud context exists, and 2. the cloud context is ready.

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
zloery: (this comment is for `Optional<String> getGcpProject(UUID workspaceUuid)`, which Github wouldn't let me comment on directly since it's unchanged)

service/src/main/java/bio/terra/workspace/service/workspace/model/AwsCloudContextFields.java
```diff
@@ -0,0 +1,166 @@
...
+          dbContext.accountId,
+          dbContext.tenantAlias,
+          dbContext.environmentAlias);
+
+    } catch (SerializationException e) {
+      throw new InvalidSerializedVersionException("Invalid serialized version", e);
+    }
+  }
+
+  /** Mask AWS version numbers so as not to collide with Azure and GCP version numbers */
```
yuhuyoyo, line 120: i don't understand the purpose of this mask..but i see you are just moving files around.

service/src/main/java/bio/terra/workspace/service/workspace/model/AzureCloudContextFields.java
```diff
@@ -0,0 +1,106 @@
...
+  @VisibleForTesting
+  public static class AzureCloudContextV100 {
+    /**
+     * Format version for serialized form of Azure cloud context - deliberately chosen to be
+     * different from the GCP cloud context version.
+     */
+    public static final long AZURE_CLOUD_CONTEXT_DB_VERSION = 100;
+
+    /** Version marker to store in the db so that we can update the format later if we need to. */
+    @JsonProperty public final long version = AZURE_CLOUD_CONTEXT_DB_VERSION;
```
yuhuyoyo, line 91: this is not used, right?

also can we also do the mask thing? confusing that this is v100

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/workspace/model/AwsCloudContext.java
ddietterich: I think they are serialized in the DB. It should be OK. The serialized form is now made and deserialized in `AwsCloudContextFields` and should be identical with what used to be in `AwsCloudContext`. That is true of all of the context objects.

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
ddietterich: Good  point. These changes will break things in flight.
We should make a plan. I am thinking that this will go into preprod next Tuesday and then go into prod on Monday, June 5. Maybe it would make sense for me to do the preprod -> prod upgrade on Sunday evening (June 4).
wdyt?

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
ddietterich: I did that for all context types, assuming that if we were requiring the context, we would also want it ready. Things broke all over the place. Turns out we "get required" in the middle of many operations, and it would find the state it was operating on (updating, deleting). I added PF-2788 to track adding operation checks on resource state. I could add it to this, but it was already a big change set...

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
ddietterich: I considered that, but how would I handle it? I thought an NPE would be a good way to know that there was a case where we were doing "get required" and things were not in the right state.

service/src/main/java/bio/terra/workspace/service/workspace/GcpCloudContextService.java
zloery: I didn't read the doc on `Optional.map` carefully enough, this comment is wrong

service/src/main/java/bio/terra/workspace/service/workspace/model/AwsCloudContextFields.java
ddietterich: This is equivalent to saying the version is 257. I don't know why it was done this way.

service/src/main/java/bio/terra/workspace/service/workspace/model/AzureCloudContextFields.java
ddietterich: It is used and it works, even though it looks odd. The object is initialized (set to 100) before JSON is deserialized into it, so whatever is in the JSON will clobber the version.

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/gcp/DeleteGcpProjectStep.java
```diff
@@ -60,18 +61,17 @@ public StepResult undoStep(FlightContext flightContext) throws InterruptedExcept
       return StepResult.getStepResultSuccess();
     }
     // Do not attempt to undo project deletions.
-    // TODO: investigate recovering projects if this happens often and recovery works as a "undo."
-    String projectId = cloudContext.get().getGcpProjectId();
-    logger.error("Unable to undo deletion of project [{}]", projectId);
+    logger.error("Unable to undo deletion of project");
     return flightContext.getResult();
```
zloery, line 65: With the stairway fix merged, this can just return a FATAL step result with the error instead of `flightContext.getResult()`

zloery (approved):

service/src/main/java/bio/terra/workspace/service/workspace/model/AwsCloudContext.java
zloery: Did the serialization store any class information? That would cause problems if we did, but if it's just by fields we should be fine

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
zloery: I'd either lean towards an off-hours release or just verifying there aren't any active flights in prod when we start the release

service/src/main/java/bio/terra/workspace/service/workspace/model/GcpCloudContext.java
zloery: I think it would be better to throw a more informative error (like an `InvalidCloudContextStateException`) instead, especially if this gets surfaced to callers. Doing that with other cases in PF-2788 seems reasonable though

zloery: Checking for null and throwing a `InvalidCloudContextStateException` would be a bit neater, but I'm a lot less concerned about this with most callers using `getRequiredReadyGcpProject` now. At this point hitting that would probably indicate a path where WSM should be calling `getRequiredReady` but isn't

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/workspace/model/AwsCloudContext.java
ddietterich: The database serializer does not store class info.

PR #1276, opened 2023-05-17
TERRA-575: Attach correct role to SageMaker notebook.

jmczerk:
The wrong role is attached to SageMaker notebook instances, so they do not have permission to log or access tags.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
@@ -84,7 +84,7 @@ public final class WorkspaceFlightMapKeys {
 
     // AWS
     public static final String AWS_CLOUD_CONTEXT = "awsCloudContext";
-    public static final String AWS_ENVIRONMENT_USER_ROLE_ARN = "awsEnvironmentUserRoleArn";
+    public static final String AWS_ENVIRONMENT_NOTEBOOK_ROLE_ARN = "awsEnvironmentUserRoleArn";
     public static final String AWS_LANDING_ZONE_KMS_KEY_ARN = "awsLandingZoneKmsKeyArn";
     public static final String AWS_LANDING_ZONE_NOTEBOOK_LIFECYCLE_CONFIG_ARN =
         "awsLandingZoneNotebookLifecycleConfigArn";
--- a/service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
@@ -84,7 +84,8 @@ public final class WorkspaceFlightMapKeys {
 
     // AWS
     public static final String AWS_CLOUD_CONTEXT = "awsCloudContext";
-    public static final String AWS_ENVIRONMENT_NOTEBOOK_ROLE_ARN = "awsEnvironmentUserRoleArn";
+    public static final String AWS_ENVIRONMENT_USER_ROLE_ARN = "awsEnvironmentUserRoleArn";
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
```diff
@@ -84,7 +84,7 @@ private ControlledResourceKeys() {}
 
     // AWS
     public static final String AWS_CLOUD_CONTEXT = "awsCloudContext";
-    public static final String AWS_ENVIRONMENT_USER_ROLE_ARN = "awsEnvironmentUserRoleArn";
```
jmczerk, line 87: Not changing the key text here, since this is persisted in the database.

dexamundsen: How about keeping both? Not many notebooks are used now, and using the correct key would help later

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
```diff
@@ -84,7 +84,7 @@ private ControlledResourceKeys() {}
...
+    public static final String AWS_ENVIRONMENT_NOTEBOOK_ROLE_ARN = "awsEnvironmentUserRoleArn";
```
dexamundsen, line 87:
```suggestion
    public static final String AWS_ENVIRONMENT_USER_ROLE_ARN = "awsEnvironmentUserRoleArn";
    public static final String AWS_ENVIRONMENT_NOTEBOOK_ROLE_ARN = "awsEnvironmentNotebookRoleArn";
```

jmczerk: I don't really know the chances of hitting some startup time deserialization error that ends up triggering a manual cleanup.  Will that only happen if a notebook is in the process of being created when we're deploying?  We know notebook creations are long running, and I'd like to de-risk this.

@ddietterich any thoughts on risk?

zloery (approved):

service/src/main/java/bio/terra/workspace/service/workspace/flight/WorkspaceFlightMapKeys.java
dexamundsen: These keys are only used during the flight from what i see. whatever notebooks we have now in devel - i can clean those workspaces and/or notebooks if needed.

jmczerk: Not sure how relevant this is, but ...

```sql
SELECT status, COUNT(status) 
FROM flight 
WHERE flightid IN (
    SELECT flightid 
    FROM flightinput 
    WHERE key = 'awsEnvironmentUserRoleArn'
) 
GROUP BY status;
```
Result:
```
 status  | count 
---------+-------
 ERROR   |    11
 FATAL   |     3
 SUCCESS |    45
(3 rows)
```

ddietterich: At this time, I am not being careful about changing flight map keys and parameters. My judgment is that the complexity of handling the case is high and the risk is very low right now. The risk is catching a running flight during a rolling upgrade and hitting a serdes failure. I have yet to see an instance of that failure.

I am trying to be careful about changing serialized metadata in the WSM database. (Although given last week's debacle, not quite careful enough).

jmczerk: Thanks Dan.

Dan and I talked offline; only flights using this key with `status` of `RUNNING` would trigger deserialization issues; the rest in the query above are kept for a period for debugging/forensics and would only be browsed in the database directly.  So it sounds like it's safe enough to do what @dexamundsen is proposing and babysit the deployment to `devel`.  I'll accept the suggested change.

PR #1279, opened 2023-05-17
[PF-2750] workspace state - take 2

ddietterich:
This adds a backfill setting any null workspace state values to READY.
I considered doing it in liquibase, but that is harder to test. So I did it in Java and made a test.

You can just review the penultimate commit to see the changes.


service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceStateBackfillTest.java
```diff
@@ -0,0 +1,42 @@
...
+    jdbcTemplate.update(
+        "UPDATE workspace SET state = NULL", new MapSqlParameterSource());
+
+    DbWorkspace dbWorkspace = workspaceDao.getDbWorkspace(workspaceId);
+    Assertions.assertNull(dbWorkspace.getState());
+
+    workspaceDao.backfillWorkspaceState();
+
+    dbWorkspace = workspaceDao.getDbWorkspace(workspaceId);
+    Assertions.assertNotNull(dbWorkspace.getState());
```
yuhuyoyo, line 33: rm, redundant

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -916,8 +919,9 @@ public List<WorkspaceUserPair> getPrivateResourceUsers() {
   }
 
   /**
-   * Temporary backfill of cloud context spend profile from workspace spend profile TODO: PF-2762
-   * remove this backfill.
+   * Temporary backfill of cloud context spend profile from workspace spend profile
+   * <p>
+   * TODO: PF-2782 remove this backfill.
```
zloery: this comment got changed, this one should be PF-2762

zloery (approved):

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
ddietterich: It really should have been PF-2763, since PF-2762 was the original work.

Anyway, I think I have all of the TODO set with the right tickets!

PR #1280, opened 2023-05-18
post-startup: Support JupyterLab provided by Docker containers

mbookman:
This PR is intended to set up the `post-startup.sh` script to support Vertex AI Workbench VMs with two different user models:

- JupyterLab provided natively on the VM
- JupyterLab provided by a Docker image (https://cloud.google.com/vertex-ai/docs/workbench/user-managed/custom-container)

To date, the `post-startup.sh` script has supported the first case well, but not the second. Key issues are tracked in https://broadworkbench.atlassian.net/browse/PF-2707.

The primary changes here are:

- All installs go to `/home/jupyter`, in particular software to `/home/jupyter/.local/bin'
  - Always install Java 17 (on-instance has Java 11) into `/home/jupyter/.local` 
- The user PATH is updated to prepend '/home/jupyter/.local/bin'
- The `ssh-agent` is managed as a daemon service with the SSH socket and environment variables written to `/home/jupyter/.ssh-agent`

Some changes here specifically for the case where JupyterLab is provided by a Docker container:

- Copy `ssh`, `ssh-add`, and `less` to `/home/jupyter/.local/bin`
- Update the JupyterLab environment inside the container image by modifying the on-host `/opt/deeplearning/jupyter/jupyter_notebook_config.py`.

The latter change is implemented via reverse engineering the way that JupyterLab is launched in the Deep Learning Docker images. This works for now, and we can consider some alternative approaches should the need arise.

I addition the use of `sudo -u jupyter -c "<command>"` is standardized with a simple shell variable:

```
# Create an alias for cases when we need to run a shell command as the jupyter user.
# Note that we deliberately use "bash -l" instead of "sh" in order to get bash (instead of dash)
# and to pick up changes to the .bash_profile.
#
# This is intentionally not a Bash function, as that can suppress error propagation.
# This is intentionally not a Bash alias as they are not supported in shell scripts.
readonly RUN_AS_JUPYTER_USER="sudo -u "${JUPYTER_USER}" bash -l -c"
```


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -48,33 +48,63 @@ set -o xtrace
 # This post startup script is not run by the same user.
 readonly JUPYTER_USER="jupyter"
 
+# Create an alias for cases when we need to run a shell command as the jupyter user.
+# Note that we deliberately use "bash -l" instead of "sh" in order to get bash (instead of dash)
+# and to pick up changes to the .bash_profile.
+#
+# This is intentionally not a Bash function, as that can suppress error propagation.
+# This is intentionally not a Bash alias as they are not supported in shell scripts.
+readonly RUN_AS_JUPYTER_USER="sudo -u "${JUPYTER_USER}" bash -l -c"
```
amygdala, line 57: I think this static code analysis is new-- how cool that it's catching easy-to-make (and miss) errors like this.

wnojopra (approved):
Nice. I couldn't find much to comment on, but I did test starting up a new cloud environment with these changes. No issues with that test.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -48,33 +48,63 @@ set -o xtrace
...
+readonly USER_HOME_LOCAL_SHARE="${USER_HOME_DIR}/.local/share"
+readonly USER_TERRA_CONFIG_DIR="${USER_HOME_DIR}/.terra"
 readonly USER_SSH_DIR="${USER_HOME_DIR}/.ssh"
+
 readonly USER_BASH_PROFILE="${USER_HOME_DIR}/.bash_profile"
-readonly USER_TERRA_CONFIG_DIR="${USER_HOME_DIR}/.terra"
 readonly POST_STARTUP_OUTPUT_FILE="${USER_TERRA_CONFIG_DIR}/post-startup-output.txt"
 
+# When JupyterLab is provided by a Docker container, the default Deep Learning images
+# pick up jupyter_notebook_config.py provided by the host VM.
```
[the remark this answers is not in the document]
mbookman: Done

rogerwangcs (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -143,98 +175,170 @@ trap 'exit_handler $? $LINENO $BASH_COMMAND' EXIT
...
-/opt/conda/bin/pip install \
+${RUN_AS_JUPYTER_USER} "pip install --user \
   dsub \
   nbdime \
   nbstripout \
   pandas_gbq \
   pre-commit \
   pylint \
-  pytest
+  pytest"
```
johntbates, line 233: any chance we could install plotnine here, too?  [PF-2798](https://broadworkbench.atlassian.net/jira/software/c/projects/PF/issues/PF-2798)

PR #1281, opened 2023-05-18
PF-2774: Load url list to gcs bucket

yuhuyoyo:


yuhuyoyo:
@harrismendell  FYI

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/SetBucketRolesStep.java
```diff
@@ -4,93 +4,54 @@
...
+    FlightMap workingMap = flightContext.getWorkingMap();
+    // Validate the outputs, so we fail fast if one goes missing.
     FlightUtils.validateRequiredEntries(
-        flightContext.getInputParameters(), ControlledResourceKeys.DESTINATION_WORKSPACE_ID);
-
-    // Gather bucket inputs and store them in the working map for the next
-    // step.
-    final BucketCloneInputs sourceInputs = getSourceInputs();
-    workingMap.put(ControlledResourceKeys.SOURCE_CLONE_INPUTS, sourceInputs);
+        workingMap,
```
ddietterich, line 36: Instead of doing the validation,  I prefer retrieving data like:
```
var storageTransferServiceSAEmail =
  FlightUtils.getRequired(workingMap, ControlledResourceKeys.STORAGE_TRANSFER_SERVICE_SA_EMAIL, String.class);
```
The advantage is, then IntelliJ and Sonar know that storageTransferServiceSAEmail cannot be null, so they don't complain.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/bucket/TransferSignedUrlsToGcsBucketStep.java
```diff
@@ -0,0 +1,109 @@
...
+
+    String signedUrlList =
+        context.getInputParameters().get(ControlledResourceKeys.SIGNED_URL_LIST, String.class);
+
+    var destinationBucketName =
+        context
+            .getInputParameters()
+            .get(ControlledResourceKeys.DESTINATION_BUCKET_NAME_FOR_SIGNED_URL_LIST, String.class);
+
+    // Look up the transfer job by name. If it's found, it means we are restarting this step and
```
ddietterich, line 59: I think this is a common code fragment. Not sure if it is reasonable, but maybe it could be shared in utils?

service/src/test/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiControllerGcsBucketTest.java
```diff
@@ -274,6 +282,46 @@ public void create() throws Exception {
...
+        sourceBucketName,
+        "raw.githubusercontent.com/yuhuyoyo/test-repo/main/helloworld2.txt",
+        "hello world again!\n");
+  }
+
+  @Test
+  public void loadSignedUrlList_403() throws Exception {
+    AuthenticatedUserRequest thirdUserAuthRequest = userAccessUtils.thirdUserAuthRequest();
+
+    // Third user has no access.
```
ddietterich, line 312: to the target bucket?

service/src/test/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiControllerGcsBucketTest.java
```diff
@@ -100,6 +105,9 @@ public class ControlledGcpResourceApiControllerGcsBucketTest extends BaseConnect
               LifecycleAction.newDeleteAction(),
               LifecycleCondition.newBuilder().setAge(3).build()));
 
+  private static final String URL_LIST =
+      "https://raw.githubusercontent.com/yuhuyoyo/test-repo/main/helloworld.tsv";
```
ddietterich, line 109: I think we will need these to be in a different place.

I also think we need some testing with real signed URLs. These are publicly available.

ddietterich (commented):
Nice and tidy. A few comments.

I think we have to make some tests with real signed URLs. The docs imply that signed URLs will not work. I don't know if we have verified that. The current tests do not verify it.

openapi/src/parts/alpha1.yaml
```diff
@@ -37,6 +83,25 @@ paths:
...
   schemas:
+    LoadSignedUrlListRequestBody:
+      description: |
+        Request to load signed urls files to a gcs bucket.
+      type: object
+      properties:
+        jobControl:
+          $ref: '#/components/schemas/JobControl'
+        urlList:
+          description: url (http or https only) to the tsv file that contains the list of files.
```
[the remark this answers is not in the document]
yuhuyoyo: oh.. i didn't realize that's not what signed urls are :) how are we suppose to access non public urls?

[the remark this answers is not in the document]
ddietterich: I am hoping that the documentation is wrong there and that with a signed URL, it will work.
The signed URL encodes some kind of token, valid for a limited time, that allows anyone holding the URL to access the file. I don't know the specifics on GCP.

Here's a link
https://cloud.google.com/storage/docs/access-control/signed-urls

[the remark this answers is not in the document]
cbookg: Well signed URLs are public URLs, so the manifest would need to contain a list of signed URLs (or otherwise public urls).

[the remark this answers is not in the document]
cbookg: but +1 to something like manifest url.

openapi/src/parts/alpha1.yaml
```diff
@@ -4,6 +4,52 @@
 ## may change without notice.
 
 paths:
+  /api/workspaces/alpha1/{workspaceId}/resources/controlled/gcp/buckets/{resourceId}/load:
+    parameters:
+      - $ref: '#/components/parameters/WorkspaceId'
+      - $ref: '#/components/parameters/ResourceId'
+    post:
+      operationId: loadSignedUrlList
```
cbookg, line 12: i suggest we avoid "signed" in the naming of these since these are really just urls.  they don't have to be signed.  they just have to be accessible.

ddietterich (approved):
Looks good to me

cbookg (approved):

PR #1284, opened 2023-05-19
[PF-2787] Fix connectedPlus tests to be happy with workspace state

ddietterich:
Fixes fir three connectedPlus tests:

1. Disabled `cloneGcpWorkspaceUndoSteps`. It is not testing what it should and we have a test case for create workspace undo, so it is of no use right now.
2. `createMcWorkspaceUndoSteps` is properly triggering the undo now
3. Retry of a step without a retry rule doesn't retry

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/workspace/GcpCloudContextConnectedTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/workspace/GcpCloudContextConnectedTest.java
@@ -60,6 +60,7 @@ import java.util.Map;
 import java.util.UUID;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledIfEnvironmentVariable;
@@ -266,6 +267,9 @@ class GcpCloudContextConnectedTest extends BaseConnectedTest {
   }
 
   @Test
+  @Disabled("PF-2259 - this test needs rewriting")
+  // TODO(PF-2259): This test is not testing the undo of CloneGcpWorkspaceFlight. It is
+  //  testing the undo of WorkspaceCreateFlight. It needs to be re
   public void cloneGcpWorkspaceUndoSteps() {
     Workspace sourceWorkspace = workspaceService.getWorkspace(workspaceId);
     // Enable an application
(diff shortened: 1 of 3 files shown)


service/src/test/java/bio/terra/workspace/service/workspace/GcpCloudContextConnectedTest.java
```diff
@@ -266,6 +267,9 @@ public void cloneGcpWorkspace() throws Exception {
   }
 
   @Test
+  @Disabled("PF-2259 - this test needs rewriting")
+  // TODO(PF-2259): This test is not testing the undo of CloneGcpWorkspaceFlight. It is
+  //  testing the undo of WorkspaceCreateFlight. It needs to be re
```
yuhuyoyo, line 272: it needs to be re?

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
```diff
@@ -543,9 +543,18 @@ void createMcWorkspaceUndoSteps() {
     retrySteps.put(CreateWorkspaceStartStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_RETRY);
     retrySteps.put(
         CreateWorkspacePoliciesStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_RETRY);
-    retrySteps.put(CreateWorkspaceAuthzStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_FATAL);
+    retrySteps.put(CreateWorkspaceAuthzStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_RETRY);
```
yuhuyoyo, line 546: what's the difference of this vs just setting fatal here?

yuhuyoyo (approved):

service/src/test/java/bio/terra/workspace/service/workspace/GcpCloudContextConnectedTest.java
ddietterich: Don't you think it needs to be re? :)

service/src/test/java/bio/terra/workspace/service/workspace/WorkspaceServiceTest.java
ddietterich: The FATAL needs to happen on the DO step, so that the flight starts undoing. Once it is undoing, then each of the steps will retry their undo method.

PR #1286, opened 2023-05-19
WOR-877 enforce protected data policy creating azure context

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-877

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/WorkspaceCreateFlight.java
+++ b/service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/WorkspaceCreateFlight.java
@@ -10,14 +10,11 @@ import bio.terra.workspace.common.utils.FlightUtils;
 import bio.terra.workspace.common.utils.RetryRules;
 import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.job.JobMapKeys;
-import bio.terra.workspace.service.policy.flight.MergePolicyAttributesStep;
-import bio.terra.workspace.service.resource.model.CloningInstructions;
 import bio.terra.workspace.service.resource.model.WsmResourceStateRule;
 import bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys;
 import bio.terra.workspace.service.workspace.model.Workspace;
 import com.fasterxml.jackson.core.type.TypeReference;
 import java.util.List;
-import java.util.UUID;
 
 public class WorkspaceCreateFlight extends Flight {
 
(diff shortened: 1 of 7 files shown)


service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/WorkspaceCreateFlight.java
```diff
@@ -62,15 +51,6 @@ public WorkspaceCreateFlight(FlightMap inputParameters, Object applicationContex
...
               serviceRetryRule);
-          if (cloningInstructions != CloningInstructions.COPY_NOTHING) {
-            addStep(
-                new MergePolicyAttributesStep(
-                    sourceWorkspaceUuid,
-                    workspace.workspaceId(),
-                    cloningInstructions,
-                    appContext.getTpsApiDispatch()),
-                serviceRetryRule);
-          }
```
dvoet, line 73: This is the workspace creation flight, it does not make sense to deal with cloning operations here. Leave that to the clone workspace flight.

@ddietterich let me know if there is a reason this was here. The clone flight also has it but I had to fix the cloning instructions (see above).

ddietterich: The create workspace flight is used by clone workspace to create the target workspace.
Please do not delete that.

aherbst-broad (approved):

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/WorkspaceCreateFlight.java
ddietterich: Ah, I see. There is a problem, but it is that the `CloneWorkspaceFlight` is doing the merge. I think the one [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/939ab7cd4c6374cdcfe212a942241ff6d8151266/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneWorkspaceFlight.java#L67) should be removed.

The step in `WorkspaceCreateFlight` is properly choosing to link or merge depending on the inputs. The one is `CloneWorkspaceFlight` is always merging.

dvoet: I agree that `CloneWorkspaceFlight` is doing the wrong thing and I fixed it in this PR. I don't think it is expected that `WorkspaceCreateFlight` has anything to do with it though. So I would opt for removing it from `WorkspaceCreateFlight` and doing it correctly in `CloneWorkspaceFlight` as proposed in this PR

ddietterich: I'm not seeing the incorrectness. Both workspace clone and create run the WorkspaceCreateFlight. I guess we could rename it CreateWorkspaceForNewOrCloneFlight, to make it more clear that it is a flight that is shared by multiple cases.

PR #1299, opened 2023-05-25
[PF-2770] Implement new async workspace and cloud context APIs

ddietterich:
This PR adds the workspace V2 REST APIs for async creation and deletion of workspaces and cloud contexts.

- Put the new APIs into their own controller module: `WorkspaceV2Api` that is dispatched from `WorkspaceApiController`
- Factored out common workspace controller methods into `WorkspaceApiUtils` and `ControllerUtils`
- Factored out common subflight waiting and metadata row waiting methods into existing modules
- Put test support for the new APIs in their own module: `MvcWorkspaceApi`
- Made a small test to exercise the new APIs

Follow-on work:
- MockMvcUtils is huge and needs to be split up. I am planning to move all of the Mvc workspace methods into `MvcWorkspaceApi`.
- Most WSM tests should be converted to use the new workspace endpoints. We should a test or two in connected and integration using the existing endpoints to make sure we don't break them. I have added [PF-2811 ](https://broadworkbench.atlassian.net/browse/PF-2811) to track that work.


service/src/main/java/bio/terra/workspace/app/controller/WorkspaceV2Api.java
```diff
@@ -0,0 +1,222 @@
...
+    this.authenticatedUserRequestFactory = authenticatedUserRequestFactory;
+    this.jobApiUtils = jobApiUtils;
+    this.jobService = jobService;
+    this.request = request;
+    this.samService = samService;
+    this.workspaceApiUtils = workspaceApiUtils;
+    this.workspaceService = workspaceService;
+  }
+
+  public ResponseEntity<ApiCreateWorkspaceV2Result> createWorkspaceV2(
```
yuhuyoyo, line 72: @Traced and @override

yuhuyoyo (approved):

yuhuyoyo (approved):

Traced:
您好，我看到以后就回复您

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/CreateWorkspaceV2Flight.java
```diff
@@ -0,0 +1,99 @@
...
+
+  public CreateWorkspaceV2Flight(FlightMap inputParameters, Object applicationContext) {
+    super(inputParameters, applicationContext);
+
+    FlightBeanBag appContext = FlightBeanBag.getFromObject(applicationContext);
+    WsmResourceStateRule wsmResourceStateRule = appContext.getFeatureConfiguration().getStateRule();
+
+    // get data from inputs that steps need
+    AuthenticatedUserRequest userRequest =
+        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
```
zloery, line 31: nit: this one should also use `getRequired`

service/src/main/java/bio/terra/workspace/app/controller/shared/WorkspaceApiUtils.java
```diff
@@ -0,0 +1,198 @@
...
+   * @param apiSpendProfile spend profile from the request body
+   * @return nullable spend profile
+   */
+  public @Nullable SpendProfile validateSpendProfilePermission(
+      AuthenticatedUserRequest userRequest, String apiSpendProfile) {
+    if (apiSpendProfile == null) {
+      return null;
+    }
+    return spendProfileService.authorizeLinking(
+        new SpendProfileId(apiSpendProfile), features.isBpmGcpEnabled(), userRequest);
```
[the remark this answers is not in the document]
ddietterich: The flag is coded as "BpmEnabled" even though the name says GCP. I believe Drew has this configured true in all Broad environments. The implementation checks the configured profiles first, so Verily usage never tests the flag.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceV2Api.java
```diff
@@ -0,0 +1,222 @@
...
+        jobId,
+        userRequest);
+
+    ApiCreateWorkspaceV2Result result = fetchCreateWorkspaceV2Result(jobId);
+    return new ResponseEntity<>(
+        result, ControllerUtils.getAsyncResponseCode(result.getJobReport()));
+  }
+
+  private ApiCreateWorkspaceV2Result fetchCreateWorkspaceV2Result(String jobId) {
+    JobApiUtils.AsyncJobResult<Workspace> jobResult =
```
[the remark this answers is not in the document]
ddietterich: Since we have no surrounding container controlling access, what permission would I check?
We will probably have one in the future, but not now. Do you see any other choice but reverting to the V1 behavior and returning just the workspace id?

openapi/src/parts/workspace.yaml
```diff
@@ -555,6 +661,62 @@ components:
...
+          $ref: '#/components/schemas/WorkspaceDescription'
+
+    DeleteWorkspaceV2Request:
+      type: object
+      required: [ jobControl ]
+      description: |
+        Request body for asynchronously deleting a workspace.
+      properties:
+        jobControl:
+          $ref: '#/components/schemas/JobControl'
```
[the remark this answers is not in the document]
ddietterich: Allowing the client to pass in the jobId gives them the ability to know if they have submitted the job; that is, to support idempotency. Imaging calling an async API from a flight step. To make the step idempotent, you would need a way to see if you had already submitted the job.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceV2Api.java
```diff
@@ -0,0 +1,222 @@
...
+
+  public ResponseEntity<ApiJobResult> deleteCloudContextV2(
+      UUID workspaceId, ApiCloudPlatform cloudContext, ApiDeleteCloudContextV2Request body) {
+    AuthenticatedUserRequest userRequest = authenticatedUserRequestFactory.from(request);
+    ControllerValidationUtils.validateCloudPlatform(cloudContext);
+    String jobId = body.getJobControl().getId();
+
+    Workspace workspace =
+        workspaceService.validateMcWorkspaceAndAction(
+            userRequest, workspaceId, SamConstants.SamWorkspaceAction.WRITE);
```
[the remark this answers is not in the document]
ddietterich: Good catch - changed to DELETE

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceV2Api.java
ddietterich: I added the `@Traced` in the WorkspaceApiController. The `@Override` was already there. I think that is the right spot for it.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceV2Api.java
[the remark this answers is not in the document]
ddietterich: I'm doing #3

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceV2Api.java
```diff
@@ -0,0 +1,222 @@
...
+
+  private ApiCreateWorkspaceV2Result fetchCreateWorkspaceV2Result(String jobId) {
+    JobApiUtils.AsyncJobResult<Workspace> jobResult =
+        jobApiUtils.retrieveAsyncJobResult(jobId, Workspace.class);
+
+    ApiWorkspaceDescription apiWorkspace = null;
+    Workspace workspace = null;
+    if (jobResult.getJobReport().getStatus().equals(ApiJobReport.StatusEnum.SUCCEEDED)) {
+      workspace = jobResult.getResult();
+      apiWorkspace = workspaceApiUtils.buildWorkspaceDescription(workspace, WsmIamRole.OWNER);
```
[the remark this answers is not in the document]
ddietterich: This goes away when I just return the workspace id

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceV2Api.java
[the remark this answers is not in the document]
ddietterich: I think I will go ahead and do 3.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceV2Api.java
```diff
@@ -0,0 +1,218 @@
...
+      if (spendProfile == null) {
+        throw new MissingRequiredFieldsException(
+            "To create a cloud context you must provide both cloudPlatform and spendProfile");
+      }
+      spendProfileId = spendProfile.id();
+    }
+
+    TpsPolicyInputs policies =
+        workspaceApiUtils.validateAndConvertPolicies(body.getPolicies(), body.getStage());
+    // WSM requires a userFacingId. Create one, if it is not provided.
```
mikenev, line 102: Probably NA for this review since it's copied from the previous code. But this seems to be on the wrong line and I don't see where we create an id - seems like we check 2 places on the body.

service/src/main/java/bio/terra/workspace/app/controller/shared/ControllerUtils.java
```diff
@@ -0,0 +1,47 @@
+package bio.terra.workspace.app.controller.shared;
+
+import bio.terra.workspace.generated.model.ApiJobReport;
+import javax.servlet.http.HttpServletRequest;
+import org.springframework.http.HttpStatus;
+
+public class ControllerUtils {
+
+  /**
+   * Returns the result endpoint corresponding to an async request, prefixed with a / character. The
```
mikenev, line 10: The first sentence here doesn't seem right. It doesn't seem to be prefixed with a / , maybe {servletpath}/

openapi/src/parts/workspace.yaml
```diff
@@ -390,6 +393,102 @@ paths:
...
+        '200':
+          $ref: '#/components/responses/CreateWorkspaceV2Response'
+        '202':
+          $ref: '#/components/responses/CreateWorkspaceV2Response'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '500':
+          $ref: '#/components/responses/NotFound'
```
mikenev, line 418: 500 is probably a server error

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerTest.java
```diff
@@ -147,7 +147,7 @@ public void createWorkspace_duplicateUuid_throws409() throws Exception {
         workspaceId,
         /*stageModel=*/ null,
         /*policies=*/ null,
-        HttpStatus.SC_CONFLICT);
+        HttpStatus.SC_BAD_REQUEST);
```
mikenev, line 150: I don't feel strongly about this but it is an interesting change. Does the response note why it's a bad request?

mikenev (approved):
Minor comments - I like the approach and splitting up some of these bigger files.

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceV2Api.java
ddietterich: I moved the comment to the correct line.

openapi/src/parts/workspace.yaml
```diff
@@ -390,6 +393,102 @@ paths:
...
+
+  /api/workspaces/v2/result/{jobId}:
+    parameters:
+      - $ref: '#/components/parameters/JobId'
+    get:
```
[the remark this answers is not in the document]
ddietterich: It will return a 202 if the flight is not complete. I'm not sure if a 409 ever happens.

openapi/src/parts/workspace.yaml
ddietterich: haha, yes!

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerTest.java
ddietterich: I removed a piece of pre-check code that duplicates the check in the DAO. The workspace DAO throws `DuplicateWorkspaceException` that derives from `BadRequestException` with the error text:

`"Workspace with id %s already exists - display name %s stage %s"`

PR #1302, opened 2023-05-25
Install plotnine library in post-startup script

johntbates:
[`plotnine`](https://pypi.org/project/plotnine/) is installed by default in CWB. Its absence in VWB makes it difficult to port workspaces from CWB.

Confirmed this works in VWB Primary Clarifier workspace environments.

Fixes [PF-2798](https://broadworkbench.atlassian.net/browse/PF-2798)

[PF-2798]: https://broadworkbench.atlassian.net/browse/PF-2798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mbookman (commented):
LGTM.

Eventually we should break this up into separate logical lists of utilities vs. libraries, but for now the list is short enough to reason over.

cbookg (approved):

PR #1303, opened 2023-05-30
Add content disposition parameter to SAS

aherbst-broad:
[Relevant ticket ](https://broadworkbench.atlassian.net/browse/WOR-1006)

## Why
We would like to more easily tie the storage access logs in Azure together with the WSM logs as captured in GCP. Today that requires a SHA256 from the SAS request (or the corresponding blob log entry in Azure). This works, but is cumbersome. We are going to start including the Sam subject ID in the SAS so we can more easily search for logs related to a particular user in both systems.

## This PR
* Pulls the request user's subject ID from Sam and sets it as the [content disposition value](https://learn.microsoft.com/en-us/dotnet/api/azure.storage.sas.sasqueryparameters.contentdisposition?view=azure-dotnet) when minting a SAS token. This is technically not what the header is for, but given that we do not serve files directly in the browser, instead giving them SAS's and an azcopy cmd, this seems like an acceptable tradeoff.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
@@ -184,7 +184,7 @@ public class AzureStorageAccessService {
 
     var samUser = samService.getSamUser(userRequest);
     logger.info(
-        "User {} [sub={}] requesting SAS token for Azure storage container {} in workspace {}",
+        "User {} [SubjectId={}] requesting SAS token for Azure storage container {} in workspace {}",
         samUser.getEmail(),
         samUser.getSubjectId(),
         storageContainerUuid.toString(),
@@ -241,7 +241,7 @@ public class AzureStorageAccessService {
             .toUpperCase();
 
     logger.info(
-        "SAS token with expiry time of {} generated for user {} [sub={}] on container {} in workspace {} [sha256 = {}]",
+        "SAS token with expiry time of {} generated for user {} [SubjectId={}] on container {} in workspace {} [sha256 = {}]",
         sasTokenOptions.expiryTime(),
         samUser.getEmail(),
         samUser.getSubjectId(),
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -182,12 +182,11 @@ public AzureSasBundle createAzureStorageContainerSasToken(
...
     features.azureEnabledCheck();
 
-    var userEmail =
-        SamRethrow.onInterrupted(
-            () -> samService.getUserEmailFromSam(userRequest), "getUserEmailFromSam");
+    var samUser = samService.getSamUser(userRequest);
     logger.info(
-        "User {} requesting SAS token for Azure storage container {} in workspace {}",
-        userEmail,
+        "User {} [sub={}] requesting SAS token for Azure storage container {} in workspace {}",
```
cahrens, line 187: I would have no idea what "sub" is. Could you make it `[SubjectId={}]`?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -241,9 +241,10 @@ public AzureSasBundle createAzureStorageContainerSasToken(
             .toUpperCase();
 
     logger.info(
-        "SAS token with expiry time of {} generated for user {} on container {} in workspace {} [sha256 = {}]",
+        "SAS token with expiry time of {} generated for user {} [sub={}] on container {} in workspace {} [sha256 = {}]",
```
cahrens, line 244: Here too.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
aherbst-broad: Done

cahrens (approved):

sergiygetlin (approved):

PR #1309, opened 2023-06-05
[PF-2788] enforce state

ddietterich:
This PR includes two things:
1. Exposes cloud context and workspace state information (state, flightid, error) through the REST API.
2. Validates workspace and cloud context state in the controller layer.

I made a unit tests to verify the state validation in the controller. As resources and/or update operations on existing resources are added, they will need to include appropriate state validation and get added to those tests.

FYI: @dexamundsen @aherbst-broad

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
@@ -47,6 +47,7 @@ import bio.terra.workspace.service.resource.model.WsmResourceType;
 import bio.terra.workspace.service.workspace.AwsCloudContextService;
 import bio.terra.workspace.service.workspace.WorkspaceService;
 import bio.terra.workspace.service.workspace.model.AwsCloudContext;
+import bio.terra.workspace.service.workspace.model.CloudPlatform;
 import bio.terra.workspace.service.workspace.model.Workspace;
 import bio.terra.workspace.service.workspace.model.WorkspaceConstants;
 import com.google.common.base.Strings;
@@ -138,6 +139,7 @@ public class ControlledAwsResourceApiController extends ControlledResourceContro
     ControlledResource awsResource =
         controlledResourceMetadataManager.validateControlledResourceAndAction(
             userRequest, workspaceUuid, resourceUuid, SamControlledResourceActions.DELETE_ACTION);
+    workspaceService.validateWorkspaceAndContextState(workspaceUuid, CloudPlatform.AWS);
 
     // sanity check that resource is of expected type
     if (awsResource.getResourceType() != wsmResourceType) {
@@ -237,6 +239,7 @@ public class ControlledAwsResourceApiController extends ControlledResourceContro
             ControllerValidationUtils.samCreateAction(
                 AccessScopeType.fromApi(body.getCommon().getAccessScope()),
                 ManagedByType.fromApi(body.getCommon().getManagedBy())));
+    workspaceService.validateWorkspaceAndContextState(workspace, CloudPlatform.AWS);
 
     String folderName = body.getAwsS3StorageFolder().getFolderName();
     if (StringUtils.isEmpty(folderName)) {
@@ -392,6 +395,7 @@ public class ControlledAwsResourceApiController extends ControlledResourceContro
             ControllerValidationUtils.samCreateAction(
                 AccessScopeType.fromApi(body.getCommon().getAccessScope()),
                 ManagedByType.fromApi(body.getCommon().getManagedBy())));
+    workspaceService.validateWorkspaceAndContextState(workspace, CloudPlatform.AWS);
 
     InstanceType instanceType =
         InstanceType.fromValue(body.getAwsSageMakerNotebook().getInstanceType());
--- a/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -4,6 +4,7 @@ import static bio.terra.workspace.app.controller.shared.PropertiesUtils.convertA
 import static bio.terra.workspace.common.utils.ControllerValidationUtils.validatePropertiesDeleteRequestBody;
 import static bio.terra.workspace.common.utils.ControllerValidationUtils.validatePropertiesUpdateRequestBody;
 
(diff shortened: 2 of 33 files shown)


service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
```diff
@@ -138,6 +139,7 @@ private ResponseEntity<ApiDeleteControlledAwsResourceResult> deleteAwsResource(
     ControlledResource awsResource =
         controlledResourceMetadataManager.validateControlledResourceAndAction(
             userRequest, workspaceUuid, resourceUuid, SamControlledResourceActions.DELETE_ACTION);
+    workspaceService.validateWorkspaceAndContextState(workspaceUuid, CloudPlatform.AWS);
```
zloery, line 142: Should we also check resource state here? I don't think it's entirely necessary since that should happen as part of the flight, but it saves launching a flight

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -323,8 +323,10 @@ public ResponseEntity<ApiWorkspaceDescription> updateWorkspace(
     if (body.getUserFacingId() != null) {
       ControllerValidationUtils.validateUserFacingId(body.getUserFacingId());
     }
-    workspaceService.validateWorkspaceAndAction(
-        userRequest, workspaceUuid, SamConstants.SamWorkspaceAction.WRITE);
+    Workspace testWorkspace =
```
zloery, line 326: nit: why `testWorkspace` instead of just `workspace`?

zloery (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -317,6 +317,7 @@ public StorageData getStorageAccountData(
       UUID workspaceUuid, UUID storageContainerUuid, AuthenticatedUserRequest userRequest) {
     // Creating an AzureStorageContainerSasToken requires checking the user's access to both the
     // storage container and storage account resource
+    // TODO: Access control checks should be done in the controller layer
```
yuhuyoyo, line 320: ticket number

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -276,6 +286,78 @@ public Workspace validateMcWorkspaceAndAction(
...
   }
 
+  public void validateWorkspaceState(UUID workspaceUuid) {
+    Workspace workspace = workspaceDao.getWorkspace(workspaceUuid);
+    validateWorkspaceState(workspace);
+  }
+
+  /**
+   * Test that the workspace is in a ready state. This is designed to be used right after
+   * validateWorkspaceAndAction
```
yuhuyoyo, line 296: so can we merge this with validateWorkspaceAndAction?

service/src/test/java/bio/terra/workspace/service/resource/statetests/AnyResourceStateFailureTest.java
```diff
@@ -0,0 +1,292 @@
...
+    when(mockSamService().getUserStatusInfo(any()))
+        .thenReturn(
+            new UserStatusInfo()
+                .userEmail(USER_REQUEST.getEmail())
+                .userSubjectId(USER_REQUEST.getSubjectId()));
+    refMetadata = makeDefaultReferencedResourceFieldsApi().name(RESOURCE_NAME);
+  }
+
+  @Test
+  void testNoContextResourceCreateValidation() throws Exception {
```
yuhuyoyo, line 85: for unit test, we usually break them into smaller pieces because they are so quick and it's more readible :)

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -276,6 +286,78 @@ public Workspace validateMcWorkspaceAndAction(
...
+   *
+   * @param workspace object describing the workspace, including its current state
+   */
+  public void validateWorkspaceState(Workspace workspace) {
```
dexamundsen, line 300: Is there a reason not to allow READ when workspace is not READY?

dexamundsen (approved):

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
ddietterich: No. I believe I have applied this only in create, update, and delete cases.
LMK if you see a spot where I got it wrong!

service/src/main/java/bio/terra/workspace/app/controller/ControlledAwsResourceApiController.java
ddietterich: I don't think we need to optimize launching flights. It is not a big cost.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
dexamundsen: grant role - https://github.com/DataBiosphere/terra-workspace-manager/blob/dd-PF-2788-enforce-state/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java#L422

clone ws - https://github.com/DataBiosphere/terra-workspace-manager/blob/dd-PF-2788-enforce-state/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java#L617

get bucket reference - https://github.com/DataBiosphere/terra-workspace-manager/blob/dd-PF-2788-enforce-state/service/src/main/java/bio/terra/workspace/app/controller/ReferencedGcpResourceController.java#L246

check on source in clone object, bucket, bq table, bq data set & data repo references - ReferencedGcpResourceController.java

- 
in all above we could still make an argument for needed a state check to handle corner cases / concurrent operations. However they are likely too uncommon to need checks.

ddietterich: I was aiming to cover create, delete, and update types of operations. I'll check grant/revoke and clone cases and make a PR as needed. Thanks for checking!

PR #1310, opened 2023-06-06
[PF-2820] post-startup: write to the .bashrc instead of .bash_profile

mbookman:
Includes a small fix where we were writing the "END: Terra-specific customizations" comment twice to the `~/.bash_profile` (instead of once to the notebook config file).


wnojopra (approved):
LGTM!

rogerwangcs (approved):
Looks good!

PR #1315, opened 2023-06-08
[PF-2825] Fix resource clone state checks and test

ddietterich:
FullIntegration failed because my new workspace/cloud context state check was requiring a cloud context for the target of a clone, even if the clone was making a reference.

This fix adds a new validation method that uses the cloning instructions to decide how much to check.

A couple of consequences of this change. Previously, the effective cloning instructions were calculated in the service layer. Now I need it in the controller. To compute it, I need to fetch the actual resource. What I found was that another validation was already returning the resource, but it was not received in a variable. So the code went and retrieved it **again** in the service layer. So I changes service APIs to take the whole resource and the computed cloning instructions.

A couple of unit tests failed, because the new error checking happens before the test. I moved those to connected tests.
I added two new unit tests for testing that the clone test is working.

p.s., this does not address missing checks for reference clones and grant/revoke.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledFlexibleResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledFlexibleResourceApiController.java
@@ -251,8 +251,8 @@ public class ControlledFlexibleResourceApiController extends ControlledResourceC
         new ApiCloneControlledFlexibleResourceResult()
             .resource(clonedFlexResource.toApiResource())
             .effectiveCloningInstructions(cloningInstructions.toApiModel())
-            .sourceWorkspaceId(resourceUuid)
-            .sourceResourceId(workspaceUuid);
+            .sourceWorkspaceId(workspaceUuid)
+            .sourceResourceId(resourceUuid);
 
     return new ResponseEntity<>(result, HttpStatus.OK);
   }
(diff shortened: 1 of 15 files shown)


dexamundsen (approved):

service/src/main/java/bio/terra/workspace/app/controller/ControlledFlexibleResourceApiController.java
```diff
@@ -210,61 +209,50 @@ public ResponseEntity<ApiCloneControlledFlexibleResourceResult> cloneFlexibleRes
...
 
     ApiCloneControlledFlexibleResourceResult result =
         new ApiCloneControlledFlexibleResourceResult()
             .resource(clonedFlexResource.toApiResource())
-            .effectiveCloningInstructions(effectiveCloningInstructions.toApiModel())
-            .sourceWorkspaceId(sourceResourceId)
-            .sourceResourceId(sourceWorkspaceId);
+            .effectiveCloningInstructions(cloningInstructions.toApiModel())
+            .sourceWorkspaceId(resourceUuid)
+            .sourceResourceId(workspaceUuid);
```
zloery, line 255: you didn't introduce this, but it looks like the resource/workspace ID arguments are swapped here

zloery (approved):

service/src/main/java/bio/terra/workspace/app/controller/ControlledFlexibleResourceApiController.java
ddietterich: Ha! And since it is setting the return, it doesn't break anything.
Good eyes! 👀 
I fixed it.

PR #1316, opened 2023-06-08
[PF-2784] Add retries to TpsApiDispatch

ddietterich:
This PR adds retries to `TpsApiDispatch`. I considered re-implementing `SamRetry` to use the `RetryUtils`, but decided it wasn't a good fit. Also, `SamRetry` has been working well - it ain't broken, so don't "fix" it.

I looked for a way to share retry code between Sam and TPS. Both are doing exactly the same thing to different instances of `ApiException`. I decided the complexity wasn't worth the effort. So...
- Copied `SamRetry` into `TpsRetry` and cleaned up naming and typing of the `ApiException`.
- Implemented the retry inside of `TpsApiDispatch`
- Copied `SamRethrow` into `Rethrow` and moved it to common utils.
- Global replace of `SamRethrow` with `Rethrow`
- Add `Rethrow` wrapping around all the places where we use TPS that are not in flights. I considered putting the `Rethrow` inside some `TpsApiDispatch` methods, but decided that could lead to developer confusion.
- Added `throws Exception` to a bunch of tests to fix compiler complaints about unhandled `InterruptedException` in tests

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/policy/TpsRetry.java
+++ b/service/src/main/java/bio/terra/workspace/service/policy/TpsRetry.java
@@ -1,0 +1,145 @@
+package bio.terra.workspace.service.policy;
+
+import static java.time.Instant.now;
+
+import bio.terra.policy.client.ApiException;
+import java.net.SocketTimeoutException;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.concurrent.TimeUnit;
+import org.apache.http.HttpStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class is a clone of the TCL SamRetry. Since the definition of ApiException is different
+ * (even though it is nearly identical), we cannot just use that code.
+ */
+public class TpsRetry {
+  private static final Logger logger = LoggerFactory.getLogger(TpsRetry.class);
+
+  // The retry function starts with INITIAL_WAIT between retries, and doubles that until it
+  // reaches MAXIMUM_WAIT, after which all retries are MAXIMUM_WAIT apart.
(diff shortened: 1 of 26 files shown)


mikenev (approved):
Appreciate the detailed description you left. Made it much simpler to read through.

service/src/main/java/bio/terra/workspace/service/policy/TpsRetry.java
```diff
@@ -0,0 +1,145 @@
...
+      throws ApiException, InterruptedException {
+    if (operationTimeout.minus(retryDuration).isBefore(now())) {
+      logger.error("TpsRetry: operation timed out after " + operationTimeout.toString());
+      // If we timed out, throw the error from Tps that caused us to need to retry.
+      throw previousException;
+    }
+    logger.info("TpsRetry: sleeping " + retryDuration.getSeconds() + " seconds");
+    TimeUnit.SECONDS.sleep(retryDuration.getSeconds());
+    retryDuration = retryDuration.plus(INITIAL_WAIT);
+    if (retryDuration.compareTo(MAXIMUM_WAIT) > 0) {
```
yuhuyoyo, line 141: nit, this can be one line: retryDuration = max(retryDuration, Maximum_WAIT). 

but also could you mention in the description that the retry duration will increase by 10 seconds each time.

ddietterich: Hmm... I couldn't find a `max` that compares Duration. Do you have one handy?
I could use `a.compareTo(b) >= 0 ? a : b`, but repeats the `retryDuration.plus(INITIAL_WAIT)`.
So I have left it as is.

I added more explanation to the description.

PR #1318, opened 2023-06-09
[PF-2812] Rewrite workspace and cloud context delete flights

ddietterich:
Rewrite based on [Better Cloud Context Delete Flight](https://docs.google.com/document/d/1KKVkNKWpnkfSJ1PAABFXzEE9gfn6x1Qke28n0rSbUE8/edit?resourcekey=0-oMIy58VST6KHxXsOSikxPA#).

The cloud context delete flights tolerate failed resource deletes and returns the cloud context to READY state.


service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/cloudcontext/BuildAndValidateResourceListStep.java
```diff
@@ -0,0 +1,72 @@
...
+    List<ResourceDeleteFlightPair> resourcePairs = new ArrayList<>();
+    for (ControlledResource resource : resources) {
+      resourcePairs.add(
+          new ResourceDeleteFlightPair(resource.getResourceId(), UUID.randomUUID().toString()));
+    }
+    context
+        .getWorkingMap()
+        .put(
+            RESOURCE_DELETE_FLIGHT_PAIR_LIST,
+            resourcePairs.toArray(new ResourceDeleteFlightPair[0]));
```
zloery, line 64: out of curiosity, why `toArray` instead of leaving this as a list? In other places I think we've serialized Lists successfully and deserialized with `TypeReference`

service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/cloudcontext/DeleteResourcesStep.java
```diff
@@ -0,0 +1,143 @@
...
+    if (flightException.isPresent()) {
+      throwable = flightException.get();
+      message = throwable.getMessage();
+    } else {
+      throwable = null;
+      message = "flight failed with no error message";
+    }
+    throw new CloudContextDeleteException(
+        String.format(
+            "Cloud context delete failed because resource %s could not be deleted. Cause was: %s",
```
zloery, line 97: Given that this is a message we'll end up returning to callers via `JobReport` error, it might be helpful to also include resource name alongside the ID, since users will actually recognize those

service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/cloudcontext/BuildAndValidateResourceListStep.java
ddietterich: Honestly, I do not understand why the TypeReference did not work. The only difference I could see was that it was a list of Record as opposed to a Class. I tried:
- original: stream `.toList()` that failed to deserialize
- explicitly populating an `ArrayList` - that failed
- trying to deserialize using `List.class` - that failed
Finally, I saw this array technique in a Baeldung page and it worked.

The other alternative I thought of would be to make a class that held a list as a member and setup an `@JsonCreator` that took a List as input. I did not try changing the Record to a Class with `@Json` tags.

service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/cloudcontext/DeleteResourcesStep.java
ddietterich: Good point. It does require doing a database op on the error path, but I wrapped that in a try/catch so we don't have nested failures flying around.

service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/cloudcontext/DeleteResourcesStep.java
```diff
@@ -0,0 +1,143 @@
...
+  // they hold a lot of data.
+  private void waitForFlight(ResourceDeleteFlightPair pair, Stairway stairway)
+      throws InterruptedException {
+    FlightState flightState;
+    try {
+      flightState =
+          RetryUtils.getWithRetry(
+              FlightUtils::flightComplete,
+              () -> stairway.getFlightState(pair.flightId()),
+              Duration.ofHours(7), /* total duration */
```
mikenev, line 119: Good to know. It's a little surprising it can take that long, but that's outside of our control.

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/gcp/DeleteCloudContextResourceFlight.java
```diff
@@ -0,0 +1,72 @@
...
+    if (controlledResources.size() != 1) {
+      throw new InternalLogicException(
+          "Unexpected number of resources: " + controlledResources.size());
+    }
+    ControlledResource resource = controlledResources.get(0);
+
+    RetryRule cloudRetry = RetryRules.cloud();
+    RetryRule dbRetry = RetryRules.shortDatabase();
+
+    addStep(
```
mikenev, line 52: Could you add a comment on why we do this one - since we have another DeleteMetadataStep below?

mikenev (approved):

PR #1321, opened 2023-06-12
PF-2831: throw 409 when we have a conflicted resource name

yuhuyoyo:
Ran connectedPlus test locally.


zloery (approved):

ddietterich (approved):
I am not confident that 409 is the right code for a name collision. I would have expected 400, bad request. But we have already been using 409 for similar cases, so I'll vote for consistency.

PR #1324, opened 2023-06-12
PF-2818: use wsm sa token to delete sam resources

yuhuyoyo:


zloery (approved):

yuhuyoyo:
> Should we do this in the individual delete resource flights as well? I'm not sure what they do.

i checked bucket and it's using wsm sa token

PR #1290, opened 2023-05-22
[PF-2791] Change serdes for state exceptions; remove exception from DeleteFailure

ddietterich:
My code to serialize and deserialize exceptions on state changes didn't work. This change adds a method of normalizing the exceptions stored. Added another method that tolerate column contents that cannot be deserialized, rather than breaking workspace retrieval altogether.

Also, this removes the ability to set any exception on a workspace delete failure. That shouldn't have been there in the first place.

PR #1330, opened 2023-06-13
[BENCH-703] Fix cloud context error deserialization

zloery:
This is a similar error to https://github.com/DataBiosphere/terra-workspace-manager/pull/1290 but for cloud contexts instead of workspaces. Errors cannot be deserialized directly to `ErrorReportException` so this uses a utility instead.

Like #1290 this also removes storing errors on cloud context deletion failure, as my understanding is that's only for the `BROKEN` state.


jmczerk (approved):
Thanks for jumping on this @zloery !

dexamundsen (approved):

PR #1337, opened 2023-06-20
[BENCH-714] Add CloudPlatform param to createCloudContext testUtil, spotbug warnings

dexamundsen:
NOTE -- rebase after https://github.com/DataBiosphere/terra-workspace-manager/pull/1335

- Address some of the spotbug warnings (unused variables, static & final declarations)
- remove exceptions that are not thrown in tests
- Add ApiCloudPlatform as a param to test util function for cloud context creation. This enables its use across all cloud platforms (specifically in AWS tests)

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneAllResourcesFlight.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneAllResourcesFlight.java
@@ -114,7 +114,7 @@ public class CloneAllResourcesFlight extends Flight {
                     resourceCloneInputs.getFlightId()),
                 RetryRules.shortDatabase());
           }
-          case CONTROLLED_GCP_AI_NOTEBOOK_INSTANCE ->
+          default ->
           // Can't throw in a flight constructor
           logger.error(
               "Unsupported controlled resource type {}",
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneAllResourcesFlight.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneAllResourcesFlight.java
@@ -61,6 +61,7 @@ public class CloneAllResourcesFlight extends Flight {
         break;
       case CONTROLLED:
         switch (resourceCloneInputs.getResource().getResourceType()) {
+            // GCP
           case CONTROLLED_GCP_GCS_BUCKET -> {
             addStep(
                 new LaunchCloneGcsBucketResourceFlightStep(
@@ -87,6 +88,9 @@ public class CloneAllResourcesFlight extends Flight {
                     resourceCloneInputs.getFlightId()),
                 RetryRules.cloudLongRunning());
           }
+            // CONTROLLED_GCP_AI_NOTEBOOK_INSTANCE: not supported
+
+            // Azure
           case CONTROLLED_AZURE_STORAGE_CONTAINER -> {
             addStep(
                 new LaunchCloneControlledAzureStorageContainerResourceFlightStep(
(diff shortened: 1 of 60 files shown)


yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneAllResourcesFlight.java
```diff
@@ -114,7 +114,7 @@ private void addFlightLaunchStepsForResource(
                     resourceCloneInputs.getFlightId()),
                 RetryRules.shortDatabase());
           }
-          case CONTROLLED_GCP_AI_NOTEBOOK_INSTANCE ->
+          default ->
```
zloery, line 117: nit: I would either leave `case CONTROLLED_GCP_AI_NOTEBOOK_INSTANCE` directly above default (so it falls through) or else leave a comment here that we intentionally don't support cloning notebooks at the moment.

service/src/test/java/bio/terra/workspace/app/controller/CreateAzureStorageContainerSasTokenTest.java
```diff
@@ -32,11 +32,11 @@ public class CreateAzureStorageContainerSasTokenTest extends BaseAzureUnitTest {
   private UUID workspaceId;
   private UUID storageContainerId;
 
-  ArgumentCaptor<SasTokenOptions> sasTokenOptionsCaptor =
+  final ArgumentCaptor<SasTokenOptions> sasTokenOptionsCaptor =
```
zloery, line 35: (here and elsewhere) I don't think we really get any value out of adding `final` to all of these test variables. Since people tend to just remove `final` if we need to modify the variable in a new test, it mostly just signals "this isn't being modified right now", not "this should never be modified".

zloery (approved):

service/src/test/java/bio/terra/workspace/app/controller/CreateAzureStorageContainerSasTokenTest.java
dexamundsen: was part of addressing spot bug warnings. agree no value add
will remove

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneAllResourcesFlight.java
dexamundsen: Adding comment, since it is not needed for AWS notebook as well

yuhuyoyo: you can also do case CONTROLLED_GCP_AI_NOTEBOOK_INSTANCE, CONTROLLED_AWS_SAGEMAKER_NOTEBOOK_INSTANCE ->

service/src/test/java/bio/terra/workspace/app/controller/CreateAzureStorageContainerSasTokenTest.java
dexamundsen: Left it as is for class level variables (instance / static)
no final spec for local (function variables)

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneAllResourcesFlight.java
dexamundsen: Adding a default case is preferred (with comment).
reason: 
the default case applies to all existing (and new resources) that don't explicitly have clone enabled. 
This way we don't miss out on changes for new resources.

dexamundsen: There are more resources that dont have clone set up, hence default makes more sense than listing all others

dexamundsen: Ignore previous replies. added separate cases with specific logs

PR #1339, opened 2023-06-22
[BENCH-470] S3 Storage Folder Unit tests

dexamundsen:
Ut for AWS S3 Storage Folder

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3StorageFolder/AwsS3StorageFolderStepTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3StorageFolder/AwsS3StorageFolderStepTest.java
@@ -57,7 +57,6 @@ public class AwsS3StorageFolderStepTest extends BaseAwsUnitTest {
   @Mock private S3Client mockS3Client;
 
   private MockedStatic<AwsUtils> mockAwsUtils;
-  private MockedStatic<DeleteAwsS3StorageFolderStep> mockDeleteAwsS3StorageFolderStep;
   private ControlledAwsS3StorageFolderResource s3FolderResource;
   private static final AwsServiceException s3Exception1 =
       S3Exception.builder().message("not authorized to perform").build();
@@ -65,7 +64,6 @@ public class AwsS3StorageFolderStepTest extends BaseAwsUnitTest {
   @BeforeAll
   public void init() {
     mockAwsUtils = Mockito.mockStatic(AwsUtils.class);
-    mockDeleteAwsS3StorageFolderStep = Mockito.mockStatic(DeleteAwsS3StorageFolderStep.class);
     s3FolderResource =
         ControlledAwsResourceFixtures.makeDefaultAwsS3StorageFolderResource(
             ControlledResourceFixtures.WORKSPACE_ID);
@@ -74,7 +72,6 @@ public class AwsS3StorageFolderStepTest extends BaseAwsUnitTest {
   @AfterAll
   public void terminate() {
     mockAwsUtils.close();
-    mockDeleteAwsS3StorageFolderStep.close();
   }
 
   @BeforeEach
@@ -94,11 +91,6 @@ public class AwsS3StorageFolderStepTest extends BaseAwsUnitTest {
         .thenReturn(ControlledAwsResourceFixtures.AWS_CREDENTIALS_PROVIDER);
     mockAwsUtils.when(() -> AwsUtils.getS3Client(any(), any())).thenReturn(mockS3Client);
     mockAwsUtils.when(() -> AwsUtils.checkException(any())).thenCallRealMethod();
-
-    mockDeleteAwsS3StorageFolderStep.clearInvocations();
-    mockDeleteAwsS3StorageFolderStep
-        .when(() -> DeleteAwsS3StorageFolderStep.executeDeleteAwsS3StorageFolder(any(), any()))
-        .thenReturn(StepResult.getStepResultSuccess());
   }
 
   @Test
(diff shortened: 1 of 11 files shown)


service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3StorageFolder/AwsS3StorageFolderStepTest.java
```diff
@@ -0,0 +1,353 @@
...
+    // do: success
+    assertThat(
+        createS3FolderStep.doStep(mockFlightContext), equalTo(StepResult.getStepResultSuccess()));
+
+    // do: error
+    assertThrows(
+        WorkspaceFixtures.API_EXCEPTION.getClass(),
+        () -> createS3FolderStep.doStep(mockFlightContext));
+
+    // undo: same as tests for DeleteAwsS3StorageFolderStep
```
nfoen, line 114: Shouldn't this check that executeDeleteAwsS3StorageFolder is called? Same for createS3FolderTestFull below

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3StorageFolder/AwsS3StorageFolderStepTest.java
```diff
@@ -0,0 +1,353 @@
...
+    mockAwsUtils = Mockito.mockStatic(AwsUtils.class);
+    s3FolderResource =
+        ControlledAwsResourceFixtures.makeDefaultAwsS3StorageFolderResource(
+            ControlledResourceFixtures.WORKSPACE_ID);
+  }
+
+  @BeforeEach
+  public void setup() {
+    when(mockFlightContext.getResult())
+        .thenReturn(new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL));
```
nfoen, line 75: Out of curiosity, why is this always mocked to return a failure?

nfoen (approved):

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3StorageFolder/AwsS3StorageFolderStepTest.java
dexamundsen: - A flight in `do direction` is still running and doesn't have a result, hence the getResult() will not have value in do().
- undo() is invoked when a flight fails and getResult() should reflect that a flight failed, hence always returning error

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3StorageFolder/AwsS3StorageFolderStepTest.java
dexamundsen: long response & more details --> 

- The only reason there is a static executeDeleteAwsS3StorageFolder function is to ensure create.undo shares the same code as delete.do (do/undo are instance methods)
- Skipped the undo test in create step since it is the same as delete step (already has tests)
- We could `verify that x is called` check, but didnt add earlier since it would add very little value here: Connected test would cover it

adding it now for the sake of completeness

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3StorageFolder/ControlledAwsS3StorageFolderResourceTest.java
```diff
@@ -0,0 +1,73 @@
...
+import bio.terra.workspace.common.BaseAwsUnitTest;
+import bio.terra.workspace.common.fixtures.ControlledAwsResourceFixtures;
+import bio.terra.workspace.common.fixtures.ControlledResourceFixtures;
+import bio.terra.workspace.service.resource.exception.InvalidNameException;
+import org.junit.jupiter.api.Test;
+
+public class ControlledAwsS3StorageFolderResourceTest extends BaseAwsUnitTest {
+
+  @Test
+  public void validateResourceTest() {
```
zloery, line 18: nit: given that these are unit tests with ~0 setup required, prefer splitting them into separate tests

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3StorageFolder/AwsS3StorageFolderStepTest.java
```diff
@@ -0,0 +1,353 @@
...
+import software.amazon.awssdk.services.s3.model.PutObjectResponse;
+import software.amazon.awssdk.services.s3.model.S3Error;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.awssdk.services.s3.model.S3Object;
+
+@TestInstance(Lifecycle.PER_CLASS)
+public class AwsS3StorageFolderStepTest extends BaseAwsUnitTest {
+
+  @Mock private FlightContext mockFlightContext;
+  @Mock private AwsCloudContextService mockAwsCloudContextService;
```
zloery, line 55: Use `@MockBean` for Spring components so they get autowired together

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3StorageFolder/AwsS3StorageFolderStepTest.java
```diff
@@ -0,0 +1,353 @@
...
+  @Mock private SamService mockSamService;
```
zloery, line 56: This mock is already available via `mockSamService()` from the base class

zloery (approved):

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/aws/s3StorageFolder/ControlledAwsS3StorageFolderResourceTest.java
dexamundsen: Retaining as a single test for now - with more creation param we may need updates later

PR #1340, opened 2023-06-22
[BENCH-306] AWS Cloud context Connected Test

dexamundsen:
Rebase & merge AFTER - https://github.com/DataBiosphere/terra-workspace-manager/pull/1339

AWS Cloud context connected test 

image


yuhuyoyo (approved):

service/src/test/java/bio/terra/workspace/service/workspace/flight/aws/CreateAwsWorkspaceFlightTest.java
```diff
@@ -0,0 +1,65 @@
...
+        awsTestUtils.getAwsCloudContext().getEnvironmentAlias());
+
+    // TODO(BENCH-715) add storage bucket
+
+    // delete workspace (with cloud context)
+    ApiJobResult deleteResult = mvcWorkspaceApi.deleteWorkspaceAndWait(userRequest, workspaceUuid);
+    assertEquals(StatusEnum.SUCCEEDED, deleteResult.getJobReport().getStatus());
+
+    // cloud context should have been deleted
+    assertTrue(awsCloudContextService.getAwsCloudContext(workspaceUuid).isEmpty());
```
nfoen, line 61: Is this a test that's actually creating aws resources? If so, does this mean that resources are not deleted if the test fails?

nfoen (approved):

service/src/test/java/bio/terra/workspace/service/workspace/flight/aws/CreateAwsWorkspaceFlightTest.java
dexamundsen: Cloud context creation in AWS does not create any AWS resources. It makes an entry in our DB as to the Bucket / region etc where resources can be created in the future

PR #1341, opened 2023-06-22
PF-2852: Make the test more resilient to failure

yuhuyoyo:
one of the test failed during update assertion so we never reached the end where we reset the bucket name. 

In the failure, the last updated timestamp is off. My guess is that in that the hook where we log the update is slightly late so we are using the previous timestamp in the updateResuorce and got the newer timestamp in the gotResource, causing the disparity.

I removed the comparison. just asserting the updated resource got the new name, new description, etc should be enough.


service/src/test/java/bio/terra/workspace/app/controller/ReferencedGcpResourceControllerGcsBucketConnectedTest.java
```diff
@@ -138,14 +155,8 @@ public void update() throws Exception {
...
             newCloningInstruction);
-    // Update the sourceResource to the updated one as all the tests are sharing
-    // the same resource.
-    ApiGcpGcsBucketResource getResource =
-        mockMvcUtils.getReferencedGcsBucket(
-            userAccessUtils.defaultUserAuthRequest(),
-            workspaceId,
-            sourceResource.getMetadata().getResourceId());
-    assertEquals(updatedResource, getResource);
+    bucketIsUpdated = true;
```
zloery, line 158: nit: this should be set right before the `updateReferencedGcsBucket` call instead to prevent the tiny edge case where the call causes a test failure but still makes the update happen

service/src/test/java/bio/terra/workspace/app/controller/ReferencedGcpResourceControllerGcsBucketConnectedTest.java
```diff
@@ -191,15 +194,10 @@ public void update_nameAndDescriptionOnly() throws Exception {
             newDescription,
             /*newBucketName=*/ null,
             newCloningInstruction);
+    bucketIsUpdated = true;
```
zloery, line 197: same here

zloery (approved):

service/src/test/java/bio/terra/workspace/app/controller/ReferencedGcpResourceControllerGcsBucketConnectedTest.java
```diff
@@ -92,6 +95,20 @@ public void cleanup() throws Exception {
...
+    if (bucketIsUpdated) {
+      mockMvcUtils.updateReferencedGcsBucket(
+          userAccessUtils.defaultUserAuthRequest(),
+          workspaceId,
+          sourceResource.getMetadata().getResourceId(),
+          sourceResourceName,
+          RESOURCE_DESCRIPTION,
+          sourceBucketName,
+          ApiCloningInstructionsEnum.NOTHING);
+    }
```
ddietterich, line 109: I think this should restore `bucketIsUpdated` to `false`, yes?

ddietterich (commented):
IIUC, the `@AfterEach` is to restore the reference resource to an initial state.

I this it would be more straightforward to create a new reference resource for each test, so that there would not be interaction between the tests.

service/src/test/java/bio/terra/workspace/app/controller/ReferencedGcpResourceControllerGcsBucketConnectedTest.java
yuhuyoyo: good catch

yuhuyoyo:
> IIUC, the `@AfterEach` is to restore the reference resource to an initial state.
> ya.
> I this it would be more straightforward to create a new reference resource for each test, so that there would not be interaction between the tests.

just slightly faster? there's only two update tests.

ddietterich:
>  just slightly faster? there's only two update tests.

I would suggest making the reference in each test.

yuhuyoyo:
> > just slightly faster? there's only two update tests.
> 
> I would suggest making the reference in each test.

alright

ddietterich (approved):

PR #1342, opened 2023-06-22
[PF-2843] Add diagnostics to catch a flaky test

ddietterich:
A quick follow-on to PF-2843. The new delete flight had a flake. I cannot reproduce it.
This code adds some logging and dumps Sam children if it seems the "children" error on workspace delete.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -40,6 +40,7 @@ import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import javax.servlet.http.HttpServletRequest;
 import okhttp3.OkHttpClient;
+import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsde.workbench.client.sam.ApiClient;
 import org.broadinstitute.dsde.workbench.client.sam.ApiException;
 import org.broadinstitute.dsde.workbench.client.sam.api.AdminApi;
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/cloudcontext/DeleteResourcesStep.java
```diff
@@ -52,12 +54,18 @@ public DeleteResourcesStep(
...
-    ResourceDeleteFlightPair[] resourcePairs =
+    ResourceDeleteFlightPair[] resourcePairsArray =
         FlightUtils.getRequired(
             context.getWorkingMap(),
             RESOURCE_DELETE_FLIGHT_PAIR_LIST,
             ResourceDeleteFlightPair[].class);
 
+    List<ResourceDeleteFlightPair> resourcePairs = Arrays.asList(resourcePairsArray);
+    logger.info(
+        "Flight {} DeleteResourcesStep found {} resources to delete",
```
yuhuyoyo, line 65: would it be helpful to list all the type and id of the resources like you did in samservice?

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -420,6 +421,28 @@ public void deleteWorkspace(AuthenticatedUserRequest userRequest, UUID uuid)
                 + "This just means the deletion was tried twice so no error thrown.");
         return;
       }
+
+      // Diagnostic: if the error was "Cannot delete a resource with children" then fetch and
+      // dump the remaining children.
+      if (apiException.getCode() == HttpStatus.BAD_REQUEST.value()
+          && StringUtils.contains(
+              apiException.getMessage(), "Cannot delete a resource with children")) {
```
mikenev, line 429: Don't like the hard coded message match, but that's coming from another service so I'm not sure how else to handle it.

mikenev (approved):

service/src/main/java/bio/terra/workspace/service/workspace/flight/delete/cloudcontext/DeleteResourcesStep.java
ddietterich: I'd have to fetch the resources from the database. That felt like more than I wanted to do for diagnostics.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
ddietterich: I totally agree. I don't know another way to do it.

PR #1343, opened 2023-06-22
[PF-2710] list workspace performance

ddietterich:
This PR should improve workspace performance. At a high level, there are two changes:
1. Change the TPS method for creating a PAO if it doesn't exist. This reduces the calls to TPS by half.
2. Move all of the code to gather workspace, cloud context, and activity log information into the WorkspaceDao. This reduces the number of Postgres transactions to 1 from (5 * N) + 1 transactions, where N is the number of workspaces.

In order to put the work into the DAO, I added DbWorkspaceContextPair as an intermediate structure to hold the result of an outer join of workspace and cloud_context tables. I added DbWorkspaceDescription to hold the combined data from workspace, cloud_context, and workspace_activity_log (last update by and last update date).

I added a batch retrieval to the activity log DAO that gets a list of the last change detail for a list of workspaces in a single self-join. I'm very proud of that query :)  In order to be able to connect the activity with the list of workspaces, I added the workspaceUuid into ActivityLogChangeDetails. That had a ripple effect through a lot of tests.

I updated the WorkspaceDescription class to have all of the data needed for the ApiWorkspaceDescription result. That helped move work out of the controller and into services.

I made variations on validateWorkspaceAndAction that would return the WorkspaceDescription for the single-workspace retrieval cases.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
@@ -487,7 +487,7 @@ public class WorkspaceDao {
   }
 
   /** Retrieves a workspace description by userFacingId. */
-  public DbWorkspaceDescription getWorkspaceByUserFacingId(String userFacingId) {
+  public DbWorkspaceDescription getWorkspaceDescriptionByUserFacingId(String userFacingId) {
     if (userFacingId == null || userFacingId.isEmpty()) {
       throw new MissingRequiredFieldException("userFacingId is required");
     }
@@ -666,7 +666,7 @@ public class WorkspaceDao {
    */
   @Traced
   @ReadTransaction
-  public Map<UUID, DbWorkspaceDescription> getWorkspacesMatchingList(
+  public Map<UUID, DbWorkspaceDescription> getWorkspaceDescriptionMapFromIdList(
       Set<UUID> idList, int offset, int limit) {
     // If the incoming list is empty, the caller does not have permission to see any
     // workspaces, so we return an empty list.
--- a/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
@@ -2,6 +2,7 @@ package bio.terra.workspace.db;
 
 import bio.terra.common.db.ReadTransaction;
 import bio.terra.common.db.WriteTransaction;
+import bio.terra.common.exception.ErrorReportException;
 import bio.terra.common.exception.MissingRequiredFieldException;
 import bio.terra.workspace.common.exception.InternalLogicException;
 import bio.terra.workspace.common.logging.model.ActivityLogChangeDetails;
--- a/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
@@ -44,7 +44,6 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.DuplicateKeyException;
 import org.springframework.dao.EmptyResultDataAccessException;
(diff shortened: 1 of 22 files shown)


service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -258,21 +266,48 @@ private UUID createWorkspaceWorker(
...
+   * @param userFacingId user facing id of the workspace in question
+   * @param action the action to authorize against the workspace
+   * @return the workspace description
+   */
+  @Traced
+  public WorkspaceDescription validateWorkspaceAndActionReturningDescription(
+      AuthenticatedUserRequest userRequest, String userFacingId, String action) {
+    DbWorkspaceDescription dbWorkspaceDescription =
+        workspaceDao.getWorkspaceByUserFacingId(userFacingId);
+    logWorkspaceAction(userRequest, dbWorkspaceDescription.getWorkspace().workspaceId(), action);
```
zloery, line 306: nit: if this is called with a ufid that doesn't exist, it will return a 404 without logging the check. can this call `logWorkspaceAction` (or a similar method with ufid instead of uuid) before the DAO?

zloery (approved):
LGTM! Did you get any performance data on `listWorkspace` calls with this change?

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -375,24 +474,67 @@ public Workspace getWorkspace(UUID uuid) {
...
+    DbWorkspaceDescription dbWorkspaceDescription =
+        getWorkspaceDescriptionWorker(WORKSPACE_CONTEXT_BY_WORKSPACE_ID_QUERY, params);
+    if (dbWorkspaceDescription == null) {
+      throw new WorkspaceNotFoundException(String.format("Workspace %s not found.", workspaceId));
+    }
+    return dbWorkspaceDescription;
+  }
+
+  /** Retrieves a workspace description by userFacingId. */
+  public DbWorkspaceDescription getWorkspaceByUserFacingId(String userFacingId) {
```
yuhuyoyo, line 490: getWorkspaceDescriptionByUserFacingId

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -520,27 +662,66 @@ public void updateWorkspaceProperties(UUID workspaceUuid, Map<String, String> pr
    * @param idList List of workspaceIds to query for
    * @param offset The number of items to skip before starting to collect the result set.
    * @param limit The maximum number of items to return.
-   * @return list of Workspaces corresponding to input IDs.
+   * @return map of Workspaces descriptions indexed by workspace id
    */
   @Traced
   @ReadTransaction
-  public List<Workspace> getWorkspacesMatchingList(Set<UUID> idList, int offset, int limit) {
+  public Map<UUID, DbWorkspaceDescription> getWorkspacesMatchingList(
```
yuhuyoyo, line 669: getWorkspaceIdToDescriptionMap

service/src/main/java/bio/terra/workspace/service/iam/model/AccessibleWorkspace.java
```diff
@@ -0,0 +1,7 @@
+package bio.terra.workspace.service.iam.model;
+
+import java.util.List;
+import java.util.UUID;
+
+public record AccessibleWorkspace(
```
yuhuyoyo, line 6: could you add a description of what is accessible workspace?

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
ddietterich: I made it `getWorkspaceDescriptionMapFromIdList`

PR #1345, opened 2023-06-22
[PF-2866] Fix compilation errors

iliadsh:
Should've merged main in first!

### DIFF
--- a/service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceGceInstanceConnectedTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceGceInstanceConnectedTest.java
@@ -735,7 +735,7 @@ public class ControlledResourceServiceGceInstanceConnectedTest extends BaseConne
         FlightDebugInfo.newBuilder().doStepFailures(retrySteps).build());
 
     controlledResourceService.deleteControlledResourceSync(
-        resource.getWorkspaceId(), resource.getResourceId(), user.getAuthenticatedRequest());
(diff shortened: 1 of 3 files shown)


service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceGceInstanceConnectedTest.java
```diff
@@ -735,7 +735,7 @@ void deleteGceInstanceDo() throws Exception {
         FlightDebugInfo.newBuilder().doStepFailures(retrySteps).build());
 
     controlledResourceService.deleteControlledResourceSync(
-        resource.getWorkspaceId(), resource.getResourceId(), user.getAuthenticatedRequest());
+        resource.getWorkspaceId(), resource.getResourceId(), false, user.getAuthenticatedRequest());
```
yuhuyoyo, line 738: this should be true.

we are force deleting the instance aren't we?

even though we are not using this variable really in the flight.

yuhuyoyo:
lol i also sent a pr. let's attach PF-2866 and you can merge this

yuhuyoyo (approved):

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceServiceGceInstanceConnectedTest.java
ddietterich: The only force applies to AWS notebooks, I think. It is irrelevant in this case.
Maybe we should have made an overload that had the old signature and filled in `false`.  Oh well

ddietterich (approved):

PR #1347, opened 2023-06-22
PF-2867: Call workspace delete v2 api in integration test

yuhuyoyo:
that one is async so we don't hit a hard timeout and retry at 2 min mark.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/DeleteWorkspaceWithControlledResource.java
+++ b/integration/src/main/java/scripts/testscripts/DeleteWorkspaceWithControlledResource.java
@@ -73,13 +73,10 @@ public class DeleteWorkspaceWithControlledResource extends WorkspaceAllocateTest
   @Override
   public void doCleanup(List<TestUserSpecification> testUsers, WorkspaceApi workspaceApi)
       throws Exception {
-    try {
-      workspaceApi.deleteWorkspaceV2(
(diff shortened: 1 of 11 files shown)


integration/src/main/java/scripts/testscripts/DeleteWorkspaceWithControlledResource.java
```diff
@@ -72,7 +74,8 @@ protected void doUserJourney(TestUserSpecification testUser, WorkspaceApi worksp
   public void doCleanup(List<TestUserSpecification> testUsers, WorkspaceApi workspaceApi)
       throws Exception {
     try {
-      workspaceApi.deleteWorkspace(getWorkspaceId());
+      workspaceApi.deleteWorkspaceV2(
+          new DeleteWorkspaceV2Request().jobControl(new JobControl().id(UUID.randomUUID().toString())), getWorkspaceId());
```
ddietterich, line 78: Does this need to wait for completion?

ddietterich (approved):
Thanks for picking this up. Looks good. One question in my comments.

integration/src/main/java/scripts/testscripts/DeleteWorkspaceWithControlledResource.java
yuhuyoyo: i changed it to assertThrow

zloery (approved):

PR #1353, opened 2023-06-26
[BENCH-471] Move test values to appropriate fixtures classes

dexamundsen:
Move workspaceId, userEmail, user subjectId to workspaceFixtures so that they can be used in other tests
Remove duplicate declaration of referenceResourceFixtures.default_Resource_Properties

PR #1356, opened 2023-06-27
Revert "[BENCH-471] Move test values to appropriate fixtures classes …

dexamundsen:
…(#1353)"

This reverts commit abfbce7a4feb82fcb246c23b07b8e1848b151192.


ddietterich:
Is this to fix the problem I tossed your way? If so, I don't think it will.

dexamundsen:
> Is this to fix the problem I tossed your way? If so, I don't think it will.

This was my most recent commit, so tried it.

No, this won't fix it. Looking into the change to be reverted

PR #1362, opened 2023-06-28
[PF-2882] Fixes to instance GPU configs

iliadsh:
Similar to machine types, guest accelerators (gpus) take a url which we should build ourselves. Also, instances with a GPU attached need their maintenance behavior set to TERMINATE

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gceinstance/CreateGceInstanceStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gceinstance/CreateGceInstanceStep.java
@@ -68,7 +68,7 @@ public class CreateGceInstanceStep implements Step {
 
   private static final String EXTERNAL_IP_TYPE = "ONE_TO_ONE_NAT";
   private static final String EXTERNAL_IP_NAME = "External IP";
-  private static final String GPU_HOST_MAINTENANCE_BEHAVIOUR = "TERMINATE";
+  private static final String HOST_MAINTENANCE_BEHAVIOUR = "TERMINATE";
 
   public CreateGceInstanceStep(
       ControlledGceInstanceResource resource,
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gceinstance/CreateGceInstanceStep.java
```diff
@@ -207,9 +210,13 @@ static Instance setFields(
...
                   accelerator ->
                       new AcceleratorConfig()
-                          .setAcceleratorType(accelerator.getType())
+                          .setAcceleratorType(
+                              GcpUtils.toAcceleratorTypeString(
+                                  projectId, zone, accelerator.getType()))
                           .setAcceleratorCount(accelerator.getCardCount()))
               .collect(Collectors.toList()));
+      // Instances with guest accelerators do not support live migration
+      instance.setScheduling(new Scheduling().setOnHostMaintenance(GPU_HOST_MAINTENANCE_BEHAVIOUR));
```
cbookg, line 219: i think we should just set this all the time to be consistent.

cbookg (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gceinstance/CreateGceInstanceStep.java
```diff
@@ -173,14 +176,14 @@ private static Instance createInstanceModel(
...
+      String projectId,
       String zone,
       String serviceAccountEmail,
       String workspaceUserFacingId,
       String cliServer,
       Instance instance,
       String gitHash) {
-    String machineType =
-        String.format("zones/%s/machineTypes/%s", zone, creationParameters.getMachineType());
+    String machineType = GcpUtils.toMachineTypeString(zone, creationParameters.getMachineType());
```
ddietterich, line 186: If this input is bogus, what kind of error comes out?

ddietterich (approved):
A question otherwise, looks OK

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gceinstance/CreateGceInstanceStep.java
iliadsh: I guess it depends how bogus the input is. If it's just some nonsense it will probably error with `Invalid value for field 'resource.machineType': 'stuff'. The URL is malformed` but if it's just not supported for the zone you get `Invalid value for field 'resource.machineType': 'stuff. Machine type with name 'stuff' does not exist in zone 'zone'`

PR #1364, opened 2023-06-28
[PF-2833] Support revoking temporary grants for GCE instances

zloery:
A handful of grant revoke flights are failing in devel as WSM was missing the logic for cleaning up GCE instance temporary grants. These are the only two places I found that switch on `WsmResourceType`, so hopefully we won't hit other cases like this.


jmczerk (approved):

service/src/main/java/bio/terra/workspace/service/grant/flight/RevokeStep.java
```diff
@@ -288,6 +296,52 @@ private void revokeResourceNotebook(
     }
   }
 
+  private void revokeResourceGceInstance(
```
dexamundsen, line 299: minor: this looks very similar to the one for Notebooks. Would a common function help? something on the lines of 

revokeResourceGrant(SetIamPolicy, grantData, ResourceType) {}

iliadsh (approved):

PR #1366, opened 2023-06-30
[BENCH-727] Ignore NotFound error in Delete functions

dexamundsen:
- Delete flight and steps are already idempotent
- Adding option in the implementation function to not error out with NotFound during delete
- Refactored AWS notebook UT class to a abstract base class + 2 concrete test classes. This prevents base class tests from executing twice


zloery (approved):
LGTM, it looks like there are a few other AWS unit tests to update though

dexamundsen:
> LGTM, it looks like there are a few other AWS unit tests to update though

Fixed the UT. 

Additional change - refactored the AWS Notebook UT (created a abstract base class) to prevent the base class tests from being executed twice

jmczerk (approved):

PR #1375, opened 2023-07-07
[BENCH-319] UT for AWS cloud context functions

dexamundsen:
- UT for AWS cloud context functions
- Bump resource discovery lib version (required for tests)


zloery (approved):

service/src/test/java/bio/terra/workspace/common/fixtures/ControlledAwsResourceFixtures.java
```diff
@@ -47,8 +47,14 @@ public class ControlledAwsResourceFixtures {
   public static final String TENANT_ALIAS = "tenant-saas";
   public static final String ENVIRONMENT_ALIAS = "unit-test-env";
   public static final String AWS_REGION = "us-east-1";
+  public static final String AWS_ENVIRONMENT_WSM_ROLE_ARN =
+      "arn:aws:iam::34567890012:role/WorkspaceManagerRole";
```
nfoen, line 51: nit: Are these account names used for anything? If not, is it possible to choose numbers that are even more obviously fake (e.g. 00000000001). As someone unfamiliar with tests, I briefly wonder if these were in an actual account before I noticed the patterns.

nfoen (approved):
lgtm

service/src/test/java/bio/terra/workspace/common/fixtures/ControlledAwsResourceFixtures.java
dexamundsen: They are dummy values, changed as suggested

PR #1383, opened 2023-07-13
[WOR-1102] Anyone with read access to a workspace should be able to read all resources in the workspace.

cahrens:
Fixes bug https://broadworkbench.atlassian.net/browse/WOR-1102, following Dan's [recommendation](https://broadinstitute.slack.com/archives/C01JVP0MLLX/p1688655016319019):

"That is a bug. The philosophy we are trying to implement is that everyone who is a reader (or above) on the workspace can see all of the resources in a workspace. Since enumeration shows the complete resource details there is no use "hiding" requests for specific resources.

I think we need to change the permission check on the get resource to check the workspace permission and not the resource permission."

I tested that this fixes the bug by:
1. Deploying this branch to a bee
2. Creating an Azure workspace on the bee
3. Sharing the workspace with a second user
4. Creating a Jupyter instance as the second user
5. And then deleting the workspace as the original owner of the workspace

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
@@ -79,8 +79,8 @@ public class ControlledResourceMetadataManager {
     ControlledResource controlledResource = resource.castToControlledResource();
     String samName = controlledResource.getCategory().getSamResourceName();
 
-    // Everyone who is a reader (or above) on the workspace can see all of the resources in a workspace.
-    // Thus the authorization check is againt the workspace and not the resource itself.
+    // Everyone who is a reader (or above) on the workspace can see all the resources in a
+    // workspace. Thus this authorization check is against the workspace and not the resource.
     if (StringUtils.equals(action, SamControlledResourceActions.READ_ACTION)) {
       checkResourceAuthz(userRequest, SamConstants.SamResource.WORKSPACE, workspaceUuid, action);
       return controlledResource;
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
@@ -79,8 +79,10 @@ public class ControlledResourceMetadataManager {
     ControlledResource controlledResource = resource.castToControlledResource();
     String samName = controlledResource.getCategory().getSamResourceName();
 
+    // Everyone who is a reader (or above) on the workspace can see all of the resources in a workspace.
+    // Thus the authorization check is againt the workspace and not the resource itself.
     if (StringUtils.equals(action, SamControlledResourceActions.READ_ACTION)) {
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
```diff
@@ -79,8 +79,10 @@ public ControlledResource validateControlledResourceAndAction(
     ControlledResource controlledResource = resource.castToControlledResource();
     String samName = controlledResource.getCategory().getSamResourceName();
 
+    // Everyone who is a reader (or above) on the workspace can see all the resources in a
+    // workspace. Thus this authorization check is against the workspace and not the resource.
     if (StringUtils.equals(action, SamControlledResourceActions.READ_ACTION)) {
-      checkResourceAuthz(userRequest, samName, resourceId, action);
+      checkResourceAuthz(userRequest, SamConstants.SamResource.WORKSPACE, workspaceUuid, action);
```
marctalbott, line 85: Should this be checking if the user can read either the workspace OR the resource? Are there cases where a user has read access to a controlled resource, but not the workspace that would start to fail after this change?

cahrens: Since it is a controlled resource, I assume the user has to have access to the workspace to be able to have any access to the resource at all.

FWIW, no tests failed.

service/src/test/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManagerTest.java
```diff
@@ -0,0 +1,125 @@
...
+import bio.terra.workspace.service.stage.StageService;
+import bio.terra.workspace.service.workspace.WorkspaceService;
+import java.util.UUID;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.mock.mockito.MockBean;
+
+public class ControlledResourceMetadataManagerTest extends BaseUnitTest {
```
cahrens, line 27: There were no unit tests for this class-- it was pulled out of someplace else for reuse, and some connected or integration tests probably hit it incidentally. I didn't want to try to add to a connected test, so created a pure unit test for the purposes of testing this specific codepath.

aherbst-broad (approved):
lgtm

aherbst-broad:
@ddietterich tagging you on this PR for visibility

service/src/main/java/bio/terra/workspace/service/resource/controlled/ControlledResourceMetadataManager.java
aherbst-broad: I think that's the case. There is the notion of a "private resource" that is managed by an application on behalf of a user (think: leo managed azure vms). The private resource user must be still be a reader on the workspace in that case. I'm not familiar with a mechanism by which we can grant access to a single resource on a workspace without first giving a user at least reader access.

marctalbott (approved):

PR #1385, opened 2023-07-14
PF-2911: Cloning reference shouldn't fail with region constraint violation

mikenev:
References are just pointers, so they don't really have a location. The clone instruction can pass the location of the source through, but policy validation shouldn't note a region violation if we're cloning a reference.


zloery (approved):

harrismendell (approved):

harrismendell:
Thanks for the quick fix!

PR #1387, opened 2023-07-14
[PF-2896]: Add dangling resource cleanup service

rogerwangcs:
Adds a service that polls all READY state controlled compute resources, checking whether their associated cloud resource exists and deleting their metadata + sam resource if it doesn't. This change is required as a part of the dataproc cluster resource work. Clusters are generally configured with scheduled deletion as a best practice. The idle deletion times are configured to be any time between 5min and 2 days. We poll every 15 minutes.


service/src/main/resources/application.yml
```diff
@@ -76,6 +76,11 @@ workspace:
     polling-interval: "15m"
     startup-wait: "5s"
 
+  dangling-resource-cleanup:
+    enabled: true
+    polling-interval: "15m"
```
yuhuyoyo, line 81: 15 min feels like a lot. is there security/cost concern with those dangling resources? can we do daily?

service/src/main/java/bio/terra/workspace/service/danglingresource/DanglingResourceCleanupService.java
```diff
@@ -0,0 +1,213 @@
...
+        resource.getResourceId(),
+        /* forceDelete= */ false,
+        /* resultPath= */ null,
+        wsmSaRequest);
+  }
+
+  /**
+   * Utility method to check if a given wsm controlled resource's associated cloud resourced exists.
+   */
+  private boolean cloudResourceExists(ControlledResource resource) {
```
yuhuyoyo, line 158: static

service/src/main/java/bio/terra/workspace/service/danglingresource/DanglingResourceCleanupService.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  /**
+   * Launches a resource delete flight for a given dangling resource. The controlled resource
+   * deletion flight is reused here to delete the sam resource and metadata entry.
+   */
+  private void launchDanglingResourceCleanupFlight(
+      ControlledResource resource, AuthenticatedUserRequest wsmSaRequest) {
+    String jobId = UUID.randomUUID().toString();
+    controlledResourceService.deleteControlledResourceAsync(
```
yuhuyoyo, line 146: this flight DeleteControlledResourcesFlight can actually takes in multiple resources. so instead of launching one flight per resources ( i think there's a limit on how many flights there can be atm), can you pass in all the resources at once?
the input param is CONTROLLED_RESOURCES_TO_DELETE.

Also if you specify operationType.system_cleanup, it will be system cleanup. currently it's just delete.

service/src/main/resources/application.yml
yuhuyoyo: curious, where is the output of the job stored?

rogerwangcs: Currently just logging resources that are cleaned up.

yuhuyoyo: sorry i mean for a job that's running in a cluster, where does the researchers see the result aftere the cluster is deleted?

rogerwangcs: Oh sorry I misinterpreted, all in the cloud console. Jobs are independent of clusters so you can view all jobs + outputs regardless of whether the clusters that ran them still exist or not,

service/src/main/java/bio/terra/workspace/service/danglingresource/DanglingResourceCleanupService.java
rogerwangcs: My initial thinking was that by using individual flights per resource deletion, we can provide flight metadata such as workspaceId + resourceId/type, etc. In fact, it seems like workspaceId is a requirement to execute the flight, so I'm not sure how we would run a flight to delete resources across multiple workspaces.

Thoughts on still cleaning up resources in individual flights but updating the operation type to SYSTEM_CLEANUP?

yuhuyoyo: ooh i forgot about that.. sure

service/src/test/java/bio/terra/workspace/service/danglingresource/DanglingResourceCleanupServiceTest.java
```diff
@@ -0,0 +1,150 @@
...
+      throw new RuntimeException("Failed to delete notebook instance", e);
+    }
+
+    // Call dangling resource cleanup
+    danglingResourceCleanupService.cleanupResourcesSuppressExceptions();
+
+    // Verify that the notebook is deleted, polling to ensure dangling resource cleanup flight to
+    // delete the notebook has completed
+    boolean isDeleted = false;
+    for (int retryCount = 0; retryCount < ASSERT_RESOURCE_CLEANUP_RETRY_COUNT; retryCount++) {
```
yuhuyoyo, line 130: i don't like we are pulling this instead of the flight.. but it's a list of flight so either way not very clean.

yuhuyoyo (approved):

service/src/test/java/bio/terra/workspace/service/danglingresource/DanglingResourceCleanupServiceTest.java
rogerwangcs: Agreed 😞

service/src/main/java/bio/terra/workspace/service/danglingresource/DanglingResourceCleanupService.java
```diff
@@ -0,0 +1,235 @@
...
+  private final ResourceDao resourceDao;
+  private final CronjobDao cronjobDao;
+  private final SamService samService;
+  private final CrlService crlService;
+  private final ScheduledExecutorService scheduler;
+
+  private final List<WsmResourceType> DANGLING_RESOURCE_TYPES =
+      List.of(
+          WsmResourceType.CONTROLLED_GCP_AI_NOTEBOOK_INSTANCE,
+          WsmResourceType.CONTROLLED_GCP_GCE_INSTANCE,
```
zloery, line 56: Is it currently possible for notebooks or GCE instances to be deleted on the cloud side yet? My understanding was this was dataproc only atm

service/src/main/java/bio/terra/workspace/service/danglingresource/DanglingResourceCleanupService.java
```diff
@@ -0,0 +1,235 @@
...
+    if (!configuration.isEnabled()) {
+      return;
+    }
+    logger.info("Beginning dangling resource cleanup cronjob");
+    // Use a one-second shorter duration here to ensure we don't skip a run by moving slightly too
+    // quickly.
+    Duration claimTime = configuration.getPollingInterval().minus(Duration.ofSeconds(1));
+    // Attempt to claim the latest run of this job to ensure only one pod runs the cleanup job.
+    if (!cronjobDao.claimJob(DANGLING_RESOURCE_CLEANUP_JOB_NAME, claimTime)) {
+      logger.info("Another pod has executed this job more recently. Ending resource cleanup.");
```
zloery, line 118: nit here and in other cronjobs: since we're running multiple cronjobs, can you include the job name in this log message? (e.g. something like "Another pod has executed dangling resource cleanup more recently, skipping this execution")

service/src/main/java/bio/terra/workspace/service/danglingresource/DanglingResourceCleanupService.java
```diff
@@ -0,0 +1,235 @@
...
+  private static boolean cloudResourceExists(ControlledResource resource, CrlService crlService) {
+    return switch (resource.getResourceType()) {
+      case CONTROLLED_GCP_AI_NOTEBOOK_INSTANCE -> aiNotebookInstanceExists(
+          resource.castByEnum(WsmResourceType.CONTROLLED_GCP_AI_NOTEBOOK_INSTANCE), crlService);
+      case CONTROLLED_GCP_GCE_INSTANCE -> gceInstanceExists(
+          resource.castByEnum(WsmResourceType.CONTROLLED_GCP_GCE_INSTANCE), crlService);
+      case CONTROLLED_GCP_DATAPROC_CLUSTER -> dataprocClusterExists(
+          resource.castByEnum(WsmResourceType.CONTROLLED_GCP_DATAPROC_CLUSTER), crlService);
+      default -> {
+        logger.info(
```
zloery, line 183: IMO this would indicate developer error, this should log at warning or error

service/src/main/java/bio/terra/workspace/service/danglingresource/DanglingResourceCleanupService.java
rogerwangcs: Not currently, should I remove notebooks and gce instances from the list here?

zloery (approved):

PR #1392, opened 2023-07-18
[WOR-1134] Fix pod name when creating DB

aherbst-broad:
The Azure database creation step is failing due to an underscore present in the pod name, which violates their naming convention.


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/CreateAzureDatabaseStep.java
```diff
@@ -75,7 +75,8 @@ public CreateAzureDatabaseStep(
   }
 
   private String getPodName(String newDbUserName) {
-    return this.workspaceId.toString() + this.resource.getDatabaseName() + newDbUserName;
+    return (this.workspaceId.toString() + this.resource.getDatabaseName() + newDbUserName)
+        .replace('_', '-');
```
cahrens, line 79: A comment explaining or linking to the naming rules could be useful.

marctalbott (approved):

dvoet (approved):

PR #1395, opened 2023-07-19
[PF-2918] - Ignore boot service child command exit codes and log its stdout, stderr

rogerwangcs:
As reported in [PF-2918](https://broadworkbench.atlassian.net/browse/PF-2918), bucket mounting in workspaces with referenced resources that the mounting user does not have access to causes the `terra-instance-boot.service` to fail. This results the gcsfuse mount points for already mounted resources to die and result in `Transport endpoint is not connected` errors.

The startup script internal runs `terra resource mount`, which returns an exit code of 1 even if any of the resources do not mount due to permissions errors or 404 bucket errrors. This behavior is intentional and useful when the command is used in scripting. ~Instead of modifying the behavior of the CLI, we address the issue by adding a `-` to the terra startup service systemd unit, which tells systemd to ignore nonzero return codes and keep the service running.~ This change updates boot script so that child command exit codes are ignored and logged in the terra config folder.

[PF-2918]: https://broadworkbench.atlassian.net/browse/PF-2918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -94,6 +94,7 @@ readonly USER_BASHRC="${USER_HOME_DIR}/.bashrc"
 readonly USER_BASH_PROFILE="${USER_HOME_DIR}/.bash_profile"
 
 readonly POST_STARTUP_OUTPUT_FILE="${USER_TERRA_CONFIG_DIR}/post-startup-output.txt"
+readonly TERRA_BOOT_SERVICE_OUTPUT_FILE="${USER_TERRA_CONFIG_DIR}/boot-output.txt"
 
 # When JupyterLab is provided by a Docker container, the default Deep Learning images
 # pick up jupyter_notebook_config.py provided by the host VM.
@@ -629,11 +630,17 @@ cat << EOF >"${TERRA_BOOT_SCRIPT}"
 #!/bin/bash
 # This script is run on instance boot to configure the instance for terra.
 
+# Send stdout and stderr from this script to a file for debugging.
+exec >> "${TERRA_BOOT_SERVICE_OUTPUT_FILE}"
+exec 2>&1
+
 # Pick up environment from the ~/.bashrc
 source "${USER_BASHRC}"
 
 # Mount terra workspace resources
 "${USER_HOME_LOCAL_BIN}/terra" resource mount
+
+exit 0
 EOF
 chmod +x "${TERRA_BOOT_SCRIPT}"
 chown ${JUPYTER_USER}:${JUPYTER_USER} "${TERRA_BOOT_SCRIPT}"
@@ -645,7 +652,7 @@ Description=Configure environment for terra
 After=jupyter.service
 
 [Service]
-ExecStart=-${TERRA_BOOT_SCRIPT}
+ExecStart=${TERRA_BOOT_SCRIPT}
 User=${JUPYTER_USER}
 RemainAfterExit=yes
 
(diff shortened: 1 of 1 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -645,7 +645,7 @@ Description=Configure environment for terra
 After=jupyter.service
 
 [Service]
-ExecStart=${TERRA_BOOT_SCRIPT}
+ExecStart=-${TERRA_BOOT_SCRIPT}
```
mbookman, line 648: My concern with this is that the TERRA_BOOT_SCRIPT could error out for many reasons, including syntax errors and other trivial bugs. We'd want to catch that and it would be better if we had some hard failures as a result. This doesn't necessarily get us there, but ignoring the exit code I think will be another mask.

Does `terra resource mount` have fine grained exit codes? It would be nice if it did map some errors to things like:

https://opensource.apple.com/source/xnu/xnu-201/bsd/sys/errno.h.auto.html

```
#define	ENOENT		2		/* No such file or directory */
...
#define	EACCES		13		/* Permission denied */
```

Then the TERRA_BOOT_SCRIPT could at least handle known things:
```
# Mount terra workspace resources
"${USER_HOME_LOCAL_BIN}/terra" resource mount
readonly MOUNT_RC="$?"

case "${MOUNT_RC}" in 
  # If a user doesn't have permissions or the bucket doesn't exist, don't error out here
  0) # EOK
  2) # ENOENT
  13) # EACCESS
    >&2 echo "Ignoring error from 'terra resource mount'"
    exit 0
    ;;
  *)
    >&2 echo "Unexpected error '${MOUNT_RC}' from 'terra resource mount'"
    exit 1
esac
```

What do you think?

mbookman (changes requested):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
rogerwangcs: Appreciate the quick review! Agree that this is the more correct approach. It would be good to add these return codes to single resource mounts. For mounting multiple resources, we aggregate the list of mount errors into a single return code. There aren't fine grained exit codes, It's essentially `1 if sum(exitCodes) > 1 else 0`. It's not clear to me how to best bubble up mount errors with varying causes. We could modify the behavior to only return a nonzero exit code if an _**unexpected**_ error occurs. Would that be a good intermediary solution?

amygdala: that seems reasonable to me fwiw

rogerwangcs: Made changes so that the boot service no longer ignores non zero exit codes from the boot script itself.

Errors in the script itself will prevent the service from running. Non zero exit codes from commands executed inside the boot script will no longer cause the service to fail, but instead be logged to `.terra/boot-output.txt` for debugging.

We don't want errors from child commands to break the service since they don't necessarily occur each vm startup and can likely be resolved by the user (in the case of mounting bad referenced buckets).

PTAL @wnojopra @amygdala

amygdala: This seems like a reasonable approach to me— if it later turns out that we want to bubble up some of the child errors differently, we can iterate.

amygdala (approved):

wnojopra (approved):
Thanks for making sure the errors don't get ignored.

PR #1396, opened 2023-07-20
Test azure usage attribute

sergiygetlin:


sergiygetlin:
This change was for testing only.

PR #1408, opened 2023-07-26
Stop deleting workspace records on duplicate create request

zloery:
Previously, workspace/cloud context/resource creation flights would always delete DB records on undo with the `DELETE_ON_FAILURE` rule, even if the flight was not the one which created those objects originally. 

This change adds a state transition to validate that the object was created by the calling flight. The method does not throw an error if this validation fails (as I believe we just want to return the 400 response from the `do` step), but no longer deletes the DB row.

Several controlled resource tests inadvertently depended on this behavior and had to be modified. Additionally, I changed the createControlledResource flight to throw a dismal failure if the flight fails between releasing the resource state lock (the last step) and flight completion to keep resource state in line with behavior for workspaces and cloud contexts. Functionally, this is a test-only change.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/db/ResourceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/ResourceDao.java
@@ -9,6 +9,7 @@ import bio.terra.common.db.ReadTransaction;
 import bio.terra.common.db.WriteTransaction;
 import bio.terra.workspace.common.exception.InternalLogicException;
 import bio.terra.workspace.common.logging.model.ActivityLogChangeDetails;
+import bio.terra.workspace.db.exception.ResourceStateConflictException;
 import bio.terra.workspace.db.model.DbResource;
 import bio.terra.workspace.db.model.DbUpdater;
 import bio.terra.workspace.db.model.UniquenessCheckAttributes;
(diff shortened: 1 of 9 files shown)


service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -805,16 +806,27 @@ public void createResourceFailure(
...
+    DbResource dbResource =
+        getDbResourceFromIds(resource.getWorkspaceId(), resource.getResourceId());
     switch (resourceStateRule) {
       case DELETE_ON_FAILURE -> {
-        deleteResourceWorker(
-            resource.getWorkspaceId(), resource.getResourceId(), /*resourceType=*/ null);
+        // There is no guarantee this is the flight which created this resource. Validate that it
+        // is before attempting to delete the workspace.
+        try {
+          stateDao.updateState(
```
ddietterich, line 816: An alternative would be to always make the state change to BROKEN. Then if it is DELETE_ON_FAILURE, do the delete. 

I am thinking we want to do the catch in both cases to avoid throwing a new error that Stairway would see as a dismal failure. Right?

ddietterich (commented):
I think we need to catch the conflict exception in the BROKEN_ON_FAILURE case also.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
zloery: Good point, we need to catch in both cases to avoid that error message.

IMO using the switch statement and setting the state to `NOT_EXISTS` for `DELETE_ON_FAILURE` is more verbose but worthwhile since it's easier to follow, agreed they're functionally equivalent though.

ddietterich (approved):
Works for me. Thanks for the fix!

aherbst-broad (approved):
Thanks for the quick fix.

PR #1409, opened 2023-07-28
[BENCH-859] Handle duplicate delete requests from applications

dexamundsen:
error: cloud context delete fails
root cause: previous delete is still running and conflicts with cloud context delete
fix: don't create duplicate delete resource flights

sample flow:

- User calls resource delete, which times out but is still running on server
- User now calls Workspace delete (internally calls cloud context delete)
- Cloud context delete created a new delete flight for the resource (since the previous delete is running)
- Flight flight completes and resource entries are deleted in DB
- Second flight attempts to update DB entry for resource (to start delete) and fails with 'resource not found'

with above fix:

- Cloud context delete will create a new resource delete flight only if there is no existing delete flight for that resource
- If there exists a flight, add it to the list and wait for it as usual


zloery (approved):

ddietterich (approved):
Very tidy

PR #1415, opened 2023-08-02
PF-2713 - Update Sam auth domains when groups change (add only)

mikenev:
During a clone operation, if WSM notices that the groups change, then we'll send the new groups to Sam to add to the auth domain. Right now, we're only able to add new groups and cannot remove groups.

There's a TODO in this review to switch to using the service account rather than the user request. That's going to require a change in Sam since the service account is a manager and not an owner of the resource. It doesn't block the user-initiated clone/data-collection import scenario. It will block background syncing of policies (where an update on the data collection policy should propagate to the workspace). However, that piece isn't implemented yet and there are other portions of the flight that currently use the user request rather than the service account so we already have that dependency.


ddietterich (approved):
LGTM

zloery (approved):

PR #1421, opened 2023-08-04
Support multiple dataproc initialization scripts

rogerwangcs:
Dataproc clusters support passing in a list of startup scripts. The wsm api should also support it to allow users to provide their own custom startup scripts to install any software they need.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/CreateDataprocClusterStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/CreateDataprocClusterStep.java
@@ -49,6 +49,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/CreateDataprocClusterStep.java
```diff
@@ -221,13 +222,14 @@ static Cluster setFields(
 
     // Set initialization script
     // TODO PF-2828: Add WSM default post-startup script
-    if (!StringUtils.isEmpty(creationParameters.getInitializationScript())) {
+    List<String> initializationScripts = creationParameters.getInitializationScripts();
+    if (initializationScripts != null) {
```
zloery, line 226: Does this need to handle an empty list being passed in? As-is this will pass an empty list to the config object, which seems different than the previous behavior of leaving it unset

rogerwangcs: Good catch. Passing in an empty list is OK and doesn't cause issues.

zloery (approved):

dexamundsen (approved):

mikenev (approved):
lgtm

PR #1422, opened 2023-08-05
BENCH-891: MergeCheck should allow non-equal group constraints

mikenev:
MergeCheck was adding a conflict if the group constraints would change at all during a merge. That was correct in M1, but in M2 we can add additional group constraints.

The only time we should fail now is if we attempt to remove group constraints. That shouldn't happen during a merge operation, so I don't see the need for this check. 

I left the tests in to cover regressions and I also needed to test this group scenario.


ddietterich (approved):
lgtm

PR #1423, opened 2023-08-06
[BENCH-895] Remove duplicated definitions of WsmResource.castByEnum

dexamundsen:
Remove duplicated definitions of WsmResource.castByEnum by moving default definition to the interface. This can still be overridden by implementing classes if needed. 
Minor cleanup and addressed warnings in files already being modified

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/ControlledAzureDatabaseResource.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/ControlledAzureDatabaseResource.java
@@ -1,6 +1,5 @@
 package bio.terra.workspace.service.resource.controlled.cloud.azure.database;
 
-import bio.terra.common.exception.BadRequestException;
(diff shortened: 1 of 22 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/ControlledAzureDatabaseResource.java
```diff
@@ -158,7 +147,7 @@ public void addCreateSteps(
 
   @VisibleForTesting
   List<Step> getAddSteps(FlightBeanBag flightBeanBag) {
-    var steps = new ArrayList<Step>();
+    ArrayList<Step> steps = new ArrayList<>();
```
zloery, line 150: nit: mostly a style issue, but I'm fine using var for cases like these where the type is clear from the declaration or like below where it's clear from the var name

zloery (approved):

nfoen (approved):
lgtm

PR #1424, opened 2023-08-07
BENCH-892: Patch workspace policies should allow updating groups

mikenev:
We have an endpoint for updating policies on workspaces, but it didn't appropriately handle adding groups. This update will let that endpoint also call Sam if the group policies on the workspace change.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/policy/TpsUtilities.java
+++ b/service/src/main/java/bio/terra/workspace/service/policy/TpsUtilities.java
@@ -2,6 +2,7 @@ package bio.terra.workspace.service.policy;
 
 import bio.terra.policy.model.TpsPolicyInputs;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 
 public class TpsUtilities {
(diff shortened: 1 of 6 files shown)


service/src/main/java/bio/terra/workspace/service/policy/TpsUtilities.java
```diff
@@ -31,6 +33,60 @@ public static List<String> getGroupConstraintsFromInputs(TpsPolicyInputs inputs)
...
+            ? new HashSet<>()
+            : new HashSet<>(
+                TpsUtilities.getGroupConstraintsFromInputs(
+                    updatedPolicies.getEffectiveAttributes()));
+
+    for (String updatedGroup : updatedGroups) {
+      if (!originalGroups.contains(updatedGroup)) {
+        result.add(updatedGroup);
+      }
+    }
```
zloery, line 58: nit: you can use `updatedGroups.removeAll(originalGroups)` here

service/src/main/java/bio/terra/workspace/service/policy/TpsUtilities.java
```diff
@@ -31,6 +33,60 @@ public static List<String> getGroupConstraintsFromInputs(TpsPolicyInputs inputs)
...
+            ? new HashSet<>()
+            : new HashSet<>(
+                TpsUtilities.getGroupConstraintsFromInputs(
+                    updatedPolicies.getEffectiveAttributes()));
+
+    for (String originalGroup : originalGroups) {
+      if (!updatedGroups.contains(originalGroup)) {
+        result.add(originalGroup);
+      }
+    }
```
zloery, line 85: Same with `originalGroups.removeAll(updatedGroups)` here

zloery (approved):

ddietterich (approved):
LGTM

PR #1425, opened 2023-08-07
WOR-1157 Add consumer pact for TPS

blakery:
[WOR-1157](https://broadworkbench.atlassian.net/browse/WOR-1157)

Adds consumer-side contract tests from WSM to TPS.

The changes in this PR don't include changes to publish all pacts to pact broker, yet. 

[WOR-1157]: https://broadworkbench.atlassian.net/browse/WOR-1157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
+++ b/service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
@@ -9,20 +9,31 @@ import au.com.dius.pact.consumer.junit5.PactConsumerTestExt;
 import au.com.dius.pact.consumer.junit5.PactTestFor;
 import au.com.dius.pact.core.model.RequestResponsePact;
 import au.com.dius.pact.core.model.annotations.Pact;
-import bio.terra.policy.model.TpsComponent;
-import bio.terra.policy.model.TpsObjectType;
-import bio.terra.policy.model.TpsPolicyInput;
-import bio.terra.policy.model.TpsPolicyInputs;
+import bio.terra.policy.model.*;
 import bio.terra.workspace.app.configuration.external.FeatureConfiguration;
 import bio.terra.workspace.app.configuration.external.PolicyServiceConfiguration;
+import bio.terra.workspace.service.iam.AuthenticatedUserRequest;
 import bio.terra.workspace.service.policy.TpsApiDispatch;
+import bio.terra.workspace.service.policy.exception.PolicyServiceDuplicateException;
+import bio.terra.workspace.service.policy.exception.PolicyServiceNotFoundException;
+import bio.terra.workspace.service.resource.exception.PolicyConflictException;
+import bio.terra.workspace.service.workspace.WorkspaceService;
+import bio.terra.workspace.service.workspace.model.CloudPlatform;
+import bio.terra.workspace.service.workspace.model.Workspace;
+import bio.terra.workspace.service.workspace.model.WorkspaceStage;
+import org.apache.http.entity.ContentType;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -0,0 +1,555 @@
...
+    var inputs =
+        new TpsPolicyInputs()
+            .inputs(List.of(new TpsPolicyInput().name("test_name").namespace("test_namespace")));
+    var tpsPao =
+        new TpsPaoGetResult()
+            .objectId(existingPolicyId)
+            .attributes(inputs)
+            .effectiveAttributes(inputs);
+    var result = dispatch.listValidRegionsForPao(tpsPao, CloudPlatform.AZURE);
+    assertNotNull(result);
```
blakery, line 443: Most of these are just asserting the result is not null, because we don't generally care about the actual values returned, just that we can deserialize and handle the result (in fact, most of the generators for the responses used will just generate random strings). 
This is largely because a lot of the time we're just passing the value back to the user.

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -0,0 +1,555 @@
...
+  @Test
+  @PactTestFor(pactMethod = "updatePaoPreexistingNoConflicts")
+  public void updatingAnExistingPaoWithNoConflicts() throws Exception {
+    var remove =
+        new TpsPolicyInputs()
+            .inputs(List.of(new TpsPolicyInput().name("test_name").namespace("test_namespace")));
+    var add =
+        new TpsPolicyInputs()
+            .inputs(List.of(new TpsPolicyInput().name("new_name").namespace("new_namespace")));
+    var result = dispatch.updatePao(existingPolicyId, add, remove, TpsUpdateMode.FAIL_ON_CONFLICT);
```
blakery, line 388: We only use the modes `FAIL_ON_CONFLICT' and 'DRY_RUN', and there isn't really any difference in request/response shape between them.
Even though we're only calling it with `FAIL_ON_CONFLICT` in our actual tests, the pact still includes all the values.

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -0,0 +1,555 @@
...
+  // since our client doesn't include the encoding in the content type header
+  static Map<String, String> contentTypeJsonHeader = Map.of("Content-Type", "application/json");
+
+  // The policy ids aren't significant - hardcoded instead of random to avoid changing the pact on
+  // every run
+  static UUID existingPolicyId = UUID.fromString("bea1edd0-a8e6-4d60-a613-e8e065f21616");
+  static String existingPolicyState = "an existing policy";
+  static String existingPolicyProviderStateValue = "${policyId}";
+  static UUID secondPolicyId = UUID.fromString("a254714b-4519-4ce4-ad87-19a7143376f4");
+  static String secondPolicyState = "a second existing policy";
```
blakery, line 52: I don't have a better name. The details of what the policy are aren't important here: just any policy, in order to test linking them.
When the provider side pacts are written, maybe someone will come up with better names, and we can go back and update this.

aherbst-broad (approved):

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
sergiygetlin: Should we care about the shape of the response? Not the values, but the structure?

sergiygetlin (approved):

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
blakery: I added a check that we can convert it to the object returned by the API.
But other than that, no - I think that would be testing TPS.

PR #1426, opened 2023-08-08
BENCH-909: Handle null more gracefully

yuhuyoyo:
deleteCloudContextResourceFlight set resource_id in the input param and we are not expecting that in the hook.


zloery (approved):

ddietterich (approved):
LGTM

PR #1427, opened 2023-08-08
Add retry to dataproc patch asserts

rogerwangcs:
Adds a missing update property `idleDeleteTtl`. Change asserts to retries that ensure that the values have been updated.

Addresses the integration test error in the downstream submodule update:
https://github.com/verily-src/terra-service-ws-manager/actions/runs/5791161286/job/15695514025

### DIFF
--- a/integration/src/main/java/scripts/testscripts/PrivateControlledDataprocClusterLifeCycle.java
+++ b/integration/src/main/java/scripts/testscripts/PrivateControlledDataprocClusterLifeCycle.java
@@ -204,24 +204,19 @@ public class PrivateControlledDataprocClusterLifeCycle extends WorkspaceAllocate
     // Update the cluster lifecycle rule through WSM. Cluster lifecycle rules cannot be updated in
     // tandem with other parameters, so we update it separately.
     String newIdleDeleteTtl = "1800s";
-    String newAutoDeleteTtl = "3600s";
     resourceUserApi.updateDataprocCluster(
         new UpdateControlledGcpDataprocClusterRequestBody()
             .updateParameters(
                 new ControlledDataprocClusterUpdateParameters()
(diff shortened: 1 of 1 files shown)


integration/src/main/java/scripts/utils/RetryUtils.java
```diff
@@ -0,0 +1,99 @@
+package scripts.utils;
```
rogerwangcs, line 1: Took this util class from the service project. Do we have a better pattern of retry + assert?

dexamundsen: for later - this might make sense in TCL or wsm- common lib(?)

integration/src/main/java/scripts/testscripts/PrivateControlledDataprocClusterLifeCycle.java
```diff
@@ -193,38 +193,32 @@ protected void doUserJourney(TestUserSpecification testUser, WorkspaceApi worksp
...
+    RetryUtils.getWithRetry(
+        cluster -> newNumPrimaryWorkers == cluster.getConfig().getWorkerConfig().getNumInstances(),
+        () -> dataproc.projects().regions().clusters().get(projectId, region, clusterId).execute(),
+        "Timed out waiting for cluster to update");
 
     // Update the cluster lifecycle rule through WSM. Cluster lifecycle rules cannot be updated in
     // tandem with other parameters, so we update it separately.
     String newIdleDeleteTtl = "1800s";
-    String newAutoDeleteTtl = "3600s";
     resourceUserApi.updateDataprocCluster(
```
mikenev, line 207: No more auto-delete? Maybe that's fine with the idle delete anyway.

zloery (approved):

PR #1433, opened 2023-08-09
[PF-2776] - Add dataproc startup script

rogerwangcs:
Adds the default dataproc startup script to configure it with the terra environment and tools (repos, bucket mounts).

Followups:
- Enable integration test in downstream repo.
- Update integration test to check for startup script completion metadata


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,781 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+# The linux user that JupyterLab will be running as. It's important to do some parts of setup in the
+# user space, such as setting Terra CLI settings which are persisted in the user's $HOME.
+# This post startup script is not run by the same user.
+readonly DATAPROC_USER="dataproc"
```
rogerwangcs, line 37: So by default, the dataproc jupyter optional component configured the jupyter service to run jupyter as root. I changed it to the dataproc user and gave it sudo (line 175).

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,781 @@
...
+# Create an alias for cases when we need to run a shell command as the dataproc user.
+# Note that we deliberately use "bash -l" instead of "sh" in order to get bash (instead of dash)
+# and to pick up changes to the .bashrc.
+#
+# This is intentionally not a Bash function, as that can suppress error propagation.
+# This is intentionally not a Bash alias as they are not supported in shell scripts.
+readonly RUN_AS_DATAPROC_USER="sudo -u ${DATAPROC_USER} bash -l -c"
+
+# Install python packages using the miniconda3 python binary
+readonly RUN_PIP="/opt/conda/miniconda3/bin/python -m pip"
```
rogerwangcs, line 48: I ran into issues where pip install commands were running a python2 binary, is there a better way to resolve this? When logged in as dataproc on the master node vm, python and pip point to the correct miniconda python3 binaries.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,781 @@
...
+
+EOF
+
+# Indicate the start of Terra customizations of the ~/.bashrc
+cat << EOF >> "${USER_BASHRC}"
+
+### BEGIN: Terra-specific customizations ###
+
+# Set the correct java installation to use
+export JAVA_HOME="${USER_HOME_LOCAL_DIR}"
```
rogerwangcs, line 210: Also had trouble getting the dataproc user pointing to the correct java binary instead of the default system java binary..

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,781 @@
...
+# Install nbstripout for the jupyter user in all git repositories.
+${RUN_AS_DATAPROC_USER} "nbstripout --install --global"
+
+# Install gcsfuse for bucket mounting
+# install gnupg and lsb-release (for gcsfuse)
+apt-get install -y \
+  gnupg \
+  lsb-release
+
+# install gcsfuse for bucket mounting
```
rogerwangcs, line 239: Install gcsfuse as it is not installed by default on the dataproc vm images

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,781 @@
...
+###################################
+# Configure Jupyter systemd service
+###################################
+
+# By default the dataproc jupyter optional component runs jupyter as the root user. We override the behavior by configuring the jupyter service to run as the dataproc user instead.
+
+emit "Configuring Jupyter systemd service..."
+
+# Modify the jupyter service configuration
+cat << EOF >${JUPYTER_SERVICE}
```
rogerwangcs, line 641: As mentioned above, I'm updating the jupyter service to run as the dataproc user. Also set the working directory to dataproc user's home folder.

mbookman (changes requested):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,771 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
rogerwangcs: Yup I strongly agree!

Though I think the modular startup script challenge is being solved as part of the large apps / app framework effort. I'm not a fan of duplicating the vertexi ai notebook script, but I feel like one large combined script could be unreadable and unweildy. I will be adding more setup for installing and configuring hail for dataproc as well. This is definitely a short term approach to meet the feature dev timeline and revisited.

I like the idea of local tooling that generates a combined script from "subscripts" depending on the environment in the form similar to what you describe. I can create a ticket for that.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,768 @@
...
+#
+# Name: startup.sh
+#
+# Description
+#   Initialization script to setup Terra configurations in a Dataproc cluster manager node
+#
+# Execution details
+#   This script is executed as root on all Dataproc nodes.
+#
+#   Note that the guest attribute is set to DONE whether the script runs successfully or not.
```
mbookman, line 11: Please list the specific guest attribute(s) and metadata used for startup orchestration. That would be super helpful to have in one place.

- attributes/dataproc-role: read by this script to determine if this is the dataproc head node or a worker node

- instance/guest-attributes/startup_script/status: set by this script to indicate STARTUP/COMPLETE/ERROR
- instance/guest-attributes/startup_script/message: set by this script with more information on any ERROR status

- instance/attributes/terra-cli-server: read by this script to configure the terra CLI server
- instance/attributes/terra-workspace-id: read by this script to configure the terra CLI workspace

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,768 @@
...
+${RUN_AS_DATAPROC_USER} "nbstripout --install --global"
+
+# Install gcsfuse for bucket mounting
+# install gnupg and lsb-release (for gcsfuse)
+apt-get install -y \
+  gnupg \
+  lsb-release
+
+# install gcsfuse for bucket mounting
+export GCSFUSE_REPO=gcsfuse-$(lsb_release -c -s) \
```
mbookman, line 237: Can the script add some detection of the existence of gcsfuse prior to trying to install it. Would be really useful to have a reusable block that could safely run elsewhere.
Was this pulled from https://cloud.google.com/storage/docs/gcsfuse-install? If so, let's add a link to that in the comments.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,768 @@
+#!/bin/bash
+#
```
mbookman, line 2: Please add a comment here (and in the Vertex AI post-startup script) to indicate that changes here should (probably) be reflected in the other script.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,768 @@
...
+# This post startup script is not run by the same user.
+readonly DATAPROC_USER="dataproc"
+
+# Create an alias for cases when we need to run a shell command as the dataproc user.
+# Note that we deliberately use "bash -l" instead of "sh" in order to get bash (instead of dash)
+# and to pick up changes to the .bashrc.
+#
+# This is intentionally not a Bash function, as that can suppress error propagation.
+# This is intentionally not a Bash alias as they are not supported in shell scripts.
+readonly RUN_AS_DATAPROC_USER="sudo -u ${DATAPROC_USER} bash -l -c"
```
mbookman, line 45: I think the need to sweep the code for RUN_AS_JUPYTER_USER -> RUN_AS_DATAPROC_USER bothers me most from a maintenance standpoint. It makes it harder to diff and keep the two scripts in sync. I'm trying to come up with the right nomenclature such that this script and the Vertex AI post-startup script can be coded consistently (but struggling a bit). A few thoughts:

- RUN_AS_NON_ROOT_USER
- RUN_AS_LOCAL_USER
- RUN_AS_LOGIN_USER
- RUN_AS_TARGET_USER

The idea would be to use the above and make the code consistent across the two scripts.
I think I like `RUN_AS_LOGIN_USER` the best of that group. It's not perfect, but feels like with a good comment and a little bit of time, it would be good enough.

mbookman (changes requested):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
rogerwangcs: Thanks for enumerating these out!

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
rogerwangcs: Not at the time of this script's execution

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -49,18 +60,18 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-# The linux user that JupyterLab will be running as. It's important to do some parts of setup in the
+# The non-root linux user that JupyterLab will be running as. It's important to do some parts of setup in the
 # user space, such as setting Terra CLI settings which are persisted in the user's $HOME.
 # This post startup script is not run by the same user.
```
mbookman, line 65: Would you mind removing this last comment "This post startup script is not run by the same user." I don't think it adds anything and I think is ambiguous.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -0,0 +1,788 @@
+#!/bin/bash
+#
+# Name: startup.sh
+#
+# NOTE FOR CONTRIBUTORS:
+#   This startup script closely mirrors the startup script used for Vertex AI Notebook instances here: /Users/rogerwangcs/local_dev/terra-workspace-manager/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh.
```
mbookman, line 6: Please remove `/Users/rogerwangcs/local_dev/` from the path.

mbookman (approved):
Two small things. Otherwise LGTM.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
rogerwangcs: Ah, thanks for the catch

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -17,9 +28,7 @@
 #     1- Get the GCS path from VM metadata (instance/attributes/post-startup-script)
 #     2- Download it to /opt/c2d/post_start.sh
 #     3- Execute /opt/c2d/post_start.sh
-#     4- Set the VM guest attribute "notebooks/handle_post_startup_script" to "DONE"
-#
-#   Note that the guest attribute is set to DONE whether the script runs successfully or not.
```
yuhuyoyo, line 22: keeo this line

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
rogerwangcs: I removed this because I thought that this was the old way of how we're setting the startup script status. I now realize that this might have been the behavior of vertexi ai notebooks after the post_start script is executed. Is the latter the case? I couldn't find where these steps are documented.

yuhuyoyo: gettign to COMPLETE doesn't mean all the execution succeeded i thought.

rogerwangcs: I thought so, assuming that everything in the script runs sync. The UI also only reads status and message attributes. https://github.com/verily-src/terra-ui/blob/master/src/components/cloud/startupScript.ts#L43-L56

yuhuyoyo: ah ok. i didn't realize it has set -o errexit at the top.

rogerwangcs: I stand corrected, UI is indeed querying this attribute as a fallback as well.

yuhuyoyo: i'm not asking about the UI fwiw. I didn't know it has `set -o e` at the top to make sure the failed command doesn't fall through.

rogerwangcs: yeah I gotcha, we do have a handler to set the status attribute wherever the script fails. But either way, setting the handle post startup script attribute is a vertex ai behavior that we do read from

PR #1434, opened 2023-08-10
[WOR-978] Remove check for DELETE on workspace resources during cloud context deletion

aherbst-broad:
## Context
[Ticket](https://broadworkbench.atlassian.net/browse/WOR-978)
[Related one-pager](https://docs.google.com/document/d/1uLT4VcIdqjo_RPZzAP0D6FsGzUbviUbnW1OtZ_k0iPc/edit)
Application controlled resources may be orphaned by their stewarding application prior to workspace deletion. When this happens, the workspace becomes undelete-able due to the Sam checks against the application controlled resources. 

## This PR
* Removes the checks in the `BuildAndValidateResourceListStep` class that is run at the start of cloud context deletion. The premise is that the principal running the deletion flight has DELETE on the workspace, and all resources should be deleted regardless of their stewardship. 
   * I had originally planned on making modifications to the `SamService` to delete the related resource Sam entries with the WSM SA but that has [already been implemented](https://github.com/DataBiosphere/terra-workspace-manager/blob/693729f4f9e2f0cd2a6184ae691e03ae35994419/service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/any/DeleteControlledSamResourcesStep.java#L19).
* Updates the `write-config.sh` dev script to emit needed landing zone users so applications may create resources in an Azure workspace during local development.

### DIFF
--- a/integration/src/main/resources/testusers/README.md
+++ b/integration/src/main/resources/testusers/README.md
@@ -5,7 +5,7 @@ Workspace Manager currently uses 3 test users for running integration tests with
 - `elijah.thunderlord@test.firecloud.org`
 - `liam.dragonmaw@test.firecloud.org`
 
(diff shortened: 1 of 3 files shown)


ddietterich (approved):
Okee dokee

sergiygetlin (approved):

PR #1435, opened 2023-08-10
[BENCH-897] Add Data Storage Volume to GCP Controlled Resource

iliadsh:


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gceinstance/CreateGceInstanceStep.java
```diff
@@ -193,7 +194,15 @@ static Instance setFields(
...
                         .setSourceImage(creationParameters.getVmImage())
-                        .setDiskSizeGb(Long.valueOf(100)))));
+                        .setDiskSizeGb(Long.valueOf(100))),
+            new AttachedDisk()
+                .setBoot(false)
+                .setDeviceName(DATA_DISK_NAME)
+                .setAutoDelete(true)
+                .setInitializeParams(
+                    new AttachedDiskInitializeParams()
+                        .setDiskType(creationParameters.getDataDiskType())
```
yuhuyoyo, line 204: should this be conditional? 

only add an AttachedDisk if the creationParameters contains datadisktype and datadisksizegb

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gceinstance/CreateGceInstanceStep.java
iliadsh: My impression from John's doc was that all instances would have an extra PD... @jmczerk confirm?

yuhuyoyo: or a default type and size?

iliadsh: Hm, but GCP will default their values if none are provided.

PR #1437, opened 2023-08-11
Do not require primary worker config in dataproc cluster creation

rogerwangcs:
Single node clusters require passing null for primary and secondary worker configs to the dataproc api. This is not possible with the current wsm create api. This change changes primary worker config to an optional property. Also removes the logic of setting secondary worker config be the same as primary worker config, the UI/CLI is already doing this.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/CreateDataprocClusterStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/CreateDataprocClusterStep.java
@@ -221,7 +221,7 @@ public class CreateDataprocClusterStep implements Step {
                         .setOptionalComponents(creationParameters.getComponents())));
 
     // Set dataproc image version
-    if(creationParameters.getImageVersion() != null) {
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/CreateDataprocClusterStep.java
```diff
@@ -195,14 +195,6 @@ static Cluster setFields(
                 .setTempBucket(tempBucketName)
                 .setMasterConfig(
                     getInstanceGroupConfig(creationParameters.getManagerNodeConfig(), false))
-                .setWorkerConfig(
-                    getInstanceGroupConfig(creationParameters.getPrimaryWorkerConfig(), false))
-                .setSecondaryWorkerConfig(
-                    creationParameters.getSecondaryWorkerConfig() != null
-                        ? getInstanceGroupConfig(
-                            creationParameters.getSecondaryWorkerConfig(), true)
```
yuhuyoyo, line 203: can you comment the field for ture false for readibility

yuhuyoyo (approved):

zloery (approved):

PR #1438, opened 2023-08-16
Add null checks to cluster metadata check

rogerwangcs:
Dataproc cluster get api doesn't return any worker configuration value like machine-type or machine uri if its node count has been created or updated to 0... Unclear why they chose this behavior but it's causing an error when updating a cluster worker count size from 0 to any value.


yuhuyoyo (approved):
please add a ticket

dexamundsen (approved):

PR #1440, opened 2023-08-16
[PF-2951] Add retry on assertBucketHasNoFiles to handle permission propagation

ddietterich:
The bucket test `clone_copyDefinition` has been flaky on and off. I think I figured out that it is a permission setting delay:
- WSM sets permissions on the bucket resource from Sam sync'd groups
- (My hypothesis is there is a propagation delay of some sort here)
- Test calls GCP immediately to check for files. Sometimes we see 403.

I added a retry on the file check. If things are actually broken, it will slow the test down by 5 minutes. But for the normal case, it will allow us to wait for propagation, the check will pass, and the test won't flake out.

### DIFF
--- a/service/src/test/java/bio/terra/workspace/common/GcsBucketUtils.java
+++ b/service/src/test/java/bio/terra/workspace/common/GcsBucketUtils.java
@@ -23,6 +23,7 @@ import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.StorageOptions;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


dexamundsen (approved):

service/src/test/java/bio/terra/workspace/common/GcsBucketUtils.java
```diff
@@ -153,17 +154,36 @@ public static void assertBucketFiles(
...
     int numFiles = 0;
     for (BlobCow blob : storageCow.get(bucketName).list().iterateAll()) {
       numFiles++;
     }
     assertEquals(0, numFiles);
   }
 
+  private Integer countFiles(StorageCow storageCow, String bucketName) {
+    int numFiles = 0;
+    for (BlobCow blob : storageCow.get(bucketName).list().iterateAll()) {
```
yuhuyoyo, line 181: .size()?

service/src/test/java/bio/terra/workspace/common/GcsBucketUtils.java
```diff
@@ -153,17 +154,36 @@ public static void assertBucketFiles(
...
+        null);
+
     int numFiles = 0;
     for (BlobCow blob : storageCow.get(bucketName).list().iterateAll()) {
       numFiles++;
     }
     assertEquals(0, numFiles);
   }
 
+  private Integer countFiles(StorageCow storageCow, String bucketName) {
```
yuhuyoyo, line 179: return int

service/src/test/java/bio/terra/workspace/common/GcsBucketUtils.java
```diff
@@ -153,17 +154,36 @@ public static void assertBucketFiles(
...
+        /*initialSleep=*/ Duration.ofSeconds(5),
+        /*factorIncrease=*/ 1.0,
+        /*sleepDurationMax=*/ Duration.ofSeconds(30),
+        null);
+
     int numFiles = 0;
     for (BlobCow blob : storageCow.get(bucketName).list().iterateAll()) {
       numFiles++;
     }
     assertEquals(0, numFiles);
```
yuhuyoyo, line 176: assertEquals(0, countFiles(storageCow, bucketName)

service/src/test/java/bio/terra/workspace/common/GcsBucketUtils.java
ddietterich: That is old code I meant to delete. The `getWithRetryOnException` returns the int from the countFiles call.

PR #1442, opened 2023-08-18
[BENCH-958] Service startup tester

ddietterich:
When we start the WSM server for integration tests, it is done in a subprocess and the output is not available in the github log. If something goes wrong, it is hard to debug. Also, if something is slow, we cannot see where the delay is.

This PR includes two things:
 1. StartupConfiguration that allows requesting immediate exit after the service is up and running
 2. A manually dispatched github workflow that sets that config and simply starts the service.

I put the new workflow into workflow-tester so I could test ahead of commit. You can ignore that file.


rogerwangcs (approved):
The workflow run looks good! Lets see the output downstream.

dexamundsen (approved):

PR #1443, opened 2023-08-18
AJ-1270: return tags from LZ resource listing

davidangb:
When listing the resources inside a landing zone, return the tags for those resources.

This will help with inspecting the landing zone to understand its configuration. The current use case is preparing to enable PgBouncer for the LZ Postgres server - we can set a tag on the Postgres resource if PgBouncer is enabled, and then Leo can read the tag (via WSM) and tell WDS and CBAS to connect to PgBouncer on port 6432 instead of Postgres on port 5432.

I am aware this will cause some duplication between tag values and elsewhere in the response, such as the WLZ-PURPOSE and WLZ-ID.

Slack discussion: https://broadinstitute.slack.com/archives/C02LMB1PQ6R/p1692376111128999

This is an additive change to an existing API. Will this cause any problems for services that use the WSM client? Are additive changes acceptable? It looks fine to me and seems to work in my bee, without updating the WSM client version in Rawls.

Tested in my bee. An example resource as returned from the API:
```json
       {
          "resourceId": "/subscriptions/0fa61a72-3a6b-4d51-a7e9-75fbeb7c7ed9/resourceGroups/mrg-terra-dev-previ-20230803171249/providers/Microsoft.DBforPostgreSQL/flexibleServers/lz6f76fb220ce486e4b7c73f8291f33e47bf2eec953e6c3b14bb3c140ebb90d",
          "resourceType": "Microsoft.DBforPostgreSQL/flexibleServers",
          "region": "eastus",
          "tags": {
            "WLZ-PURPOSE": "SHARED_RESOURCE",
            "WLZ-ID": "1b6ad2db-7c71-4b0b-8e63-bc9fac4e66eb",
            "mock-pgbouncer-tag": "enabled"
          }
        },
```

### DIFF
--- a/openapi/src/parts/azure_landing_zone.yaml
+++ b/openapi/src/parts/azure_landing_zone.yaml
@@ -333,6 +333,11 @@ components:
         region:
           description: A region where an Azure resource deployed.
           type: string
+        tags:
+          description: Tags for this Azure resource.
+          type: object
+          additionalProperties:
+            type: string
 
     AzureLandingZoneDefinitionList:
       type: object
--- a/service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
+++ b/service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
@@ -239,6 +239,7 @@ public class LandingZoneApiDispatch {
       return new ApiAzureLandingZoneDeployedResource()
           .resourceId(resource.resourceId())
           .resourceType(resource.resourceType())
+          .tags(resource.tags())
           .region(resource.region());
     }
     if (purpose.getClass().equals(SubnetResourcePurpose.class)) {
@@ -246,6 +247,7 @@ public class LandingZoneApiDispatch {
           .resourceParentId(resource.resourceParentId().orElse(null)) // Only available for subnets
           .resourceName(resource.resourceName().orElse(null)) // Only available for subnets
           .resourceType(resource.resourceType())
+          .tags(resource.tags())
           .region(resource.region());
     }
     throw new LandingZoneUnsupportedPurposeException(
--- a/service/src/test/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatchTest.java
+++ b/service/src/test/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatchTest.java
@@ -363,13 +363,13 @@ public class LandingZoneApiDispatchTest extends BaseAzureUnitTest {
                 .resourceId("Id32")
                 .resourceType("fooType32")
                 .region("fooRegion2")
-                .tags(Map.of("resourceTag1", "resourceValue1", "resourceTag3", "resourceValue4"))
+                .tags(Map.of("resourceTag1", "resourceValue1", "resourceTag3", "resourceValue3"))
                 .build(),
             LandingZoneResource.builder()
                 .resourceId("Id33")
                 .resourceType("fooType33")
                 .region("fooRegion1")
-                .tags(Map.of("resourceTag1", "resourceValue1", "resourceTag3", "resourceValue4"))
+                .tags(Map.of("resourceTag1", "resourceValue1", "resourceTag4", "resourceValue4"))
                 .build());
 
     when(landingZoneService.listResourcesByPurpose(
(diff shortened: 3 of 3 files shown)


openapi/src/parts/azure_landing_zone.yaml
```diff
@@ -333,6 +333,11 @@ components:
         region:
           description: A region where an Azure resource deployed.
           type: string
+        tags:
+          description: Tags for this Azure resource.
+          type: object
+          additionalProperties:
+            type: string
```
davidangb, line 340: since the `ApiAzureLandingZoneDeployedResource` model class is generated from the OpenApi definition, changing this definition adds a `tags` member to that model.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -239,13 +239,15 @@ private ApiAzureLandingZoneDeployedResource toApiAzureLandingZoneDeployedResourc
...
           .resourceType(resource.resourceType())
+          .tags(resource.tags())
           .region(resource.region());
     }
     if (purpose.getClass().equals(SubnetResourcePurpose.class)) {
       return new ApiAzureLandingZoneDeployedResource()
           .resourceParentId(resource.resourceParentId().orElse(null)) // Only available for subnets
           .resourceName(resource.resourceName().orElse(null)) // Only available for subnets
           .resourceType(resource.resourceType())
+          .tags(resource.tags())
```
davidangb, line 250: the `LandingZoneResource` model already had `tags`; this change copies those tags from the `LandingZoneResource` to the `ApiAzureLandingZoneDeployedResource`.

service/src/test/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatchTest.java
```diff
@@ -149,6 +150,39 @@ void createAzureLandingZone_LandingZoneDoesNotExistsExceptionIsHandledAndStartsJ
...
+                      res -> {
+                        assertNotNull(
+                            res.getTags(), "tags null for resource id " + res.getResourceId());
+                        assertEquals(
+                            2,
+                            res.getTags().size(),
+                            "tags size for resource id " + res.getResourceId());
+                      });
+            });
+  }
```
davidangb, line 184: new unit test specifically to check that tags are returned in the api response. I could have added these asserts to an existing test, but that felt like it broke a bit of test encapsulation.

sergiygetlin (commented):
This change is definitely required, but I think we also need to confirm that corresponding tags in LandingZoneResource are defined (not null).

aherbst-broad (approved):
Looks like a reasonable approach.

rtitle (approved):
LGTM -- is the idea that Leo would read the tags when it queries the LZ service to list resources? (And then make some decisions depending on whether a tag is present, like `pgbouncer`.)

davidangb:
> is the idea that Leo would read the tags when it queries the LZ service

@rtitle yeah, exactly that. https://github.com/DataBiosphere/leonardo/pull/3614 is very much draft/work-in-progress, but you can get the gist of what I'm thinking there.

service/src/test/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatchTest.java
```diff
@@ -320,16 +357,19 @@ private void setupLandingZoneResources() {
...
                 .resourceId("Id32")
                 .resourceType("fooType32")
                 .region("fooRegion2")
+                .tags(Map.of("resourceTag1", "resourceValue1", "resourceTag3", "resourceValue4"))
                 .build(),
             LandingZoneResource.builder()
                 .resourceId("Id33")
                 .resourceType("fooType33")
                 .region("fooRegion1")
+                .tags(Map.of("resourceTag1", "resourceValue1", "resourceTag3", "resourceValue4"))
```
jladieu, line 372: Just pattern matching here, but do you want this to be `resourceTag4` instead of `3`?

service/src/test/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatchTest.java
```diff
@@ -320,16 +357,19 @@ private void setupLandingZoneResources() {
                 .resourceId("Id31")
                 .resourceType("fooType31")
                 .region("fooRegion1")
+                .tags(Map.of("resourceTag1", "resourceValue1", "resourceTag2", "resourceValue2"))
                 .build(),
             LandingZoneResource.builder()
                 .resourceId("Id32")
                 .resourceType("fooType32")
                 .region("fooRegion2")
+                .tags(Map.of("resourceTag1", "resourceValue1", "resourceTag3", "resourceValue4"))
```
jladieu, line 366: Just pattern matching here, but do you want this to be `resourceValue3` instead of `4`?

davidangb: good catch, fixing both cases in the name of cleanliness

yuliadub (approved):

PR #1445, opened 2023-08-22
Dataproc Update Fixes

rogerwangcs:
This change includes the following fixes:
- Fix a bad property set on cluster secondary workers.
- Add a check to to skip the update call if there are no parameters to update. This caused a dismal failure in the update undo step when there were no previous parameters to update to, resulting in an empty update call error.
- Add a null check on `idleDeleteTtl`, though it should never be null if its parent `lifecycleConfig` exists.

Note: I intentional rely on the gcp api to handle validation errors. There are a few parameters that cannot be fully validated locally:
- num secondary workers have dynamic constraints depending on the autoscaling policy attached to the cluster
- the autoscaling policy provided may or may not exist in the project.
Ideally these validation errors cause the flight to fail normally and report the error back to the caller.

### DIFF
--- a/integration/src/main/java/scripts/testscripts/PrivateControlledDataprocClusterLifeCycle.java
+++ b/integration/src/main/java/scripts/testscripts/PrivateControlledDataprocClusterLifeCycle.java
@@ -246,7 +246,7 @@ public class PrivateControlledDataprocClusterLifeCycle extends WorkspaceAllocate
 
     // Verify that WSM throws a bad request exception
     assertThat(
-        "Error from GCP is 403 or 404",
+        "WSM throws a bad request exception",
         badClusterUpdateParameter.getCode(),
         equalTo(HttpStatus.SC_NOT_FOUND));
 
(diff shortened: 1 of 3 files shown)


dexamundsen (approved):

ddietterich (commented):
IIUC, this will tolerate a bad TTL input because it will fail the update and then restore the known-good update.

Would you add a test to exercise that error path?

rogerwangcs:
> IIUC, this will tolerate a bad TTL input because it will fail the update and then restore the known-good update.

Not quite. TTL and the other lifecycle fields are omitted from being "restored" in the undo step.

In general, it's unclear to me why we need to restore anything in the undo step for direct resource attribute patch calls. My understanding is that any failure in the patch call means no attributes changed, so theres no need to restore attributes to their original values if they are unchanged. Perhaps it was the case for vertex ai metadata, but it's not the case for updating this set of cluster attributes.

ddietterich:
> it's unclear to me why we need to restore anything in the undo step for direct resource attribute patch calls.
It may not be needed in this flight. If there were other steps after the update step and those could fail, then you would need to be able to undo to the previous state.

rogerwangcs:
> It may not be needed in this flight. If there were other steps after the update step and those could fail, then you would need to be able to undo to the previous state.

Good point. I just added a test

ddietterich (approved):
LGTM, thanks!

PR #1446, opened 2023-08-23
BENCH-979: fix flaky test DeleteWorkspaceWithControlledResource

yuhuyoyo:
call the deleteWorkspaceAsync endpoint in the test.


mikenev (approved):
lgtm. You added polling to which is probably helpful.

ddietterich (approved):
Maybe make a common method.

PR #1450, opened 2023-08-25
PF-2957: Group policy not enforced after adding from a data collection

mikenev:
The problem was that we were trying to get the original group list from tps but by the time this step is called, tps is already updated so the current group list == merged group list. Both lists are already stored in the flight, so the solution is to get the original list from there rather than querying tps.
I believe this happened because originally we had the steps in a different order, so querying tps for the original list worked.

Still thinking how to test this. We do have coverage for the tps updates and merge logic. We need to add in validating the actual Sam groups but we don't do anything like that currently.


ddietterich (approved):
LGTM - maybe make a ticket to add the Sam validation?

mikenev:
> LGTM - maybe make a ticket to add the Sam validation?

Good idea, added PF-2598

zloery (approved):

PR #1452, opened 2023-08-26
[BENCH-822] Switch default notebook instance type to ml.t3.medium

dexamundsen:
This is recommended by AWS support to improve notebook startup times

ml.t3.medium is of newer generation, comparison for reference - https://www.cloudzero.com/advisor/t2-vs-t3
the cost difference is 1/3 of one cent per hour ($0.0036).

note: VWB ui already using this in UI / integration tests. This may be the reason we have mostly seen the delay in WSM tests


jmczerk (approved):
LGTM.  Do we need to change the UI default as well?

dexamundsen:
> LGTM.  Do we need to change the UI default as well?

ui is already using the updated instance types

yuhuyoyo (approved):

PR #1456, opened 2023-08-29
[BENCH-992] Incorrect categorization of flight states

ddietterich:
A small piece of code that mapped Stairway flight status to async job state had a bug. It was categorizing 
QUEUED, WAITING, READY, and READY_TO_RESTART as FAILED states. Those are actually RUNNING states - the flight is not complete so wouldn't have a complete time.

In this particular case, the flight enters the QUEUED state. In this case, I think the QUEUED state happened because the Stairway thread pool had too many pending flights in its queue. At that point, Stairway pushes flights to its pubsub queue for other pods to deal with. Even though we don't have that properly configured, the state changes still happen.


dexamundsen (approved):

yuhuyoyo (approved):

gdlex4015:
Dan, you referred to Jira ticket BENCH-992 in this PR, I do not see project BENCH in my Jira, is that a project that lives outside Broad DSP?

ddietterich:
Yes, it is.

PR #1457, opened 2023-08-30
[PF-2967] Add retries for assertBucketFiles cloud call

zloery:
This test utility method makes a call directly to GCS, which sometimes fails due to permissions propagation issues or files not being loaded into the bucket. We should retry this call briefly.

This is currently causing the `loadSignedUrlList_succeedsAndFileLoadedIntoBucket` to flake intermittently, as there can be a very short delay between the job loading files into the bucket and GCS acknowledging those files exist.


ddietterich (approved):
Nice find!

PR #1463, opened 2023-09-06
[BENCH-970] Fetch user email from sam using userRequest

dexamundsen:
- Fetch user email from sam using userRequest
- Check if aws is enabled for the user during clone workspace

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -67,6 +67,7 @@ import bio.terra.workspace.service.resource.controlled.model.ControlledResource;
 import bio.terra.workspace.service.spendprofile.SpendProfile;
 import bio.terra.workspace.service.spendprofile.SpendProfileId;
 import bio.terra.workspace.service.spendprofile.SpendProfileService;
+import bio.terra.workspace.service.workspace.AwsCloudContextService;
 import bio.terra.workspace.service.workspace.WorkspaceService;
 import bio.terra.workspace.service.workspace.exceptions.MissingSpendProfileException;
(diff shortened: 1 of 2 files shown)


rogerwangcs (approved):

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -485,7 +489,9 @@ public ResponseEntity<ApiCreateCloudContextResult> createCloudContext(
 
     CloudPlatform cloudPlatform = CloudPlatform.fromApiCloudPlatform(apiCloudPlatform);
     if (cloudPlatform == CloudPlatform.AWS) {
-      featureService.featureEnabledCheck(FeatureService.AWS_ENABLED, userRequest.getEmail());
+      featureService.featureEnabledCheck(
+          FeatureService.AWS_ENABLED,
+          samService.getUserEmailFromSamAndRethrowOnInterrupt(userRequest));
```
yuhuyoyo, line 494: do we have a situation when the petSA create a workspace?

zloery (approved):

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
dexamundsen: I checked the created_by_email field in devel and did not find any pet sa emails.

yuhuyoyo: then why do we need to query sam for user email?

dexamundsen: User Request may not have the mail address filled, if it only has a access token. email address is however needed for feature check

PR #1464, opened 2023-09-07
BENCH-1063: Remove spark monitor

rogerwangcs:
Due to package version incompatibilities with jupyter, [hail's spark monitor](https://github.com/hail-is/sparkmonitor) doesn't work out of the box. This change remove the default hail installation script installs the `sparkmonitor` nbextension.  For now, we expect users rely on the inline text based spark job output logs rendered in cell output. Users can further monitor larger jobs in cloud console.

Removing spark monitor fixes also fixes the spark monitor initialization error with `hl.init()` when using the hail kernel.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
@@ -733,8 +733,9 @@ if [[ "${SOFTWARE_FRAMEWORK}" == "HAIL" ]]; then
   # Create the Hail install script. The script is based off of Hail's init_notebook.py
   # script that is executed by 'hailctl dataproc start'. This modified script omits
   # the following steps:
-  # - Configuring and enabling hail's spark monitor nbextension
   # - Configuring and starting a custom jupyter systemd service
+  # - Configuring and enabling hail's spark monitor nbextension.
+  # TODO: BENCH-1094: Consider whether we need spark monitor
   cat << EOF >"${HAIL_SCRIPT_PATH}"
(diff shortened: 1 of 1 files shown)


amygdala (approved):
that LGTM.  I'll create a ticket for when we have more breathing room to investigate the installation issues & think about whether we should try to add support for it (possibly not)

update: created https://verily.atlassian.net/browse/BENCH-1094

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh
```diff
@@ -731,8 +731,11 @@ if [[ "${SOFTWARE_FRAMEWORK}" == "HAIL" ]]; then
...
-  # script that is executed by 'hailctl dataproc start'. This modified script excludes
-  # the step of starting a jupyter service.
+  # script that is executed by 'hailctl dataproc start'. This modified script omits
+  # the following steps:
+  # - Configuring and starting a custom jupyter systemd service
+  # - Configuring and enabling hail's spark monitor nbextension.
+  # TODO: BENCH-1094: Consider whether we need spark monitor
   cat << EOF >"${HAIL_SCRIPT_PATH}"
 #!${RUN_PYTHON}
 # This modified Hail installation script installs the necessary Hail packages and jupyter extensions,
```
rogerwangcs, line 741: > Please add some indication of "why".

Updated PR description. I added a comment on omitting spark monitor, but I left it outside of the user facing hail install script. Perhaps I could add something like: `# Also skip installing spark monitor due to version incompatibility with the latest version of jupyter`?

mbookman (approved):

PR #1465, opened 2023-09-07
delete k8s namespaces first

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-1249

### DIFF
--- a/service/src/main/java/bio/terra/workspace/db/ResourceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/ResourceDao.java
@@ -345,7 +345,7 @@ public class ResourceDao {
 
   /**
    * Returns a list of all controlled resources in a workspace, optionally filtering by cloud
-   * platform.
+   * platform, ordered by created date newest to oldest.
    *
    * @param workspaceUuid ID of the workspace to return resources from.
    * @param cloudPlatform Optional. If present, this will only return resources from the specified
@@ -364,6 +364,8 @@ public class ResourceDao {
       sql += " AND cloud_platform = :cloud_platform";
       params.addValue("cloud_platform", cloudPlatform.toSql());
     }
+
+    sql += " ORDER BY created_date desc";
 
     List<DbResource> dbResources = jdbcTemplate.query(sql, params, DB_RESOURCE_ROW_MAPPER);
     return dbResources.stream()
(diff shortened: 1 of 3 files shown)


aherbst-broad (approved):

service/src/main/java/bio/terra/workspace/service/workspace/AzureCloudContextService.java
```diff
@@ -111,11 +111,16 @@ public CloudContext makeCloudContextFromDb(DbCloudContext dbCloudContext) {
...
-    // Delete VMs first because they use other resources like disks, networks, etc.
+    // Delete VMs and k8s namespaces first because they use other resources like disks, identities,
+    // databases, etc.
     List<ControlledResource> orderedList =
         new ArrayList<>(
             unorderedList.stream()
-                .filter(r -> r.getResourceType() == WsmResourceType.CONTROLLED_AZURE_VM)
+                .filter(
+                    r ->
+                        r.getResourceType() == WsmResourceType.CONTROLLED_AZURE_VM
```
cahrens, line 121: What about line 138? Maybe there is a way to rewrite this code so that the "special" resources don't have to be listed in multiple places.

rtitle:
Just double checking -- in Leo I had to delete the database before the namespace, because deleting the namespace dropped the DB role that tables were depending on. Is that not the case anymore?

dvoet:
> Just double checking -- in Leo I had to delete the database before the namespace, because deleting the namespace dropped the DB role that tables were depending on. Is that not the case anymore?
> 
> So the working Leo deletion order is currently: database, namespace, identity

I will fix this. I am changing the code to delete in reverse of creation.

service/src/main/java/bio/terra/workspace/db/ResourceDao.java
```diff
@@ -365,6 +365,8 @@ public List<ControlledResource> listControlledResources(
       params.addValue("cloud_platform", cloudPlatform.toSql());
     }
 
+    sql += " ORDER BY created_date desc";
```
cahrens, line 368: It would be good to add a comment to the code about why reverse order is being returned.

rtitle (approved):
LGTM -- I also tend to think we can ensure/test that apps are doing the "right thing" w.r.t. database roles.

cahrens (approved):

PR #1467, opened 2023-09-08
WOR-1250 support names as well as ids creating databases and k8s namespaces

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-1250


service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -791,6 +801,28 @@ public ResponseEntity<ApiAzureManagedIdentityResource> getAzureManagedIdentity(
...
+   * of the resources to be passed in. But names are easier to use and port nicely during clone
+   * operations. So currently the api supports both names and ids. But we convert to names before
+   * storing the resource.
+   *
+   * @param workspaceId
+   * @param maybeUuid a uuid or a name
+   * @return if maybeUuid is a uuid and exists in the workspace, return the name of the resource
+   *     with that uuid. Otherwise, return maybeUuid.
+   */
+  private String maybeLookupName(UUID workspaceId, String maybeUuid) {
```
sergiygetlin, line 815: We are trying to locate resource when 'id' is passed, but In case 'maybeUuid' represents name of a resource should we validate that such resource exists?

dvoet: the validation happens later on in one of the stairway steps

openapi/src/parts/controlled_azure_database.yaml
```diff
@@ -88,9 +88,8 @@ components:
           type: string
           pattern: ^[a-zA-Z][a-zA-Z0-9_]{0,31}$
         owner:
-          description: Resource id of an azure managed identity in this workspace to be the owner of the database, required for shared databases. For private databases, the owner will be determined from privateResourceUser.
+          description: Resource id or name of an azure managed identity in this workspace to be the owner of the database, required for shared databases. For private databases, the owner will be determined from privateResourceUser.
```
cjllanwarne, line 91: How can the system distinguish names from IDs? Is it possible to get a collision between a name and an ID on different databases?

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -791,6 +801,28 @@ public ResponseEntity<ApiAzureManagedIdentityResource> getAzureManagedIdentity(
         fetchCreateControlledKubernetesNamespaceResult(jobId), HttpStatus.OK);
   }
 
+  /**
+   * Initial implementations of createAzureKubernetesNamespace and createAzureDatabase required ids
+   * of the resources to be passed in. But names are easier to use and port nicely during clone
+   * operations. So currently the api supports both names and ids. But we convert to names before
+   * storing the resource.
```
cjllanwarne, line 808: Is this going to create a change in query response type (from ID to name) too? Is that going to break anything downstream?

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -791,6 +801,28 @@ public ResponseEntity<ApiAzureManagedIdentityResource> getAzureManagedIdentity(
...
+   */
+  private String maybeLookupName(UUID workspaceId, String maybeUuid) {
+    if (maybeUuid == null) {
+      return null;
+    }
+    try {
+      return wsmResourceService.getResource(workspaceId, UUID.fromString(maybeUuid)).getName();
+    } catch (IllegalArgumentException | ResourceNotFoundException e) {
+      return maybeUuid;
+    }
```
cjllanwarne, line 823: Exception throwing as a form of validation is a little clunky

cjllanwarne (approved):
Looks good, a few nit picks

openapi/src/parts/controlled_azure_database.yaml
dvoet: if it is a uuid and exists in the workspace, then it is treated as an id, otherwise a name

sergiygetlin (approved):

rtitle (approved):
Changes LGTM.

Not fully tested, but here is the Leo side of this: https://github.com/DataBiosphere/leonardo/pull/3681

(As a bonus I also wired up the OpenCensus `span` to the WSM client)

blakery (approved):

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
dvoet: I merged [this PR](https://github.com/DataBiosphere/terra-workspace-manager/pull/1470) to have a phased rollout and avoid a breaking change. I don't think we should use ids at all so I would like to eventually remove that option. The problem comes with cloning resources because the id is not portable across clones but the name is.

PR #1468, opened 2023-09-11
Check that cluster is in running state before updating

rogerwangcs:
Currently WSM doesn't check cluster state before performing update calls. Since updating non wsm managed cluster attributes require the cluster to be running, calling the endpoint while the cluster is in a different state will result in a dismal failure due to the update call in the undo step failing.

This change:
- validates that cluster is `RUNNING` before proceeding with the gcp update.
- Apply google json error handling to retrieve metadata steps (to avoid returning unneeded fatal failure on bad requests).
- Also require autoscaling policy attribute to be updated alone.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
@@ -1,6 +1,8 @@
 package bio.terra.workspace.app.controller;
 
+import bio.terra.cloudres.google.dataproc.ClusterName;
 import bio.terra.common.exception.BadRequestException;
+import bio.terra.common.exception.InternalServerErrorException;
 import bio.terra.common.exception.ValidationException;
 import bio.terra.workspace.app.configuration.external.FeatureConfiguration;
 import bio.terra.workspace.app.controller.shared.JobApiUtils;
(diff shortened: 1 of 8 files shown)


service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
```diff
@@ -1021,6 +1035,26 @@ public ResponseEntity<ApiGcpDataprocClusterResource> updateDataprocCluster(
       }
     }
 
+    // Ensure that the cluster is in a "RUNNING" state
+    try {
```
yuhuyoyo, line 1039: can you do this as part of the flight?

rogerwangcs: Yeah I originally was gonna add this as a flight step but moved it here to keep all the validation checks together. I don't have a strong opinion so I can change if you want

dexamundsen: You recommend creating a new step for this. 
reason - makes code idempotent if its a step (see validation steps in AWS), handles retries and rollback if necessary.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/ValidateDataprocClusterCreateStep.java
```diff
@@ -0,0 +1,56 @@
...
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import bio.terra.stairway.exception.RetryException;
+import bio.terra.workspace.service.crl.CrlService;
+import bio.terra.workspace.service.resource.GcpFlightExceptionUtils;
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
+import com.google.api.services.dataproc.model.Cluster;
+import java.io.IOException;
+
+public class ValidateDataprocClusterCreateStep implements Step {
```
dexamundsen, line 15: minor: Should this be named updateStep?

service/src/main/java/bio/terra/workspace/app/controller/ControlledGcpResourceApiController.java
rogerwangcs: That makes a lot of sense. done.

dexamundsen (approved):

ddietterich (approved):
lgtm

PR #1473, opened 2023-09-14
Fix dataproc integration test flake

rogerwangcs:
This change:
- Add a check that the dataproc resource user has access to the created cluster before directly querying it. This catches the occaisonal `Not authorized for requested resource` permission flake.
- Wraps `userHasProxyAccess()` with retry to address the integration test sometimes failing at that step.
- Adds logic to properly test stop / start.
- Move `getWithRetryOnException` from `ClientTestUtils` to `RetryUtils`.

Note: The dataproc integration test suite not enabled in the broad environment (this repo's test config). This only applies to the downstream verily integration test suite.


zloery (approved):

yuhuyoyo (approved):

rogerwangcs:
I still see a failure happening occaisonally while I'm running locally. I'm going to merge this and keep monitoring the nightly tests.

PR #1475, opened 2023-09-18
enable apps on workspace create

dvoet:
fix for https://broadworkbench.atlassian.net/browse/WOR-1256 by removing the dao to dao function call which triggered nested retries.

partial fix for https://broadworkbench.atlassian.net/browse/WOR-732 by implementing app enablement correctly on workspace clone and create

### DIFF
--- a/service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
@@ -193,19 +193,6 @@ public class ApplicationDao {
   }
 
   /**
-   * Enable a list of applications. Used as part of create workspace
-   *
-   * @param workspaceUuid workspace to enable applications in
-   * @param applicationIds list of application ids
-   */
-  @WriteTransaction
-  public void enableWorkspaceApplications(UUID workspaceUuid, List<String> applicationIds) {
-    for (String applicationId : applicationIds) {
-      enableWorkspaceApplication(workspaceUuid, applicationId);
-    }
-  }
-
-  /**
    * Enable an application in a workspace - do not perform the application state check. This is only
    * used in testing.
    *
--- a/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
+++ b/service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
@@ -208,17 +208,14 @@ public class WorkspaceDao {
       };
 
   private final NamedParameterJdbcTemplate jdbcTemplate;
-  private final ApplicationDao applicationDao;
   private final StateDao stateDao;
   private final WorkspaceActivityLogDao workspaceActivityLogDao;
 
   @Autowired
   public WorkspaceDao(
-      ApplicationDao applicationDao,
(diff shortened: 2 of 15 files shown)


service/src/main/java/bio/terra/workspace/db/ApplicationDao.java
```diff
@@ -192,19 +192,6 @@ public WsmWorkspaceApplication enableWorkspaceApplication(
...
-  /**
-   * Enable a list of applications. Used as part of create workspace
-   *
-   * @param workspaceUuid workspace to enable applications in
-   * @param applicationIds list of application ids
-   */
-  @WriteTransaction
-  public void enableWorkspaceApplications(UUID workspaceUuid, List<String> applicationIds) {
-    for (String applicationId : applicationIds) {
-      enableWorkspaceApplication(workspaceUuid, applicationId);
```
dvoet, line 204: this line was the main problem for https://broadworkbench.atlassian.net/browse/WOR-1256. This is called in a function annotated with @WriteTransaction and then it calls a function in another dao that is also @WriteTransaction. The transaction propagation is REQUIRED so the same transaction is used but another retry interceptor is layered in! So we do retries within the same transaction instead of retrying the entire transaction.

service/src/main/java/bio/terra/workspace/db/WorkspaceDao.java
```diff
@@ -297,11 +293,6 @@ INSERT INTO workspace (workspace_id, user_facing_id, display_name, description,
       }
     }
 
-    // If we have applicationIds to create, do that now.
-    if (applicationIds != null) {
-      applicationDao.enableWorkspaceApplications(workspaceUuid, applicationIds);
-    }
-
```
dvoet, line 304: this was the main problem for https://broadworkbench.atlassian.net/browse/WOR-1256. This is called in a function annotated with @WriteTransaction and then it calls a function in another dao that is also @WriteTransaction. The transaction propagation is REQUIRED so the same transaction is used but another retry interceptor is layered in! So we do retries within the same transaction instead of retrying the entire transaction.

I removed this code because it was causing problems and it is not right anyway. To enable apps we need to do the database work _and_ call sam. Since there is already a flight to do so, the solution is to change the create workspace flight to launch a subflight to enable apps if appropriate.

service/src/test/java/bio/terra/workspace/common/fixtures/WorkspaceFixtures.java
```diff
@@ -98,7 +98,7 @@ public static Workspace buildMcWorkspace(@Nullable UUID workspaceUuid) {
 
   public static void createWorkspaceInDb(Workspace workspace, WorkspaceDao workspaceDao) {
     var flightId = UUID.randomUUID().toString();
-    workspaceDao.createWorkspaceStart(workspace, /* applicationIds */ null, flightId);
+    workspaceDao.createWorkspaceStart(workspace, /* applicationIds */ flightId);
```
marctalbott, line 101: nit: remove old comment here and where duplicated in other tests

marctalbott (approved):

PR #1477, opened 2023-09-19
[DNM] [BENCH-1202]: Update Vertex AI and Dataproc cluster resource startup scripts to install new CLI

rogerwangcs:
With this change, startup scripts will no longer install from the DataBiosphere/terra-cli repo and instead from the forked verily terra-cli repo. The CLI installation is now environment dependent, and will install the latest supported CLI version defined in AFS for that environment.

Note: installing the new CLI using the `verily` prod environment currently does not work, but will be resolved once the AFS CLI version changes are promoted to prod next week.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -376,15 +376,17 @@ ${RUN_AS_LOGIN_USER} "\
 # Install & configure the Terra CLI
 emit "Installing the Terra CLI ..."
 
+# Fetch the Terra CLI server environment from the metadata server to install appropriate CLI version
+readonly TERRA_SERVER="$(get_metadata_value "instance/attributes/terra-cli-server")"
+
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -376,15 +376,17 @@ ${RUN_AS_LOGIN_USER} "\
 # Install & configure the Terra CLI
 emit "Installing the Terra CLI ..."
 
+# Fetch the Terra CLI server environment from the metadata server to install appropriate CLI version
+readonly TERRA_SERVER="$(get_metadata_value "instance/attributes/terra-cli-server")"
+
 ${RUN_AS_LOGIN_USER} "\
-  curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash && \
+  curl -L https://storage.googleapis.com/devcontainerpkg/workbench-cli/download-install.sh | TERRA_CLI_SERVER=${TERRA_SERVER} bash && \
```
cbookg, line 383: i feel like this should only be used in the verily repo/deployment?  config param?

PR #1478, opened 2023-09-19
Cache storage data per workspace and container

aherbst-broad:
We're bumping into Azure throttling under load to the getSasToken endpoint. This is because we are making a live call to Azure to get storage account information for workspaces, from which we mint a sas token locally. 

This PR introduces a basic cache keyed on workspaceId and storageContainerUuid.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
@@ -29,10 +29,14 @@ import com.azure.storage.common.sas.SasProtocol;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.time.OffsetDateTime;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


dvoet (commented):
This is fine as it for not but a couple future comments
1. would `java.util.concurrent.ConcurrentMap` be better?
2. perhaps use `@Cacheable` from spring

dvoet (approved):

rtitle (approved):
:shipit:

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -78,6 +84,8 @@ public AzureStorageAccessService(
     this.storageAccountKeyProvider = storageAccountKeyProvider;
     this.azureConfiguration = azureConfiguration;
     this.workspaceService = workspaceService;
+    this.storageAccountCache =
+        new ConcurrentHashMap<>(new PassiveExpiringMap<>(1, TimeUnit.HOURS));
```
dvoet, line 88: the implementation of this`ConcurrentHashMap` constructor is
```
    public ConcurrentHashMap(Map<? extends K, ? extends V> m) {
        this.sizeCtl = DEFAULT_CAPACITY;
        putAll(m);
    }
```
The use of `PassiveExpiringMap` is ignored.

PR #1483, opened 2023-09-21
[BENCH-1202]  - Support installing CLI version based on environment

rogerwangcs:
Read the `terra-server` metadata and install the CLI from the appropriate broad/verily distribution url. For verily, fetch the distribution path from afs, for broad, install from the open source repo.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -382,8 +382,26 @@ readonly TERRA_SERVER="$(get_metadata_value "instance/attributes/terra-cli-serve
 # If the server environment is a verily server, use the verily download script.
 # Otherwise, install the latest DataBiosphere CLI release.
 if [[ $TERRA_SERVER == *"verily"* ]]; then
+  # Map the CLI server to the appropriate AFS service environment
+  declare -A serverMap=(
+      ["verily-devel"]="devel"
+      ["verily-autopush"]="autopush"
+      ["verily-staging"]="staging"
+      ["verily-preprod"]="preprod"
+      ["verily"]=""
+  )
+  environment="${serverMap[$TERRA_SERVER]}"
+  if [[ -z "$environment" && "$TERRA_SERVER" != "verily" ]]; then
+      >&2 echo "ERROR: $TERRA_SERVER is not a known verily server."
+      exit 1
+  fi
+  # Build AFS service path and fetch the CLI distribution path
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -376,15 +376,43 @@ ${RUN_AS_LOGIN_USER} "\
...
-  curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash && \
-  cp terra '${TERRA_INSTALL_PATH}'"
+# Fetch the Terra CLI server environment from the metadata server to install appropriate CLI version
+readonly TERRA_SERVER="$(get_metadata_value "instance/attributes/terra-cli-server")"
+
+# If the server environment is a verily server, use the verily download script.
+# Otherwise, install the latest DataBiosphere CLI release.
+if [[ $TERRA_SERVER == *"verily"* ]]; then
+  # Map the CLI server to the appropriate AFS service environment
+  declare -A serverMap=(
```
ddietterich, line 386: You could use a case statement here. Then the default case can generate the error.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -376,15 +376,43 @@ ${RUN_AS_LOGIN_USER} "\
...
 
-${RUN_AS_LOGIN_USER} "\
-  curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash && \
-  cp terra '${TERRA_INSTALL_PATH}'"
+# Fetch the Terra CLI server environment from the metadata server to install appropriate CLI version
+readonly TERRA_SERVER="$(get_metadata_value "instance/attributes/terra-cli-server")"
+
+# If the server environment is a verily server, use the verily download script.
+# Otherwise, install the latest DataBiosphere CLI release.
+if [[ $TERRA_SERVER == *"verily"* ]]; then
```
ddietterich, line 384: Never seen that style of pattern match. There is another one, but if that works, that's cool

ddietterich (approved):
I'd prefer using a `case`, but this looks like it works

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
ddietterich: There is a `=~` operator that matches a regex.

I didn't realize that the `==` operator would match a bash pattern. TIL

zloery (approved):

PR #1485, opened 2023-09-21
Fix logging during postgres database deletion

sergiygetlin:
https://broadworkbench.atlassian.net/browse/WOR-1237
Postgres server name used twice instead of database name.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/CreateAzureDatabaseStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/CreateAzureDatabaseStep.java
@@ -105,7 +105,7 @@ public class CreateAzureDatabaseStep implements Step {
     try {
       logger.info(
           "Attempting to delete database {} in server {} of resource group {}",
-          getResourceName(databaseResource),
+          resource.getDatabaseName(),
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/CreateAzureDatabaseStep.java
```diff
@@ -105,7 +105,7 @@ public StepResult undoStep(FlightContext context) throws InterruptedException {
     try {
       logger.info(
           "Attempting to delete database {} in server {} of resource group {}",
-          getResourceName(databaseResource),
```
aherbst-broad, line 108: Why does getResourceName not work in this case?

aherbst-broad (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/CreateAzureDatabaseStep.java
sergiygetlin: because it is a name of postgres instance.

cahrens (approved):

cahrens:
Link the ticket please @sergiygetlin

PR #1487, opened 2023-09-26
Return resource id for subnets

jsaun:
TES expects the batch subnet as a full resource id.

Related PR: https://github.com/DataBiosphere/leonardo/pull/3432 is blocked until we can return the resource id. 

https://broadworkbench.atlassian.net/browse/TOAZ-335


aherbst-broad (approved):

aherbst-broad:
@jsaun is there an associated ticket?

rtitle (approved):
LGTM.. longer term though I think we want to remove the direct dependency on TES to the LZ Service.

https://broadinstitute.slack.com/archives/C02GJ782BAA/p1695993003954919?thread_ts=1695934429.002639&cid=C02GJ782BAA

jsaun:
> @jsaun is there an associated ticket?

No, but I'll write one quick.

PR #1492, opened 2023-09-28
[BENCH-1270] - Added --allow-releaseinfo-change to apt-get update

johntbates:
Resolves error when running `post-startup.sh` on newly created environments in dev:

```
E: Repository 'http://packages.cloud.google.com/apt gcsfuse-buster InRelease' changed its 'Origin' value from 'gcsfuse-jessie' to 'namespaces/gcs-fuse-prod/repositories/gcsfuse-buster'
E: Repository 'http://packages.cloud.google.com/apt gcsfuse-buster InRelease' changed its 'Label' value from 'gcsfuse-jessie' to 'namespaces/gcs-fuse-prod/repositories/gcsfuse-buster'
```

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -225,7 +225,7 @@ fi
 emit "Resynchronizing apt package index..."
 
 # The apt package index may not be clean when we run; resynchronize
-apt-get update
+apt-get update --allow-releaseinfo-change
 
 # Create the target directories for installing into the HOME directory
 ${RUN_AS_LOGIN_USER} "mkdir -p '${USER_BASH_COMPLETION_DIR}'"


wnojopra (commented):
1) At the top of this file there's this comment:
```
# NOTE FOR CONTRIBUTORS:
#   This startup script closely mirrors the startup script used for Vertex AI Notebook instances here: service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh.
#   Please ensure that changes to shared logic are reflected in both scripts.
```
So I think you'll also want to update the [dataproc startup script](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/dataproccluster/startup.sh).

2) IIUC, the flag allows downloading from repos that have updated their releases. I _think_ this change is fine, but its not 100% clear to me if there are any other side effects/consequences. I'm going to add Matt as a reviewer instead of me his opinion.

mbookman:
Thanks for digging into this @johntbates!

Went into this deeper on an internal discussion thread.
This may be a good short term (maybe long term even) fix, though we might also want to consider building and publishing our own VM images to avoid some of the post-startup steps in production.

FWIW, I tested to see if the problem is due to a new DeepLearning image (20230822) and it does *not* appear to be. I tested with the same image as another VM I already have successfully booted and it failed as well:

```
$ terra resource create gcp-notebook \
  --name test_pinned \
  --vm-image-name r-latest-cpu-experimental-v20230807 \
  --vm-image-project deeplearning-platform-release
```

resulted in:

> Something went wrong. Try creating a new environment. Contact Verily for more support. Error output:
> Error on line 1, command "apt-get". See /home/jupyter/.terra/post-startup-output.txt for more information.

wnojopra (approved):
Thanks, Matt. I'm approving as this seems to be a good immediate fix for prod.

PR #1493, opened 2023-09-29
[BENCH-1270] - Dataproc startup script:  added --allow-releaseinfo-ch…

johntbates:
…ange to apt-get update

Adds the same flag to `apt-get update' to resolve the issue reported in [BENCH-1270](https://verily.atlassian.net/browse/BENCH-1270) when creating dataproc cluster nodes.


johntbates:
This does not seem to be blocking dataproc development or use, so I'm going to close this PR.

PR #1494, opened 2023-09-29
[BENCH-1264] No retries on more errors creating GCE instance

ddietterich:
Return errors instead of dropping into a dismal failure.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/GcpFlightExceptionUtils.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/GcpFlightExceptionUtils.java
@@ -1,6 +1,8 @@
 package bio.terra.workspace.service.resource;
 
 import bio.terra.common.exception.BadRequestException;
+import bio.terra.common.exception.ForbiddenException;
+import bio.terra.common.exception.UnauthorizedException;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import org.springframework.http.HttpStatus;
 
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/workspace/service/resource/GcpFlightExceptionUtils.java
```diff
@@ -19,4 +21,18 @@ public static void handleGcpBadRequestException(GoogleJsonResponseException e)
...
     }
   }
+
+  /**
+   * Map GCP status codes to WSM exceptions. These are errors that we DO NOT want to retry in the
+   * flight.
+   */
+  public static void handleGcpNonRetryableException(GoogleJsonResponseException e) {
+    int statusCode = e.getStatusCode();
+    switch (HttpStatus.valueOf(statusCode)) {
```
ddietterich, line 31: I'm not 100% sure about this list. I wanted parallel behavior to `pollAndRetry`.

For GCE instance we definitely want to bail on FORBIDDEN and we don't ever expect UNAUTHORIZED.

I _think_ we want to bail on TOO_MANY_REQUESTS, since retrying will be making more requests.

I _think_ we do not want to bail on CONFLICT. I'm not sure what it means in this case. If it means, "wait a minute, I'm busy on that thing" (like it does for WSM and Sam), we would want to retry.


Question

rogerwangcs (approved):

zloery (approved):

PR #1495, opened 2023-10-03
[BENCH-1257] Change AWS token expiration to 30mins

dexamundsen:


zloery (approved):

dexamundsen:
Holding this off until the RC of google token refresh is looked into

PR #1496, opened 2023-10-04
WOR-1282 Fix error deleting old namespaces

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-1282

There are a couple dozen k8s namespaces in production that have the managed identity resource id stored in the database instead of the resource name. This change allows us to tolerate that case.


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/managedIdentity/GetWorkspaceManagedIdentityStep.java
```diff
@@ -44,18 +45,7 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
             .getWorkingMap()
             .get(ControlledResourceKeys.AZURE_CLOUD_CONTEXT, AzureCloudContext.class);
     var msiManager = crlService.getMsiManager(azureCloudContext, azureConfig);
-    ControlledAzureManagedIdentityResource managedIdentityResource =
-        resourceDao
-            .getResourceByName(workspaceId, managedIdentityName)
-            .castByEnum(WsmResourceType.CONTROLLED_AZURE_MANAGED_IDENTITY);
-    if (managedIdentityResource == null) {
```
dvoet, line 51: managedIdentityResource can never be null

marctalbott (approved):

blakery (approved):

PR #1497, opened 2023-10-09
BENCH-994: Create/delete Security Groups at AWS Cloud Context create/delete time.

jmczerk:
[AWS Security Groups](https://docs.aws.amazon.com/vpc/latest/userguide/security-groups.html) are roughly analogous to GCP Firewall Rules.

The intent of this PR is to create a Security Group scoped to the Workspace Cloud Context in each supported AWS region's landing zone at AWS Cloud Context creation time, so that compute resources within the same Workspace have IP reachability with one another (but not across workspaces) within the landing zone's private VPC subnet (assuming they are placed in the Workspace Security Group).  These Security Groups will be deleted at Cloud Context deletion time, after any referencing controlled resources have been destroyed.


service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/CreateWorkspaceApplicationSecurityGroupsStep.java
```diff
@@ -0,0 +1,219 @@
...
+      regionSecurityGroups.put(region.toString(), groupId);
+
+      logger.info(
+          "Mapped Security Group ID '{}' to Workspace {} (Landing Zone {})",
+          groupId,
+          workspaceUuid.toString(),
+          region.toString());
+    }
+
+    flightContext
```
yuhuyoyo, line 185: when i was adding firewall rules in RBS, I updated the map inside the for loop so that if only one region failed, I don't retry on all the other regions too. idk how fast it is for AWS to query for security group.

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/DeleteWorkspaceApplicationSecurityGroupsStep.java
```diff
@@ -0,0 +1,88 @@
...
+      AwsCloudContextService awsCloudContextService,
+      SamService samService,
+      UUID workspaceUuid) {
+    this.crlService = crlService;
+    this.samService = samService;
+    this.awsCloudContextService = awsCloudContextService;
+    this.workspaceUuid = workspaceUuid;
+  }
+
+  private void deleteSecurityGroup(
```
yuhuyoyo, line 39: share this method with the undo step above?

service/src/test/java/bio/terra/workspace/service/workspace/flight/aws/CreateDeleteAwsContextFlightTest.java
```diff
@@ -0,0 +1,238 @@
...
+  private UUID createWorkspace() {
+    Workspace request =
+        WorkspaceFixtures.defaultWorkspaceBuilder(UUID.randomUUID())
+            .spendProfileId(getSpendProfile())
+            .build();
+    return workspaceService.createWorkspace(
+        request, null, null, null, userAccessUtils.defaultUserAuthRequest());
+  }
+
+  private Map<String, StepStatus> getCreateStepNameToStepStatusMap() {
```
yuhuyoyo, line 214: nit static

service/src/test/java/bio/terra/workspace/service/workspace/flight/aws/CreateDeleteAwsContextFlightTest.java
```diff
@@ -0,0 +1,238 @@
...
+    retrySteps.put(
+        SetWorkspaceApplicationEgressIngressStep.class.getName(),
+        StepStatus.STEP_RESULT_FAILURE_RETRY);
+    retrySteps.put(MakeAwsCloudContextStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_RETRY);
+    retrySteps.put(
+        CreateCloudContextFinishStep.class.getName(), StepStatus.STEP_RESULT_FAILURE_RETRY);
+    return retrySteps;
+  }
+
+  private Map<String, StepStatus> getDeleteStepNameToStepStatusMap() {
```
yuhuyoyo, line 229: nit static 

also there's Map.of(key, value)

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/DeleteWorkspaceApplicationSecurityGroupsStep.java
jmczerk: Thanks, added a utility method to reuse and clean up the call site a bit.

service/src/test/java/bio/terra/workspace/service/workspace/flight/aws/CreateDeleteAwsContextFlightTest.java
jmczerk: Streamlined with Map.of(), thanks!

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/SetWorkspaceApplicationEgressIngressStep.java
```diff
@@ -0,0 +1,118 @@
...
+    try {
+      regionCow.authorizeIngress(
+          AuthorizeSecurityGroupIngressRequest.builder()
+              .groupId(securityGroupId)
+              .ipPermissions(
+                  IpPermission.builder()
+                      .fromPort(0)
+                      .toPort(0)
+                      .ipProtocol("-1")
+                      .userIdGroupPairs(UserIdGroupPair.builder().groupId(securityGroupId).build())
```
zloery, line 73: iiuc, this allows ingress from other hosts within a security group - given that the security groups are regional, does this mean that compute resources in a workspace will only be able to reach each other if they're in the same region?

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/SetWorkspaceApplicationEgressIngressStep.java
```diff
@@ -0,0 +1,118 @@
...
+
+  private void authorizeIngress(EC2SecurityGroupCow regionCow, String securityGroupId) {
+    try {
+      regionCow.authorizeIngress(
+          AuthorizeSecurityGroupIngressRequest.builder()
+              .groupId(securityGroupId)
+              .ipPermissions(
+                  IpPermission.builder()
+                      .fromPort(0)
+                      .toPort(0)
```
zloery, line 71: Does this allow all ports, or none? I don't see anything about using `0` in the AWS documentation

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/CreateWorkspaceApplicationSecurityGroupsStep.java
```diff
@@ -0,0 +1,212 @@
...
+        .getWorkingMap()
+        .put(WorkspaceFlightMapKeys.AWS_APPLICATION_SECURITY_GROUP_ID, regionSecurityGroups);
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
```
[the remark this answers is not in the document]
dexamundsen: +1

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/DeleteWorkspaceApplicationSecurityGroupsStep.java
```diff
@@ -0,0 +1,79 @@
...
+  private final CrlService crlService;
+  private final SamService samService;
+  private final AwsCloudContextService awsCloudContextService;
+  private final UUID workspaceUuid;
+
+  public DeleteWorkspaceApplicationSecurityGroupsStep(
+      CrlService crlService,
+      AwsCloudContextService awsCloudContextService,
+      SamService samService,
+      UUID workspaceUuid) {
```
dexamundsen, line 32: Same here - this can be fetched from flightmap

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/DeleteWorkspaceApplicationSecurityGroupsStep.java
```diff
@@ -0,0 +1,79 @@
...
+          securityGroupId,
+          region.toString(),
+          workspaceUuid.toString());
+    }
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    Optional.ofNullable(
```
dexamundsen, line 60: this entire function could be the delete util, called fom create.undo. It handle the other comment on deleting all groups for the workspace

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/CreateWorkspaceApplicationSecurityGroupsStep.java
```diff
@@ -0,0 +1,212 @@
...
+    for (Region region : awsEnvironment.getSupportedRegions()) {
+      Optional<LandingZone> landingZone = awsEnvironment.getLandingZone(region);
+
+      // ID's of security groups are not known until after the security group is created.  Thus, in
+      // order to make this step idempotent and not leak security groups, we need to check whether a
+      // security group has already been created for this workspace in a prior invocation by looking
+      // for a security group tagged for this workspace before we actually create a new one.
+
+      String groupId =
+          findOrCreateSecurityGroupInRegion(
```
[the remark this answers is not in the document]
jmczerk: Correct .... the WSM IAM role is the only entity allowed to create Security Groups in the LZ, so we can depend on these tags being authoritative.

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/SetWorkspaceApplicationEgressIngressStep.java
jmczerk: Yeah, the docs are super sketchy about this, this is the best ref I could find: https://repost.aws/questions/QUVWll2XoIRB6J5JqZipIwZQ/what-is-mean-fromport-is-0-and-toport-is-0-in-security-groups-ippermission-ippermissionegress#ANlQylxlBvSaqrIip2SAFajQ

I'll add some verbiage in comments for egress and ingress being explicit about what these rules accomplish.

service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java
```diff
@@ -939,4 +944,41 @@ public static void checkException(SdkException ex, String altMessage, boolean ig
...
+
+      logger.info(
+          "Deleted Security Group ID '{}' for Workspace {} (Landing Zone {})",
+          securityGroupId,
+          workspaceUuid.toString(),
+          awsRegion.toString());
+
+    } catch (Ec2Exception e) {
+      if (e.awsErrorDetails().errorCode().equals("InvalidGroup.NotFound")) {
+        throw new NoSuchElementException(
```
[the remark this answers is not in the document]
jmczerk: This is specifically here to hide the AWS internals of parsing the AWS error code so that callers can be explicitly catch/log/ignore `NoSuchElementException` for idempotent deletes, and any other AWS exceptions will propagate.  Non-idempotent deletes (rollback case) will let the NSE propagate.

service/src/main/java/bio/terra/workspace/service/workspace/AwsCloudContextService.java
```diff
@@ -86,20 +96,47 @@ public void addCreateCloudContextSteps(
       UUID workspaceUuid,
       SpendProfile spendProfile,
       AuthenticatedUserRequest userRequest) {
+    if (featureService.isFeatureEnabled(
+        FeatureService.AWS_APPLICATIONS_ENABLED,
+        flightBeanBag.getSamService().getSamUser(userRequest).getEmail())) {
```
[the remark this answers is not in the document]
jmczerk: Actually, this will not be a per-user feature flag, as it is only here to gate.  We will only gate by user at the AWS platform level.  I will fix this up.

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/SetWorkspaceApplicationEgressIngressStep.java
jmczerk: Yes, that's correct.

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/SetWorkspaceApplicationEgressIngressStep.java
```diff
@@ -0,0 +1,118 @@
...
+                      .ipProtocol("-1")
+                      .userIdGroupPairs(UserIdGroupPair.builder().groupId(securityGroupId).build())
+                      .build())
+              .build());
+      logger.info(
+          "Allowed ingress only from other hosts within Security Group {} on Security Group {}",
+          securityGroupId,
+          securityGroupId);
```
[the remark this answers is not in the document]
jmczerk: I struggled a little with clarity here.  If I change the wording to `Allowed ingress only from other hosts internally on Security Group X`, does that make it clear what it's accomplishing?

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/DeleteWorkspaceApplicationSecurityGroupsStep.java
jmczerk: See above.

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/CreateWorkspaceApplicationSecurityGroupsStep.java
```diff
@@ -0,0 +1,212 @@
...
+  private final CrlService crlService;
+  private final SamService samService;
+  private final AwsCloudContextService awsCloudContextService;
+  private final UUID workspaceUuid;
+
+  public CreateWorkspaceApplicationSecurityGroupsStep(
+      CrlService crlService,
+      AwsCloudContextService awsCloudContextService,
+      SamService samService,
+      UUID workspaceUuid) {
```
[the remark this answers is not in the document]
dexamundsen: Fetching it from flighmap - only helps to reduce another param passed to this flight. 
the class member could still be present - we initialize it in the do / undo by fetching it from flight map. 
not a blocker.

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/CreateWorkspaceApplicationSecurityGroupsStep.java
```diff
@@ -0,0 +1,212 @@
...
+    this.awsCloudContextService = awsCloudContextService;
+    this.workspaceUuid = workspaceUuid;
+  }
+
+  private TagSpecification generateTagSpecification(LandingZone landingZone, UUID workspaceUuid) {
+    Metadata metadata = landingZone.getMetadata();
+    return TagSpecification.builder()
+        .resourceType(ResourceType.SECURITY_GROUP)
+        .tags(
+            List.of(
```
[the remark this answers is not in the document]
jmczerk: So the EC2 `Tag` class is different from the STS `Tag` class (which those particular methods use, for session tagging).  I feel like I wasted a lot of time trying to find a good solution for generic tagging methods, I'd rather not revisit this until we end up with another class that's tagging something resources with EC2 tags (likely the upcoming EC2 instance controlled resource type) and factor out the commonality then.

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/CreateWorkspaceApplicationSecurityGroupsStep.java
```diff
@@ -0,0 +1,212 @@
...
+      EC2SecurityGroupCow regionCow, UUID workspaceUuid, LandingZone landingZone) {
+
+    CreateSecurityGroupResponse response =
+        regionCow.create(
+            CreateSecurityGroupRequest.builder()
+                .vpcId(landingZone.getApplicationVpcId().orElseThrow())
+                .description(
+                    String.format(
+                        "Application Security Group for Workspace %s", workspaceUuid.toString()))
+                .groupName(String.format("ws-app-%s", workspaceUuid.toString()))
```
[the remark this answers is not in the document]
jmczerk: They only are unique within region, but appended anyway.

service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/aws/CreateWorkspaceApplicationSecurityGroupsStep.java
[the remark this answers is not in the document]
jmczerk: Changed the logic to always do lookup by tags in undo step.  Note that we intentionally let `AwsUtils.deleteWorkspaceSecurityGroup` thrown exceptions propagate here, as we should not expect a security group we just found to fail deletion in the absence of an error (unlike the idempotency case in delete do step).

Not crazy about the fact that this instantiates nested cows, maybe we should back out the util function... That said I think we might be getting to the point of over-fussing this PR, so unless anyone has strong feelings I'll leave it like this.

dexamundsen (approved):

PR #1498, opened 2023-10-10
BENCH-1330: add permission iap tunneling on gce VM owners

yuhuyoyo:


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gceinstance/GceInstanceCloudSyncStep.java
```diff
@@ -91,8 +87,7 @@ public StepResult doStep(FlightContext flightContext)
    * works in com.google.cloud.Policy objects, but these are not used by the compute engine API.
    * Transform the com.google.cloud.Policy into a list of bindings to use for the GCP GCE API.
    */
-  private List<Binding> createBindings(GcpCloudContext cloudContext, FlightMap workingMap)
-      throws InterruptedException {
+  private List<Binding> createBindings(GcpCloudContext cloudContext) throws InterruptedException {
```
yuhuyoyo, line 90: if we want to flag-guard this, we will create an optional policy binding here instead of in the mapping

zloery (approved):

jmczerk (approved):

PR #1499, opened 2023-10-10
[WOR-1294] Add tag permissions to the get SAS token API

giventocode:
In this PR:
- Added the `t` (tag) permission to the API that generates SAS tokens.
- The `t` permission is allowed when the user has write access to the workspace.

### DIFF
--- a/openapi/src/common/parameters.yaml
+++ b/openapi/src/common/parameters.yaml
@@ -128,6 +128,7 @@ components:
         |`w`         | Write       |
         |`d`         | Delete      |
         |`l`         | List        |
+        |`t`         | Tags        |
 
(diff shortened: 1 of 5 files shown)


aherbst-broad (approved):

sergiygetlin (approved):
LGTM. Just to clarify what does this 't' permission would deny/allow?

giventocode:
> LGTM. Just to clarify what does this 't' permission would deny/allow?

Thanks! It allows setting [tags for blobs](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-index-how-to?tabs=azure-portal).

PR #1503, opened 2023-10-13
BENCH-1371: Support for user provided startup script for vertex AI notebooks

rogerwangcs:
https://verily.atlassian.net/browse/BENCH-1371

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -772,7 +772,7 @@ if [[ -n "${USER_STARTUP_SCRIPT}" ]]; then
   # Copy the user's startup script to the user's .terra directory
   curl -L "${USER_STARTUP_SCRIPT}" > "${USER_STARTUP_SCRIPT_FILE}"
 
-  # Run the user's startup script as the login user
+  # Run the user's startup script as the root user so that they made install packages
   chmod +x "${USER_STARTUP_SCRIPT_FILE}"
   source "${USER_STARTUP_SCRIPT_FILE}"
 fi
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
@@ -16,6 +16,7 @@
 #   - instance/guest-attributes/startup_script/message: Set by this script, storing the message of this script's execution. If the status is "ERROR", this message will contain an error message, otherwise it will be empty.
 #   - instance/attributes/terra-cli-server: Read by this script to configure the Terra CLI server.
 #   - instance/attributes/terra-workspace-id: Read by this script to configure the Terra CLI workspace.
+#   - instance/attributes/terra-user-startup-script: Read by this script to optionally run a user-provided startup script.
 #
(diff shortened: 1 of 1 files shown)


rogerwangcs:
@ddietterich an upstream wsm change.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -756,6 +757,26 @@ fi
...
+if [[ -n "${USER_STARTUP_SCRIPT}" ]]; then
+  emit "Running user provided startup script..."
+  readonly USER_STARTUP_SCRIPT_FILE="${USER_TERRA_CONFIG_DIR}/user-startup-script.sh"
+
+  # Copy the user's startup script to the user's .terra directory
+  curl -L "${USER_STARTUP_SCRIPT}" > "${USER_STARTUP_SCRIPT_FILE}"
+
+  # Run the user's startup script as the login user
+  chmod +x "${USER_STARTUP_SCRIPT_FILE}"
+  source "${USER_STARTUP_SCRIPT_FILE}"
```
yuhuyoyo, line 777: how are you running it as login user?

rogerwangcs: oh ty, the comment should be "run as root user".

yuhuyoyo (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
rogerwangcs: I thought it was intentional that we give root access? This is the current behavior:
```
User jupyter may run the following commands on vm-name:
    (ALL : ALL) ALL
    (ALL : ALL) NOPASSWD: ALL
```

yuhuyoyo: interesting! didn't realize we allow sudo there. 

so it's not a change in behavior, sgtm

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh
```diff
@@ -756,6 +757,26 @@ fi
...
+if [[ -n "${USER_STARTUP_SCRIPT}" ]]; then
+  emit "Running user provided startup script..."
+  readonly USER_STARTUP_SCRIPT_FILE="${USER_TERRA_CONFIG_DIR}/user-startup-script.sh"
+
+  # Copy the user's startup script to the user's .terra directory
+  curl -L "${USER_STARTUP_SCRIPT}" > "${USER_STARTUP_SCRIPT_FILE}"
+
+  # Run the user's startup script as the root user so that they made install packages
+  chmod +x "${USER_STARTUP_SCRIPT_FILE}"
+  source "${USER_STARTUP_SCRIPT_FILE}"
```
mbookman, line 777: I don't think we want to source the script as that can produce side effects for this script.
I think we just want to execute it.

We should also log output from the script to a known location, such as `~/.terra/user-startup-output.txt` (too bad our log file is ".txt" instead of ".log").

Would be good to update the above to:

```
  emit "Running user provided startup script, output to ${USER_STARTUP_OUTPUT_FILE}..."
```

mbookman (changes requested):

PR #1507, opened 2023-10-17
Wm 2274 pg dump image

dvoet:
https://broadworkbench.atlassian.net/browse/WM-2274

changing the docker build process for azureDatabaseUtils to build a base image containing pg_dump then have jib use that image to add the java bits

### DIFF
--- a/azureDatabaseUtils/Dockerfile
+++ b/azureDatabaseUtils/Dockerfile
@@ -5,8 +5,8 @@
 ### Sourced from https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-debian
 FROM us.gcr.io/broad-dsp-gcr-public/base/jre:17-debian
 
-# refresh the repository
-RUN apt-get update
-
-# Add postgres client for pg_dump command
-RUN apt-get install postgresql-client-14 -y
+# Add postgres client for pg_dump command and clean up
+RUN apt-get update  \
+    && apt-get install postgresql-client-14 -y --no-install-recommends \
+    && apt-get clean  \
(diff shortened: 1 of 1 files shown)


azureDatabaseUtils/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+### for pg_dump commands to backup/restore the database.
+
+### Sourced from https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-debian
+FROM us.gcr.io/broad-dsp-gcr-public/base/jre:17-debian
+
+# refresh the repository
+RUN apt-get update
+
+# Add postgres client for pg_dump command
+RUN apt-get install postgresql-client-14 -y
```
mspector, line 12: what do you think about pinning this to a specific version (as Codacy suggests)? it doesn't seem that important to me, but...

dvoet: it is already pinned to postgres 14 so I am inclined to ignore

blakery (approved):

aherbst-broad (approved):
Building using jib locally works, lgtm.

davidangb (approved):
fwiw this is a slightly more detailed version of what's been working for WDS already: https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/Dockerfile … so :+1:

dvoet:
> fwiw this is a slightly more detailed version of what's been working for WDS already: https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/Dockerfile … so 👍

yeah, I copied from there and updated to cleanup :)

PR #1513, opened 2023-10-26
[WM-2278] add pgDump azureDatabaseUtils command

mspector:
Adds a pgDump command to azureDatabaseUtils. This is needed for [workflow cloning](https://docs.google.com/document/d/1m135VjGM7kdZxaxQAXeKIhmCP64cM7bXtUE67Q6CgNI/edit#bookmark=kix.7nqrewpayoh5). This command will eventually be called during `CloneAllResourcesFlight` to enable generic Azure database cloning.


azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/storage/AzureBlobStorage.java
```diff
@@ -0,0 +1,106 @@
...
+    try (toStream) {
+      blobContainerClient.getBlobClient(blobName).downloadStream(toStream);
+    } catch (IOException ioEx) {
+      throw new LaunchProcessException("Error streaming input to child process", ioEx);
+    }
+  }
+
+  private BlobContainerClient constructBlockBlobClient(
+      String workspaceId, String blobstorageDetails) {
+    // TODO: determine where blobstorageDetails should come from.
```
mspector, line 74: seeking reviewer opinions on how to handle this. I've been manually testing this by passing it into the azureDatabaseUtils command. Ultimately, it seems like it should be passed in during CloneAllResourcesFlight (which will be worked on as part of [WM-2236](https://broadworkbench.atlassian.net/browse/WM-2236)).

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +89,97 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+      String sourceDbHost,
+      String sourceDbPort,
+      String sourceDbUser,
+      String pgDumpFilename,
+      String destinationWorkspaceId,
+      String blobstorageDetails) {
+    logger.info("running DatabaseService.pgDump against {}", sourceDbName);
+    logger.info("destinationWorkspaceId: {}", destinationWorkspaceId);
+    try {
+      // Grant the database role (sourceDbName) to the workspace identity (sourceDbUser).
```
dvoet, line 104:
```suggestion
      // Grant the database role (sourceDbName) to the landing zone identity (sourceDbUser).
```

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +89,97 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+          localProcessLauncher.getInputStream(),
+          pgDumpFilename,
+          destinationWorkspaceId,
+          blobstorageDetails);
+
+      String output = checkForError(localProcessLauncher);
+      logger.info("pg_dump output: {}", output);
+
+    } catch (PSQLException ex) {
+      logger.error("process error: {}", ex.getMessage());
```
dvoet, line 129: why ignore this exception?

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +89,97 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+      LocalProcessLauncher localProcessLauncher = new LocalProcessLauncher();
+      localProcessLauncher.launchProcess(commandList, envVars);
+
+      storage.streamOutputToBlobStorage(
+          localProcessLauncher.getInputStream(),
+          pgDumpFilename,
+          destinationWorkspaceId,
+          blobstorageDetails);
+
+      String output = checkForError(localProcessLauncher);
```
dvoet, line 125: I feel like an exception should be thrown if there is an error

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/runners/PgDumpDatabaseRunner.java
```diff
@@ -0,0 +1,53 @@
...
+  private String sourceDbUser;
+
+  @Value("${env.params.dumpfileName}")
+  private String dumpfileName;
+
+  @Value("${env.params.destinationWorkspaceId}")
+  private String destinationWorkspaceId;
+
+  @Value("${env.params.blobstorageDetails}")
+  private String blobstorageDetails;
```
dvoet, line 33: what's in blobstorageDetails?

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/storage/AzureBlobStorage.java
```diff
@@ -0,0 +1,106 @@
...
+    // ultimately calls the pgDump function.
+
+    BlobServiceClient blobServiceClient =
+        new BlobServiceClientBuilder().endpoint(blobstorageDetails).buildClient();
+
+    try {
+      // the way storage containers are set up in a workspace are as follows:
+      // billing project gets a single azure storage account
+      // each workspace gets a container inside of that storage account to keep its data
+      return blobServiceClient.getBlobContainerClient("sc-" + workspaceId);
```
dvoet, line 87: I don't like this string concat being here. I think this utility should just be told where to put it and let the intricacies of where to put it be determined upstream

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/storage/AzureBlobStorage.java
```diff
@@ -0,0 +1,106 @@
...
+    } catch (BlobStorageException e) {
+      // if the default workspace container doesn't exist, something went horribly wrong
+      logger.error("Default storage container missing for workspace id {}", workspaceId);
+      throw (e);
+    }
+  }
+
+  @Override
+  public void deleteBlob(String blobFile, String workspaceId, String authToken) {
```
dvoet, line 96: is this used?

azureDatabaseUtils/gradle/dependencies.gradle
```diff
@@ -3,6 +3,8 @@ dependencies {
   // Azure dependencies
   implementation 'com.azure:azure-identity:1.9.1'
   implementation 'com.azure:azure-identity-extensions:1.1.4'
+  implementation 'org.postgresql:postgresql'
```
dvoet, line 6: this is already on line 9 below

dvoet (commented):
first pass

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/runners/PgDumpDatabaseRunner.java
mspector: this is the blob storage URL with a SAS token included as a query string. WDS makes a couple of WSM API calls to get it (see [here](https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/main/java/org/databiosphere/workspacedataservice/workspacemanager/WorkspaceManagerDao.java#L50)) - it seems like in this case it should be passed in as an argument, but I'm not certain about that. What do you think?

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/storage/AzureBlobStorage.java
mspector: good call... not yet, but I think it will be. I'll remove for now and reintroduce it when needed

azureDatabaseUtils/gradle/dependencies.gradle
mspector: thanks - I didn't recognize it as being equivalent

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
mspector: good question. I have to admit that I lifted this part from WDS's code, and didn't question it - I'll talk to someone on that team and see if there's some non-obvious reason for ignoring it

mspector: I got an answer: they implemented it this way because, for WDS, this process is part of app startup, and they wanted it to continue even if some part of the dump/restore failed, so that they could gracefully fall back to an empty database.

I think that reasoning shouldn't apply in our case, because (as part of our Leonardo changes) we're going to have apps use databases *if they exist* - and if they don't exist, create them from scratch. 

Something I'm not sure about: if this process throws an exception, and the flight step fails as a result, does that cancel the whole flight? or can the flight continue to the next step?

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
mspector: I agree, I'll make that change (see [my comment above](https://github.com/DataBiosphere/terra-workspace-manager/pull/1513#discussion_r1374877561))

PR #1514, opened 2023-10-30
[WM-2278, WM-2279] azureDatabaseUtils db dump and restore

mspector:
includes changes from https://github.com/DataBiosphere/terra-workspace-manager/pull/1513/files


azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,157 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
 
     databaseDao.restoreLoginPrivileges(namespaceRole);
   }
+
+  public void pgDump(
```
mspector, line 94: most of this is copied from WDS's implementation

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,157 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+    storage.streamOutputToBlobStorage(
+        localProcessLauncher.getInputStream(),
+        blobFileName,
+        blobContainerName,
+        blobContainerUrlAuthenticated);
+
+    checkForError(localProcessLauncher);
+  }
+
+  public void pgRestore(
```
mspector, line 135: same here - most of this is copied from WDS's implementation

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,157 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+      } catch (IOException e) {
+        errorMsg =
+            "process failed with exit code "
+                + exitCode
+                + ", but encountered an exception reading the error output: "
+                + e.getMessage();
+      }
+      throw new LaunchProcessException(errorMsg);
+    }
+    logger.info("localProcessLauncher terminated with exit code {}", exitCode);
```
dvoet, line 238: exit code will always be 0

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,157 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+    // (2) the workspace identity can grant itself access at any time, so revoking the role
+    // doesn't protect us.
+    databaseDao.grantRole(adminUser, dbName);
+
+    List<String> commandList = generateCommandList("pg_dump", dbName, dbHost, dbPort, adminUser);
+    Map<String, String> envVars = null;
+    try {
+      envVars = Map.of("PGPASSWORD", determinePassword());
+    } catch (PSQLException e) {
+      logger.error(e.getMessage());
```
dvoet, line 117: why not just let this error bubble up? As it is, the message is logged but envVars is left null probably leading to a NPE later

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,157 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+          blobContainerName,
+          blobContainerUrlAuthenticated);
+
+      // doCleanup isn't set to true until checkForError completes without throwing an exception
+      checkForError(localProcessLauncher);
+      databaseDao.reassignOwner(adminUser, dbName);
+      doCleanup = true;
+    } finally {
+      if (doCleanup) {
+        storage.deleteBlob(blobFileName, blobContainerName, blobContainerUrlAuthenticated);
```
dvoet, line 181: I don't think cleanup should be in this utility. I think it should be a step in the stairway flight after the step that calls pg restore

dvoet (changes requested):

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,157 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+      String adminUser,
+      String blobFileName,
+      String blobContainerName,
+      String blobContainerUrlAuthenticated) {
+
+    // Grant the database role (dbName) to the landing zone identity (adminUser).
+    // In theory, we should be revoking this role after the operation is complete.
+    // We are choosing to *not* revoke this role for now, because:
+    // (1) we could run into concurrency issues if multiple users attempt to clone the same
+    // workspace at once;
```
yuliadub, line 107: users should be able to clone at once thought, so I am not sure I understand how identity impacts that but maybe I missed another piece here?

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
yuliadub: +1

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
yuliadub: I'd argue to just do it here (since we wont need it for anything else, might as well keep it all contained in one place) if we do this somewhere else - aren't we are more likely to end up with blobs that may not get cleaned up?

azureDatabaseUtils/src/test/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseServiceTest.java
```diff
@@ -14,6 +18,7 @@ public class DatabaseServiceTest extends BaseUnitTest {
 
   @MockBean private DatabaseDao databaseDao;
   @MockBean private Validator validator;
+  @MockBean private BlobStorage storage;
```
yuliadub, line 21: are you streaming to a storage container in azure when runing these tests?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/AzureDatabaseUtilsRunner.java
```diff
@@ -120,6 +129,64 @@ public void createDatabaseWithDbRole(
         aksNamespace);
   }
 
+  public void pgDumpDatabase(
```
yuliadub, line 132: do you have any tests to make sure that the pg_dump that you created is what you exception? same for when you restore?

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,157 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+      } catch (PSQLException e) {
+        logger.error(e.getMessage());
+      }
+      LocalProcessLauncher localProcessLauncher = new LocalProcessLauncher();
+      localProcessLauncher.launchProcess(commandList, envVars);
+
+      logger.info("running DatabaseService.pgRestore: {}", String.join(" ", commandList));
+      logger.info("blobContainerName: {}", blobContainerName);
+
+      storage.streamInputFromBlobStorage(
```
yuliadub, line 169: in the WDS case, we cant just restore the [schema ](https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/main/java/org/databiosphere/workspacedataservice/service/BackupRestoreService.java#L192)from the pg_dump and leave it as is -- we actually need to change the schema definition from old workspace id to new workspace id (since the pg_dump we take comes from the old workspace). In this code, I dont see either of those values (workspaceId or sourceWorkspaceId) passed into these functions -- would you expect that this needs to happen outside of this code? My preference would be to keep it here since if an error does happen, this is contained in the restore code where we wont leave the user with a non usable wds db schema. 

not sure if cbas needs similar alteration since as far as I know it doesnt have a concept of instances?

dvoet: Any transformations that need to happen after the pg_dump are app specific. So anything WDS needs to do will have to be up to WDS and we will need to figure out a way for WDS to know that it is starting up from a clone. CBAS wants to clear out any workflow runs from the source workspace. I think their plan is to delete any record older than the database creation time which will come from WSM and passed in by Leo.

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,157 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+      String blobContainerUrlAuthenticated) {
+
+    // Grant the database role (dbName) to the landing zone identity (adminUser).
+    // In theory, we should be revoking this role after the operation is complete.
+    // We are choosing to *not* revoke this role for now, because:
+    // (1) we could run into concurrency issues if multiple users attempt to clone the same
+    // workspace at once;
+    // (2) the workspace identity can grant itself access at any time, so revoking the role
+    // doesn't protect us.
+    databaseDao.grantRole(adminUser, dbName);
```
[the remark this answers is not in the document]
dvoet: because we have not thus far and granting on the fly is easier than back filling

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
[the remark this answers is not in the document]
dvoet: I think success or failure is all WSM will need

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
dvoet: The problem is when there are concurrent clone operations. If one operation removes access, it may cause the other process to fail. Cloning operations are independent so understanding when to remove access is difficult.

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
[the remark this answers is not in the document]
yuliadub: does this mean these operation wont be async? i.e if one user is cloning and another cloning request comes in, would the second request be kicked off before the first one completes or no? or would the two requests not even know about each other?

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
yuliadub: ah like you mean if someone changes permissions to a given workspace or something like that? would it make sense to give the cloning operation the necessary permission for the whole e2e it needs to complete as long as it had them at cloning request?

azureDatabaseUtils/src/test/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseServiceTest.java
mspector: no - these are unit tests - but that's a good reminder that I should have some kind of connected test here too

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/AzureDatabaseUtilsRunner.java
mspector: no, and I'm not sure how I would do that - do you have a suggestion?

this isn't exactly the same thing, but we do have a plan to generate an md5sum of the dump file, and verify that it's the same when we restore (to let us know that the dump file hadn't been altered). but that will be part of the overall db clone flight.

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
mspector: > not sure if cbas needs similar alteration since as far as I know it doesnt have a concept of instances?

that's right - we won't need to do this kind of thing for cbas. Doug's comment about our plan is correct

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
mspector: (sorry for my slow reply...)

> aren't we are more likely to end up with blobs that may not get cleaned up?

maybe, I'm not sure. but doing the cleanup in a separate step fits the stairway model better, because each step needs to be idempotent. (If this step completes successfully, it'll delete the dump file, making subsequent runs fail)

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
mspector: I'm not sure how to actually implement that idea in a way that would help avoid the case where some independent operation removes access at an inopportune moment...

azureDatabaseUtils/src/test/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseServiceTest.java
mspector: now I think it might be better to test this in the context of the stairway steps that call these utilities, which will come in a later PR

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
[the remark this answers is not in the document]
dvoet: that is correct, they are completely independent operations

dvoet (approved):

ashanhol (approved):
The pgdump/pgrestore functions look good to me :)

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
yuliadub: Im sorry but I still dont quite understand why 2 users cloning at the same time is not possible. With smaller data sets this wont be an issue since cloning happens fast but with larger ones (say where the db dump is more then say 1GB, possibly even smaller) not being to have two users clone at once may cause issues that may cause cloning to fail or cause the user to wait a long time for their workspace to be ready

azureDatabaseUtils/src/test/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseServiceTest.java
yuliadub: got it, my suggestion was going to be to use a mock for this, something like local file storage (so you could test that whatever is written is what you expect) without needing to connect to blob storage/etc

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
mspector: No worries, this thread has become a bit confusing: two users cloning simultaneously *would* be a problem if we revoke the role after the operation is complete, because there's a chance that one user's cloning process would remove the role's permission while another one is trying to operate on the database. 

That's why we're choosing not to revoke the role.

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,145 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+    // workspace at once;
+    // (2) the workspace identity can grant itself access at any time, so revoking the role
+    // doesn't protect us.
+
+    databaseDao.grantRole(adminUser, dbName);
+
+    List<String> commandList = generateCommandList("psql", dbName, dbHost, dbPort, adminUser);
+    Map<String, String> envVars = null;
+    envVars = Map.of("PGPASSWORD", determinePassword());
+    LocalProcessLauncher localProcessLauncher = new LocalProcessLauncher();
```
sergiygetlin, line 157: Does it make sense to inject LocalProcessLauncher, so we can mock it?

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,145 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+    commandList.add("-v");
+    commandList.add("-w");
+
+    return commandList;
+  }
+
+  private void checkForError(LocalProcessLauncher localProcessLauncher)
+      throws LaunchProcessException {
+    // materialize only the first 1024 bytes of the error stream to ensure we don't DoS ourselves
+    int errorLimit = 1024000;
```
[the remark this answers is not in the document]
mspector: oh wow, thanks for catching this

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/storage/BlobStorage.java
```diff
@@ -0,0 +1,18 @@
...
+/* Allow for blob storage mocking. */
+public interface BlobStorage {
+  default void streamOutputToBlobStorage(
+      InputStream fromStream,
+      String blobName,
+      String blobContainerName,
+      String blobContainerUrlAuthenticated) {}
+
+  default void streamInputFromBlobStorage(
+      OutputStream toStream, String blobName, String blobContainerName, String authToken) {}
```
blakery, line 15: Why provide an empty default implementation, instead of just the definition?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/AzureDatabaseUtilsRunner.java
yuliadub: the way we do this in WDS today is we have a docker container that has a db that we can actually take a backup off, and then we have a backup file checked into the repo that we compare the dump to. 

for the restore, we take the checked in dump and restore, then actually run a query to check that the schema was set up correctly and WDS has the expected number of rows that is expected.

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
yuliadub: originally I didn't like the idea of having a database with a schema that was tied to another workspace Id, but I chatted with David more about this and mostly resolved all my concerns (around WDS somehow becoming unusable due to this or us having unnecessary data in the db/etc)

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
yuliadub: aha, okay. I think I finally follow :) thanks for the clarification! I missed the part where this only realted to revoking the role somehow.

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/storage/BlobStorage.java
yuliadub: we had to do this in WDS since we need to set up BlobClient to point at the right blob first, you cant just stream from the blobName directly

yuliadub (approved):

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,145 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+    // We are choosing to *not* revoke this role for now, because:
+    // (1) we could run into concurrency issues if multiple users attempt to clone the same
+    // workspace at once;
+    // (2) the workspace identity can grant itself access at any time, so revoking the role
+    // doesn't protect us.
+
+    databaseDao.grantRole(adminUser, dbName);
+
+    List<String> commandList = generateCommandList("psql", dbName, dbHost, dbPort, adminUser);
+    Map<String, String> envVars = null;
```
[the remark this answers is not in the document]
mspector: thanks -- I think this was a holdover from WDS, in which there were likely intervening lines that made this separation make more sense

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -80,4 +90,145 @@ public void restoreNamespaceRoleAccess(String namespaceRole) {
...
+                + exitCode
+                + ", but encountered an exception reading the error output: "
+                + e.getMessage();
+      }
+      throw new LaunchProcessException(errorMsg);
+    }
+  }
+
+  private String determinePassword() throws PSQLException {
+    return new String(
```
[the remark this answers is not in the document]
mspector: looks like it returns `char[]`

PR #1515, opened 2023-10-30
[WOR-1318] Update java-json library

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1318

I tried removing the pinning in the service `dependencies.gradle` file, but then we end up with the older versions similar to what was in the integration lockfile. Therefore I updated the pinned version to be the one with the fixes.

https://nvd.nist.gov/vuln/detail/CVE-2023-5072


integration/build.gradle
```diff
@@ -46,6 +46,7 @@ dependencies {
     // Google dependencies
     constraints {
         implementation 'com.google.guava:guava:30.1.1-jre'
+        implementation 'org.json:json:20231013' // security vulnerability
```
cahrens, line 49: The integration directory likely is not scanned for vulnerabilities, but we might as well update its version of the library to match the service directory.

marctalbott (approved):

aherbst-broad (approved):
Thanks for taking this.

PR #1511, opened 2023-10-25
Add first pact verification for WDS integration.

jladieu:
This pairs with https://github.com/DataBiosphere/terra-workspace-data-service/pull/372 to establish a contract between WDS and WSM, initially focused on `ReferencedGcpResourceController`.

I tested this running a local pact-broker by following the steps in this [doc](https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2829680649/Contract+Test+Local+Development) and using the displayed IntelliJ run configuration: 
image

PR #1517, opened 2023-11-01
AJ-1235: Bump `dius.pact` from 4.1.7 -> 4.3.19

jladieu:
In prep for introducing new provider tests, this gets WSM up to the minimum recommended version to take advantage of the consumer version selector feature.  For more background, see [thread](https://broadinstitute.slack.com/archives/C043YJ40719/p1698429494520969?thread_ts=1698259062.481639&cid=C043YJ40719)

This is a prerequesite PR for: https://github.com/DataBiosphere/terra-workspace-manager/pull/1511, which adds the first provider test for WSM's contracts with WDS.


service/src/main/java/bio/terra/workspace/service/policy/TpsRetry.java
```diff
@@ -20,9 +20,9 @@ public class TpsRetry {
 
   // The retry function starts with INITIAL_WAIT between retries, and doubles that until it
   // reaches MAXIMUM_WAIT, after which all retries are MAXIMUM_WAIT apart.
-  private static final Duration MAXIMUM_WAIT = Duration.ofSeconds(30);
-  private static final Duration INITIAL_WAIT = Duration.ofSeconds(10);
-  private static final Duration OPERATION_TIMEOUT = Duration.ofSeconds(300);
+  private static final Duration MAXIMUM_WAIT = Duration.ofSeconds(1);
```
jladieu, line 23: I temporarily bumped these retries down to 1 second, which was helpful with iteration speed, but plan to revert this.  Leaving this comment here as a reminder to my future self that this was helpful.

service/gradle/dependencies.gradle
```diff
@@ -115,7 +115,7 @@ dependencies {
   testImplementation "org.testcontainers:postgresql:1.17.5"
 
   // pact
-  testImplementation 'au.com.dius.pact.consumer:junit5:4.1.7'
+  testImplementation 'au.com.dius.pact.consumer:junit5:4.3.19'
```
jladieu, line 118: This is the key version bump that triggered all the other changes in this PR.

service/src/test/java/bio/terra/workspace/pact/ProfileApiTest.java
```diff
@@ -41,6 +41,7 @@ public RequestResponsePact existingAzureBillingProfile(PactDslWithProvider build
             .stringType("managedResourceGroupId")
             .uuid("id");
     return builder
+        .usingLegacyDsl()
```
jladieu, line 44: `usingLegacyDsl()`, `PactBuilder` and returning `V4Pact` are all weird artifacts which _I believe_ are transitional and may be able to go away if we bump to an _even later_ version of Pact (as we have done in WDS).  However, WSM is currently using a version of `gradle` which is incompatible with the newer versions of Pact JVM and so here they remain.

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -35,6 +35,8 @@
 @Tag("pact-test")
 @ExtendWith(PactConsumerTestExt.class)
 public class TpsApiTest {
+  private static final String UUID_REGEX =
```
jladieu, line 38: The `UUID_REGEX` was previously provided by the `dsl.DslPart` static import.  However, the new version of pact seems to have rewritten a bunch of the Kotlin innards, and the `UUID_REGEX` is no hidden in a companion object which hasn't been made visible to the Java runtime.

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -132,17 +132,18 @@ public void testCreatingAPolicyWithNoExistingPolicy() throws Exception {
   }
 
   @Pact(consumer = "wsm", provider = "tps")
-  public RequestResponsePact createPaoWithAPreexistingPolicy(PactDslWithProvider builder) {
+  public V4Pact createPaoWithAPreexistingPolicy(PactBuilder builder) {
     return builder
-        .given(existingPolicyState)
+        .usingLegacyDsl()
+        .given("a policy with the given id exists", Map.of("id", existingPolicyId.toString()))
```
jladieu, line 138: This is an important change that I made throughout.

Several tests were _expecting_ the provider to return the specific `existingPolicyId` and failing, since the newer version of the pact JVM deliberately randomizes this stuff unless explicitly told otherwise.

So, in order for me to not to have to rewrite those expectations, this provides a mechanism for the consumer test to communicate to the provider test the exact ID it's expecting to exist.  The provider can access this using the optional `Map` arg that is provided to the `@State` annotated method that will eventually be implemented for this, and set up its stubbed/mocked responses accordingly to match the consumer expectations.

jladieu: Note: the provider tests don't exist yet, which is why I'm able to make this change without a corresponding provider fix (see https://broadinstitute.slack.com/archives/C02LMB1PQ6R/p1698849691034179?thread_ts=1698761219.655949&cid=C02LMB1PQ6R).

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -182,26 +182,25 @@ public void deletingAnExistingPolicy() throws Exception {
   }
 
   @Pact(consumer = "wsm", provider = "tps")
-  public RequestResponsePact getPaoWithAnExistingPolicy(PactDslWithProvider builder) {
+  public V4Pact getPaoWithAnExistingPolicy(PactBuilder builder) {
     var policyResponseShape =
         new PactDslJsonBody()
-            .valueFromProviderState("objectId", existingPolicyProviderStateValue, existingPolicyId)
+            .stringValue("objectId", existingPolicyId.toString())
```
jladieu, line 188: Since the consumer is providing the `policyId` in the `getPao` of the test that uses this pact, it doesn't actually make sense that the provider would provide a value for this.  Instead, we accomplish communicating the expected policy ID using the parameterized `given` on line 196.

I made this change throughout.

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -290,19 +286,18 @@ public RequestResponsePact mergePaoWhenBothExist(PactDslWithProvider builder) {
             .object("resultingPao");
 
     return builder
-        .given(existingPolicyState)
-        .given(secondPolicyState)
+        .usingLegacyDsl()
+        .given("a policy with the given id exists", Map.of("id", existingPolicyId.toString()))
```
jladieu, line 290: This is how you tell the provider that a policy with a specific ID is expected to exist.

On the provider side tests in TPS, you'll end up with code like:

```
 @State({"a policy with the given id exists"})
  public void policyExists(Map<?, ?> parameters) throws InterruptedException {
    var existingPolicyId = parameters.get("id").toString();
    // TODO: mock/stub out DAO to return a hardcoded policy whenever `existingPolicyId` is requested
  }
```

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -444,27 +437,30 @@ public void listingValidRegionsOnAPolicy() throws Exception {
...
     var explainResponse =
         new PactDslJsonBody()
             .numberType("depth")
-            .valueFromProviderState(
-                "objectId", existingPolicyProviderStateValue, existingPolicyId.toString())
+            .stringValue("objectId", existingPolicyId.toString())
             .object(
                 "policyInput",
-                new PactDslJsonBody().stringType("namespace").stringType("name").closeObject());
+                newJsonBody(
```
jladieu, line 447: I switched over to use the new DSL-sugary version of `PactdslJsonBody()` because the new Kotlin implementation has strict expectations about `null` which aren't satisfied by the old style.  I'm not sure why I didn't have to rewrite _all_ of the `PactDslJsonBody()` stuff in the file, but am happy that I got to avoid doing so.

service/src/test/java/bio/terra/workspace/pact/ProfileApiTest.java
jladieu: Oops, I guess I didn't need to do this at all.  Turns out, we just needed to explicitly specify `PactVersion` as a param to avoid this.  Unrolled this in 70403afc97c79a52e3a3bb81263fed1a2cd8f1d3

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -253,23 +248,21 @@ public RequestResponsePact linkPaoWhenBothExist(PactDslWithProvider builder) {
...
-        .given(existingPolicyState)
-        .given(secondPolicyState)
+        .given("a policy with the given id exists", Map.of("id", existingPolicyId.toString()))
+        .given("a policy with the given id exists", Map.of("id", secondPolicyId.toString()))
         .uponReceiving("A request to link the policies")
         .method("POST")
         .headers(contentTypeJsonHeader)
         .body(linkRequestShape)
-        .pathFromProviderState(
-            "/api/policy/v1alpha1/pao/$s/link".formatted(existingPolicyProviderStateValue),
```
jladieu, line 263: I just want to point out there was a `$s` here instead of a `%s`, which I assumed to be a bug, so I fixed it.

nawatts (approved):

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
blakery: The map passed in `given` is specifying values the pact expects to exist with a particular provider state.

It allows _more_ flexibility on the provider side, because it's requiring that the value matches what the provider says it is on the provider side, not an exact value.
In other words, the provider and consumer only have to agree on the names of things, not their exact value.

So if you have defined in your tests `existingPolicyId` as `some_id`:

`.path("/api/policy/v1alpha1/pao/%s".formatted(existingPolicyId))`

will generate a pact with the path as:
`"/api/policy/v1alpha1/pao/some_id"`

where: 
``` 
.pathFromProviderState(
            "/api/policy/v1alpha1/pao/%s".formatted(existingPolicyProviderStateValue),
            "/api/policy/v1alpha1/pao/%s".formatted(existingPolicyId))
```
will generate a pact with a path:
`"/api/policy/v1alpha1/pao/{{existingPolicyId}}"`

Then the provider can set `existingPolicyId` to whatever they want on their side, so long as it matches the expected type (string, uuid, etc).

I think pact might also run through some randomly generated matching values on the consumer side, to ensure your code can handle any value of the type you specified.

It's plausible that the correct way to do this changed with the version update, but the pattern shouldn't need to be changed.

blakery (changes requested):

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -185,27 +188,25 @@ public void deletingAnExistingPolicy() throws Exception {
...
 
     return builder
-        .given(existingPolicyState)
+        .given("a policy with the given id exists", Map.of("id", existingPolicyId.toString()))
         .uponReceiving("A request to retrieve a policy")
         .method("GET")
-        .pathFromProviderState(
-            "/api/policy/v1alpha1/pao/%s".formatted(existingPolicyProviderStateValue),
-            "/api/policy/v1alpha1/pao/%s".formatted(existingPolicyId))
+            .path("/api/policy/v1alpha1/pao/%s".formatted(existingPolicyId))
```
jladieu, line 201: This is the one spot where I couldn't get the existing test to pass while maintaining `pathFromProviderState`, so it's still just expecting the existingPolicyId that is being explicitly passed in by the consumer.

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -185,27 +188,25 @@ public void deletingAnExistingPolicy() throws Exception {
...
-            .valueFromProviderState("objectId", existingPolicyProviderStateValue, existingPolicyId)
+            .stringValue("objectId", existingPolicyId.toString())
             .stringMatcher("component", tpsComponentRegex)
             .stringMatcher("objectType", tpsObjectTypeRegex)
             .object("effectiveAttributes", tpsPolicyInputsObjectShape)
             .object("attributes", tpsPolicyInputsObjectShape);
 
     return builder
-        .given(existingPolicyState)
+        .given("a policy with the given id exists", Map.of("id", existingPolicyId.toString()))
```
jladieu, line 198: This is how you tell the provider that a policy with a specific ID is expected to exist.

On the provider side tests in TPS, you'll end up with code like:

 @State({"a policy with the given id exists"})
  public void policyExists(Map<?, ?> parameters) throws InterruptedException {
    var existingPolicyId = parameters.get("id").toString();
    // TODO: mock/stub out DAO to return a hardcoded policy whenever `existingPolicyId` is requested
  }

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
blakery: yup, definitely a bug

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
jladieu: I put things back the way they were (for the most part).  One [stubborn test](https://github.com/DataBiosphere/terra-workspace-manager/pull/1517#discussion_r1378911452) resisted, and I had to omit the `valueFromProviderState` calls for it.

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -185,27 +188,25 @@ public void deletingAnExistingPolicy() throws Exception {
   public RequestResponsePact getPaoWithAnExistingPolicy(PactDslWithProvider builder) {
     var policyResponseShape =
         new PactDslJsonBody()
-            .valueFromProviderState("objectId", existingPolicyProviderStateValue, existingPolicyId)
+            .stringValue("objectId", existingPolicyId.toString())
```
blakery, line 191: This works with:
`.valueFromProviderState("objectId", existingPolicyProviderStateValue, existingPolicyId.toString())`
instead of
`.valueFromProviderState("objectId", existingPolicyProviderStateValue, existingPolicyId)`

Apparently the uuid isn't getting converted to a string automatically.

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
```diff
@@ -185,7 +188,8 @@ public void deletingAnExistingPolicy() throws Exception {
   public RequestResponsePact getPaoWithAnExistingPolicy(PactDslWithProvider builder) {
     var policyResponseShape =
         new PactDslJsonBody()
-            .valueFromProviderState("objectId", existingPolicyProviderStateValue, existingPolicyId)
+            .valueFromProviderState(
+                "objectId", existingPolicyProviderStateValue, existingPolicyId.toString())
```
jladieu, line 192: @blakery this did the trick!  Thanks for figuring out that the `toString()` was essential.

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
jladieu: Fixed by e419a3404c191f2765a7411e50cb2a1861a05763

service/src/test/java/bio/terra/workspace/pact/TpsApiTest.java
jladieu: Didn't end up needing to do this after all, but the comment is still valid if this construct creeps back in down the road.

blakery (approved):

PR #1520, opened 2023-11-06
TOAZ-336 phase 4: Send Janitor messages for Azure Database and KubernetesNamespace

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/TOAZ-336

Depends on: https://github.com/DataBiosphere/terra-resource-janitor/pull/169

Bumps CRL dependency and adds Janitor handling for Azure Database and KubernetesNamespace. Tested on a BEE with a local Janitor (see above Janitor PR).


service/gradle/dependencies.gradle
```diff
@@ -22,9 +22,6 @@ dependencies {
   // Azure dependencies
   implementation 'com.azure:azure-storage-blob:12.21.1'
   implementation group: 'com.azure.resourcemanager', name: 'azure-resourcemanager-monitor', version: '2.31.0'
-  implementation group: 'com.azure.resourcemanager', name: 'azure-resourcemanager-containerservice', version: '2.31.0'
```
rtitle, line 25: These Azure libraries now come from CRL.

aherbst-broad (approved):
lgtm

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/CreateAzureDatabaseStep.java
```diff
@@ -58,13 +59,34 @@ private String getPodName() {
...
+    var landingZoneId =
+        landingZoneApiDispatch.getLandingZoneId(
+            bearerToken, workspaceService.getWorkspace(workspaceId));
+    var databaseResource =
+        landingZoneApiDispatch
+            .getSharedDatabase(bearerToken, landingZoneId)
+            .orElseThrow(() -> new RuntimeException("No shared database found"));
+
+    // Record resource for cleanup in Janitor
+    crlService.recordAzureCleanup(
```
dvoet, line 82: shouldn't we record cleanup before creating the database?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/kubernetesNamespace/CreateKubernetesNamespaceStep.java
```diff
@@ -44,6 +51,16 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
           null,
           null,
           null);
+
+      // Record namespace for cleanup in Janitor
+      crlService.recordAzureCleanup(
```
dvoet, line 56: register cleanup before resource creation

dvoet (approved):
overall looks good with a couple points about ordering

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/CreateAzureDatabaseStep.java
rtitle: Hm yeah -- will fix.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/KubernetesClientProvider.java
```diff
@@ -148,4 +144,14 @@ public StepResult stepResultFromException(ApiException exception, HttpStatus...
...
   }
+
+  public ApiAzureLandingZoneDeployedResource getClusterResource(UUID workspaceId) {
+    var bearerToken = new BearerToken(samService.getWsmServiceAccountToken());
+    UUID landingZoneId =
+        landingZoneApiDispatch.getLandingZoneId(
+            bearerToken, workspaceService.getWorkspace(workspaceId));
+    return landingZoneApiDispatch
+        .getSharedKubernetesCluster(bearerToken, landingZoneId)
+        .orElseThrow(() -> new RuntimeException("No shared cluster found"));
```
sergiygetlin, line 155: Not critical, but probably would be better to throw more specific exception.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/CreateAzureDatabaseStep.java
```diff
@@ -58,13 +59,35 @@ private String getPodName() {
...
```
sergiygetlin, line 78: Is it possible that this would be recorded twice in case of Stairway retry?

rtitle: Yes, but Janitor does handle duplicate messages. It records each event in its DB and if there is a duplicate it puts the record in DUPLICATE status and does not try to handle it. In fact I've noticed de-duplication happens fairly often for VM messages.

PR #1521, opened 2023-11-07
[WM-2354] Azure Managed Identity cloning

dillydally414:
https://broadworkbench.atlassian.net/browse/WM-2354

Adds support for cloning of the Azure Managed Identity resource. Also provides a layer of abstraction for Azure cloning, to share some common functionality such as retrieving cloud context, checking if resource already exists in destination workspace, etc... Azure Storage Container cloning was refactored to utilize this.


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/common/CloneControlledAzureResourceFlight.java
```diff
@@ -0,0 +1,186 @@
...
+
+    List<StepRetryRulePair> resourceCloningSteps = new ArrayList<>();
+
+    if (cloningInstructions != CloningInstructions.COPY_NOTHING) {
+      // Get the cloud context and store it in the working map
+      resourceCloningSteps.add(
+          new StepRetryRulePair(
+              new GetAzureCloudContextStep(
+                  destinationWorkspaceId, flightBeanBag.getAzureCloudContextService()),
+              RetryRules.shortDatabase()));
```
dvoet, line 74: other than this cloud context, is there any reason this is azure specific?

dvoet (approved):
I wish there were less boilerplate but that is a problem for another time.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/common/CloneControlledAzureResourceFlight.java
dillydally414: I didn't look much at the other cloning implementations and naively assumed that they differed more across different cloud contexts, though it also makes sense that the cloud context may be the primary differentiator. It also feels like a much larger lift to refactor existing GCP clones, and much higher likelihood I accidentally break something I don't understand.

kpierre13 (approved):
LGTM. I may ask a few more questions on the side, unrelated to this PR in particular.

PR #1524, opened 2023-11-11
[WM-2349] Database clone using shared container

mspector:
This PR is in a draft state with comments to help my team pick up where I left off while I'm on vacation for the week of 11/13.

There are several `TODO` comments (in code, not in GitHub) that leave hints/pointers/suggestions, which I wrote after asking for Doug's advice on what to do next.


service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/CloneControlledAzureDatabaseResourceFlightTest.java
```diff
@@ -0,0 +1,168 @@
...
+        destinationWorkspace, userAccessUtils.defaultUserAuthRequest());
+  }
+
+  @Test
+  void cloneControlledAzureDatabaseFlightTest() throws InterruptedException {
+    FlightMap inputs = new FlightMap();
+
+    // To fill in this value:
+    // Go to your workspace dashboard -> cloud information -> copy the "Storage SAS URL"
+    inputs.put("BLOB_CONTAINER_URL_AUTHENTICATED", "https://...");
```
mspector, line 68: You'll need to fill in this value to make this test work. There's no `WorkspaceFlightMapKeys` entry for this because we shouldn't be passing the SAS token around via the flight map. It should be generated on the fly.

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/CloneControlledAzureDatabaseResourceFlightTest.java
```diff
@@ -0,0 +1,168 @@
...
+
+    // To fill in this value:
+    // Go to your workspace dashboard -> cloud information -> copy the value starting with "sc-"
+    //   from either the "Storage SAS URL" or the "Storage Container URL".
+    // Or: use "sc-${WORKSPACE_ID}" (our conventional format for storage containers names)
+    inputs.put(
+        WorkspaceFlightMapKeys.ControlledResourceKeys.AZURE_STORAGE_CONTAINER_NAME, "sc-...");
```
mspector, line 75: You'll also need to fill in this value until the flight is updated to include storage container creation and deletion (See [WM-2366](https://broadworkbench.atlassian.net/browse/WM-2366)

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/CloneControlledAzureDatabaseResourceFlightTest.java
```diff
@@ -0,0 +1,168 @@
...
+        StairwayTestUtils.blockUntilFlightCompletes(
+            jobService.getStairway(),
+            CloneControlledAzureDatabaseResourceFlight.class,
+            inputs,
+            Duration.ofMinutes(1),
+            null);
+
+    assertEquals(FlightStatus.SUCCESS, result.getFlightStatus());
+
+    ClonedAzureResource resultDatabase =
```
mspector, line 162: I don't have any assertions here because we shouldn't merge this connected test. The workflow cloning project should have only one connected test in the end - a test that clones an entire workspace and demonstrates that a database has been cloned as part of that process.

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/CloneControlledAzureDatabaseResourceFlightTest.java
```diff
@@ -0,0 +1,215 @@
...
+    // To fill in this value:
+    // Go to your workspace dashboard -> cloud information -> copy the "Storage SAS URL"
+    // inputs.put("BLOB_CONTAINER_URL_AUTHENTICATED", "https://...");
+
+    // To fill in this value:
+    // Go to your workspace dashboard -> cloud information -> copy the value starting with "sc-"
+    //   from either the "Storage SAS URL" or the "Storage Container URL".
+    // Or: use "sc-${WORKSPACE_ID}" (our conventional format for storage containers names)
+    // inputs.put(
+    //    WorkspaceFlightMapKeys.ControlledResourceKeys.AZURE_STORAGE_CONTAINER_NAME, "sc-...");
```
cjllanwarne, line 84: Reminder to remove this at some point

dillydally414:
Marking as ready for initial review, I still may add additional tests but the core functionality is all here.

dillydally414:
Test runs:
- https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/6934791437

kpierre13 (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/LaunchCloneControlledAzureDatabaseResourceFlightStep.java
salonishah11, line 1: nit: Can you remove the comment from L68-L69. I had added that in case there were any other parameters that needed to be set for sub-flight and is no longer needed.

salonishah11 (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/CopyControlledAzureDatabaseDefinitionStep.java
```diff
@@ -0,0 +1,223 @@
...
+                        || wsmResource
+                            .getResourceId()
+                            .toString()
+                            .equals(sourceDatabase.getDatabaseOwner()))
+            .findFirst()
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
+                        "Could not find managed identity that owns database %s"
+                            .formatted(sourceDatabase.getResourceId())));
```
dvoet, line 106: why not use [getResourceByName](https://github.com/DataBiosphere/terra-workspace-manager/blob/ad7ebaffadb34f5d87291de438ecaec34e8799c5/service/src/main/java/bio/terra/workspace/db/ResourceDao.java#L567), below too. Shouldn't the destination managed identity WSM resource have the same name as the source?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/DumpAzureDatabaseStep.java
```diff
@@ -0,0 +1,156 @@
...
+    var userRequest =
+        getRequired(
+            inputParameters,
+            JobMapKeys.AUTH_USER_INFO.getKeyName(),
+            AuthenticatedUserRequest.class);
+
+    // TODO WM-2366: replace with storage container created in WM-2371
+    ControlledAzureStorageContainerResource destinationContainer =
+        wsmResourceService
+            .enumerateResources(
```
dvoet, line 78: perhaps the [enumerateResources in ResourceDao](https://github.com/DataBiosphere/terra-workspace-manager/blob/ad7ebaffadb34f5d87291de438ecaec34e8799c5/service/src/main/java/bio/terra/workspace/db/ResourceDao.java#L269) would be better

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/DumpAzureDatabaseStep.java
```diff
@@ -0,0 +1,156 @@
...
+        blobFileName,
+        destinationContainer.getStorageContainerName(),
+        blobContainerUrlAuthenticated);
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    // Nothing to undo in this step (dump file will be cleaned up by storage container deletion)
```
dvoet, line 153: is this still true?

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/CloneControlledAzureDatabaseResourceFlight.java
```diff
@@ -1,12 +1,85 @@
...
+                flightBeanBag.getWsmResourceService()),
+            RetryRules.cloud()),
+        new StepRetryRulePair(
+            new RestoreAzureDatabaseStep(
+                flightBeanBag.getLandingZoneApiDispatch(),
+                flightBeanBag.getSamService(),
+                flightBeanBag.getWorkspaceService(),
+                flightBeanBag.getAzureStorageAccessService(),
+                flightBeanBag.getAzureDatabaseUtilsRunner()),
+            RetryRules.cloud()));
```
dvoet, line 83: do we want to delete the dump file?

dillydally414: From my understanding this will be deleted by followup ticket WM-2372 which will delete the container the dumpfile is in

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/DumpAzureDatabaseStep.java
dillydally414: Not currently, but it will be (hence why we are merging into feature branch).

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/CopyControlledAzureDatabaseDefinitionStep.java
dillydally414: The database ownership seems to refer to either the managed identity resource name or resource id, which is why the custom check is needed for the sourceIdentity. For destination, I suppose we could assume that it will have the same name but this is only necessarily true in workspace clones, and would not hold for eventual individual identity/database clones where names can be specified different from the original, whereas they would still have the correct resource lineage.

From the swagger:
> Resource id or name of an azure managed identity in this workspace to be the owner of the database, required for shared databases. For private databases, the owner will be determined from privateResourceUser.

dvoet: only the name of the resource is stored in the database. Any ids that may be in the database currently are old and can be ignored. The api accepts id for backwards compatibility but translates to name before storing.

PR #1526, opened 2023-11-13
[WM-2236][WM-2237] Add support for cloning Controlled Azure database flight

salonishah11:
Testing:
We were able to test this in a BEE. We manually created a cbas database with cloning instructions set to COPY_RESOURCE in source workspace. Cloning the workspace did call the LaunchCloneDatabase classes as expected in logs:
```
{"timestampSeconds":1699990357,"timestampNanos":473000000,"severity":"INFO","message":"Launching flight class: 
bio.terra.workspace.service.resource.controlled.flight.clone.azure.database.CloneControlledAzureDatabaseResourceFlight stairway: stairwayec9a9a4e- .... }
```
We had also added a println in `CloneControlledAzureDatabaseResourceFlight` sub-flight class to check if it was being called as expected and the println did show up in logs (Note: the println will be removed before merging. It is only for testing purposes):
Screenshot 2023-11-14 at 4 02 41 PM

Closes:
- https://broadworkbench.atlassian.net/browse/WM-2237
- https://broadworkbench.atlassian.net/browse/WM-2236


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneControlledAzureDatabaseResourceFlightStep.java
```diff
@@ -0,0 +1,99 @@
...
+
+      // generate cloneDetails
+      WsmResourceCloneDetails cloneDetails = new WsmResourceCloneDetails();
+      WsmCloneResourceResult cloneResult =
+          WorkspaceCloneUtils.flightStatusToCloneResult(subflightState.getFlightStatus(), resource);
+      cloneDetails.setResult(cloneResult);
+
+      FlightMap resultMap = FlightUtils.getResultMapRequired(subflightState);
+      var clonedDatabase =
+          resultMap.get(JobMapKeys.RESPONSE.getKeyName(), ControlledAzureDatabaseResource.class);
```
dillydally414, line 53: This will return a `ClonedAzureResource`

dillydally414:
I am not sure what is included in this ticket vs. in other/follow-up tickets, but I wanted to mention there were a couple more changes I had to make to get the managed identity cloning functional:
- Adding resource type to FindResourcesToClone - https://github.com/DataBiosphere/terra-workspace-manager/pull/1521/files#diff-456c3140ed43be51fbc37c5b3330eda9926efc43fa49ba287b18e4bcaedd3907R102-R103
- Adding new flight to ActivityFlight - https://github.com/DataBiosphere/terra-workspace-manager/pull/1521/files#diff-aadc81e8c4616a4082fe7b272a189b3d8442720fb314f0792408ee72294c7362R106-R108

salonishah11:
@dillydally414 thanks for pointing that out. I think it makes sense be included in this ticket. I have update the code 👍

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneControlledAzureDatabaseResourceFlightStep.java
```diff
@@ -0,0 +1,102 @@
...
+  private final String subflightId;
+
+  public AwaitCloneControlledAzureDatabaseResourceFlightStep(
+      ControlledAzureDatabaseResource resource, String subflightId) {
+    this.resource = resource;
+    this.subflightId = subflightId;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
```
cjllanwarne, line 41: I don't see any "wait" logic here, where does that happen? Am I missing something?

TOL: Could we possibly split out a "handle results" method to make it clearer which parts of this are "waiting for it to end" and which parts are finding and handling the outcomes?

cjllanwarne (approved):
LGTM, with one small question

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneControlledAzureDatabaseResourceFlightStep.java
salonishah11: I think the AwaitCloneX class has a polling mechanism set up using `context.getStairway().waitForFlight(...)` method which keeps polling for X seconds and then extract results after the polling finishes or when the flight succeeds. Based on the other classes and the implementation I am not really sure how to split it up 🤔

dillydally414 (approved):

salonishah11:
Regarding code smells:
- TODO comment will be addressed in a follow up PR
- Refactoring of the duplicate code will be a larger effort for AwaitX classes and out of scope of this PR
- It seems other tests classes do have public modifier and public methods so I am leaving it as is

PR #1531, opened 2023-11-17
Enable workflow to run on branch for troubleshooting

jladieu:


.github/workflows/verify_consumer_pacts.yaml
```diff
@@ -90,20 +115,23 @@ jobs:
...
-          if [[ "$GITHUB_EVENT_NAME" == "push" || "$GITHUB_EVENT_NAME" == "workflow_dispatch"]]; then
-            GITHUB_REF=${{ github.ref }} # The Git Ref that this workflow runs on
-            GITHUB_SHA=${{ github.sha }} # The Git Sha that this workflow runs on
+          if [[ "$GITHUB_EVENT_NAME" == "push" ]]; then
+            GITHUB_REF=${{ github.ref }}
+            GITHUB_SHA=${{ github.sha }}
           elif [[ "$GITHUB_EVENT_NAME" == "pull_request" ]]; then
             GITHUB_REF=refs/heads/${{ github.head_ref }}
             GITHUB_SHA=${{ github.event.pull_request.head.sha }}
+          elif [[ "$GITHUB_EVENT_NAME" == "workflow_dispatch" ]]; then
```
jladieu, line 124: I got previous feedback to collapse this conditional block with the one for push since they are identical, however I introduced a syntax error when doing that, and put it back to the way it was before to also make it more identical to the upstream PR from which I copied this stuff.  I'm hoping that will make convergence easier, at the expense of a little more duplication for the time being.

jladieu:
Note for reviewers, most of the GHA config code is copied from @ichengchang's PR and purposefully left as identical as possible to make it easier to converge later.

service/src/test/java/bio/terra/workspace/pact/WdsContractVerificationTest.java
```diff
@@ -76,7 +77,7 @@
 })
 @Tag("pact-verification")
 @Provider("workspacemanager") // should match the terra chart name
-@PactBroker()
+@PactBroker(enablePendingPacts = "true", providerTags = "main")
```
jladieu, line 80: Doing this because it seems like we should to keep WSM build from falsely failing: https://docs.pact.io/pact_broker/advanced_topics/pending_pacts

davidangb (commented):
Am I correct in understanding that the new tag.yml is used to get the semver for contract tests, but tag-publish.yml is used when actually bumping and releasing? I confess I don't know the WSM CI workflow very well.

I'm worried about the behavior of tag.yml vs. the actual bumping in tag-publish.yml. For instance, they use different versions of `actions/bumper@bumper` and have different values for `VERSION_LINE_MATCH`.  Is there any way to consolidate this?

I defer to the WSM team on this, since I'm shaky on these workflows.

.github/workflows/tag.yml
```diff
@@ -0,0 +1,61 @@
+name: Tag
+on:
+  workflow_dispatch:
+  workflow_call:
+    inputs:
+      dry-run:
+        description: "Determine the next version without tagging the branch. The workflow can use the outputs new_tag and tag in subsequent steps. Possible values are true and false (default)"
```
cahrens, line 7: Did this workflow get copied from BPM? If so, could you move it to https://github.com/broadinstitute/terra-github-workflows instead of duplicating it?

jladieu:
@davidangb Yes, your understanding is similar to mine which is that the `tag.yml` workflow is all about coming up with the next tag that will be used for semantic versioning during deployment & release, without actually committing anything to that effect.  This allows pacts to register against the version with which they will be released to the wild.

I agree having two implementations of tagging logic isn't ideal, but I am reluctant to touch any of the existing tagging logic for WSM as part of this task as I think the chance of messing something important up is a lot higher and I'm probably not the person who should be doing that.

@blakery @cahrens do you have any thoughts on how best to handle the fact that there are essentially two different ways a semver tag gets generated for WSM?

I think ideally the `tag.yml` which I'm introducing as part of this PR would be owned & managed via the terra-github-workflows repo as @cahrens [suggested](https://github.com/DataBiosphere/terra-workspace-manager/pull/1531#discussion_r1399636802)

.github/workflows/tag.yml
jladieu: Yes I did just copy it over from there.  I agree that terra-github-workflows should be the desired resting place for this GHA logic, but the original author @ichengchang (from whom I copied this) mentioned being on vacation this week and is probably the best person to shepherd (or at least influence) that sort of consolidation.

I see this change as a tactical "fix a broken thing that I broke" + "get this logic some early and additional mileage to inform if any other tweaks are needed before it can be made shareable", which is why it's landing here in pure copypasta format.

jladieu:
FYI Ivan is making more upstream changes in: https://github.com/DataBiosphere/terra-billing-profile-manager/pull/345, which I plan to pull into this PR.

.github/workflows/verify_consumer_pacts.yaml
ichengchang: Good to go on this one. The `push` and `workflow_dispatch` conditions have been collapsed into one condition.

service/src/test/java/bio/terra/workspace/pact/WdsContractVerificationTest.java
ichengchang: Looks good

service/src/test/java/bio/terra/workspace/pact/WdsContractVerificationTest.java
```diff
@@ -103,7 +108,17 @@ public static SelectorBuilder consumerVersionSelectors() {
     // runs, and your consumer is publishing such Pacts under their feature branch name, you can add
     // the following to the SelectorBuilder:
     //   .branch("consumer-feature-branch-name")
-    return new SelectorBuilder().mainBranch().deployedOrReleased();
+    // Updated comments
```
jladieu, line 111: TODO merge the two blocks of comments here

jladieu:
Woohoo: 
image

jladieu:
@ichengchang I think this good -- when you're happy with it, can you mark it as approved and I'll see if I can get one of our workspaces friends to give a thumb as well?

ichengchang (approved):
LGTM

PR #1543, opened 2023-12-01
[WM-2398] sanitize process launcher inputs

mspector:
1. is it ok that I'm calling this just before the call to localProcessLauncher.launchProcess? the alternative is to put the validation inside the LocalProcessLauncher class, but then it'd have to be more general.
2. what kind of checks should I be doing? I can add a check for each expected item in the list of arguments... would that be sufficient?


mspector:
closing following the outcome of this discussion: https://broadinstitute.slack.com/archives/C01JVP0MLLX/p1701458266479929?thread_ts=1701380894.702229&cid=C01JVP0MLLX

PR #1552, opened 2023-12-11
[WM-2392] workflows app / workflow cloning backwards compatibility

mspector:


service/src/main/resources/db/changesets/20231208_clone-workflows-app-resources.yaml
```diff
@@ -0,0 +1,34 @@
...
+    author: mspector
+    changes:
+    - update:
+        tableName: resource
+        columns:
+            - column:
+                name: cloning_instructions
+                value: "COPY_RESOURCE"
+        where:
+          or:
```
salonishah11, line 13: Maybe I am missing something, but what is the purpose of this `or`? 🤔

service/src/main/resources/db/changesets/20231208_clone-workflows-app-resources.yaml
```diff
@@ -0,0 +1,34 @@
...
+              and:
+                - column:
+                    name: resource_type
+                    value: "AZURE_DATABASE"
+                - column:
+                    name: name
+                    value: "cbas"
```
salonishah11, line 20: It seems that this `and` condition would also update the cloning instructions of `cbas` database for `CROMWELL` app type. Should we add additional check to ensure it only does so for `WORKFLOWS_APP` app type?

service/src/main/resources/db/changesets/20231208_clone-workflows-app-resources.yaml
mspector: it's part of the "where" clause - it applies "or" to the two "and" conditions below.

@dvoet just suggested I write this in sql for clarity - this is probably why

service/src/main/resources/db/changesets/20231208_clone-workflows-app-resources.yaml
mspector: I'm not 100% sure if it matters, but I can add that check. better to be safe

salonishah11: I think WSM might start cloning the `cbas` database in `CROMWELL` app if its set to `COPY_RESOURCE` since in WSM I don't think there is a check to see which app it belongs too. And if the user somehow in the workspace spun up a `CROMWELL` app then Leo will attach the cloned `cbas` database new app and I am not sure how things will work after this. So yeah I agree its better to be safe by adding that additional check.

service/src/main/resources/db/changesets/20231208_clone-workflows-app-resources.yaml
mspector: @salonishah11 , I'm sorry, my explanation was wrong - it's a typo, leftover from when I did implement it in the way I described

service/src/main/resources/db/changesets/20231208_clone-workflows-app-resources.yaml
```diff
@@ -0,0 +1,34 @@
+databaseChangeLog:
+- changeSet:
+    id: Set cloneInstructions to COPY_RESOURCE for WORKFLOWS_APP resources.
```
cjllanwarne, line 3: This is updating the CBAS database, not changing the WORKFLOWS_APP directly

mspector: I guess I was thinking... "update resources associated with WORKFLOWS_APP". but you're right, it's more precise to describe them as being associated with the CBAS database

cjllanwarne (approved):
Looks good to me. How have you tested the changes?

dvoet (approved):
so much better as sql, thanks

mspector:
@cjllanwarne , I tested the non-sql version in a BEE as follows:
1. deployed a BEE with Leo version set to a version in which `cbas` and `idworkflows_app` are created with `COPY_NOTHING` instructions
2. created a workspace, launched workflows app, verified that both resources have `COPY_NOTHING`
3. redeploy WSM in my BEE with a version containing this changeset
4. verify both resources have `COPY_RESOURCE`

PR #1567, opened 2023-12-15
[NO TICKET] Fix bumper.

cahrens:
No ticket needed because this does not impact any production code (github workflows only).

This is to fix the problem that dependabot PRs trigger a new version to be created.


ichengchang (approved):
Looks good!

aherbst-broad (approved):

PR #1568, opened 2023-12-15
[WM-2420] cloning across mrgs

mspector:
This PR changes the workspace ID used in the pgDump azure database utils command from the source workspace to the destination workspace.

We need to use the destination workspace ID for all azure database utils pods so that the AKS cluster used matches the MRG that is provided in the clone flight's azure cloud context.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/AzureDatabaseUtilsRunner.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/AzureDatabaseUtilsRunner.java
@@ -375,20 +375,24 @@ public class AzureDatabaseUtilsRunner {
   }
 
   private void runAzureDatabaseUtils(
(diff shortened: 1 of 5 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/AzureDatabaseUtilsRunner.java
```diff
@@ -163,8 +163,8 @@ public void pgDumpDatabase(
 
     runAzureDatabaseUtils(
         azureCloudContext,
-        sourceWorkspaceId,
-        createPodDefinition(sourceWorkspaceId, podName, envVars, secretStringData),
+        destinationWorkspaceId,
+        createPodDefinition(destinationWorkspaceId, podName, envVars, secretStringData),
```
mspector, line 167: the changes in `pgDumpDatabase` are the only meaningful ones. the rest are renamed variables.

PR #1569, opened 2023-12-15
[WM-2420] use source workspace's cloud context in pgdump pod

mspector:
fixes a bug in workflow cloning when trying to clone a workspace across MRGs

this has been tested in a BEE

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/DumpAzureDatabaseStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/DumpAzureDatabaseStep.java
@@ -2,7 +2,6 @@ package bio.terra.workspace.service.resource.controlled.flight.clone.azure.datab
 
 import static bio.terra.workspace.common.utils.FlightUtils.getRequired;
 import static bio.terra.workspace.service.resource.controlled.cloud.azure.AzureUtils.getResourceName;
-import static bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.ControlledResourceKeys.AZURE_CLOUD_CONTEXT;
 
 import bio.terra.common.iam.BearerToken;
 import bio.terra.stairway.FlightContext;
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/DumpAzureDatabaseStep.java
```diff
@@ -111,8 +113,16 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
+        azureCloudContextService
+            .getAzureCloudContext(sourceDatabase.getWorkspaceId())
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
+                        "Cloud context for the workspace containing the source database could not be found."));
+
     this.azureDatabaseUtilsRunner.pgDumpDatabase(
-        workingMap.get(AZURE_CLOUD_CONTEXT, AzureCloudContext.class),
+        sourceAzureContext,
```
cjllanwarne, line 125: Interesting that this wasn't already what was in the working map (should we have source and destination contexts available there?). And is there any reason why we don't also need an equivalent change in the restore step?

cjllanwarne (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/DumpAzureDatabaseStep.java
salonishah11: Same question regarding if this change also needs to be added to the restore step

salonishah11 (approved):
Just one question otherwise LGTM

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/azure/database/DumpAzureDatabaseStep.java
mspector: I didn't check exhaustively, but I think the source context isn't in the working map because this pgDump step is the only step (at the moment) that needs to run azureDatabaseUtils pods in the source workspace context.

We don't need to change anything in the restore step because the azureDatabaseUtils pod that it deploys runs in the destination workspace context already.

cahrens (approved):

mspector:
@dvoet , @cahrens , @aherbst-broad , in the interest of getting this merged in time for release on Tuesday: we're merging with only WM reviews. This feels to us like a small / safe enough change that should only affect the workflow cloning journey anyway.

I'll followup on Monday to see if you'd like us to make any changes that will make it out in future releases

mspector:
thanks @cahrens !

PR #1576, opened 2023-12-19
[WOR-477] Remove unneeded checkout-helm-versions

cahrens:
When `checkout-helm-versions` is executed by a workflow triggered by the dependabot user, it fails.

image

Since this action seems to be no longer needed (it was introduced in 2021), deleting it.


.github/workflows/full-integration-tests.yml
```diff
@@ -28,14 +28,6 @@ jobs:
...
 
-    # Sparse check out terra-helmfile repo version directories
-    - uses: ./.github/actions/checkout-helm-versions
-      with:
-        repository: broadinstitute/terra-helmfile
-        ref: master
-        token: ${{ secrets.BROADBOT_TOKEN }}
-        path: integration/terra-helmfile
-
     - name: Initialize Postgres DB for local server test run
```
cahrens, line 31: Successful run of `full-integration-tests` without this: https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/7265787352/job/19796247576

.github/workflows/nightly-tests.yml
```diff
@@ -22,14 +22,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      # Sparse check out terra-helmfile repo version directories
```
cahrens, line 25: Nightly tests haven't been since "last year": https://github.com/DataBiosphere/terra-workspace-manager/actions/workflows/nightly-tests.yml

.github/workflows/pr-integration.yml
```diff
@@ -42,14 +42,6 @@ jobs:
...
     - uses: actions/checkout@v3
 
-    # Sparse check out terra-helmfile repo version directories
-    - uses: ./.github/actions/checkout-helm-versions
-      with:
-        repository: broadinstitute/terra-helmfile
-        ref: master
-        token: ${{ secrets.BROADBOT_TOKEN }}
-        path: integration/terra-helmfile
-
```
cahrens, line 52: Successful run can be seen on this PR.

.github/workflows/full-integration-tests.yml
sergiygetlin: What was this action used for?

sergiygetlin (approved):

.github/workflows/full-integration-tests.yml
cahrens: I have no idea. It was added 2 years ago by Ivan (who is on vacation). I thought that deleting it would point out to me what it was needed for… but then everything passed.

Most likely the need for it went away with the various changes to how our services are configured/built/deployed. If deleting it causes some problem in the future, we can do more investigation.

marctalbott (approved):

PR #1577, opened 2023-12-20
WM-2418: Wait for database table to exist before reassigning

cjllanwarne:
Makes sure the pgrestore command finishes before we try to update table ownership.

Otherwise in the worst case (for very small dumps being restored), we can end up running the table reassignment before the table creation command has run.

### DIFF
--- a/azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
+++ b/azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
@@ -272,7 +272,8 @@ public class DatabaseService {
       }
     }
     localProcessLauncher.getOutputStream().flush();
-    Thread.sleep(5);
+    localProcessLauncher.getOutputStream().close();
+    localProcessLauncher.waitForTerminate();
 
     databaseDao.reassignOwner(adminUser, dbName);
   }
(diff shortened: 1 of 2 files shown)


kpierre13 (approved):

salonishah11 (approved):

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
```diff
@@ -272,6 +273,8 @@ public void pgRestore(
       }
     }
     localProcessLauncher.getOutputStream().flush();
+    localProcessLauncher.getOutputStream().close();
+    localProcessLauncher.waitForTerminate();
```
aherbst-broad, line 277: Is there any kind of timeout that can be added here? i.e., how long are we willing to wait for this process to die before giving up?

aherbst-broad (approved):
lgtm, had a comment but not a blocker.

azureDatabaseUtils/src/main/java/bio/terra/workspace/azureDatabaseUtils/database/DatabaseService.java
cjllanwarne: Not here but I believe there is an overall pod time limit asserted by the wrapping task

PR #1589, opened 2024-01-04
Do not fail out azure k8s namespace deletion when managed identity resource is not present

aherbst-broad:
[Related ticket](https://broadworkbench.atlassian.net/browse/WOR-1418)

Clients can delete the managed identity an Azure kubernetes namespace uses out of band. This causes the k8s namespace deletion flight to fail as it expects the managed identity resource to be present.

This is an attempt at making the namespace deletion flight resilient to these situations.

This PR:

* Modifies the logic in the GetWorkspaceManagedIdentityStep to allow the identity resource to be missing.
* Refactors the azure logic to a separate helper to ease testability.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/kubernetesNamespace/ControlledAzureKubernetesNamespaceResource.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/kubernetesNamespace/ControlledAzureKubernetesNamespaceResource.java
@@ -21,6 +21,7 @@ import bio.terra.workspace.service.resource.controlled.cloud.azure.managedIdenti
 import bio.terra.workspace.service.resource.controlled.cloud.azure.managedIdentity.GetFederatedIdentityStep;
 import bio.terra.workspace.service.resource.controlled.cloud.azure.managedIdentity.GetPetManagedIdentityStep;
 import bio.terra.workspace.service.resource.controlled.cloud.azure.managedIdentity.GetWorkspaceManagedIdentityStep;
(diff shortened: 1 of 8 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/kubernetesNamespace/ControlledAzureKubernetesNamespaceResource.java
```diff
@@ -213,14 +215,17 @@ private List<Step> getCreateFederatedCredentialsSteps(FlightBeanBag flightBeanBa
...
     return switch (getAccessScope()) {
       case ACCESS_SCOPE_SHARED -> new GetWorkspaceManagedIdentityStep(
-          flightBeanBag.getAzureConfig(),
-          flightBeanBag.getCrlService(),
           getWorkspaceId(),
-          flightBeanBag.getResourceDao(),
-          getManagedIdentity());
+          getManagedIdentity(),
+          missingIdentityBehavior,
+          new ManagedIdentityHelper(
```
dvoet, line 225: should this be a spring managed bean?

dvoet (approved):

blakery (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/kubernetesNamespace/ControlledAzureKubernetesNamespaceResource.java
aherbst-broad: I'm going to leave this as-is for simplicity; if this gets to more than a single callsite, it'd be a good idea.

PR #1602, opened 2024-01-12
[NO TICKET] Try grouping dependabot updates.

cahrens:
Let's try out the grouping feature. It seems like it will be particularly useful in this repo given that dependabot PRs currently require some hand-holding (as dependabot user doesn't have access to all the necessary secrets).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,19 @@ updates:
   # Enable version updates for Gradle
   - package-ecosystem: "gradle"
     directory: "/"
-    open-pull-requests-limit: 10
+    open-pull-requests-limit: 5
+    groups:
+      npm-major-dependencies:
+        patterns:
+          - "*"
+        update-types:
+          - "major"
+      npm-minor-patch-dependencies:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


okotsopoulos (approved):

.github/dependabot.yml
```diff
@@ -3,7 +3,19 @@ updates:
...
   - package-ecosystem: "gradle"
     directory: "/"
-    open-pull-requests-limit: 10
+    open-pull-requests-limit: 5
+    groups:
+      npm-major-dependencies:
+        patterns:
+          - "*"
+        update-types:
+          - "major"
```
marctalbott, line 12: I would think we wouldn't want to group all major dependency updates together as those PRs would be potentially difficult/risky to merge

.github/dependabot.yml
```diff
@@ -3,7 +3,19 @@ updates:
   # Enable version updates for Gradle
   - package-ecosystem: "gradle"
     directory: "/"
-    open-pull-requests-limit: 10
+    open-pull-requests-limit: 5
+    groups:
+      npm-major-dependencies:
```
marctalbott, line 8: nit: remove `npm` since this is a Gradle project. I was confused at first about whether this would only look for npm updates before realizing it was just the group name

cahrens: Oops! I will update.

PR #1605, opened 2024-01-12
[NO TICKET] Feedback from grouping dependabot updates.

cahrens:
Follow-up to #1602


marctalbott (approved):

okotsopoulos (approved):
Good comments from @marctalbott -- on TDR UI's side, we may end up ungrouping major upgrades too.

PR #1607, opened 2024-01-12
[WOR-1439] Micrometer: publish percentile histograms for per-endpoint latencies

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1439

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -72,6 +72,12 @@ management:
     web:
       exposure:
         include: "prometheus"
+  metrics:
+    distribution:
+      # Used to publish a histogram suitable for computing aggregable (across dimensions) percentile
+      # latency approximations in Prometheus (by using histogram_quantile)
+      # For more information: https://micrometer.io/docs/concepts#_histograms_and_percentiles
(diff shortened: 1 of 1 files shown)


service/src/main/resources/application.yml
```diff
@@ -72,6 +72,12 @@ management:
     web:
       exposure:
         include: "prometheus"
+  metrics:
+    distribution:
+      # Used to publish a histogram suitable for computing aggregable (across dimensions) percentile
+      # latency approximations in Prometheus (by using histogram_quantile)
+      # For more information: https://micrometer.io/docs/concepts#_histograms_and_percentiles
+      percentiles-histogram[http.server.requests]: true
```
okotsopoulos, line 81: Some extra info on what's possible here:

We can narrow down the metrics we collect by setting lower and/or upper bounds on the default buckets collected or by picking our own buckets based on established SLAs.  But Jack tells me that these metrics are cheap, and I see that in dire times average workspace resolution latency has taken minutes:

Screenshot 2024-01-12 at 5 53 10 PM

So I'm sticking with the defaults, and we can tune collection later if desired.

aherbst-broad (approved):

PR #1614, opened 2024-01-16
[NO TICKET] Add ticket number and prevent problematic updates (dependabot)

cahrens:
Follow-up to https://github.com/DataBiosphere/terra-workspace-manager/pull/1611 to exclude problematic upgrades (we have tickets for them), include the long-lived JIRA ticket on dependabot PRs, and make the total number of allowable PRs consistent with other repos.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,7 +27,7 @@ updates:
       prefix: "[WOR-1448]"
     ignore:
       - dependency-name: "org.springframework.boot:spring-boot-gradle-plugin"
-        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
+        update-types: [ "version-update:semver-major", "version-update:semver-minor", "version-update:semver-patch"]
       - dependency-name: "com.google.apis:google-api-services-iam"
         update-types: [ "version-update:semver-patch", "version-update:semver-minor" ]
       - dependency-name: "com.google.apis:google-api-services-notebooks"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,11 +3,13 @@ updates:
   # Enable version updates for Gradle
   - package-ecosystem: "gradle"
     directory: "/"
-    open-pull-requests-limit: 5
+    open-pull-requests-limit: 10
     groups:
       minor-patch-dependencies:
         patterns:
           - "*"
+        exclude-patterns:
+          - "spotless-plugin-gradle" # likely to require reformatting of code
         update-types:
           - "minor"
           - "patch"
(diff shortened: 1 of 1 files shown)


aherbst-broad (approved):

cahrens:
The patch version upgrade of spring-boot from 3.1.2-> 3.1.7 is [failing in LZS](https://github.com/DataBiosphere/terra-landing-zone-service/pull/389/files). Ignore patch also until this issue can be investigated (which is ticked by https://broadworkbench.atlassian.net/browse/WOR-1449) .

aherbst-broad (approved):

blakery (approved):

.github/dependabot.yml
```diff
@@ -22,4 +24,11 @@ updates:
...
     commit-message:
-      prefix: "[No Ticket]"
+      prefix: "[WOR-1448]"
+    ignore:
+      - dependency-name: "org.springframework.boot:spring-boot-gradle-plugin"
+        update-types: [ "version-update:semver-major", "version-update:semver-minor", "version-update:semver-patch"]
+      - dependency-name: "com.google.apis:google-api-services-iam"
+        update-types: [ "version-update:semver-patch", "version-update:semver-minor" ]
+      - dependency-name: "com.google.apis:google-api-services-notebooks"
+        update-types: [ "version-update:semver-patch", "version-update:semver-minor" ]
```
okotsopoulos, line 34: I'm sorry, I'm still not understanding why we want to let Dependabot create major version updates for any of these deps.  Please feel free to Slack me if it's easier to hash out there.

.github/dependabot.yml
```diff
@@ -22,4 +24,11 @@ updates:
       - "dependency"
       - "gradle"
     commit-message:
-      prefix: "[No Ticket]"
+      prefix: "[WOR-1448]"
+    ignore:
+      - dependency-name: "org.springframework.boot:spring-boot-gradle-plugin"
+        update-types: [ "version-update:semver-major", "version-update:semver-minor", "version-update:semver-patch"]
```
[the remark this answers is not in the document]
cahrens: You are saying to remove `update-types` entirely? If that means it ignores all updates, we could certainly do that. I don't feel strongly about it though… I would hope we can remove the ignoring of patch and minor versions fairly easily, but major may continue to be prolematic.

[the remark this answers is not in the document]
okotsopoulos: Correct, the `update-types` line is not required.  If unspecified, all updates will be ignored, we do this already in TDR UI.  But same here, I don't feel that strongly about it.

.github/dependabot.yml
cahrens: The way the versioning works with these repos, I suspect that Dependabot would never realize the version was a major version update. The [specific case](https://github.com/DataBiosphere/terra-workspace-manager/pull/1611/commits/d5ee76a6357fe01ced60dc935f3fffd5f2a8bd2a) I added these for was lumped with the "minor release upgrades", but it was from v1->v2 with breaking package name changes.

cahrens: So yeah, I can see ignoring major version updates too for completeness, but I thought that these ignores would solve the specific pain I was feeling.

cahrens: Ticket for further investigation: https://broadworkbench.atlassian.net/browse/WOR-1446

cahrens: @okotsopoulos I simplified the configuration to just ignore all updates for those 3 libraries. Thanks for the suggestion!

PR #1619, opened 2024-01-23
[WOR-1426] Shift to Repo secrets for integration tests

aherbst-broad:
## Context
The [write-config.sh](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/scripts/write-config.sh) script is liberally used throughout WSM CI. This means that GitHub needs to pull secrets from our Vault instance to run ITs, etc. We want to move away from this and leverage GH repo secrets instead (wired up via [Terraform](https://github.com/broadinstitute/terraform-ap-deployments/pull/1338).

The original plan was to shift to pure workload identity federation via the google/auth github action. This proved unworkable for integration tests as there is a significant amount of machinery that relies on SA keys being placed in the `/config` directory with well-known names. 

As initially mentioned, I've instead pursued a middle path of wiring the secrets into the repo, pulling them in the GitHub actions and rendering the keys at the needed well-known paths.

### Still TODO in a separate PR
* Ticketed [here](https://broadworkbench.atlassian.net/browse/WOR-1463)
   * Wire up release promotion 
   * Wire up managed application publisher credentials and the consuming test action


.github/actions/start-local-server/action.yml
```diff
@@ -27,7 +27,7 @@ runs:
         # Give it some time before starting to poll.
         sleep 10
         started=1
-        for i in `seq 1 50`;
+        for i in `seq 1 120`;
```
aherbst-broad, line 30: Poking through various workflow runs, this consistently times out but we never see it because there is no exit() call. I've added one here and bumped this to a more generous timeout. I am unsure, but I would not be surprised if this contributed to some test instability in the past.

.github/actions/write-credentials/action.yml
```diff
@@ -0,0 +1,61 @@
...
+runs:
+  using: "composite"
+  steps:
+
+    - name: Write credentials
+      id: 'setup-user-delegated-creds'
+      run: |
+        mkdir -p config
+        USER_DELEGATED_SA_B64=${{ inputs.user-delegated-sa-b64 }}
+        echo ::add-mask::$USER_DELEGATED_SA_B64
```
aherbst-broad, line 32: Probably being paranoid here but I want to make sure these secrets (either base64 or decoded). are never erroneously logged either here or in downstream actions.

.github/workflows/nightly-tests.yml
```diff
@@ -1,208 +0,0 @@
-# This workflow runs the WSM full regression tests, including perf, integration, and
```
aherbst-broad, line 1: This workflow hasn't run in 2 years.

.github/actions/write-credentials/action.yml
```diff
@@ -0,0 +1,79 @@
...
+        POLICY_CLIENT_SA=$(echo $POLICY_CLIENT_SA_B64 | base64 --decode)
+        echo ::add-mask::$POLICY_CLIENT_SA
+        echo $POLICY_CLIENT_SA > config/policy-client-sa.json
+
+        WSM_SA_B64=${{ inputs.wsm-sa-b64 }}
+        echo ::add-mask::$WSM_SA_B64
+        WSM_SA=$(echo $WSM_SA_B64 | base64 --decode)
+        echo ::add-mask::$WSM_SA
+        echo $WSM_SA > config/wsm-sa.json
+      shell: bash
```
aherbst-broad, line 79: Noted in the readme but: this will need to thread in managed app credentials as well to let us fully switch over. I've cut that off for this PR and will ticket up the rest.

okotsopoulos (approved):
Looks good -- do we want to consider the follow-on ticket for managed app publisher credentials for our next sprint?

blakery (approved):

aherbst-broad:
@okotsopoulos yes indeed, I've pulled it up.

PR #1629, opened 2024-02-01
[IA-4632] Add disk create-result

jdcanas:
Hey all, not super familiar with WSM.

I added this code then ran then ran all gradle tasks to `generateSwagger[xxx]`, and they all succeeded and I see all the parallel constructs to what appears for other `create-result` endpoints in the generated code directories. I figure its better to post here to see if I'm missing anything obvious while attempting to verify with a locally running WSM in parallel (which is proving to be a bit of a learning curve and something I think would go smoother to pair on).

Seems a bit fishy this was so straight forward, but the resource itself already exists. There's nothing I have to do to make WSM actually track the result, does it do this already for controlled resources and just wasn't exposing it?

### DIFF
--- a/openapi/src/parts/controlled_azure_disk.yaml
+++ b/openapi/src/parts/controlled_azure_disk.yaml
@@ -145,8 +145,8 @@ components:
           $ref: '#/components/schemas/ControlledResourceCommonFields'
         azureDisk:
           $ref: '#/components/schemas/AzureDiskCreationParameters'
-        jobReport:
-          $ref: '#/components/schemas/JobReport'
+        jobControl:
+          $ref: '#/components/schemas/JobControl'
 
     CreatedControlledAzureDiskResult:
       description: Api result class for creating an Azure disk
--- a/openapi/src/parts/controlled_azure_disk.yaml
+++ b/openapi/src/parts/controlled_azure_disk.yaml
@@ -24,6 +24,24 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/workspaces/v1/{workspaceId}/resources/controlled/azure/disks/create-result/{jobId}:
+    parameters:
+      - $ref: '#/components/parameters/WorkspaceId'
+      - $ref: '#/components/parameters/JobId'
+    get:
+      summary: Retrieve information about an Azure Disk create job.
+      operationId: getCreateAzureDiskResult
+      tags: [ ControlledAzureResource ]
+      responses:
+        '200':
+          $ref: '#/components/responses/CreateControlledAzureDiskResponse'
+        '202':
+          $ref: '#/components/responses/CreateControlledAzureDiskResponse'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '500':
+          $ref: '#/components/responses/ServerError'
+
   /api/workspaces/v1/{workspaceId}/resources/controlled/azure/disks/{resourceId}:
     parameters:
       - $ref: '#/components/parameters/WorkspaceId'
@@ -127,3 +145,24 @@ components:
           $ref: '#/components/schemas/ControlledResourceCommonFields'
         azureDisk:
           $ref: '#/components/schemas/AzureDiskCreationParameters'
+
+    CreatedControlledAzureDiskResult:
+      description: Api result class for creating an Azure disk
+      type: object
+      properties:
+        azureDisk:
+          $ref: '#/components/schemas/AzureDiskResource'
+        jobReport:
+          $ref: '#/components/schemas/JobReport'
+        errorReport:
+          $ref: '#/components/schemas/ErrorReport'
+
+
+  responses:
+    CreateControlledAzureDiskResponse:
+      description: Response to create controlled Azure disk
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/CreatedControlledAzureDiskResult'
+
--- a/openapi/src/parts/controlled_azure_disk.yaml
+++ b/openapi/src/parts/controlled_azure_disk.yaml
@@ -145,6 +145,8 @@ components:
           $ref: '#/components/schemas/ControlledResourceCommonFields'
         azureDisk:
           $ref: '#/components/schemas/AzureDiskCreationParameters'
+        jobReport:
+          $ref: '#/components/schemas/JobReport'
 
     CreatedControlledAzureDiskResult:
       description: Api result class for creating an Azure disk
--- a/openapi/src/parts/controlled_azure_disk.yaml
+++ b/openapi/src/parts/controlled_azure_disk.yaml
@@ -18,13 +18,38 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/CreatedControlledAzureDiskResult'
+                $ref: '#/components/schemas/CreatedControlledAzureDisk'
         '403':
           $ref: '#/components/responses/PermissionDenied'
         '500':
           $ref: '#/components/responses/ServerError'
 
-  /api/workspaces/v1/{workspaceId}/resources/controlled/azure/disks/create-result/{jobId}:
+  /api/workspaces/v2/{workspaceId}/resources/controlled/azure/disks:
+    parameters:
+      - $ref: '#/components/parameters/WorkspaceId'
+    post:
+      summary: Create a new controlled Azure Disk
+      operationId: createAzureDiskV2
+      tags: [ ControlledAzureResource ]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateControlledAzureDiskRequestV2Body'
+      responses:
+        '200':
+          description: Response to create controlled Azure disk
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreateControlledAzureResourceResult'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '500':
+          $ref: '#/components/responses/ServerError'
+
+  /api/workspaces/v2/{workspaceId}/resources/controlled/azure/disks/create-result/{jobId}:
     parameters:
       - $ref: '#/components/parameters/WorkspaceId'
       - $ref: '#/components/parameters/JobId'
@@ -34,9 +59,9 @@ paths:
       tags: [ ControlledAzureResource ]
       responses:
         '200':
-          $ref: '#/components/responses/CreateControlledAzureDiskResponse'
+          $ref: '#/components/responses/CreateControlledAzureResourceResponse'
         '202':
-          $ref: '#/components/responses/CreateControlledAzureDiskResponse'
+          $ref: '#/components/responses/CreateControlledAzureResourceResponse'
         '403':
           $ref: '#/components/responses/PermissionDenied'
         '500':
@@ -126,7 +151,27 @@ components:
             A valid size in GB integer representation
           type: integer
 
+    CreatedControlledAzureDisk:
+      description: Response payload for requesting a new Azure disk
+      type: object
+      properties:
+        resourceId:
+          description: UUID of a newly-created resource.
+          type: string
+          format: uuid
+        azureDisk:
+          $ref: '#/components/schemas/AzureDiskResource'
+
     CreateControlledAzureDiskRequestBody:
+      description: Payload for requesting a new controlled Azure Disk resource.
+      type: object
+      properties:
+        common:
+          $ref: '#/components/schemas/ControlledResourceCommonFields'
+        azureDisk:
+          $ref: '#/components/schemas/AzureDiskCreationParameters'
+
+    CreateControlledAzureDiskRequestV2Body:
       description: Payload for requesting a new controlled Azure Disk resource.
       type: object
       properties:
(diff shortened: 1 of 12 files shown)


openapi/src/parts/controlled_azure_disk.yaml
```diff
@@ -24,6 +24,24 @@ paths:
...
+      - $ref: '#/components/parameters/JobId'
+    get:
+      summary: Retrieve information about an Azure Disk create job.
+      operationId: getCreateAzureDiskResult
+      tags: [ ControlledAzureResource ]
+      responses:
+        '200':
+          $ref: '#/components/responses/CreateControlledAzureDiskResponse'
+        '202':
+          $ref: '#/components/responses/CreateControlledAzureDiskResponse'
```
cpate4, line 39:
```suggestion
        '200':
          $ref: '#/components/responses/CreateControlledAzureResourceResponse'
        '202':
          $ref: '#/components/responses/CreateControlledAzureResourceResponse'
```
I am curious if we want to follow [this pattern](https://github.com/DataBiosphere/terra-workspace-manager/blob/67e624e81cadc5a975c04aab7661f1047a6d5d82/openapi/src/parts/controlled_azure_disk.yaml#L101-L104) re: naming the response.

Thoughts? (cc: @aherbst-broad)

cpate4 (commented):
Curiosity question ... hoping the workspaces team can way in on this

LizBaldo (commented):
That looks very straightforward indeed, and so the autogenerated code does not get committed to github? 
Does this work on your BEE?

jdcanas:
Hmm.. not sure how to push these changes to a bee, it seems artifacts don't get generated by branch name? 
image

openapi/src/parts/controlled_azure_disk.yaml
aherbst-broad: @cpate4 yes, we haven't enforced this but following the defacto pattern would be good.

aherbst-broad:
@jdcanas the autogenerated controller will create a stub of `getCreateAzureDiskResult` on the `ControlledAzureResourceApi` interface. I think you'll want to override that in `ControlledAzureResourceApiController` to actually get the result.

openapi/src/parts/controlled_azure_disk.yaml
jdcanas: Hmm... they do differ slightly in the sense that they echo back the resource, with some additional fields that aren't on the request. Namely `AzureDiskCreationParameters` is transformed into the WSM-internal `AzureDiskResource` in the response.

I can see an argument that this echoing doesn't provide much value, but I would like this to be a separate ticket since it involves non-trivial testing on resources this PR doesn't touch and is technically a breaking API change insofar that it removes fields from a response body.

jdcanas:
Was able to test this in a bee

openapi/src/parts/controlled_azure_disk.yaml
jdcanas: I did it for disk creation, but not any other resource creation

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -135,7 +106,7 @@ public ControlledAzureResourceApiController(
 
   @WithSpan
   @Override
-  public ResponseEntity<ApiCreatedControlledAzureDisk> createAzureDisk(
```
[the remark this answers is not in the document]
jdcanas: Ah yeah, good point. I made the request body and a new v2 endpoint

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -160,18 +131,30 @@ public ResponseEntity<ApiCreatedControlledAzureDisk> createAzureDisk(
             .size(body.getAzureDisk().getSize())
             .build();
 
-    // TODO: make createDisk call async once we have things working e2e
-    final ControlledAzureDiskResource createdDisk =
-        controlledResourceService
-            .createControlledResourceSync(
-                resource, commonFields.getIamRole(), userRequest, body.getAzureDisk())
-            .castByEnum(WsmResourceType.CONTROLLED_AZURE_DISK);
+    final String jobId =
```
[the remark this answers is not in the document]
jdcanas: investigating this still, want to test manually a bit more. Thanks for the examples, would have done this to begin with but wasn't sure if this level of code was something we unit tested.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/ControlledAzureDiskResource.java
```diff
@@ -170,6 +170,9 @@ public void validate() {
       throw new MissingRequiredFieldException(
           "Missing required region field for ControlledAzureDisk.");
     }
+    if (!(getSize() > 0)) {
```
jdcanas, line 173: Not really part of this PR, but found it was a bit odd we were allowing 0 when I was digging through the API's validation for testing

openapi/src/parts/controlled_azure_disk.yaml
```diff
@@ -24,6 +24,49 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/workspaces/v2/{workspaceId}/resources/controlled/azure/disks:
```
lucymcnatt, line 27: thanks for adding this as a v2 async request!

openapi/src/parts/controlled_azure_disk.yaml
```diff
@@ -24,6 +24,49 @@ paths:
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateControlledAzureDiskRequestV2Body'
+      responses:
+        '200':
+          description: Response to create controlled Azure disk
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreateControlledAzureResourceResult'
```
lucymcnatt, line 46: Question here, currently the createDisk request to WSM returns the wsmResourceId in the response, I might be missing something, but I don't see that supplied in the v2 create or create-result

lucymcnatt: We can get that ID from getting the VM information, but that doesn't work for apps...

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/ControlledAzureDiskResource.java
aherbst-broad: Nit, this may be a bit simpler: 

```suggestion
    if (getSize() <= 0) {
```

openapi/src/parts/controlled_azure_disk.yaml
aherbst-broad: There was this [discussion](https://github.com/DataBiosphere/terra-workspace-manager/pull/1629/files/67e624e81cadc5a975c04aab7661f1047a6d5d82#diff-663a06eccffd61fc3303d37e8aa796c277376804a18f2cfa765d78db683cd734)

@jdcanas maybe the resource is needed after all? I am not sure about leo's use, but if it is I suggest mimic-ing the rest of the resource result endpoints and leaving the refactoring of the responses to exclude resources to a later effort as there are likely other consumer use cases we need to consider further before shifting to this new pattern.

aherbst-broad (commented):
Looks good, there is an open question about the structure of the result and whether it needs to include the resource on success.

openapi/src/parts/controlled_azure_disk.yaml
lucymcnatt: The thing is, we don't need the whole resource in most cases in Leo, so could we just add the resource ID to the `CreateControlledAzureResourceResult`? Then when we do need more info, we can `get` it.

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -15,36 +15,7 @@
...
-import bio.terra.workspace.generated.model.ApiCreatedControlledAzureKubernetesNamespaceResult;
-import bio.terra.workspace.generated.model.ApiCreatedControlledAzureManagedIdentity;
-import bio.terra.workspace.generated.model.ApiCreatedControlledAzureStorageContainer;
-import bio.terra.workspace.generated.model.ApiCreatedControlledAzureVmResult;
-import bio.terra.workspace.generated.model.ApiDeleteControlledAzureResourceRequest;
-import bio.terra.workspace.generated.model.ApiDeleteControlledAzureResourceResult;
-import bio.terra.workspace.generated.model.ApiJobControl;
-import bio.terra.workspace.generated.model.ApiJobReport;
-import bio.terra.workspace.generated.model.ApiResourceQuota;
+import bio.terra.workspace.generated.model.*;
```
cpate4, line 18: do you need to make sure your code formatter is setup with the google format options? this change feels odd to me given the amount of models that were previously explicitly listed (which may have been the code's standard)

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/ControlledAzureDiskResource.java
cpate4: do we want to add a test (couldn't find one - but I may have missed it) to validate this is doing what we think?

openapi/src/parts/controlled_azure_disk.yaml
jdcanas: Ah, good catch Lucy, missed this requirement for downstream consumption by Leo. Should be able to add `resourceId` to `CreateControlledAzureResourceResult` with no problems as I see it and accomplish both goals of it being generic and satisfying Leo's use case. Will add this and update tests to verify the field's presence.

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
jdcanas: This was done by the pre-configured gradle formatter not my IDE, the same gradle formatter is used as a check on PRs for consistency. 

Should be good in this regard since the same config for ensuring PR consistency is how I formatted it, not just an IDE garble.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/ControlledAzureDiskResource.java
jdcanas: There are a few tests that validate the happy path here by creating valid disk resource that failed and needed updating from 0 -> 1 when this constraint was added. 

It seems the general practice for validation testing is that more complex validation that is contained in `Util` files (i.e. `AzureResourceValidationUtils` has a complex `validate` method for azure vms) is unit tested, but basic assertions are included in the constructor for these controlled resources (i.e. this  method in `ControlledAzureDiskResource`) and we don't test the unhappy path for these. Could have this assumption wrong, what is your recommendation @aherbst-broad ?

service/src/test/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiControllerAzureDiskTest.java
```diff
@@ -0,0 +1,96 @@
...
+    mockMvc
+        .perform(
+            addJsonContentType(
+                addAuth(
+                    post(String.format(CREATE_CONTROLLED_AZURE_DISK_V2_PATH_FORMAT, workspaceId))
+                        .content(objectMapper.writeValueAsString(diskRequestV2Body)),
+                    USER_REQUEST)))
+        .andExpect(status().is(HttpStatus.SC_OK))
+        .andExpect(MockMvcResultMatchers.jsonPath("$.jobReport").exists())
+        .andExpect(MockMvcResultMatchers.jsonPath("$.resourceId").exists());
```
jdcanas, line 94: cc @lucymcnatt , added test coverage for this too so future updates on `CreateControlledAzureResource` won't remove the leo requirement of resourceId

lucymcnatt: nice thanks!

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/ControlledAzureDiskResource.java
aherbst-broad: We have not been diligent in getting unit tests around the build+validate code for our azure resources. A unit test would be good here. I imagine you could grab an instance of a disk from the `ControlledAzureResourceFixtures` class and make assertions to get the needed coverage, i.e,  `ControlledAzureResourceFixtures.getAzureDisk("foo", "bar", 0)`, etc.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/AzureDiskValidationTest.java
```diff
@@ -0,0 +1,20 @@
...
+public class AzureDiskValidationTest extends BaseAzureUnitTest {
+
+  @Test
+  void validateInvalidDiskSize() {
+    assertThrows(
+        InconsistentFieldsException.class,
+        () ->
+            ControlledAzureResourceFixtures.getAzureDisk(
+                "test", ControlledAzureResourceFixtures.DEFAULT_AZURE_RESOURCE_REGION, 0));
+  }
```
cpate4, line 19: Thank you!

cpate4 (approved):

PR #1626, opened 2024-01-29
[WOR-1454] Populate MDC for Stairway flight and its steps

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1454

**Problem**

We've had a hard time debugging the dismal failures of our services' Stairway flights when [alerts](https://console.cloud.google.com/monitoring/alerting/incidents/0.n8b4q18pyjgu?channelType=slack&project=broad-dsde-staging) are emitted to #dsp-workspaces-alerts.  We'd resort to searching across fields for the flight ID, roughly around the time of the failure, which is a fuzzy process liable to miss relevant logs.

Previously, we tried to propagate the calling thread's MDC to each Stairway step.  At present, this only contains the `requestId` set via TCL's [RequestIdFilter](https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/logging/RequestIdFilter.java).

But this left a few gaps in context:
- When starting the step, we logged before setting the MDC: we want to set the MDC before logging for it to be useful.
- Stairway flight execution outside of the steps themselves, notably when handling dismal failures.

**Changes**

MdcHooks now does the following:
- Propagates flight-specific mapped diagnostic context (MDC) across a Stairway flight's thread(s) for the duration of the flight. This includes context provided by the caller in the input map, intended to store the MDC from the calling thread to preserve context across the duration of a request.
- Propagates step-specific MDC across a Stairway flight's thread(s) for the duration of each step.

Updated unit tests.

**Concrete example**

My [Bee](https://beehive.dsp-devops.broadinstitute.org/environments/okotsopo/chart-releases/workspacemanager) is up to date with these changes.

I modified the workspace creation flight to fail dismally.  Here are the [logs for the duration of such a request](https://cloudlogging.app.goo.gl/kvvx7Krn6hLSrzx67).

I've exposed some of the new MDC additions in the log summary lines to help visualize the lifespan of each piece of context:
- `jsonPayload.requestId` - available from the calling thread's MDC via TCL's [RequestIdFilter](https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/logging/RequestIdFilter.java)
- `jsonPayload.flightId` - available on the Stairway thread(s) for the duration of the flight
- `jsonPayload.flightStepClass` - available on the Stairway thread(s) for the duration of the step
- `jsonPayload.flightStepDirection` - available on the Stairway thread(s) for the duration of the step's execution in the specified direction

[...]

PR #1636, opened 2024-02-09
[WOR-1509] Stairway subflight logs should clear step context from parent flight

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1509

**Problem**

Some of the flights in WSM kick off subflights.  The initial logs emitted by subflights before the first step errantly included step-specific context from the calling flight, interspersed with flight-specific context from the new subflight.

**Fix**

This PR intentionally clears the step-specific context in `MdcHook` when starting a flight.  (One line of code)

**Change in Unit Testing Approach**

When I first updated `MdcHook`, I observed that the existing unit testing approach in `MdcHookTest` involved constructing, submitting, and polling Stairway flights to completion.  But in this approach, much of what is being tested is Stairway code rather than that of our `MdcHook`.  We were also unable to test the behavior of `MdcHook.startHook` and `MdcHook.endHook`, and I wanted to make sure that this bugfix was reflected in unit tests.

I followed through on the testing strategy I proposed in this [PR comment](https://github.com/DataBiosphere/terra-workspace-manager/pull/1626#discussion_r1476240262):

- `MdcHookTest` rewritten as a Unit test with strict stub mocking and targeted component class loading, using a dummy `TestFlightContext` to cover all of our state transitions (flights starting and ending, steps starting and ending)
- `StairwayInitializationServiceTest` introduced as a pure Unit test with strict stub mocking
- `StartupInitializerTest` introduced as a pure Unit test with strict stub mocking

Together, this unit test chain answers the following questions with precision:
- Does my `MdcHook` do what we expect it to do, including where subflights are concerned?
- Does WSM add my `MdcHook` to Stairway on initialization?
  - And more generally, when WSM initializes Stairway, does it do so with all expected hooks, exception handling, etc?
- Does WSM initialize Stairway?
  - And more generally, does WSM perform all other initialization that we expect?

A failure at any point in the chain will narrow the scope of investigation.  If bugs are identified in the future or behavior changes, the building blocks are in place to expand testing further with ease.

**Previous Pattern: All Unit Tests are SpringBootTests**

This was more work than I expected: the prevailing pattern in WSM is that by default, all unit tests are `SpringBootTest`s which load a full `ApplicationContext` in each time and require careful handling of `MockBean`s to avoid maxing out database connections.

But the majority of unit tests shouldn't need to be `SpringBootTest`s at all, and the few that do shouldn't need to load the full `ApplicationContext` and can instead use context slicing to pull in only the components under test.  Overuse of `@SpringBootTest` with full contexts yields long-running test suites with resource-intensive initialization and disincentivizes writing unit-testable code which can be tested layer by layer.

As such, it makes sense that as we move to implement a new testing strategy in our code, we'll find that there are places where our code is hard to test under that new strategy.  As much as we're able, we should consider this code malleable: if the code is difficult to unit test, this likely points to a design smell and opportunity for refactoring.

More reading on DSP's unit testing philosophy here: [All Roads Lead to Unit Tests](https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2498494502/All+Roads+Lead+to+Unit+Tests)

I hope that this PR can serve as an example for untangling ourselves from this prior default behavior and a springboard for further discussion.  From experience, this untangling is the biggest lift / most daunting at the start but over time the application code grows easier to maintain and its tests easier to expand as a result.

**Advice for Reviewers**

I recommend stepping through the PR commit-by-commit and reading the commit messages.  I also recommend viewing `MdcHookTest` in its entirety rather than in diff view… while the name hasn't changed, the approach to testing has changed so much that the diff view harms rather than helps.


service/src/main/java/bio/terra/workspace/app/StartupInitializer.java
```diff
@@ -1,64 +1,49 @@
...
         applicationContext.getBean(FeatureConfiguration.class);
-    StartupConfiguration startupConfiguration =
-        applicationContext.getBean(StartupConfiguration.class);
 
     // Log the state of the feature flags
     featureConfiguration.logFeatures();
 
     // Migrate the database
     DataSource workspaceDataSource =
-        DataSourceInitializer.initializeDataSource(workspaceDatabaseConfiguration);
```
okotsopoulos, line 38: `DataSourceInitializer` docstring says:
```
Prefer using @DataSourceManager instead of this code so that connection pools are closed when the Spring application context is deleted.
```

Which also has the benefit of making the code easier to test.

service/src/main/java/bio/terra/workspace/common/utils/MdcHook.java
```diff
@@ -111,7 +112,10 @@ public HookAction startFlight(FlightContext flightContext) {
     String serializedMdc = flightContext.getInputParameters().get(MDC_FLIGHT_MAP_KEY, String.class);
     // Note that this destroys any previous context on this thread.
     MDC.setContextMap(deserializeMdc(serializedMdc));
+    // If the serializedMdc itself contains flight and step context from a parent flight, this will
+    // be overridden below:
     addFlightContextToMdc(flightContext);
+    removeStepContextFromMdc(flightContext);
```
okotsopoulos, line 118: This one line of code was the actual point of the PR. :)

service/src/test/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHookTest.java
```diff
@@ -564,121 +518,21 @@ private FlightMap buildInputParams(UUID workspaceUuid, OperationType operationTy
     return inputParams;
   }
 
-  /**
```
okotsopoulos, line 567: I got my inspiration for using a `FlightContext` constructed specifically for Stairway hook tests from here, so I refactored it lightly to share our new `TestFlightContext`.

service/src/test/java/bio/terra/workspace/app/StartupInitializerTest.java
```diff
@@ -0,0 +1,99 @@
...
+  @Mock private DataSource wsmDataSource;
+  @Mock private DataSource lzsDataSource;
+
+  private enum DatabaseInitializationInstruction {
+    INITIALIZE,
+    UPGRADE,
+    CONTINUE
+  }
+
+  @ParameterizedTest
```
cahrens, line 48: Why is this a parameterized test?

service/src/test/java/bio/terra/workspace/app/StartupInitializerTest.java
```diff
@@ -0,0 +1,99 @@
...
+        when(landingZoneDatabaseConfiguration.getDataSource()).thenReturn(lzsDataSource);
+      }
+      case CONTINUE -> {
+        // No additional mocking.
+      }
+    }
+
+    StartupInitializer.initialize(context);
+
+    verify(featureConfiguration).logFeatures();
```
cahrens, line 83: What does this verify do?

service/src/test/java/bio/terra/workspace/app/StartupInitializerTest.java
okotsopoulos: Because when `StartupInitializer` runs, it may initialize the database, upgrade the database, or take no extra action.  Making this a parameterized test allows us to verify all of those pathways without duplicating code.

cahrens: Oh never mind, it's going to iterate over the enum values.

service/src/test/java/bio/terra/workspace/app/StartupInitializerTest.java
okotsopoulos: Exactly.

cahrens (approved):

service/src/test/java/bio/terra/workspace/common/logging/WorkspaceActivityLogHookTest.java
cahrens: Nice!

service/src/main/java/bio/terra/workspace/app/StartupInitializer.java
```diff
@@ -1,64 +1,49 @@
...
     // Initialize Terra Landing Zone library
     LandingZoneMain.initialize(applicationContext, migrateService);
-
-    // NOTE:
-    // Fill in this method with any other initialization that needs to happen
-    // between the point of having the entire application initialized and
-    // the point of opening the port to start accepting REST requests.
-    // TODO: PF-2763 remove after 2023/05/26
-    WorkspaceDao workspaceDao = applicationContext.getBean(WorkspaceDao.class);
-    workspaceDao.backfillCloudContextSpendProfile();
```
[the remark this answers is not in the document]
okotsopoulos: Great shout -- I'll take on that initial clean-up in this PR, and agree with you about designating the rest OOS.

service/src/test/java/bio/terra/workspace/common/BaseUnitTest.java
```diff
@@ -1,12 +1,10 @@
...
 /**
- * Base class for unit tests: not connected to cloud providers. Includes GCP unit tests & others
- * cloud-agnostic WSM component unit tests
+ * Base class for Spring Boot unit tests: not connected to cloud providers. Includes GCP unit tests
+ * & others cloud-agnostic WSM component unit tests
  */
-@Tag("unit")
-@ActiveProfiles("unit-test")
+@Unit
 public class BaseUnitTest extends BaseUnitTestMocks {}
```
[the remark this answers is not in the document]
okotsopoulos: I'm glad you asked this.  Yes, this is currently reflective of the state of WSM unit tests as a whole: it's a non-trivial lift to convert a test suite that currently uses Spring Boot to one that doesn't.  But ultimately, we do not want the base unit test to include Spring.

To make this more clear, I'll make some additional naming changes up the chain of base tests.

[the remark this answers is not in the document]
okotsopoulos: Actually, fully propagating these name changes touched 100+ files.  I walked them back and will handle in a follow-on PR if we'd like, to avoid obscuring the meat of this one.

PR #1648, opened 2024-02-20
Update sam client, remove json lib constraint

aherbst-broad:
[Ticket](https://broadworkbench.atlassian.net/browse/WOR-1520)

We should do the same thing as we did in BPM and remove this constraint and bump the sam lib.


okotsopoulos:
@aherbst-broad would you mind adding a ticket and description? Thanks!

marctalbott (approved):

PR #1651, opened 2024-02-21
[WOR-1529] Rename WSM test bases to indicate that they're SpringBootTests

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1529

To make the existing state of WSM unit tests more clear, this PR renames the Spring Boot base test classes to indicate that they’re Spring Boot tests.  That way it’s less likely that new unit tests will extend them without realizing that they’re pulling in a full WSM context.

This is a follow-on PR from https://github.com/DataBiosphere/terra-workspace-manager/pull/1636#discussion_r1489736673 as it touches a high number of files (nearly all renames) and would have obscured the intent of the original PR.


blakery (approved):

marctalbott (approved):
Thanks for doing this!

PR #1654, opened 2024-02-23
WOR-787 add interface for resource deletion steps

blakery:
Changes to support common error handling for missing resources in Azure
* Adds an interface for a `Step` involved in deleting a `ControlledResource`
* remove `ControlledResource.addDeleteSteps()` in favor of `ControlledResource.getDeleteSteps()` to enforce typing of steps

The majority of the changes in this PR are either changing a class to implementing `DeleteControlledResourceStep` instead of just `Step`, or changing a method in a `ControlledResource` from `addDeleteSteps()` to `getDeleteSteps`.


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/any/flexibleresource/ControlledFlexibleResource.java
blakery, line 1: This changes in this file are just part of the change from `addDeleteSteps` to `getDeleteSteps`

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/kubernetesNamespace/ControlledAzureKubernetesNamespaceResource.java
```diff
@@ -237,19 +229,43 @@ private Step getGetManagedIdentityStep(
...
 
+  /**
+   * TODO: when implementing the
+   *
+   * @param flightBeanBag
+   * @param missingIdentityBehavior
+   * @return
+   */
+  @NotNull
+  private DeleteControlledResourceStep getGetManagedIdentityForDeleteStep(
```
blakery, line 240: This is the one resource where requiring steps that are explicitly typed for delete doesn't fit cleanly.

The main options I could see are:
1.  Switch to a different approach (give up).
2. Add the `DeleteControlledResourceStep` interface to `GetWorkspaceManagedIdentityStep`. This could add undesired behavior on other steps.
3. Add a simple extension for the delete step.

Going with 3 adds a bit more complexity here, but less in the long run.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/DeleteAzureControlledResourceStep.java
```diff
@@ -0,0 +1,5 @@
+package bio.terra.workspace.service.resource.controlled.cloud.azure;
+
+import bio.terra.workspace.service.resource.controlled.flight.delete.DeleteControlledResourceStep;
+
+public abstract class DeleteAzureControlledResourceStep implements DeleteControlledResourceStep {}
```
aherbst-broad, line 5: A note explaining this class' reason for being would be good.

Also, will there be downstream type checking in the azure deletion code to ensure all steps are tagged w/this interface?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/kubernetesNamespace/ControlledAzureKubernetesNamespaceResource.java
```diff
@@ -237,19 +229,43 @@ private Step getGetManagedIdentityStep(
     };
   }
 
+  /**
+   * TODO: when implementing the
```
aherbst-broad, line 233: This TODO should get cleaned up

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/kubernetesNamespace/ControlledAzureKubernetesNamespaceResource.java
aherbst-broad: Can you explain more what the undesired behavior would be on the other steps?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/managedIdentity/GetWorkspaceManagedIdentityForDeleteStep.java
```diff
@@ -0,0 +1,21 @@
...
+
+import bio.terra.workspace.service.resource.controlled.flight.delete.DeleteControlledResourceStep;
+import java.util.UUID;
+
+/**
+ * A simple type extension for GetWorkspaceManagedIdentityForDeleteStep, to type as a
+ * DeleteControlledResourceStep. TODO: figure out if this needs functionality from
+ * DeleteControlledAzureResourceStep (TBD, WOR-787), and adjust as needed
+ */
+public class GetWorkspaceManagedIdentityForDeleteStep extends GetWorkspaceManagedIdentityStep
```
aherbst-broad, line 11: Same question, this strikes me as awkward -- what's the downside in adding the marker interface to the base class?

aherbst-broad (commented):
Looking good, had a couple questions.

Also, given that this is adding expectations for resource developers, an update to either the [ADDING_RESOURCES ](https://github.com/dataBiosphere/terra-workspace-manager) readme or an additional readme in `bio.terra.workspace.service.resource.controlled.cloud.azure` is in order so developers don't have to guess what's needed.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/kubernetesNamespace/ControlledAzureKubernetesNamespaceResource.java
blakery: For example, if we were creating a resource, we would probably _need_ to fail if we can't access a particular resource.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/DeleteAzureControlledResourceStep.java
blakery: This class will become much more obvious with a future PR adding behavior here.
I just didn't want to have to change it twice for all the azure resources.

Typechecking for `DeleteControlledResourceStep` is included in this PR. Going further to require `DeleteAzureControlledResourceStep` doesn't really work without also changing the Azure `ControlledResource`s to extend a common class, and I'm not sure it's worth it yet. I was going to start with adding more description here and and also some specifics in `DeleteControlledResourceStep`, that will also direct to here.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/managedIdentity/GetWorkspaceManagedIdentityForDeleteStep.java
blakery: I went back and forth on this quite a bit.

Now that I'm thinking of the marker interface as simply indicating usage, as opposed to behavior, I think simply adding a brief comment to the `GetWorkspaceManagedIdentityStep`  is probably a better approach.

I think I was thinking too far ahead, and was framing it as the `DeleteControlledAzureResourceStep`.

okotsopoulos (approved):

PR #1655, opened 2024-02-23
[IA-4584] Add a delete database result endpoint

LizBaldo:
JIRA Ticket: https://broadworkbench.atlassian.net/browse/IA-4633

This PR adds the following endpoints:
- An async delete database post to rerun the job ID
- A get delete-result database endpoint to retrieve the status of the delete database job

We are doing this so that Leonardo can properly poll for the status of the database deletion job, instead of assuming that the operation is really done sync (e.g. Leo marks the database as `deleted` immediately, but something goes wrong with the deletion on the WSM side)

Tested on a BEE with the following steps:
- Create a database
- Delete the database using the new async endpoint and make sure a job id is in the response body
- Use the get delete-result endpoint with this job id until the status is `SUCCESS`

### DIFF
--- a/openapi/src/parts/controlled_azure_database.yaml
+++ b/openapi/src/parts/controlled_azure_database.yaml
@@ -72,6 +72,30 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
+    post:
+      summary: |
+        Delete a controlled Azure Database resource. This is async, because the 
+        operation can take a few minutes. OpenAPI does not support request body 
+        in DELETE, but async state requires it. Hence this is a POST.
+      operationId: deleteAzureDatabaseAsync
+      tags: [ ControlledAzureResource ]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/DeleteControlledAzureResourceRequest'
+      responses:
+        '200':
+          $ref: '#/components/responses/DeleteControlledAzureResourceResponse'
+        '202':
+          $ref: '#/components/responses/DeleteControlledAzureResourceResponse'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '500':
+          $ref: '#/components/responses/ServerError'
 
   /api/workspaces/v1/{workspaceId}/resources/controlled/azure/databases/delete-result/{jobId}:
     parameters:
@@ -88,6 +112,8 @@ paths:
           $ref: '#/components/responses/DeleteControlledAzureResourceResponse'
         '403':
           $ref: '#/components/responses/PermissionDenied'
+        '404':
+          $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
 
--- a/openapi/src/parts/controlled_azure_database.yaml
+++ b/openapi/src/parts/controlled_azure_database.yaml
@@ -73,6 +73,24 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/workspaces/v1/{workspaceId}/resources/controlled/azure/databases/delete-result/{jobId}:
+    parameters:
+      - $ref: '#/components/parameters/WorkspaceId'
+      - $ref: '#/components/parameters/JobId'
+    get:
+      summary: Retrieve information about an Azure Database delete job.
+      operationId: getDeleteAzureDatabaseResult
+      tags: [ ControlledAzureResource ]
+      responses:
+        '200':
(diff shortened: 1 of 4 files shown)


service/src/test/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiControllerAzureDatabaseTest.java
```diff
@@ -0,0 +1,142 @@
+package bio.terra.workspace.app.controller;
```
LizBaldo, line 1: I tried to be a good scout and add some unit test coverage for the main create and delete database operations. But I am very new to WSM so please lmk if something important is missing here.

aherbst-broad: This seems like a reasonable amount of coverage for this change; fwiw `deleteDatabase` + `createAzureDatabaseWithNoParameters` NPE when I run locally so perhaps there is some mocking that needs cleaning up?

aherbst-broad (approved):
Looks like a good approach in keeping with the API idioms present in WSM, thanks for doing this.

service/src/test/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiControllerAzureDatabaseTest.java
LizBaldo: Yeah, I trying to understand what I am missing in terms of mocks. I was naively following the examples for disks and VMs as a first pass

openapi/src/parts/controlled_azure_database.yaml
```diff
@@ -72,6 +72,50 @@ paths:
...
+      tags: [ ControlledAzureResource ]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/DeleteControlledAzureResourceRequest'
+      responses:
+        '200':
+          $ref: '#/components/responses/DeleteControlledAzureResourceResponse'
```
jdcanas, line 90: hmm... is this a valid response for a POST?

For disk I think I had to use `CreateControlledAzureResourceResult` over `CreateControlledAzureResourceResponse` for the `post`, and then `...Response` for the subsequent job `get`
```
responses:
        '200':
          description: Response to create controlled Azure disk
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/CreateControlledAzureResourceResult'
```

jdcanas (approved):
Looks good with a few nits, assuming it runs fine in a bee

openapi/src/parts/controlled_azure_database.yaml
LizBaldo: 🤔 I am following the example of the deleteDisk async POST that is using `Response` so I do not think making a `Result` is needed. @aherbst-broad is there an established best practices that you would like us to follow for WSM?

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -339,6 +339,19 @@ ControlledAzureDiskResource buildControlledAzureDiskResource(
         .build();
   }
 
+  @VisibleForTesting
+  ControlledAzureDatabaseResource buildControlledAzureDatabaseResource(
+      ApiAzureDatabaseCreationParameters creationParameters,
+      ControlledResourceFields commonFields) {
+    return ControlledAzureDatabaseResource.builder()
```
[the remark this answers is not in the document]
LizBaldo: It looks like the constructor is already tested here: https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/ControlledAzureDatabaseResourceTest.java

LizBaldo:
Just tested on my BEE and it worked just fine (I had some trouble creating the database for a while and kept hitting a timeout with `exception waiting for azure database utils container`, but slack suggested to retry and it worked after a couple of times)

One weird thing though is that the azureConnectedTests started failing but the PR is not toching them at all. Are they known to be flaky @aherbst-broad?

openapi/src/parts/controlled_azure_database.yaml
lucymcnatt: yeah I think create is different than delete in that the create result returns more information about the resource itself than the response on delete (when it isn't necessary to have the resourceId)

lucymcnatt: TLDR; using the response should be fine

lucymcnatt (approved):
LGTM as long as its working on a BEE, which it sounds like it is!

PR #1664, opened 2024-03-06
[WOR-1561] Modifications to cloneWorkspace to support Rawls stage workspaces

marctalbott:
Ticket: [WOR-1561](https://broadworkbench.atlassian.net/browse/WOR-1561)
- make spend profile id an optional parameter when cloning a workspace
- move policy-specific steps in `CreateWorkspaceFlight` out of `MC_WORKSPACE` case so cloned Rawls stage workspaces will inherit policy from their source workspace

[WOR-1561]: https://broadworkbench.atlassian.net/browse/WOR-1561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -546,8 +546,6 @@ components:
         Request body for cloning an entire workspace. Cloning instructions
         are taken from individual resources.
       type: object
-      required:
-        - spendProfile
       properties:
         destinationWorkspaceId:
           description: |
--- a/openapi/src/parts/workspace.yaml
+++ b/openapi/src/parts/workspace.yaml
@@ -561,7 +561,9 @@ components:
           description: A description of the cloned workspace
           type: string
         spendProfile:
-          description: ID of provided spend profile
+          description: >-
+            ID of provided spend profile. If not present, source workspace spend profile will be used. 
+            If source workspace spend profile is also not present, destination workspace will have no spend profile
           type: string
         location:
           description: >-
--- a/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
@@ -620,21 +620,20 @@ public class WorkspaceApiController extends ControllerBase implements WorkspaceA
           samService.getUserEmailFromSamAndRethrowOnInterrupt(getAuthenticatedInfo()));
     }
 
-    // TODO: PF-2694 REST API part
-    //  When we make the REST API changes, the spend profile will come with the source cloud context
-    //  and we will take it from there. Only if missing, will we take it from the workspace. For
-    //  this part, we do the retrieval and permission check using the workspace spend profile.
-    Optional<SpendProfileId> spendProfileIdOptional =
-        Optional.ofNullable(body.getSpendProfile()).map(SpendProfileId::new);
-
-    if (spendProfileIdOptional.isEmpty()) {
-      spendProfileIdOptional = sourceWorkspace.getSpendProfileId();
+    // Use the requested spend profile if present. If not present, use the source workspace's
+    // spend profile. If source workspace also has no spend profile, the clone operation will
+    // proceed and the destination workspace will have no spend profile.
+    SpendProfileId workspaceSpendProfileId = sourceWorkspace.getSpendProfileId().orElse(null);
+    SpendProfileId spendProfileId =
+        Optional.ofNullable(body.getSpendProfile())
+            .map(SpendProfileId::new)
+            .orElse(workspaceSpendProfileId);
+    SpendProfile spendProfile = null;
+    if (spendProfileId != null) {
+      spendProfile =
+          spendProfileService.authorizeLinking(
+              spendProfileId, features.isBpmGcpEnabled(), petRequest);
     }
-    Optional<SpendProfile> spendProfileOptional =
-        spendProfileIdOptional.map(
-            spendProfileId ->
-                spendProfileService.authorizeLinking(
-                    spendProfileId, features.isBpmGcpEnabled(), petRequest));
 
     // Accept a target workspace id if one is provided. This allows Rawls to specify an
     // existing workspace id. WSM then creates the WSMspace supporting the Rawls workspace.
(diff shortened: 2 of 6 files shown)


service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerTest.java
```diff
@@ -369,11 +374,7 @@ public void cloneWorkspace_rawls() throws Exception {
 
     ApiCloneWorkspaceResult cloneWorkspace =
         mockWorkspaceV1Api.cloneWorkspace(
-            USER_REQUEST,
-            sourceWorkspaceId,
-            DEFAULT_SPEND_PROFILE_NAME,
-            null,
-            destinationWorkspaceId);
+            USER_REQUEST, sourceWorkspaceId, null, null, destinationWorkspaceId);
```
cahrens, line 377: This tests the case of not having a spend profile on clone. Do we have coverage elsewhere for having a profile (and linking the policy attributes)?

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -626,17 +626,15 @@ public ResponseEntity<ApiCloneWorkspaceResult> cloneWorkspace(
...
-        spendProfileService.authorizeLinking(
-            spendProfileId, features.isBpmGcpEnabled(), petRequest);
+    if (spendProfileIdOptional.isEmpty()) {
+      spendProfileIdOptional = sourceWorkspace.getSpendProfileId();
+    }
+    Optional<SpendProfile> spendProfileOptional =
+        spendProfileIdOptional.map(
+            spendProfileId ->
+                spendProfileService.authorizeLinking(
+                    spendProfileId, features.isBpmGcpEnabled(), petRequest));
```
okotsopoulos, line 637: I like Optionals, especially for defining default values inline, and I see that they are widely used through the repo. But sometimes I find that they can obscure the intent of a block of code, which is the case for me here.

The intent is to fall back to the source workspace's spend profile if none is specified.  We are okay with passing a null profile ID to the `Workspace.Builder` and a null profile to `workspaceservice.cloneWorkspace`.

I found the following to better convey the intent, and then you don't have to unpack the Optionals further down in the method.  Thoughts?

```
    SpendProfileId workspaceSpendProfileId = sourceWorkspace.getSpendProfileId().orElse(null);
    SpendProfileId spendProfileId =
            Optional.ofNullable(body.getSpendProfile()) 
                    .map(SpendProfileId::new)
                    .orElse(workspaceSpendProfileId);
    SpendProfile spendProfile = null;
    if (spendProfileId != null) {
      spendProfile = spendProfileService.authorizeLinking(spendProfileId, features.isBpmGcpEnabled(), petRequest);
    }
```

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -626,17 +626,15 @@ public ResponseEntity<ApiCloneWorkspaceResult> cloneWorkspace(
...
-                .getSpendProfileId()
-                .orElseThrow(
-                    () ->
-                        MissingSpendProfileException.forWorkspace(sourceWorkspace.workspaceId())));
 
-    SpendProfile spendProfile =
-        spendProfileService.authorizeLinking(
-            spendProfileId, features.isBpmGcpEnabled(), petRequest);
+    if (spendProfileIdOptional.isEmpty()) {
+      spendProfileIdOptional = sourceWorkspace.getSpendProfileId();
```
okotsopoulos, line 631: Just curious -- under what scenarios would a workspace not have a spend profile?  I see that this field is an Optional.

openapi/src/parts/workspace.yaml
```diff
@@ -546,8 +546,6 @@ components:
         Request body for cloning an entire workspace. Cloning instructions
         are taken from individual resources.
       type: object
-      required:
-        - spendProfile
```
okotsopoulos, line 550: Is it worth expanding the documentation of the `spendProfile` property below to state the fallback behavior when unspecified?

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
marctalbott: I like it, I felt this was all a bit clunky so thank you for the suggestion! I'm very accustomed to using Scala Options and avoiding null at all costs, but I haven't found Java Optionals to be quite as nice.

openapi/src/parts/workspace.yaml
marctalbott: Yes, I think it is. I'll add to it

service/src/test/java/bio/terra/workspace/app/controller/WorkspaceApiControllerTest.java
marctalbott: The test above covers cloning with a profile. I also added a verify to this test to ensure that we're merging the policy from source -> destination correctly.

okotsopoulos (approved):

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
```diff
@@ -620,23 +620,20 @@ public ResponseEntity<ApiCloneWorkspaceResult> cloneWorkspace(
...
-                .getSpendProfileId()
-                .orElseThrow(
-                    () ->
-                        MissingSpendProfileException.forWorkspace(sourceWorkspace.workspaceId())));
-
-    SpendProfile spendProfile =
-        spendProfileService.authorizeLinking(
-            spendProfileId, features.isBpmGcpEnabled(), petRequest);
+        Optional.ofNullable(body.getSpendProfile())
+            .map(SpendProfileId::new)
```
cahrens, line 629: I'm having trouble understanding this-- what is the .map operation doing?

marctalbott: If there's a spend profile passed in the body of the request, this will create a `SpendProfileId` from the string value. If there's no value passed in the request, then this will be an empty `Optional`

cahrens (approved):

service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java
cahrens: Ah, so `SpringProfileId::new` is basically a constructor call.

PR #1666, opened 2024-03-07
[WOR-1560] WSM uses Stairway 1.0.0's native MDC support

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1560

**Objective**

WSM uses [Stairway 1.0.0's native MDC (mapped diagnostic context) support](https://github.com/DataBiosphere/stairway/pull/125), rather than populating the log MDC itself on flight state transitions.

**How?**

To achieve this, I first upgraded to the latest terra-common-lib, from which we get our Stairway dependency.  Then I was able to drastically simplify our `MdcHook`: it is now `StairwayLoggingHook`, and does what it says on the tin (supplement logging on notable state transitions).  I removed now-unused code, including cleaning up `JobBuilder`'s job initialization which no longer has to serialize the calling thread's MDC and write it to the input map.

Because this [TCL version got rid of Flagsmith](https://github.com/DataBiosphere/terra-common-lib/pull/158), and WSM used Flagsmith in its user-level feature flagging (a Verily artifact), I excised Flagsmith from our repo to allow this upgrade and ticketed further clean-up here (we discussed in refinement that we wanted to fully remove AWS code from the repo): https://broadworkbench.atlassian.net/browse/WOR-1559

**Other Dependency Upgrades**

I had some trouble with dependencies straight out of the box: OTEL dependencies, and a Google BOM conflict within our repo.

I learned that Spring Boot Dependency Manager for Spring Boot 3.1.2 pulls in [opentelemetry-bom:1.25.0](https://github.com/spring-projects/spring-boot/blob/v3.1.2/spring-boot-project/spring-boot-dependencies/build.gradle#L1068-L1074) as a dependency.

Using the [BOM (bill of materials)](https://github.com/open-telemetry/opentelemetry-java) to manage OTEL dependencies is preferred to managing them individually: the BOM makes sure that they remain compatible with one another.

The only OTEL dependencies that WSM needs to define directly are:
1. `opentelemetry-api` so that we can construct `OpenTelemetry` objects (versioned by the BOM, which is itself versioned by Spring Boot dependency manager)
2. `opentelemetry-instrumentation-annotations` so that we can use the `@WithSpan` annotation (not versioned by the BOM)

[terra-common-lib defines the OTEL deps it needs](https://github.com/DataBiosphere/terra-common-lib/blob/3f898a95a10e2c8cee2bab719ca09647b2390040/build.gradle#L92-L109), they are available as runtime dependencies so we don't need to redefine them.  **But** it does not use the BOM to version them, or [Spring Boot dependency manager](https://github.com/DataBiosphere/terra-common-lib/blob/3f898a95a10e2c8cee2bab719ca09647b2390040/build.gradle#L11-L15)... it versions them directly at 1.34.1 (and 2.0.0 for instrumentation).

I found when removing the OTEL deps that WSM didn't need, I also needed to instruct our Spring Boot dependency manager to pin opentelemetry-bom to a higher version compatible with TCL's OTEL dependencies.

**Demonstration**

I deployed the [latest version of this feature branch](https://broadworkbench.atlassian.net/wiki/spaces/WOR/pages/2916679772/How+to+test+a+WSM+branch+on+a+BEE) to my [Bee](https://workspace.okotsopo.bee.envs-terra.bio/swagger-ui.html#), and triggered a [flight](https://workspace.okotsopo.bee.envs-terra.bio/swagger-ui.html#/Workspace/deleteWorkspace).  The [resulting logs](https://cloudlogging.app.goo.gl/X9kDfLy5oAhhhj8WA) have their `jsonPayload` populated as we'd expect:

- Calling thread's context over the entirety of the flight (that's where `jsonPayload.requestId` comes from)
- Flight-specific context over the entirety of the flight
- Step-specific context over the entirety of each step

A `jsonPayload` for a log emitted during a step:
Screenshot 2024-03-18 at 1 00 39 PM

All logs for the request, with `flightId` and `flightStepNumber` added to the summary line:

Screenshot 2024-03-18 at 1 02 08 PM


service/gradle/dependencies.gradle
```diff
@@ -42,24 +42,24 @@ dependencies {
...
-  implementation "io.opentelemetry.instrumentation:opentelemetry-spring-boot:${openTelemetryVersion}-alpha"
+  implementation "io.opentelemetry.instrumentation:opentelemetry-spring-webmvc-6.0:${openTelemetryInstrumentationVersion}-alpha"
+  implementation "io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:${openTelemetryInstrumentationVersion}"
+  implementation "io.opentelemetry.instrumentation:opentelemetry-spring-boot:${openTelemetryInstrumentationVersion}-alpha"
 
-  var gcpOpenTelemetryExporterVersion = '0.25.2'
+  // Google cloud open telemetry exporters
+  var gcpOpenTelemetryExporterVersion = '0.27.0'
   implementation "com.google.cloud.opentelemetry:exporter-trace:${gcpOpenTelemetryExporterVersion}"
   implementation "com.google.cloud.opentelemetry:exporter-metrics:${gcpOpenTelemetryExporterVersion}"
```
okotsopoulos, line 59: Naively mimicking what I saw in https://github.com/DataBiosphere/terra-common-lib/pull/131/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7 -- but I want to understand whether there's a better way deal with these transitive deps.

aherbst-broad (approved):

blakery (approved):
Sorry - I went through everything yesterday, but forgot to hit the button!

service/gradle/dependencies.gradle
okotsopoulos: @blakery this was a funny one.  There is an opentelemetry-bom (bill of materials) that ideally we'd use to version OTEL dependencies rather than versioning them directly, so that they remain in compatibility with one another.  I was surprised to find that this is a dependency that Spring Boot Dependency Manager brings in, and that plug-in clobbers the versions we see from other sources.

Once I instructed the Spring Boot dependency manager to use an OTEL BOM version compatible with latest TCL, I was able to dramatically simplify our OTEL direct dependencies:

https://github.com/DataBiosphere/terra-workspace-manager/pull/1666/commits/8aef8d631b7f853a0f6d14e7120b068031e03a70

PR #1675, opened 2024-03-20
[WOR-1566] Upgrade io.kubernetes:client-java to 20.0.1

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1566

This change was motivated by a desire to address a security vulnerability with jose4j, a transitive dependency of the Kubernetes client. Upgrading the client brought this dependency past its vulnerability without needing to explicitly pin it.

Breaking changes were introduced in v20.0.0: optional parameters are now consolidated into a single object, and we must explicitly execute our requests (the client now supports different flavors of execution such as async).

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/managedIdentity/CreateFederatedIdentityStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/managedIdentity/CreateFederatedIdentityStep.java
@@ -23,7 +23,6 @@ import com.azure.resourcemanager.msi.fluent.models.FederatedIdentityCredentialIn
 import com.google.common.annotations.VisibleForTesting;
 import io.kubernetes.client.openapi.ApiException;
 import io.kubernetes.client.openapi.apis.CoreV1Api;
-import io.kubernetes.client.openapi.models.V1DeleteOptions;
 import io.kubernetes.client.openapi.models.V1ObjectMeta;
 import io.kubernetes.client.openapi.models.V1ServiceAccount;
 import java.util.List;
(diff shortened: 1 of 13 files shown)


blakery (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/managedIdentity/CreateFederatedIdentityStep.java
```diff
@@ -157,12 +156,11 @@ private void createK8sServiceAccount(CoreV1Api aksApi, String uamiClientId) thro
...
                     .namespace(k8sNamespace));
 
-    aksApi.createNamespacedServiceAccount(k8sNamespace, k8sServiceAccount, null, null, null, null);
+    aksApi.createNamespacedServiceAccount(k8sNamespace, k8sServiceAccount).execute();
   }
 
   private void deleteK8sServiceAccount(CoreV1Api aksApi, String k8sNamespace) throws ApiException {
-    aksApi.deleteNamespacedServiceAccount(
-        ksaName, k8sNamespace, null, null, null, null, null, new V1DeleteOptions());
+    aksApi.deleteNamespacedServiceAccount(ksaName, k8sNamespace).execute();
```
cahrens, line 163: You lost the deleteOptions here.

cahrens: Example: https://github.com/DataBiosphere/terra-landing-zone-service/pull/423/files#diff-dc845e0597a636315d1033f7e01368dfcc4af46bf7d9e7f6b4ba21a3bec5281dR102

okotsopoulos: Thanks for the example.

okotsopoulos: I can add it back in, but do you know what the purpose of an empty `V1DeleteOptions()` object is?

cahrens: Ha, no I do not! I was just trying to make sure all the original arguments still were maintained. Since we weren't passing null for it, I naively assumed it was somehow important.

okotsopoulos: All good!  I think even if it's not buying us anything, there's value in maintaining the original arguments within the scope of this upgrade, I added it back.

cahrens (approved):

PR #1678, opened 2024-03-21
[WOR-1586] Handle processing exception in TpsRetry

aherbst-broad:
These exceptions can occur when there are low-level IO issues (socket timeout, etc.). They bypass the retry mechanism due to the exception filter on ApiException, causing flakiness with tests (and presumably in actual usage).

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/policy/TpsRetry.java
+++ b/service/src/main/java/bio/terra/workspace/service/policy/TpsRetry.java
@@ -103,7 +103,7 @@ public class TpsRetry {
         }
       } catch (ProcessingException ws) {
         logger.info("TpsRetry: caught retry-able ProcessingException: ", ws);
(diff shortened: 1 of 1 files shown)


service/src/main/java/bio/terra/workspace/service/policy/TpsRetry.java
```diff
@@ -100,6 +101,9 @@ private <T> T perform(TpsRetry.TpsFunction<T> function)
         } else {
           throw ex;
         }
+      } catch (ProcessingException ws) {
```
okotsopoulos, line 104: Confirming my understanding -- all `ProcessingExceptions` are retryable (unlike `ApiExceptions`, only a subset of which are retryable).

okotsopoulos (approved):
Thanks for making this improvement!

marctalbott (approved):

PR #1685, opened 2024-03-26
WOR-1538 Azure delete disk flight fails dismally

sergiygetlin:
JIRA: https://broadworkbench.atlassian.net/browse/WOR-1538

Before deleting an Azure disk, we need to make sure that a disk is detached from a virtual machine. Otherwise, disk deletion will fail. 

This PR introduces additional steps for the disk deletion flight:
1) GetAzureDiskAttachedVmStep - to save identifier of a virtual machine to which disk is attached. This is needed in case detach process succeeded, but disk deletion failed, and we need to get back to the previous state and attach disk to the original virtual machine.
2) DetachAzureDiskFromVmStep - to detach a disk from a virtual machine if needed.

Validated following scenarios:

Happy path:
1) Disk attached to a VM and deleted successfully.
2) Disk is not attached to a VM and deleted successfully.

Unhappy path:
1) Disk is attached to a virtual machine. During deletion it was detached, but deletion itself failed. As a result disk was attached back to the original virtual machine.


service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DeleteAzureDiskStepTest.java
```diff
@@ -14,22 +15,28 @@
...
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
 import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
 
-class DeleteAzureDiskStepTest extends BaseAzureSpringBootUnitTest {
+@Tag("unit")
```
sergiygetlin, line 37: This class implements pure unit tests. There is no reason to have Spring context here.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DeleteAzureDiskStepUnitTest.java
sergiygetlin, line 1: The reason this file removed is because we already have similar tests implemented in this test class -DeleteAzureDiskStepTest.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DetachAzureDiskFromVmStep.java
```diff
@@ -0,0 +1,190 @@
...
+            resource.getWorkspaceId());
+        return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, ex);
+      }
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, ex);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  protected StepResult deleteResource(FlightContext context) throws InterruptedException {
```
blakery, line 102: I don't think the naming convention of `deleteResource` should prevent using it here.

Most of the error cases in the `doStep` of this class are already handled in `DeleteAzureControlledResourceStep`.

Then if there is additional error handling, you can override the exception handling method for just the cases you care about.

As it's currently written, you might as well use the `DeleteControlledResourceStep`.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/GetAzureDiskAttachedVmStepTest.java
```diff
@@ -0,0 +1,123 @@
...
+  @Test
+  void undoStep_Success() throws InterruptedException {
+    // always return success
+    assertThat(
+        getAzureDiskAttachedVmStep.undoStep(flightContextMock),
+        equalTo(StepResult.getStepResultSuccess()));
+  }
+
+  private void setupCommonMocks() {
+    when(controlledAzureDiskResourceMock.getDiskName()).thenReturn(DISK_NAME);
```
blakery, line 103: If this is happening in every test, why not just include it in the setup method?

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/GetAzureDiskAttachedVmStepTest.java
```diff
@@ -0,0 +1,123 @@
...
+  void doStep_VmIdentifierSaved() throws InterruptedException {
+    final String vmId = "vmId";
+    setupCommonMocks();
+    when(disksMock.getById(anyString())).thenReturn(diskMock);
+    when(diskMock.isAttachedToVirtualMachine()).thenReturn(true);
+    when(diskMock.virtualMachineId()).thenReturn(vmId);
+    StepResult stepResult = getAzureDiskAttachedVmStep.doStep(flightContextMock);
+
+    assertThat(stepResult, equalTo(StepResult.getStepResultSuccess()));
+    verify(flightWorkingMapMock, times(1))
```
blakery, line 68: Nit/FYI: the default verification mode is `times(1)`. So you don't need to specify it unless you want > 1.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/GetAzureDiskAttachedVmStep.java
```diff
@@ -0,0 +1,83 @@
...
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    // nothing to undo here
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  protected StepResult deleteResource(FlightContext context) throws InterruptedException {
+    /*
+    This step follows the contract for a step which deletes an Azure resource, but the implementation diverges from "standard".
```
blakery, line 76: If this step isn't using _any_ of the functionality of `DeleteAzureControlledResourceStep`, I think it would probably be better to just use the `DeleteControlledResourceStep`, since that was intended to be a basic indicator interface anyway. 
If we think there's a way to improve `DeleteAzureControlledResourceStep` that will make it more applicable here, we should do that. 
We might also consider making `DeleteAzureControlledResourceStep.doStep()` final, just so it's clear what the pattern is.

I realize other opinions on this may vary - I'm not really intending to ask for specific changes here, just to prompt some examination and debate.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DetachAzureDiskFromVmStep.java
```diff
@@ -0,0 +1,190 @@
...
+      if (AzureManagementExceptionUtils.isExceptionCode(
+          ex, AzureManagementExceptionUtils.RESOURCE_NOT_FOUND)) {
+        logger.info(
+            "Vm to which disk was attached does not exist in workspace {}.",
+            resource.getWorkspaceId());
+        return Optional.empty();
+      }
+      throw ex;
+    }
+    return Optional.of(vm);
```
aherbst-broad, line 168: 👍 I like that this returns an optional rather than a null.

aherbst-broad (approved):
lgtm, had a few questions.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DetachAzureDiskFromVmStep.java
```diff
@@ -0,0 +1,190 @@
...
+      if (AzureManagementExceptionUtils.isExceptionCode(
+          ex, AzureManagementExceptionUtils.RESOURCE_NOT_FOUND)) {
+        logger.info(
+            "Disk '{}' does not exist in workspace {}.",
+            resource.getResourceId(),
+            resource.getWorkspaceId());
+        return StepResult.getStepResultSuccess();
+      }
+      if (AzureManagementExceptionUtils.isExceptionCode(
+          ex, AzureManagementExceptionUtils.OPERATION_NOT_ALLOWED)) {
```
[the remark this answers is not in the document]
sergiygetlin: Yes, this is to handle potential detaching issue and fail fast in this case. I don't have good example of a potential cause, but concurrency might be the case.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DetachAzureDiskFromVmStep.java
```diff
@@ -0,0 +1,190 @@
...
+    } catch (ManagementException ex) {
+      if (AzureManagementExceptionUtils.isExceptionCode(
+          ex, AzureManagementExceptionUtils.RESOURCE_NOT_FOUND)) {
+        logger.info(
+            "Disk '{}' does not exist in workspace {}.",
+            resource.getResourceId(),
+            resource.getWorkspaceId());
+        return StepResult.getStepResultSuccess();
+      }
+      if (AzureManagementExceptionUtils.isExceptionCode(
```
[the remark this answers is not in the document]
sergiygetlin: The same but for attaching.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DetachAzureDiskFromVmStep.java
```diff
@@ -0,0 +1,190 @@
...
+            "Disk {} is attached, but not found in the list of attached disk for vm {}.",
+            resource.getResourceId(),
+            disk.virtualMachineId());
+        return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY);
+      }
+    } catch (ManagementException ex) {
+      if (AzureManagementExceptionUtils.isExceptionCode(
+          ex, AzureManagementExceptionUtils.RESOURCE_NOT_FOUND)) {
+        logger.info(
+            "Disk '{}' does not exist in workspace {}.",
```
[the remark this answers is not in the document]
sergiygetlin: Agree. Probably make sense to make this consistent across all similar logging.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DeleteAzureDiskFlightUtils.java
```diff
@@ -0,0 +1,31 @@
...
+public class DeleteAzureDiskFlightUtils {
+  public static final String DISK_ATTACHED_VM_ID_KEY = "DISK_ATTACHED_VM_ID";
+
+  private static final String DISK_RESOURCE_ID_FORMAT =
+      "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/disks/%s";
+
+  private DeleteAzureDiskFlightUtils() {}
+
+  public static AzureCloudContext getAzureCloudContext(FlightContext context) {
+    return context
```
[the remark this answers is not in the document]
sergiygetlin: Good catch! Forgot about that.

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/GetAzureDiskAttachedVmStepTest.java
sergiygetlin: @blakery not sure what exactly do you mean here. It looks like I already moved all common code into `setupCommonMocks`. No?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DetachAzureDiskFromVmStep.java
```diff
@@ -0,0 +1,190 @@
...
+      if (vm.isEmpty()) {
+        // this is unlikely event - disk is attached, but the vm cannot be found; let's retry
+        logger.warn(
+            "Disk {} is attached to vm in workspace {}, but vm cannot be found. Retrying.",
+            resource.getResourceId(),
+            resource.getWorkspaceId());
+        return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY);
+      }
+      var attachedDisk =
+          vm.get().dataDisks().entrySet().stream()
```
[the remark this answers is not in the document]
sergiygetlin: In this context `entrySet` is a part of HashMap implementation and it could not be null. According to the Java SDK `dataDisk()` also returns empty collection and it is unlikely that SDK behavior in this context will change. But I agree that we need to protect ourselves.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DetachAzureDiskFromVmStep.java
sergiygetlin: In this case, I definitely intended to use an interface, but started using the abstract class as if the interface didn't exist.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/GetAzureDiskAttachedVmStep.java
sergiygetlin: The same here. I indented to use interface.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/DetachAzureDiskFromVmStep.java
```diff
@@ -0,0 +1,196 @@
...
+        return Optional.empty();
+      }
+      throw ex;
+    }
+    return Optional.of(vm);
+  }
+
+  private void detachDisk(VirtualMachine vm, Integer lun) {
+    var vmUpdate = vm.update();
+    vmUpdate.withoutDataDisk(lun);
```
cahrens, line 169: This methods don't return the mutated state? I'm accustomed to seeing chaining, like

`vm.update().withoutDataDisk(lun).apply()`

sergiygetlin: That's valid observation. Current implementation is mutating its state. Here is vm [implementation](https://github.com/Azure/azure-libraries-for-java/blob/22f90e7c258c304eafe25c2558ddf7de81563198/azure-mgmt-compute/src/main/java/com/microsoft/azure/management/compute/implementation/VirtualMachineImpl.java#L1096). It is not necessary, but I will update it.

cahrens (approved):

PR #1687, opened 2024-03-27
WOR-1519: add method to retrieve a resource of a workspace by name

blakery:
[WOR-1519](https://broadworkbench.atlassian.net/browse/WOR-1519)

Adds an endpoint to look up reasources by name, instead of enumerating all the resources in a workspace.
If the user does not have access to the entire workspace, access is checked for the specific resource instead.

Using the new endpoint, I was able to get leo to create apps a workspace with an auth domain, in a BEE.
With the first attempt, the app creation still failed, because of a new call to retrieve the workspace description. After pushing a version that didn't have that call, leo was able to successfully create the apps.
Fixing that call will require a separate solution.

[WOR-1519]: https://broadworkbench.atlassian.net/browse/WOR-1519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ResourceApiController.java
@@ -4,6 +4,7 @@ import static bio.terra.workspace.app.controller.shared.PropertiesUtils.convertA
 import static bio.terra.workspace.common.utils.ControllerValidationUtils.validatePropertiesDeleteRequestBody;
 import static bio.terra.workspace.common.utils.ControllerValidationUtils.validatePropertiesUpdateRequestBody;
 
+import bio.terra.common.exception.ForbiddenException;
 import bio.terra.workspace.app.configuration.external.FeatureConfiguration;
 import bio.terra.workspace.app.controller.shared.JobApiUtils;
 import bio.terra.workspace.generated.controller.ResourceApi;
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/app/controller/ResourceApiController.java
```diff
@@ -111,6 +116,54 @@ public ResponseEntity<ApiResourceDescription> getResource(UUID workspaceUuid, UU
...
+        case CONTROLLED -> {
+          try {
+            return controlledResourceMetadataManager.validateControlledResourceAndAction(
+                userRequest, workspaceUuid, resource.getResourceId(), readAction);
+          } catch (ForbiddenException resourceException) {
+            // if no access, throw original exception to avoid leaking information on resource
+            throw workspaceException;
+          }
+        }
+        case REFERENCED -> {
```
marctalbott, line 151: I'm not sure this is exactly what we want for a referenced resource. Checking access on the referenced resource involves a permissions check on the underlying cloud resource that's referenced. If a user doesn't have access to a workspace, then access to the underlying resource that's referenced shouldn't grant access to read the reference itself too IMO. 

Not really sure what the downside of allowing the access would be, but still best prevent it

marctalbott (approved):

okotsopoulos (approved):
Looks great!

PR #1688, opened 2024-03-27
Replace Stairway example waitForFlight with more robust method (WOR-1…

cahrens:
…591).

### DIFF
--- a/service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
+++ b/service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
@@ -29,7 +29,7 @@ public final class FlightUtils {
   public static final int FLIGHT_POLL_CYCLES = 360;
 
   // Parameters for waiting for subflight completion
-  private static final Duration SUBFLIGHT_TOTAL_DURATION = Duration.ofHours(1);
+  public static final Duration SUBFLIGHT_TOTAL_DURATION = Duration.ofHours(1);
   private static final Duration SUBFLIGHT_INITIAL_SLEEP = Duration.ofSeconds(10);
   private static final double SUBFLIGHT_FACTOR_INCREASE = 0.7;
   private static final Duration SUBFLIGHT_MAX_SLEEP = Duration.ofMinutes(2);
@@ -220,6 +220,11 @@ public final class FlightUtils {
    */
   public static StepResult waitForSubflightCompletion(Stairway stairway, String flightId)
       throws InterruptedException {
+    return waitForSubflightCompletionAndReturnFlightInfo(stairway, flightId).getStepResult();
+  }
+
+  public static StepResultWithFlightInfo waitForSubflightCompletionAndReturnFlightInfo(
+      Stairway stairway, String flightId) throws InterruptedException {
     try {
       FlightState flightState =
           waitForFlightCompletion(
@@ -230,19 +235,22 @@ public final class FlightUtils {
               SUBFLIGHT_FACTOR_INCREASE,
               SUBFLIGHT_MAX_SLEEP);
       if (flightState.getFlightStatus() == FlightStatus.SUCCESS) {
-        return StepResult.getStepResultSuccess();
+        return new StepResultWithFlightInfo(flightState, StepResult.getStepResultSuccess());
       }
-      return new StepResult(
-          StepStatus.STEP_RESULT_FAILURE_FATAL,
-          flightState
-              .getException()
-              .orElse(new RuntimeException("Flight failed with an empty exception")));
+      return new StepResultWithFlightInfo(
+          flightState,
+          new StepResult(
+              StepStatus.STEP_RESULT_FAILURE_FATAL,
+              flightState
+                  .getException()
+                  .orElse(new RuntimeException("Flight failed with an empty exception"))));
     } catch (InterruptedException ie) {
       // Propagate the interrupt
       Thread.currentThread().interrupt();
       throw ie;
     } catch (Exception e) {
-      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+      return new StepResultWithFlightInfo(
+          null, new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e));
     }
   }
 
--- a/service/src/main/java/bio/terra/workspace/common/utils/StepResultWithFlightInfo.java
+++ b/service/src/main/java/bio/terra/workspace/common/utils/StepResultWithFlightInfo.java
@@ -1,0 +1,65 @@
+package bio.terra.workspace.common.utils;
+
+import bio.terra.stairway.*;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+public class StepResultWithFlightInfo {
+  @Nullable private FlightState flightState;
+  private StepResult stepResult;
+
+  /**
+   * FlightState and StepResult are returned together from a step execution.
+   *
+   * @param flightState may be null if the flight failed with a runtime exception
+   * @param stepResult the result of the step
+   */
+  public StepResultWithFlightInfo(FlightState flightState, StepResult stepResult) {
+    this.flightState = flightState;
+    this.stepResult = stepResult;
+  }
+
+  public StepResult getStepResult() {
+    return stepResult;
+  }
+
+  /** Returns true if the step was successful. */
+  public boolean isSuccess() {
+    return stepResult.isSuccess();
+  }
+
+  /** Returns the overall status of the step. */
+  public StepStatus getStepStatus() {
+    return stepResult.getStepStatus();
+  }
+
+  public Optional<Exception> getStepException() {
+    return stepResult.getException();
+  }
+
+  /** Returns the last status of the flight that was executed during the step. */
+  public FlightStatus getFlightStatus() {
+    if (flightState == null) {
+      // A runtime error occurred in the flight.
+      return FlightStatus.FATAL;
+    }
+    return flightState.getFlightStatus();
+  }
+
+  public String getFlightErrorMessage() {
+    if (flightState == null) {
+      return String.format(
+          "Subflight had unexpected status %s. No exception for subflight found.", getStepStatus());
+    }
+    return FlightUtils.getFlightErrorMessage(flightState);
+  }
+
+  /** Will be null if the flight failed in such a way that no map was saved. */
+  @Nullable
+  public FlightMap getFlightMap() {
+    if (flightState == null) {
+      return null;
+    }
+    return FlightUtils.getResultMapRequired(flightState);
+  }
+}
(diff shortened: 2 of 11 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
```diff
@@ -21,8 +21,6 @@ public class BlobCopier {
   private final AzureStorageAccessService storageAccessService;
   private final AuthenticatedUserRequest userRequest;
 
-  public static final Duration MAX_BLOB_COPY_POLL_TIMEOUT = Duration.ofMinutes(60);
```
cahrens, line 24: SUBFLIGHT_TOTAL_DURATION is 60 minutes, and it is what the utility method uses.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneAllResourcesFlightStep.java
```diff
@@ -34,8 +32,6 @@ public class AwaitCloneAllResourcesFlightStep implements Step {
 
   public AwaitCloneAllResourcesFlightStep() {}
 
-  private static final int AWAIT_CLONE_RESOURCES_POLL_CYCLES = 720;
```
cahrens, line 37: This would poll for a total of 12 minutes because we polled every second. We now poll for an hour maximum, using the constants defined in https://github.com/DataBiosphere/terra-workspace-manager/pull/1688/files#diff-af300abcc49577c64d79724ef7af4288bbe77f0e6f5ed6c4e42e2a9a323379adR32.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneAllResourcesFlightStep.java
```diff
@@ -53,37 +49,26 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
-            StepStatus.STEP_RESULT_FAILURE_FATAL,
-            subflightState
-                .getException()
-                .orElseGet(
-                    () ->
-                        new RuntimeException(
-                            String.format(
-                                "Subflight had unexpected status %s. No exception for subflight found.",
-                                subflightState.getFlightStatus()))));
+        return subflightResult.getStepResult();
```
cahrens, line 58: The constructed StepResult will include exception information.
https://github.com/DataBiosphere/terra-workspace-manager/pull/1688/files#diff-af300abcc49577c64d79724ef7af4288bbe77f0e6f5ed6c4e42e2a9a323379adR240

cahrens:
To get an idea of test coverage, I inverted the return value of [isSuccess](https://github.com/DataBiosphere/terra-workspace-manager/pull/1689/commits/2146237b9118e8e0f357180e6c06b1bc6ce423b6) to see what tests failed.

Unittests:
WorkspaceApiControllerTest » cloneWorkspace_rawls()FAILED | 20.293s |  
WorkspaceApiControllerTest » cloneWorkspace()FAILED | 19.126s

service/src/main/java/bio/terra/workspace/common/utils/StepResultWithFlightInfo.java
```diff
@@ -0,0 +1,52 @@
+package bio.terra.workspace.common.utils;
+
+import bio.terra.stairway.*;
+import javax.annotation.Nullable;
+
+public class StepResultWithFlightInfo {
```
cahrens, line 6: I didn't write a unittest for this class because it is mostly just pass-throughs, but I can add one if folks think it's worthwhile.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneAllResourcesFlightStep.java
```diff
@@ -53,37 +49,24 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
-      //noinspection deprecation
-      FlightState subflightState =
-          context
-              .getStairway()
-              .waitForFlight(
-                  cloneAllResourcesFlightId,
-                  FLIGHT_POLL_SECONDS,
-                  AWAIT_CLONE_RESOURCES_POLL_CYCLES);
-      if (FlightStatus.SUCCESS != subflightState.getFlightStatus()) {
+      StepResultWithFlightInfo subflightResult =
```
cahrens, line 52: We do have unit test coverage for this usage.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/BlobCopier.java
```diff
@@ -77,7 +75,7 @@ public BlobCopierResult copyBlobs(
         destStorageData.storageContainerResource().getResourceId(),
         destStorageData.storageContainerResource().getWorkspaceId());
     var pollResults =
-        blobPollers.map(blobPoller -> blobPoller.waitForCompletion(MAX_BLOB_COPY_POLL_TIMEOUT));
+        blobPollers.map(blobPoller -> blobPoller.waitForCompletion(SUBFLIGHT_TOTAL_DURATION));
```
aherbst-broad, line 78: ~Can you explain a bit more the motivation behind combining constants here? It seems like we'd want to allow these durations to vary independently.~ nm, I see that the AwaitCloneControlledAzureStorageContainerResourceFlightStep is polling on a subflight using this code, I think this makes sense.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/AwaitCloneAllResourcesFlightStep.java
aherbst-broad: 👍 an hour seems like a better idea, I'm a bit surprised it was 12 minutes.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
```diff
@@ -220,6 +220,11 @@ public static FlightState waitForJobFlightCompletion(Stairway stairway, String f
    */
   public static StepResult waitForSubflightCompletion(Stairway stairway, String flightId)
       throws InterruptedException {
+    return waitForSubflightCompletionAndReturnFlightInfo(stairway, flightId).getStepResult();
+  }
+
+  public static StepResultWithFlightInfo waitForSubflightCompletionAndReturnFlightInfo(
```
aherbst-broad, line 226: 🤔 so the tradeoff with the shift to this method is that we won't get the ability to customize the timeouts at the callsites and instead enforce a universal 60 minute timeout. Scanning through the changes I don't see any timeouts that will be lowered.

What do you think about getting a unit test around this code? I'm not sure we have any and it'd be good to get regression coverage on the retry and completion logic to avoid situations that caused this ticket (i.e., misinterpreting QUEUED as a terminal state).

aherbst-broad (commented):
Looking good, what do you think about a unit test for the waitForSubflight logic?

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
cahrens: If we wish to customize the timeouts, we could add the ability to pass in those parameters. I started to do that, but then decided we should just wait until we decide it's actually needed.

I'll see about adding some unit test coverage for the `waitForSubflightCompletionAndReturnFlightInfo ` utility method.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
```diff
@@ -276,9 +284,7 @@ public static boolean flightComplete(FlightState flightState) {
         "Testing flight {} completion; state is {}",
         flightState.getFlightId(),
         flightState.getFlightStatus());
-    return (flightState.getFlightStatus() == FlightStatus.ERROR
-        || flightState.getFlightStatus() == FlightStatus.FATAL
-        || flightState.getFlightStatus() == FlightStatus.SUCCESS);
+    return !flightState.isActive();
```
cahrens, line 287: Updated in the new version of Stairway (brought in through TCL change):

https://github.com/DataBiosphere/stairway/pull/137/files#r1540026406

service/src/main/java/bio/terra/workspace/common/utils/StepResultWithFlightInfo.java
```diff
@@ -0,0 +1,52 @@
+package bio.terra.workspace.common.utils;
```
marctalbott, line 1: Is there a case where this is used outside of the context of subflights? If not, it might be best to rename it to indicate that this class only makes sense in the context of subflights. Otherwise, I find the pairing of a step and a flight a bit confusing since it's a hierarchical relationship

marctalbott: Actually looking around a bit more, I'm not sure if we need this pairing at all. The steps that call `waitForSubflightCompletionAndReturnFlightInfo` tend to generate their own step results and just rely on the `flightState` included here, which makes it a bit confusing that they're dropping a `StepResult`. The steps that call `waitForSubflightCompletion` do use the step result, but that's all they use. 

if `waitForSubflightCompletionAndReturnFlightInfo` returned `FlightState`, `waitForSubflightCompletion` could then use that return value to determine the appropriate `StepResult` to return.

marctalbott (commented):
Generally looking good, just have some questions about the return types for the FlightUtils methods

service/src/main/java/bio/terra/workspace/common/utils/StepResultWithFlightInfo.java
cahrens: @marctalbott can you take a look at https://github.com/DataBiosphere/terra-workspace-manager/pull/1688/commits/4634c4e87af27af64e370c776baf22a9b0d0c3c4 and let me know what you think? I collapsed the code into a single method, but kept a return object class as it seemed like the cleanest way do the various data/error conversions. But I _think_ that now the relationship between the subflight `FlightState` and `StepStatus` is clearer.

marctalbott (approved):

service/src/main/java/bio/terra/workspace/common/utils/StepResultWithFlightInfo.java
marctalbott: I like it, thanks for refactoring that!

cahrens: Great! I'm going to work now on some unittest coverage.

service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java
cahrens: @aherbst-broad I have added a test and did some refactoring per Marc's suggestion (https://github.com/DataBiosphere/terra-workspace-manager/pull/1688/commits/d12eff57e3d8bb750c882d5e6abc250afd00b6fc). I'm going to test in a bee, but this is ready for review again.

aherbst-broad (approved):
Test looks good, thanks for adding it.

aherbst-broad (approved):

marctalbott (approved):

PR #1690, opened 2024-03-28
Use the WSM SA token when pulling a landing zone's region.

aherbst-broad:


okotsopoulos (approved):
This looks great -- I have a Friday brain and want my Monday brain to look at the testing a little more closely.  Or, if you can point me towards the tests that cover attempted retrievals failing when the caller is missing permissions, they are not jumping out in the diff view for me.

aherbst-broad:
@okotsopoulos good call out, we were lacking coverage there. I've added tests that assert we get a 403 if the user is lacking the requisite auth in [this](https://github.com/DataBiosphere/terra-workspace-manager/pull/1690/commits/1576da49012239e15f92628bc614305d44290a2d) commit.

sergiygetlin (approved):

PR #1691, opened 2024-03-28
[WOR-1425] Remove last vault dependency and shift to setup-gradle

aherbst-broad:
This PR removes the last vault dependency in the CI Github actions. 

It also shifts to using the `setup-gradle` action. 
Using the `setup-gradle` action instead of bare `gradle` setup steps brings a few [advantages](https://github.com/gradle/actions/tree/main/setup-gradle#why-use-the-setup-gradle-action) that I think are worthwhile:
* Build scans are now embedded in the workflow summary instead of having to dig through the log output
* Handles [caching](https://github.com/gradle/actions/tree/main/setup-gradle#caching-build-state-between-jobs) and allows us to move away from actions/cache


.github/actions/start-local-server/action.yml
```diff
@@ -27,7 +27,7 @@ runs:
         # Give it some time before starting to poll.
         sleep 10
         started=1
-        for i in `seq 1 120`;
+        for i in `seq 1 240`;
```
aherbst-broad, line 30: I had to bump this out in my testing; I suspect it's taking longer than usual here due to cache misses. The cache should be populated on merge to main.

.github/workflows/full-service-tests.yml
```diff
@@ -7,12 +7,6 @@ on:
   schedule:
     - cron: '0 8 * * *' # Run nightly at 3AM ET
 
-  workflow_dispatch:
-    inputs:
-      testEnv:
```
aherbst-broad, line 12: I found no evidence of this workflow running in anything but a 'local' context.

.github/workflows/full-service-tests.yml
```diff
@@ -32,59 +26,47 @@ jobs:
...
+        user-delegated-sa-b64: ${{ secrets.USER_DELEGATED_SA_DEV }}
+        buffer-client-sa-b64: ${{ secrets.BUFFER_CLIENT_SA_DEV }}
+        testrunner-sa-b64: ${{ secrets.TESTRUNNER_SA_DEV }}
+        testrunner-k8s-sa-b64: ${{ secrets.TESTRUNNER_K8S_SA_DEV }}
+        wsm-sa-b64: ${{ secrets.WSM_SA_DEV }}
+        janitor-sa-b64: ${{ secrets.JANITOR_SA_DEV }}
+        policy-client-sa-b64: ${{ secrets.POLICY_CLIENT_SA_DEV }}
+
+      # Turn on TPS for integration tests
+    - name: Write config
```
cahrens, line 51: Where was TPS enabled previously?

aherbst-broad: Via the `write-config.sh` script invocation in the now dead `write-config` action. It might be worth it to universally enable TPS via a config change as I don't think there are any envs that don't have it enabled any more.

cahrens (approved):

marctalbott (approved):

PR #1698, opened 2024-04-03
Proof-of-concept port of `ApplicationAbleFlight` using annotations

pshapiro4broad:
This POC shows how using `StepInput` and `StepOutput` annotations could simplify stairway flight and steps.

If you're not familiar with this code, compare these versions with the code in `ApplicationAbleDaoStep`, `ApplicationAbleIamStep`, etc.

These annotations support a form of static data flow analysis. I added a test that generates this output to show what information the annotations provide. Note that without running a step you can't tell which outputs are actually written to by the code, so this prints out all outputs as being potential step outputs.

Test output:
```
StepData[stepName=ApplicationAblePrecheckStepV3, inputs=[authUserInfo, workspaceId, applicationAbleEnum], outputs=[wsmApplication, applicationAbleDao, applicationAbleSam, response, statusCode]]
StepData[stepName=ApplicationAbleIamStepV3, inputs=[authUserInfo, workspaceId, applicationAbleEnum, wsmApplication, applicationAbleSam], outputs=[response, statusCode]]
StepData[stepName=ApplicationAbleDaoStepV3, inputs=[workspaceId, applicationAbleEnum, applicationAbleDao], outputs=[response, statusCode]]
```

### DIFF
--- a/service/src/main/java/bio/terra/workspace/common/stairway/StepUtils.java
+++ b/service/src/main/java/bio/terra/workspace/common/stairway/StepUtils.java
@@ -1,0 +1,119 @@
+package bio.terra.workspace.common.stairway;
+
+import bio.terra.stairway.Flight;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+
+// Suppress sonar warnings about reflection API usage. Reflection APIs must be used to read and
+// write fields in a Step.
+@SuppressWarnings({"java:S3011"})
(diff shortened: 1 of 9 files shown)


service/src/main/java/bio/terra/workspace/common/stairway/StepUtils.java
```diff
@@ -0,0 +1,119 @@
...
+
+  public static class IllegalGetException extends RuntimeException {
+    public IllegalGetException(Throwable cause) {
+      super(cause);
+    }
+  }
+
+  private StepUtils() {}
+
+  public static String keyFromField(Field field) {
```
pshapiro4broad, line 38: By default, the inputs and outputs are mapped to flight map (or flight input map) elements by using the annotated field's name. You can also provide an annotation argument to use a different name instead.

PR #1699, opened 2024-04-04
[WOR-1546] Ensure cloud context creation happens as part of the synchronous clone workspace creation operation

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1546

I transitioned the create and clone workspace methods to use the `CreateWorkspaceV2Flight`, which has the ability to create the cloud context within the flight (as a subflight). I deleted the v1 workspace creation flight.

Tested in a bee. There are also some connected tests (both Azure and GCP) of the `cloneWorkspace` method.


service/build.gradle
```diff
@@ -51,7 +51,7 @@ jacoco {
 
 sonar {
     properties {
-        property 'sonar.projectName', 'terra-workspace-manager'
+        property 'sonar.projectName', "${rootProject.name}_${projectDir.name}"
```
cahrens, line 54: I'm not sure if this is helping, though I did get a coverage report with 98% coverage after I made this change (which was higher than I had gotten previously). Unfortunately it still seems that only the last report shows on the PR, when really we would want to show one per project.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -716,6 +717,21 @@ public String cloneWorkspace(
         cloningInstructions,
         userRequest);
 
+    // Create the cloud context synchronously so that we can display a shell workspace.
```
marctalbott, line 720: The `CreateWorkspaceV2Flight` creates the cloud context as part of the flight. Would it make sense to move the `createWorkspaceForClone` call above over to the v2 flight instead?

cahrens: Possibly… let me go through them both carefully and make sure it wouldn't change any behavior. I would need to determine the spend profile to use before calling the create workspace flight.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneCreateCloudContextFlight.java
```diff
@@ -0,0 +1,107 @@
...
+
+    // Only create a cloud context if the source workspace has a cloud context
+    Optional<GcpCloudContext> gcpCloudContextOptional =
+        flightBeanBag.getGcpCloudContextService().getGcpCloudContext(sourceWorkspaceId);
+    if (gcpCloudContextOptional.isPresent()) {
+      SpendProfile gcpSpendProfile =
+          getSpendProfile(spendProfile, gcpCloudContextOptional.get(), flightBeanBag, userRequest);
+
+      addStep(
+          new LaunchCreateCloudContextFlightStep(
```
aherbst-broad, line 52: 🤔 Probably for the future, but I want to note: I am unsure why we're spawning a subflight for the creation of a cloud context. At base, we're validating the cloud context info (policy compatibility, cloud availability) which seem like lightweight operations. 

Given the drawbacks of subflights we've encountered, it worth considering refactoring these into plain steps and pulling them in here.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneCreateCloudContextFlight.java
```diff
@@ -0,0 +1,107 @@
...
+              flightBeanBag.getWorkspaceService(),
+              CloudPlatform.GCP,
+              gcpSpendProfile,
+              ControlledResourceKeys.CREATE_GCP_CLOUD_CONTEXT_FLIGHT_ID),
+          cloudRetryRule);
+      addStep(
+          new AwaitCreateCloudContextFlightStep(
+              ControlledResourceKeys.CREATE_GCP_CLOUD_CONTEXT_FLIGHT_ID),
+          longCloudRetryRule);
```
aherbst-broad, line 61: Curious why this is the long cloud retry rule vs the "regular" one for Azure on L80.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneCreateCloudContextFlight.java
```diff
@@ -0,0 +1,107 @@
...
+          cloudRetryRule);
+      addStep(
+          new AwaitCreateCloudContextFlightStep(
+              ControlledResourceKeys.CREATE_AZURE_CLOUD_CONTEXT_FLIGHT_ID),
+          cloudRetryRule);
+    }
+  }
+
+  /*
+   * There is an API mismatch in the V1 APIs. If a new spend profile is passed with
```
aherbst-broad, line 85: Another future looking note: it may be time to start making stronger assertions about having a single CC in a workspace. Technically it's possible to have more than one but in practice that is a very unsupported and likely to lead to strange bugs.

service/src/test/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneCreateCloudContextFlightTest.java
```diff
@@ -0,0 +1,131 @@
...
+                CreateCloudContextIdsForFutureStepsStep.class,
+                LaunchCreateCloudContextFlightStep.class,
+                AwaitCreateCloudContextFlightStep.class)));
+  }
+
+  @Test
+  void testUsesSourceSpendProfileIfNoDestinationSpendProfile() {
+    when(flightBeanBag.getSpendProfileService()).thenReturn(spendProfileService);
+    FeatureConfiguration config = new FeatureConfiguration();
+    // This is a misnomer... it is not specific to GCP (and in reality, always on).
```
aherbst-broad, line 90: Good call out -- can you drop a ticket to remove this flag? I don't think it's needed anymore.

aherbst-broad (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneCreateCloudContextFlight.java
```diff
@@ -0,0 +1,107 @@
...
+  private SpendProfile getSpendProfile(
+      SpendProfile spendProfile,
+      CloudContext cloudContext,
+      FlightBeanBag flightBeanBag,
+      AuthenticatedUserRequest userRequest) {
+    if (spendProfile != null) {
+      return spendProfile;
+    }
+
+    var spendProfileService = flightBeanBag.getSpendProfileService();
```
cahrens, line 100: @marctalbott It looks like all this code is no longer necessary because you introduced it further upstream, [here](https://github.com/DataBiosphere/terra-workspace-manager/blob/cfdaa8e366001b33b2cfa0a8c63710f1fd8e37af/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java#L623).

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -232,36 +275,23 @@ private UUID createWorkspaceWorker(
...
-            .addParameter(
-                WorkspaceFlightMapKeys.WORKSPACE_STAGE, workspace.getWorkspaceStage().name())
-            .addParameter(WorkspaceFlightMapKeys.POLICIES, policies)
-            .addParameter(WorkspaceFlightMapKeys.APPLICATION_IDS, applications)
-            .addParameter(
-                WorkspaceFlightMapKeys.ResourceKeys.CLONING_INSTRUCTIONS, cloningInstructions)
-            .addParameter(SOURCE_WORKSPACE_ID, sourceWorkspaceUuid)
-            .addParameter(WorkspaceFlightMapKeys.PROJECT_OWNER_GROUP_ID, projectOwnerGroupId);
-
-    workspace
```
cahrens, line 258: SPEND_PROFILE_ID is no longer referenced anywhere (we pass the actual spend profile if we have it).

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/CheckSpendProfileStep.java
```diff
@@ -1,73 +0,0 @@
-package bio.terra.workspace.service.workspace.flight.create.workspace;
-
-import static bio.terra.workspace.service.workspace.flight.WorkspaceFlightMapKeys.SPEND_PROFILE;
```
cahrens, line 3: I noticed that this is unused.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
```diff
@@ -145,16 +146,31 @@ public UUID createWorkspaceForClone(
       @Nullable TpsPolicyInputs policies,
       @Nullable List<String> applications,
       @Nullable String projectOwnerGroupId,
+      @Nullable SpendProfile spendProfile,
       UUID sourceWorkspaceId,
       CloningInstructions cloningInstructions,
       AuthenticatedUserRequest userRequest) {
+
+    // If there is a cloud platform associated with the source workspace, get it here.
+    Optional<DbCloudContext> optionalAzureCloudContext =
```
cahrens, line 155: This is what was going on in the old [CloneWorkspace](https://github.com/DataBiosphere/terra-workspace-manager/pull/1699/files#diff-461cf5a8cc6b98c856722e9336dd65621b9e0dbf776b15d3d419bcbeca17db64L62) flight to determine the cloud context to create.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneWorkspaceFlight.java
```diff
@@ -58,68 +32,7 @@ public CloneWorkspaceFlight(FlightMap inputParameters, Object applicationContext
 
     addStep(new CreateIdsForFutureStepsStep());
 
-    // Only create a cloud context if the source workspace has a cloud context
-    Optional<GcpCloudContext> gcpCloudContextOptional =
```
cahrens, line 62: Cloud context creation now happens before this flight is called, when we create the destination workspace.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneWorkspaceFlight.java
```diff
@@ -58,68 +32,7 @@ public CloneWorkspaceFlight(FlightMap inputParameters, Object applicationContext
...
-      CloudContext cloudContext,
-      FlightBeanBag flightBeanBag,
-      AuthenticatedUserRequest userRequest) {
-    if (spendProfile != null) {
-      return spendProfile;
-    }
-
-    var spendProfileService = flightBeanBag.getSpendProfileService();
-    var features = flightBeanBag.getFeatureConfiguration();
-    return spendProfileService.authorizeLinking(
```
cahrens, line 122: This checking was actually also happening [earlier](https://github.com/DataBiosphere/terra-workspace-manager/blob/cfdaa8e366001b33b2cfa0a8c63710f1fd8e37af/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java#L623) in the code execution path.

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/CreateWorkspaceV2Flight.java
```diff
@@ -57,30 +69,40 @@ public CreateWorkspaceV2Flight(FlightMap inputParameters, Object applicationCont
           new CreateWorkspacePoliciesStep(
               workspace, policyInputs, appContext.getTpsApiDispatch(), userRequest),
           serviceRetryRule);
+
+      addStep(
```
cahrens, line 73: I moved this up earlier, to match the original `[WorkspaceCreateFlight](https://github.com/DataBiosphere/terra-workspace-manager/blob/47dcbd5918e76349165a16230ec9369cf0de9819/service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/WorkspaceCreateFlight.java#L65)`. It's not clear to me that this should ONLY be done for a MC_WORKSPACE.

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/CreateWorkspaceV2Flight.java
```diff
@@ -57,30 +69,40 @@ public CreateWorkspaceV2Flight(FlightMap inputParameters, Object applicationCont
...
+          new LinkSpendProfilePolicyAttributesStep(
+              workspace.workspaceId(), workspace.spendProfileId(), appContext.getTpsApiDispatch()),
+          serviceRetryRule);
+
+      // If we're cloning, we need to copy the policies from the source workspace.
```
cahrens, line 78: This code was in the original [WorkspaceCreateFlight](https://github.com/DataBiosphere/terra-workspace-manager/blob/47dcbd5918e76349165a16230ec9369cf0de9819/service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/WorkspaceCreateFlight.java#L74) and is needed now that this flight is also used for cloning. There was test coverage that failed when I did not have this in the right place.

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/CreateWorkspaceV2Flight.java
```diff
@@ -57,30 +69,40 @@ public CreateWorkspaceV2Flight(FlightMap inputParameters, Object applicationCont
...
     }
 
     // Workspace authz is handled differently depending on whether WSM owns the underlying Sam
     // resource or not, as indicated by the workspace stage enum.
     switch (workspace.getWorkspaceStage()) {
-      case MC_WORKSPACE -> {
-        if (appContext.getFeatureConfiguration().isTpsEnabled()) {
+      case MC_WORKSPACE ->
           addStep(
-              new LinkSpendProfilePolicyAttributesStep(
```
cahrens, line 68: Moved this up out of the switch, as noted above.

service/src/test/java/bio/terra/workspace/service/workspace/GcpCloudContextConnectedTest.java
```diff
@@ -331,114 +306,4 @@ public void cloneGcpWorkspace() throws Exception {
             .listFoldersInWorkspace(destinationWorkspace.getWorkspaceId())
             .contains(sourceFolder));
   }
-
-  @Test
-  @Disabled("PF-2259 - this test needs rewriting")
```
cahrens, line 336: This test was disabled, and it was the only remaining places calling `LaunchCreateCloudContextFlightStep` and `AwaitCreateCloudContextFlightStep`, so I'm deleting it.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneCreateCloudContextFlight.java
cahrens: I have switched to using the v2 workspace creation flight. It still uses a subflight, reusing`createCloudContext`, from WorkspaceService. At this point, I would rather just stay with the subflights here, and tackle removing subflights if necessary later.

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/CreateWorkspaceV2Flight.java
cahrens: @marctalbott @aherbst-broad I'm particularly interested on your thoughts here-- if this should stay within the `MC_WORKSPACE` case. A RAWLS stage workspace won't have a spendProfileId, will it?

aherbst-broad: Rawls workspaces don't have spend profiles or cloud contexts. If the step is robust to the spend profile ID and cloud context not being present it's probably OK, but it'd be best to avoid that check altogether to avoid spamming as you propose. 

I am unsure about the OOO of linking spend profile attributes before the source workspace attributes. My instinct is that it doesn't matter -- if they are incompatible the merge step will fail out. We do want to ensure that rawls workspaces are having their policies merged from the source workspace, but the OOO doesn't matter in that case since they don't have a spend profile.

service/src/main/java/bio/terra/workspace/common/logging/model/ActivityFlight.java
```diff
@@ -33,8 +32,6 @@
  */
 public enum ActivityFlight {
   // Workspace
-  WORKSPACE_CREATE_FLIGHT(
```
aherbst-broad, line 36: 🤔 Will this break any in progress flights at the time of rollout?

aherbst-broad (commented):
Had a question about backwards compatibility.

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CreateIdsForFutureStepsStep.java
```diff
@@ -18,21 +18,13 @@ public class CreateIdsForFutureStepsStep implements Step {
   @Override
   public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
     FlightMap workingMap = context.getWorkingMap();
-    workingMap.put(
-        ControlledResourceKeys.CREATE_GCP_CLOUD_CONTEXT_FLIGHT_ID,
```
cahrens, line 22: I will want to remove these keys from `ControlledResourceKeys`… will do so in the cleanup PR.

service/src/main/java/bio/terra/workspace/service/workspace/flight/create/workspace/CreateWorkspaceV2Flight.java
cahrens: Added the "if" check, like we had in the original `WorkspaceCreateFlight`.

service/src/main/java/bio/terra/workspace/common/logging/model/ActivityFlight.java
cahrens: As discussed in standup, I have pulled out the flight deletion and will put it in a separate PR after this one gets to production.

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
marctalbott: This would require further refactoring to address, but this is a change in behavior in that the old CloneWorkspace flight could clone an Azure cloud context and a GCP cloud context if a workspace had both. This will prioritize the Azure cloud context over the GCP cloud context. In practice, we only have workspaces with a single cloud context, but WSM does generally permit multiple cloud contexts.

I don't feel particularly strongly about continuing to support multiple cloud contexts in a workspace, but it might be worth looking at how difficult it would be to change this to continue to support it until we decide we officially want to drop support

cahrens: I’m not inclined to support multiple cloud contexts. We don’t use the functionality, and the new workspace creation flight didn’t support creating multiple clouds.

I could log here if there happen to be 2, but in reality we know we won’t hit that case. 

Note that I did not choose to support AWS, as it is my understanding that we will remove the AWS code.

marctalbott: Fair enough, just wanted to make sure we were doing this intentionally

marctalbott (approved):

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
cahrens: I'll create a ticket for us to enforce a single cloud context per workspace (@aherbst-broad had also mentioned this).

service/src/main/java/bio/terra/workspace/service/resource/controlled/flight/clone/workspace/CloneCreateCloudContextFlight.java
cahrens: Created https://broadworkbench.atlassian.net/browse/WOR-1626

service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java
cahrens: Created https://broadworkbench.atlassian.net/browse/WOR-1626

PR #1709, opened 2024-04-16
[WOR-1127] Extract an interface for landing zone service operations (step 1)

aherbst-broad:
## Context
We want to be able to replace our calls to the "amalgamated" landing zone library with HTTP calls to a live landing zone service API. This is the first in a series of PRs that will allows us to switch between using the amalgamated library and live calls. The size of this change is pushing up on what my limit for what's a digestible change, so I'm PRing this first step.

## This PR
* Extracts a [common interface](https://github.com/DataBiosphere/terra-workspace-manager/pull/1709/files#diff-b94670fab36c36a4570005adfc7170b0589b5320871e9311ba466358baffcaf6) that abstracts operations against landing zones
  * The sole implementation in this PR uses the existing amalgamated library calls
  * The interface consumes (mostly) Java primitives for its parameters rather than amalgamated or api client library types. It returns WSM Api* types
  * The LandingZoneApiDispatch class has been [hardcoded](https://github.com/DataBiosphere/terra-workspace-manager/pull/1709/files#diff-dd0817ce027b40a0fe2bdf92e6984cc3763a1aee40a84e856ae348c83e612493R53) to use an instance of the amalgamated implementation. A subsequent PR will introduce the ability to switch to an HTTP Client API version via configuration.
* Introduces an adapter class that is focused on the conversion between types (i.e., all the `toApiFooBar` code)
   * It is limited to conversion between LZ library internal types <-->  and WSM API types at the moment. I foresee it also handling the conversion between LZ API Client types <--> WSM API types in a subsequent PR.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/AmalgamatedLandingZoneService.java
+++ b/service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/AmalgamatedLandingZoneService.java
@@ -1,6 +1,5 @@
 package bio.terra.workspace.amalgam.landingzone.azure;
 
-import bio.terra.common.exception.ConflictException;
 import bio.terra.common.iam.BearerToken;
 import bio.terra.landingzone.library.landingzones.deployment.LandingZonePurpose;
 import bio.terra.landingzone.service.landingzone.azure.LandingZoneService;
@@ -101,20 +100,9 @@ public class AmalgamatedLandingZoneService implements WorkspaceLandingZoneServic
     ApiAzureLandingZoneList result = new ApiAzureLandingZoneList();
     List<LandingZone> landingZones =
         landingZoneService.getLandingZonesByBillingProfile(bearerToken, billingProfileId);
-    if (!landingZones.isEmpty()) {
-      // The enforced logic is 1:1 relation between Billing Profile and a Landing Zone.
-      // The landing zone service returns one record in the list if landing zone exists
-      // for a given billing profile.
-      if (landingZones.size() == 1) {
-        result.addLandingzonesItem(typeAdapter.toApiAzureLandingZone(landingZones.get(0)));
-      } else {
-        throw new ConflictException(
-            String.format(
-                "There are more than one landing zone found for the given billing profile: '%s'. Please"
-                    + " check the landing zone deployment is correct.",
-                billingProfileId));
-      }
-    }
+    landingZones.forEach(
+        landingZone -> result.addLandingzonesItem(typeAdapter.toApiAzureLandingZone(landingZone)));
+
     return result;
   }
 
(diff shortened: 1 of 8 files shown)


service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/AmalgamatedLandingZoneService.java
```diff
@@ -0,0 +1,192 @@
...
+    return typeAdapter.toApiAzureLandingZone(landingZoneRecord);
+  }
+
+  @Override
+  public ApiAzureLandingZoneList listLandingZonesByBillingProfile(
+      BearerToken bearerToken, UUID billingProfileId) {
+    ApiAzureLandingZoneList result = new ApiAzureLandingZoneList();
+    List<LandingZone> landingZones =
+        landingZoneService.getLandingZonesByBillingProfile(bearerToken, billingProfileId);
+    if (!landingZones.isEmpty()) {
```
blakery, line 104: Why enforce this here?
If LZS is returning a list, and this is returning a list, why not just do a straight map operation?
And if it _is_ enforced, why return a list at all?

Maybe the correct question to ask is "why is LZS returning a list here?"

Either way, I'm not sure it's correct for WSM to be enforcing LZS constraints.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/LandingZoneApiDispatch.java
```diff
@@ -85,27 +66,25 @@ public ApiCreateLandingZoneResult createAzureLandingZone(
...
-            .definition(body.getDefinition())
-            .version(body.getVersion())
-            .parameters(
-                MapperUtils.LandingZoneMapper.landingZoneParametersFrom(body.getParameters()))
-            .billingProfileId(body.getBillingProfileId())
-            .build();
-    return toApiCreateLandingZoneResult(
-        landingZoneService.startLandingZoneCreationJob(
-            bearerToken, body.getJobControl().getId(), landingZoneRequest, asyncResultEndpoint));
+    return amalgamated.startLandingZoneCreationJob(
```
blakery, line 69: Why not start using the `WorkspaceLandingZoneService` interface type now?

It will always be `AmalgamatedLandingZoneService` until the other implementation is created and configured, but then this class shouldn't need further adaptation, unless the interface changes.

aherbst-broad: Good call, I'll make the change.

service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/AmalgamatedLandingZoneService.java
aherbst-broad: True. The LZS actually verifies this itself [here](https://github.com/DataBiosphere/terra-landing-zone-service/blob/ceafbf7d5bf25554dafbc25c1f4f282edd8eba4b/service/src/main/java/bio/terra/lz/futureservice/app/service/LandingZoneAppService.java#L67), so this logic is redundant.  Removed.

blakery (approved):

PR #1710, opened 2024-04-17
Wire up HTTP LZS client (step 2)

aherbst-broad:
## Context
This is the followup to #1709.
We want to be able to switch over to using the new Landing Zone web service when we are ready. The previous PR (linked above) introduced a new interface that abstracted over the direct library calls ("amalgamated"). The next step is to introduce an implementation of that interface for the LZS HTTP API and the needed configuration that will allows switching to that API when ready.

## This PR
* Introduces an `HttpLandingZoneService` implementation of the `WorkspaceLandingZoneService` interface. This is a "vanilla" WSM HTTP client layer, with the requisite `LzsRetry` class and internal exception hierarchy.
* Adds new type marshaling methods to the `LandingZoneApiTypeAdapter` to help map between Landing Zone Service API client types and the needed WSM API types for `LandingZoneApiDispatch`
* Adds a test for `LandingZoneApiTypeAdapter`
* I have not implemented a unit test for the `HttpLandingZoneService` class. My plan is to introduce coverage for that code via pact tests as described in WOR-1119.


service/src/main/java/bio/terra/workspace/amalgam/landingzone/azure/HttpLandingZoneService.java
```diff
@@ -0,0 +1,265 @@
...
+    } catch (ApiException e) {
+      throw convertApiException(e);
+    }
+
+    return result;
+  }
+
+  @Override
+  public ApiAzureLandingZoneList listLandingZones(BearerToken bearerToken) {
+    throw new RuntimeException("todo");
```
aherbst-broad, line 154: This is problematic, it looks like Leo is using this endpoint for some reason. I'm investigating a resolution.

sergiygetlin (approved):

blakery (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/LzsRetry.java
```diff
@@ -0,0 +1,153 @@
...
+      } catch (ProcessingException ws) {
+        logger.info("LzsRetry: caught retry-able ProcessingException: ", ws);
+        sleepOrTimeoutBeforeRetrying(ws);
+      }
+    }
+  }
+
+  private boolean isRetryable(ApiException apiException) {
+    return isTimeoutException(apiException)
+        || apiException.getCode() == HttpStatus.SC_FORBIDDEN
```
aherbst-broad, line 113: This is problematic in my testing; lzs returns an (expected?) 403 during the poll for a landing zone during creation, which causes gateway timeouts as this goes into an extensive retry. I will need to think about this a bit.

marctalbott: Should 403 be a retry-able status code? I guess if we're seeing intermittent 403s that eventually go away then we might need it to be, but I'm generally not inclined to retry 403s as long as we trust LZS and Sam to return 403s correctly.

marctalbott (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/LzsRetry.java
aherbst-broad: Yeah, I think I agree. This was originally installed in WSM for TPS calls as we still get occasional spurious 403s from TPS from the oidc proxy. Might not be worth it in this case.  (see also my investigations into unknown error -102 AH01095)

PR #1714, opened 2024-04-23
[WOR-1119] Pact tests for LZS 

aherbst-broad:
## Context
As part of the extraction of landing zone code to a dedicated service, we need to assert the contracts between WSM and the LZS. 

## This PR
* Implements consumer pact tests for the WSM -> LZS relationship
   * These tests are split between a "Jobs" test suite that is concerned with the LZ job creation and polling APIs and a "Read" test suite that is concerned with reading LZ data


service/src/test/java/bio/terra/workspace/pact/LzApiTest.java
```diff
@@ -0,0 +1,612 @@
...
+  static final UUID landingZoneId = UUID.fromString("a051c5f8-9de8-4b4f-a453-8f9645df9c7b");
+  static final UUID asyncJobId = UUID.fromString("3b0db775-1916-42d2-aaeb-ebc7aea69eca");
+  static final UUID billingProfileId = UUID.fromString("4955757a-b027-4b6c-b77b-4cba899864cd");
+  static final UUID azureResourceId = UUID.fromString("e3327645-1e8f-454d-bdae-fe06b4762542");
+
+  static DslPart jobReportShape =
+      new PactDslJsonBody()
+          .stringType("id")
+          .stringType("description")
+          .stringMatcher("status", "RUNNING|SUCCEEDED|FAILED")
```
blakery, line 59: Normally, I would say they should be looser, and only declare what the service actually uses, but since WSM is basically acting as a proxy for LZS, defining all the fields makes sense.

If any fields are optional though, we should write separate validations for when they are present, and omit them otherwise.

For example, `completed` won't be present in jobs that are still running.
So for the default case, we should omit it.
We still want to validate the correct type of `completed`, so for that we should write a test that checks for it in a case where it's required to be present.

service/src/test/java/bio/terra/workspace/pact/LzApiTest.java
```diff
@@ -0,0 +1,612 @@
...
+
+  @Pact(consumer = "workspacemanager", provider = "lzs")
+  public RequestResponsePact startCreateLandingZone(PactDslWithProvider builder) {
+    var createResponseShape =
+        new PactDslJsonBody()
+            .uuid("landingZoneId")
+            .stringType("definition")
+            .stringType("version")
+            .object("jobReport", jobReportShape)
+            .object("errorReport", errorReportShape);
```
blakery, line 112: Same thing as above, about optional fields.
I wouldn't expect the `errorReport` to be present in the initial create response - so it shouldn't be here.

service/src/test/java/bio/terra/workspace/pact/LzApiTest.java
```diff
@@ -0,0 +1,612 @@
...
+  }
+
+  @Pact(consumer = "workspacemanager", provider = "lzs")
+  public RequestResponsePact createLandingZoneResult_notAuthorized(PactDslWithProvider builder) {
+    return builder
+        .given("An existing landing zone creation job")
+        .uponReceiving("A request to get the landing zone creation job result")
+        .method("GET")
+        .pathFromProviderState(
+            "/api/landingzones/v1/azure//create-result/${asyncJobId}",
```
blakery, line 174: Good use of provider state.

service/src/test/java/bio/terra/workspace/pact/LzApiTest.java
```diff
@@ -0,0 +1,612 @@
...
+  @Test
+  @PactTestFor(pactMethod = "createLandingZone_notAuthorized", pactVersion = PactSpecVersion.V3)
+  void startCreateLandignZone_notAuthorized() {
+    assertThrows(
+        LandingZoneServiceAuthorizationException.class,
+        () ->
+            landingZoneService.startLandingZoneCreationJob(
+                new BearerToken("fake"),
+                "jobId",
+                UUID.randomUUID(),
```
blakery, line 425: We should avoid using random UUIDs here.

The value will end up in the generated contract, which will change on every run, and cause the broker to re-run validation every time it's pushed, since it looks like the contract changed.

service/src/test/java/bio/terra/workspace/pact/LzApiTest.java
aherbst-broad: Good point; I've taken a whack at that [here](https://github.com/DataBiosphere/terra-workspace-manager/pull/1714/files#diff-ed7c73cdb1082bca22090bc2b30b66a80fe68e3f9f02179ff4e2b10975ffa5c9R45-R76), let me know what you think.

marctalbott (approved):

service/src/test/java/bio/terra/workspace/pact/landingzones/LzReadApiTest.java
```diff
@@ -0,0 +1,321 @@
...
+    var result =
+        landingZoneService.listLandingZonesByBillingProfile(
+            new BearerToken("fake"), BILLING_PROFILE_ID);
+
+    assertNotNull(result);
+    assertThat(result.getLandingzones(), is(not(empty())));
+  }
+
+  @Test
+  @PactTestFor(pactMethod = "listLandingZones", pactVersion = PactSpecVersion.V3)
```
blakery, line 227: One thing that some people have found to improve readability, is to locate the `@Pact` method next to the corresponding test(s).

I don't think it's necessary or anything like that, just something you might prefer.

service/src/test/java/bio/terra/workspace/pact/landingzones/LzReadApiTest.java
```diff
@@ -0,0 +1,321 @@
...
+            .eachLike("landingzones")
+            .uuid("landingZoneId")
+            .uuid("billingProfileId")
+            .stringType("definition")
+            .stringType("version")
+            .stringType("region")
+            .date("createdDate");
+
+    return builder
+        .uponReceiving("A request to list all landing zones")
```
blakery, line 145: Do we need some sort of state here?

aherbst-broad: @blakery I don't follow, what kind of state?

blakery (approved):

PR #1717, opened 2024-04-28
[DDO-3622] Use new IAP and Sherlock URL

jack-r-warren:
Unfortunately another instance of this workflow failing (it was also impacted by the Sherlock v3 migration not too long ago) -- this seems to be one of the only IAP/Sherlock-accessors that DevOps doesn't have direct control over. My bad for not preparing a PR here in advance, I always forget that this exists until I'm trying to clean-up after off-hours maintenance. (It seems this repo's automation has automatically @'ed people for review off-hours -- sorry!)

I had a successful run [here](https://github.com/DataBiosphere/terra-workspace-manager/actions/runs/8871845520/job/24355698223) (at least, it got past the IAP change). The first attempt did fail, I tweaked config on our end and re-ran.


aherbst-broad (approved):
Thanks for catching this

PR #1719, opened 2024-04-30
IA-4732: Add support for spot Azure VMs

rtitle:
https://broadworkbench.atlassian.net/browse/IA-4732

Add support for Spot Azure VMs in WSM. This is to support Hail Batch, which makes use of Spot VMs. 

Introduced a new optional field `priority` with possible values `REGULAR` and `SPOT`. Defaults to `REGULAR` if not specified. I moduled this after the [Azure API](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?view=rest-compute-2024-03-01&tabs=HTTP#virtualmachineprioritytypes), but let me know if there are alternative suggestions. (Note Azure also includes a `Low` priority which is deprecated in favor of Spot).

Tested in a BEE by manually creating a Spot VM, and also running a Hail pipeline.

### DIFF
--- a/openapi/src/parts/controlled_azure_vm.yaml
+++ b/openapi/src/parts/controlled_azure_vm.yaml
@@ -61,8 +61,8 @@ paths:
           $ref: '#/components/responses/ServerError'
     post:
       summary: |
-        Delete a controlled Azure VM resource. This is async, because the 
-        operation can take a few minutes. OpenAPI does not support request body 
+        Delete a controlled Azure VM resource. This is async, because the
+        operation can take a few minutes. OpenAPI does not support request body
         in DELETE, but async state requires it. Hence this is a POST.
       operationId: deleteAzureVm
       tags: [ ControlledAzureResource ]
@@ -140,6 +140,10 @@ components:
           description: Supply base64 encoded custom data for vm. https://docs.microsoft.com/en-us/azure/virtual-machines/custom-data
           type: string
           format: base64
+        priority:
+          description: Priority of the VM (Regular or Spot). Defaults to Regular.
+          type: string
+          enum: [ 'REGULAR', 'SPOT' ]
 
     AzureVmImage:
       description: >-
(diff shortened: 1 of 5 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/CreateAzureVmStep.java
```diff
@@ -267,4 +269,21 @@ private VirtualMachine.DefinitionStages.WithFromImageCreateOptionsManaged addIma
...
+
+  private VirtualMachine.DefinitionStages.WithCreate addSizeAndPriorityStep(
+      VirtualMachine.DefinitionStages.WithManagedCreate priorSteps,
+      ApiAzureVmCreationParameters creationParameters) {
+    var priority =
+        Optional.ofNullable(creationParameters.getPriority())
+            .map(
+                creationPriority ->
+                    switch (creationPriority) {
+                      case SPOT -> VirtualMachinePriorityTypes.SPOT;
```
cahrens, line 281: Can you add unit test coverage of this?

rtitle: Will do 👍

rtitle: Added unit tests.

cahrens (approved):

openapi/src/parts/controlled_azure_vm.yaml
```diff
@@ -140,6 +140,10 @@ components:
           description: Supply base64 encoded custom data for vm. https://docs.microsoft.com/en-us/azure/virtual-machines/custom-data
           type: string
           format: base64
+        priority:
+          description: Priority of the VM (Regular or Spot). Defaults to Regular.
+          type: string
+          enum: [ 'REGULAR', 'SPOT' ]
```
dvoet, line 146: What happens to the WSM resource when a spot instance disappears?

rtitle: In WSM: nothing. The resource continues to exist.

We are using the default `deallocate` eviction policy though, which means the Azure VM does not completely disappear: it moves into a stopped-deallocated state, and can be redeployed (or deleted) later. So there is still an Azure resource even if eviction happens.

aherbst-broad (approved):

PR #1723, opened 2024-05-02
add cache for sam actions

marctalbott:


service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -487,14 +493,24 @@ public boolean isAuthorized(
...
+    return samPermissionsCache.computeIfAbsent(
+        key,
+        v -> {
+          try {
+            return SamRetry.retry(
+                () -> resourceApi.resourcePermissionV2(iamResourceType, resourceId, action));
+          } catch (ApiException apiException) {
+            throw SamExceptionFactory.create(
+                "Error checking resource permission in Sam", apiException);
+          } catch (InterruptedException e) {
```
marctalbott, line 510: I'm a bit worried about this no longer being an `InterruptedException` since there are places where we catch `InterruptedException` upstream and rethrow them (see [Rethrow](https://github.com/DataBiosphere/terra-workspace-manager/blob/f4e51dd768a8d3277ae6e62540cb971c09fb5492/service/src/main/java/bio/terra/workspace/common/utils/Rethrow.java), but I couldn't figure out how to make it stop complaining about uncaught exceptions without it

aherbst-broad: Yeah this is a bit concerning; I believe the need for Interruptedexception is so stairway can catch it. I'm wondering if we can dumb this down a bit and change the logic to something like:

* check for the key in the map -- if present, return it
* if not present, do the try { SamRetry () } business
* Put the result of ^^ in the map
* Return the result
 
That way we preserve the exception semantics.

aherbst-broad (approved):
lgtm, had a comment about preserving the interruptedexception semantics.

service/src/main/java/bio/terra/workspace/app/configuration/external/SamConfiguration.java
```diff
@@ -11,11 +12,21 @@ public class SamConfiguration {
...
+
   public String getBasePath() {
     return basePath;
   }
 
   public void setBasePath(String basePath) {
     this.basePath = basePath;
   }
+
+  public Duration getPermissionsCacheTtl() {
```
cahrens, line 25: What does Ttl stand for?

cahrens: Having read the rest of the PR, it's the length of the length of time the cache is good for. So… "time to ??"?

Can we change this to something more self-documenting, like "duration"?

marctalbott: Time to live. Maybe `permissionsCacheLifetime` would be clearer?

cahrens (approved):

PR #1730, opened 2024-05-08
Remove UAMI from Create Batchpool API

THWiseman:
- Old behavior: take a list of UAMI from the user request and assign those to the Azure Batch pool when it's created
- New behavior: Assign exactly one UAMI to the Azure Batch Pool, which is retrieved from Sam based on the user sending the request. 

`userAssignedIdentities` is now a deprecated (but not fully removed) field in the `AzureBatchPoolCreationParameters`. While this field is no longer used, its presence is necessary for the WSM API to remain backwards compatible.


service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -380,6 +383,20 @@ public ResponseEntity<ApiCreatedControlledAzureBatchPool> createAzureBatchPool(
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
     var workspace =
         validateWorkspaceResourceCreationPermissions(userRequest, workspaceUuid, body.getCommon());
+    String userEmail = samService.getSamUser(userRequest).getEmail();
```
rtitle, line 386: The logic here should likely be in a stairway step, instead of the API controller. For reference, [this](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/AssignManagedIdentityAzureVmStep.java) is the step which requests a Sam pet and assigns it to Azure VMs.

service/src/test/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiControllerBatchPoolTest.java
```diff
@@ -128,7 +129,7 @@ public void createBatchPool400WithInconsistentUAMI() throws Exception {
                     .characterEncoding("UTF-8")
                     .content(objectMapper.writeValueAsString(batchPoolRequest)),
                 USER_REQUEST))
-        .andExpect(status().is(HttpStatus.SC_BAD_REQUEST));
+        .andExpect(status().is(HttpStatus.SC_INTERNAL_SERVER_ERROR));
```
marctalbott, line 132: nit: rename the test too to indicate that this is a 500 now not a 400

marctalbott (approved):

jgainerdewar (approved):

PR #1740, opened 2024-05-20
[WX-1593] Update Sam Client

THWiseman:
Bump the version of the Sam java client to `v0.0.229`.

Necessary to take advantage of some new endpoints (`actionManagedIdentity`) in a future branch.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/workspace/service/iam/SamService.java
@@ -49,15 +49,7 @@ import org.broadinstitute.dsde.workbench.client.sam.api.GoogleApi;
 import org.broadinstitute.dsde.workbench.client.sam.api.ResourcesApi;
 import org.broadinstitute.dsde.workbench.client.sam.api.StatusApi;
 import org.broadinstitute.dsde.workbench.client.sam.api.UsersApi;
-import org.broadinstitute.dsde.workbench.client.sam.model.AccessPolicyMembershipRequest;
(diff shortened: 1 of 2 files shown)


aherbst-broad (approved):

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -207,10 +199,10 @@ public UserStatusInfo getUserStatusInfo(AuthenticatedUserRequest userRequest)
   public String getOrCreateUserManagedIdentityForUser(
       String userEmail, String subscriptionId, String tenantId, String managedResourceGroupId)
       throws InterruptedException {
-    AzureApi azureApi = samAzureApi(getWsmServiceAccountToken());
 
-    GetOrCreatePetManagedIdentityRequest request =
-        new GetOrCreatePetManagedIdentityRequest()
+    AzureApi azureApi = samAzureApi(getWsmServiceAccountToken());
+    GetOrCreateManagedIdentityRequest request =
```
cahrens, line 204: So is this actually no longer a pet identity?

THWiseman: Pet identities (a Terra concept) are a subset of User Assigned Managed Identities (an Azure concept). This type encompasses the Azure concept, not the Terra one. 

The new Sam client renamed this type since now Sam knows about more than just pet identities, e.g. 'action identities.' Renaming the type allows it to use the same logic for things that are identical on the Azure side of things. 

This method is still retrieving the user's pet identity and calling the Sam pet endpoint.

cahrens (approved):

PR #1741, opened 2024-05-21
[WOR-1667] Cache Sam responses when creating SAS tokens

marctalbott:
Ticket: [WOR-1667](https://broadworkbench.atlassian.net/browse/WOR-1667)
* The SAS token API receives heavy bursts of traffic that have brought WSM down before. This PR introduces caching for all four Sam calls in this code path to alleviate the load that these heavy bursts of traffic place on WSM and Sam.
    * A fix for this performance issue was initially attempted with https://broadworkbench.atlassian.net/browse/WOR-1335, but that was rolled back due to failures in our tests. This PR takes a more targeted approach that will have a smaller impact on upstream services.

[WOR-1667]: https://broadworkbench.atlassian.net/browse/WOR-1667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
@@ -214,7 +214,7 @@ public class ControlledAzureResourceApiController extends ControlledResourceCont
     // are determined below.
     Workspace workspace =
         sasTokenWorkspaceCache.computeIfAbsent(
-            new WorkspaceCacheKey(userRequest, workspaceUuid),
+            new WorkspaceCacheKey(userRequest.getSubjectId(), workspaceUuid),
             v ->
                 workspaceService.validateWorkspaceAndAction(
                     userRequest, workspaceUuid, SamConstants.SamWorkspaceAction.READ));
@@ -1002,4 +1002,4 @@ public class ControlledAzureResourceApiController extends ControlledResourceCont
   }
 }
 
-record WorkspaceCacheKey(AuthenticatedUserRequest userRequest, UUID workspaceUuid) {}
+record WorkspaceCacheKey(String userSubjectId, UUID workspaceUuid) {}
--- a/service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
+++ b/service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
@@ -52,8 +52,12 @@ import com.google.common.annotations.VisibleForTesting;
 import io.opentelemetry.instrumentation.annotations.WithSpan;
 import jakarta.servlet.http.HttpServletRequest;
 import java.time.OffsetDateTime;
+import java.util.Collections;
+import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+import org.apache.commons.collections4.map.PassiveExpiringMap;
 import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -60,6 +64,10 @@ public class AzureStorageAccessService {
   private final AzureConfiguration azureConfiguration;
   private final WorkspaceService workspaceService;
   private final Map<StorageAccountCoordinates, StorageData> storageAccountCache;
+  private final Map<String, SamUser> samUserCache;
```
aherbst-broad, line 67: Yeah, this is a lot of caches. But, looking through the code, this all feels like essential complexity:
* We need the sam user info to emit the audit log entry. This is general WSM tech debt I think (moving away from authenticateduserrequest)
* We need the storage container resource, which has an implicit sam check against the controlled resource, so we can get the container resource name for building the sas URL
* We need the permissions on the container to ensure the actions the user has available and that we're sanely building the sas token permissions

That middle one strikes me as awkward but I can't think of a better approach.

aherbst-broad (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
```diff
@@ -82,19 +90,31 @@ public AzureStorageAccessService(
...
-            () ->
-                samService.listResourceActions(
-                    userRequest, samResourceName, storageContainerUuid.toString()),
-            "listResourceActions");
+    List<String> containerActions;
+    var cacheKey = new StorageContainerCacheKey(userRequest.getSubjectId(), storageContainerUuid);
+    if (storageContainerPermissionsCache.containsKey(cacheKey)) {
+      containerActions = storageContainerPermissionsCache.get(cacheKey);
+    } else {
+      containerActions =
```
cahrens, line 110: Do you know why this one in particular has the `Rethrow.onInterrupted` handling?

cahrens (approved):

service/src/main/java/bio/terra/workspace/app/controller/ControlledAzureResourceApiController.java
```diff
@@ -206,8 +213,11 @@ public ResponseEntity<ApiCreateControlledAzureResourceResult> getCreateAzureDisk
     // You must have at least READ on the workspace to use this method. Actual permissions
     // are determined below.
     Workspace workspace =
-        workspaceService.validateWorkspaceAndAction(
-            userRequest, workspaceUuid, SamConstants.SamWorkspaceAction.READ);
+        sasTokenWorkspaceCache.computeIfAbsent(
```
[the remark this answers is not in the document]
marctalbott: Turns out we do have unit test coverage in `CreateAzureStorageContainerSasTokenTest` so I added a test there. I didn't realize we had unit coverage for any of these controller methods -- thanks for pointing me in the right direction

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/AzureStorageAccessService.java
marctalbott: `listResourceActions` throws an `InterruptedException` so this is needed. In other cases, we either aren't interacting with the `SamService` directly (although the underlying Sam call does have the `Rethrow.onInterrupted` handling) or the `SamService` call doesn't throw an `InterruptedException` (see `getSamUser`)

PR #1753, opened 2024-06-05
[WOR-1507] Capture metrics and status for flights and steps

aherbst-broad:
## Context
We want to capture the success/failure rate of flights. 

## This PR
Introduces a `FlightMetricsHook` class. This is a stairway lifecycle hook that emits a micrometer timer at the end of steps and flights. The metrics are namespaced in the following manner:

For flight-level data:
```
stairway_flight_<flight_class_name>_duration_seconds_<metric>
```
For step-level data:
```
stairway_flight_<flight_class_name>_step_<step_class_name>_duration_seconds_<metric>
```
The following tags are attached to both types of metrics:
* Status: Final status of the flight
* Resource type: Cloud resource type; this is only emitted in the case of flights with a `RESOURCE_TYPE` value in their input parameters (i.e,. `CreateControlledCloudResourceFlight`). It is omitted otherwise.

[Example](https://grafana.dsp-devops.broadinstitute.org/explore?orgId=1&left=%7B%22datasource%22:%228ZLL-_cnk%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22editorMode%22:%22builder%22,%22expr%22:%22rate%28stairway_flight_WorkspaceDeleteFlight_duration_seconds_count%7Bcluster%3D%5C%22terra-qa-bees%5C%22,%20service%3D%5C%22workspacemanager-appmetrics-service%5C%22%7D%5B5m%5D%29%22,%22legendFormat%22:%22%7B%7Bstatus%7D%7D%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-6h%22,%22to%22:%22now%22%7D%7D)

### DIFF
--- a/scripts/build_and_push.sh
+++ b/scripts/build_and_push.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+
+# Builds the current WSM branch and pushes container images
+# for AzureDatabaseUtils and WSM to GCR. Useful for testing
+# on BEEs.
+
+SHA=$(git rev-parse HEAD)
+./gradlew :service:jibDockerBuild --image="gcr.io/broad-dsp-gcr-public/terra-workspace-manager:$SHA"
+docker push "gcr.io/broad-dsp-gcr-public/terra-workspace-manager:$SHA"
+./gradlew :azureDatabaseUtils:jibDockerBuild --image="us.gcr.io/broad-dsp-gcr-public/azure-database-utils:$SHA"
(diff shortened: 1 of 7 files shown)


cahrens (approved):

scripts/build_and_push.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
marctalbott, line 1: Thanks for writing this

PR #1754, opened 2024-06-05
ID-1266: Assign azure_private_storage_account action managed identities to VMs

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/ID-1266

Design doc: https://docs.google.com/document/d/1VT__hmPH1cyYYm643cJzlsrVIyKeWPgEnjpC0fjV63E/edit

This adds an optional field `userAssignedManagedIdentities` to Azure VM creation. It is only accepted for application-managed VMs. If present, WSM will assign the VM to the specified identities _in addition_ to the pet managed identity for the owning user (current behavior). A corresponding Leo PR will request the identities from Sam and pass them to WSM.

This is to support the `enterprise data access` team-sized problem -- we are provisioning action managed identities for `private_azure_storage_account/read`, and we want to auto-assign these identities to notebook VMs so customers can access their external data.

Tested in a BEE with WSM and Leo: created an AAMI in Sam, created a workspace, created a VM, verified the VM is assigned to the right identities. Also verified that VMs for users without AAMIs continue to work with no change.


openapi/src/resources/azure_vm.yaml
```diff
@@ -21,6 +21,10 @@ components:
           description: A valid WSM identifier for an Azure Disk that corresponds to a valid azure resource
           type: string
           format: uuid
+        priority:
```
rtitle, line 24: I figured these 2 things made sense to be in the Azure VM response, instead of just creation parameters.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/AssignManagedIdentityAzureVmStep.java
```diff
@@ -37,17 +40,23 @@ public AssignManagedIdentityAzureVmStep(
...
+
+    ComputeManager computeManager = crlService.getComputeManager(azureCloudContext, azureConfig);
+    MsiManager msiManager = crlService.getMsiManager(azureCloudContext, azureConfig);
 
-      ComputeManager computeManager = crlService.getComputeManager(azureCloudContext, azureConfig);
-      MsiManager msiManager = crlService.getMsiManager(azureCloudContext, azureConfig);
+    Set<String> userAssignedIdentities = new HashSet<>();
 
+    // Add user assigned identities from the request, if any
+    if (CollectionUtils.isNotEmpty(resource.getUserAssignedIdentities())) {
```
aherbst-broad, line 54: It seems like we're consuming the list of identities and adding them to the VM without first checking with Sam, is that something that should be done? That is, can a user add any arbitrary user assigned identity to the vm using this api?

aherbst-broad (commented):
Had a question about validating the identities.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/AssignManagedIdentityAzureVmStep.java
rtitle: I think the key factor is that the identities are only accepted for application-managed resources (WSM throws a Bad Request otherwise). The calling application (Leo) is requesting these identities from Sam as the user. WSM trusts the application to the validation.

Application-managed resources are tricky, because WSM does not have access to the user's token -- so we were thinking to do the Sam validation upstream of WSM for the application case.

rtitle: Added a comment to this effect as well

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/ControlledAzureVmResource.java
```diff
@@ -137,6 +163,13 @@ public void addCreateSteps(
             flightBeanBag.getSamService(),
             flightBeanBag.getWorkspaceService()),
         cloudRetry);
+    flight.addStep(
+        new GetPetManagedIdentityStep(
```
rtitle, line 167: This step is used in batch pool and kubernetes namespace creation as well.

service/dependencies.gradle
```diff
@@ -41,7 +41,7 @@ dependencies {
   implementation group: "bio.terra", name: "terra-resource-buffer-client", version: "0.198.42-SNAPSHOT"
 
   // Cloud Resource Library
-  implementation group: 'bio.terra', name: 'terra-cloud-resource-lib', version: "1.2.30-SNAPSHOT"
+  implementation group: 'bio.terra', name: 'terra-cloud-resource-lib', version: "1.2.30-rt-bump-SNAPSHOT"
```
rtitle, line 44: TODO: update when https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/182 is merged

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/vm/ControlledAzureVmResource.java
```diff
@@ -137,6 +163,13 @@ public void addCreateSteps(
...
+            flightBeanBag.getAzureConfig(),
+            flightBeanBag.getCrlService(),
+            flightBeanBag.getSamService(),
+            getAssignedUser().orElseThrow()),
```
rtitle, line 171: VMs are always private and have assigned users

aherbst-broad (approved):

PR #1756, opened 2024-06-11
AJ-1869 WIP Add limits to spend profile and workspace description

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1869

Adds the `limits` map to the `SpendProfile` and returns it in the `AzureCloudContext` of a `WorkspaceDescription`.  This necessitates adding Azure-specific fields to the `SpendProfileModel` and distinguishing between GCP and Azure spend profiles throughout the texting fixtures.

Tested in a bee using https://github.com/broadinstitute/terra-helmfile/pull/5626 .

### DIFF
--- a/service/dependencies.gradle
+++ b/service/dependencies.gradle
@@ -23,7 +23,7 @@ dependencies {
 
   // Terra deps
   implementation group: "bio.terra", name: "datarepo-client", version: "2.13.0-SNAPSHOT"
-  implementation group: "bio.terra", name:"billing-profile-manager-client", version: "0.1.536-SNAPSHOT"
+  implementation group: "bio.terra", name:"billing-profile-manager-client", version: "0.1.548-SNAPSHOT"
   implementation group: "bio.terra", name:"terra-policy-client", version:"1.0.9-SNAPSHOT"
   implementation group: "bio.terra", name:"terra-aws-resource-discovery", version:"v0.6.3-SNAPSHOT"
 
--- a/service/src/main/java/bio/terra/workspace/app/configuration/external/SpendProfileConfiguration.java
+++ b/service/src/main/java/bio/terra/workspace/app/configuration/external/SpendProfileConfiguration.java
@@ -52,9 +52,6 @@ public class SpendProfileConfiguration {
 
     private Map<String, String> limits;
     private CloudPlatform cloudPlatform;
-    private UUID tenantId;
-    private UUID subscriptionId;
-    private String managedResourceGroupId;
 
     public String getId() {
       return id;
@@ -86,30 +83,6 @@ public class SpendProfileConfiguration {
 
     public void setCloudPlatform(CloudPlatform cloudPlatform) {
       this.cloudPlatform = cloudPlatform;
-    }
-
-    public UUID getTenantId() {
-      return tenantId;
-    }
-
-    public void setTenantId(UUID tenantId) {
-      this.tenantId = tenantId;
-    }
-
-    public UUID getSubscriptionId() {
-      return subscriptionId;
-    }
-
-    public void setSubscriptionId(UUID subscriptionId) {
-      this.subscriptionId = subscriptionId;
-    }
-
-    public String getManagedResourceGroupId() {
-      return managedResourceGroupId;
-    }
-
-    public void setManagedResourceGroupId(String managedResourceGroupId) {
-      this.managedResourceGroupId = managedResourceGroupId;
     }
   }
 }
--- a/service/src/main/java/bio/terra/workspace/app/configuration/external/SpendProfileConfiguration.java
+++ b/service/src/main/java/bio/terra/workspace/app/configuration/external/SpendProfileConfiguration.java
@@ -6,12 +6,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
-
-import javax.annotation.Nullable;
 
 /**
  * Configuration for Spend Profiles.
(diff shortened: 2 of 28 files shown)


service/dependencies.gradle
```diff
@@ -23,7 +23,7 @@ dependencies {
 
   // Terra deps
   implementation group: "bio.terra", name: "datarepo-client", version: "2.13.0-SNAPSHOT"
-  implementation group: "bio.terra", name:"billing-profile-manager-client", version: "0.1.536-SNAPSHOT"
+  implementation group: "bio.terra", name:"billing-profile-manager-client", version: "0.1.548-SNAPSHOT"
```
calypsomatic, line 26: I don't think this is actually necessary since we don't actually use BPM's models, but it shouldn't hurt to update

service/src/main/java/bio/terra/workspace/app/configuration/external/SpendProfileConfiguration.java
```diff
@@ -62,5 +71,45 @@ public String getBillingAccountId() {
...
+
+    public Map<String, String> getLimits() {
+      return limits;
+    }
+
+    public void setLimits(Map<String, String> limits) {
+      this.limits = limits;
+    }
+
+    public CloudPlatform getCloudPlatform() {
```
cahrens, line 83: I apologize… this has probably been discussed already, but I am confused as to the duplication here of information that is already in the billing profile. Why do we need to do this instead of getting information directly from the billing profile like we do here:

````
spend =
          new SpendProfile(
              spendProfileId,
              getProfileCloudPlatform(profile),
              profile.getBillingAccountId(),
              profile.getTenantId(),
              profile.getSubscriptionId(),
              profile.getManagedResourceGroupId(),
              Optional.ofNullable(profile.getOrganization())
                  .map(SpendProfileOrganization::new)
                  .orElse(null));

Comment (cahrens) [reply service/src/main/java/bio/terra/workspace/app/configuration/external/SpendProfileConfiguration.java:83]: Since the list of SpendProfiles is only used by tests, I think one could just mock the return value of BPM in tests (for tests that care about limits, which will just be Azure tests). Aren't we getting the limits from BPM itself?

Comment (calypsomatic) [reply service/src/main/java/bio/terra/workspace/app/configuration/external/SpendProfileConfiguration.java:83]: We won't be able to get the limits from BPM since we can't rely on having permission to get that information, hence adding it directly to WSM config.  That said, i'm open to a different or better method of getting the SpendProfile with limits from the config.

File: service/src/main/resources/application.yml [hunk 3]
```diff
@@ -152,6 +152,14 @@ workspace:
       id: wm-default-spend-profile
       # The billing account workspace-dev has access to.
       billing-account-id: 01A82E-CA8A14-367457
+      cloud-platform: GCP
+    - id: facade00-0000-4000-a000-000000000000
```
aherbst-broad, line 156: 🤔 I want to make sure I'm grokking this -- is the idea to config a full billing/spend profile model here for every profile we want to have limits? So, BPM will have a full copy of the BP and WSM will as well? If that's the case, I worry that we'll end up in situations where WSM and BPM disagree about both the data in a billing profile and it's state (i.e, deleted, not deleted ,etc.).

I wonder if this could be made more targeted. Instead of config'ing out the entire BP, we could create a map of BP IDs to limit configs and layer those in just-in-time when the spend profile service returns its results.

aherbst-broad (commented):
Had a question about the approach

davidangb (commented):
questions/comments inline!

service/src/main/resources/application.yml
cahrens: This is what I'm thinking as well-- I would rather pull the information from BPM the first time it is read and then store that in the map. This avoids duplication and also reduces the amount of code changes considerably.

calypsomatic: I think that's a good idea - I was mostly trying to conform to what the pre-existing methods and classes expected, which was requiring the full information.  I'll do another pass-through.

cahrens: So to understand… with @aherbst-broad 's suggestion we still duplicate the limits in both repos, but we don't duplicate the MRG information, and we can tell if the billing profile has actually been deleted or the user no longer has permissions on it, which seems like an excellent idea.

I do still wonder if there is a way to store the limits from BPM by storing thing them during workspace creation and thus avoid duplicating the limits also, but it seems like folks don't want want to figure out how to store them in a WSM database?

calypsomatic: Yeah I think we didn't want to deal with storing in a database.  But even if I cut it down to just the ID and take out all the MRG info, I don't see how that will let us tell if it's been deleted or not.

cahrens: I believe it's because when we are authorizing against a spend profile, we will still call BPM get the actual profile (which does permission checks) instead of just using a hardcoded copy of information from the config. But the limits themselves will always be available because they are in config.

service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfileService.java
```diff
@@ -133,22 +133,29 @@ private static List<SpendProfile> adaptConfigurationModels(
...
-            spendModel ->
-                new SpendProfile(
-                    new SpendProfileId(spendModel.getId()),
-                    CloudPlatform.GCP,
-                    spendModel.getBillingAccountId(),
-                    null,
-                    null,
-                    null,
-                    null))
+            spendModel -> !Strings.isNullOrEmpty(spendModel.getId()))
```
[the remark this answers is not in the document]
calypsomatic: It would be empty for a non-GCP spend profile, so I didn't want to filter those out.

service/src/main/resources/application.yml
calypsomatic: [This](https://github.com/DataBiosphere/terra-workspace-manager/blob/0ce9424ef54582128ef9b73fa58341c9f748ec7a/service/src/main/java/bio/terra/workspace/service/spendprofile/SpendProfileService.java#L60) is the part that's tripping me up - when we read in from the config, we convert the `SpendProfileModel` to a `SpendProfile` which requires all the extra azure coordinate information.  I tried to minimize changes to this since presumably in other parts of the code we need a real `SpendProfile`, but it seems like if I need a dummy `SpendProfileModel` that doesn't contain all the information it'll need a different config, or other awkward changes to accommodate both the "real" and "limits" spendprofile, or instead of initializing from the config simply read from it as needed.

PR #1761, opened 2024-06-20
[WOR-1719] Copy source stewarding application to destination resource's policy

aherbst-broad:
### Context
When cloning application controlled resources that have a COPY_RESOURCE definition, we need to ensure that the "owning" application's SA from the source resoure is set on the resource's policy, not that of the cloning user. 

### This PR
* Grabs the application ID from the source resource and uses that to resolve the application's service account name, rather than assuming the incoming user request has it.

### TODO
* Any previously created application managed resource that was copied with COPY_RESOURCE will likely need a cleanup which will be captured in a forthcoming ticket. This will involve:
   * Setting the associated_app field to the correct value instead of null
   * Granting the needed Sam permissions


service/src/main/java/bio/terra/workspace/service/resource/controlled/model/ControlledResource.java
```diff
@@ -305,6 +306,8 @@ public ControlledResourceFields buildControlledCloneResourceCommonFields(
     // override name and description if provided
     cloneResourceCommonFields.name(name == null ? getName() : name);
     cloneResourceCommonFields.description(description == null ? getDescription() : description);
+    cloneResourceCommonFields.applicationId(
+        applicationId == null ? getApplicationId() : applicationId);
```
cahrens, line 310: What does `getApplicationId` do in this case? I'm trying to understand why there would be an application ID if one was not passed in.

service/src/main/java/bio/terra/workspace/service/workspace/WsmApplicationService.java
```diff
@@ -119,7 +120,16 @@ public WsmWorkspaceApplication launchApplicationAbleJobAndWait(
 
   public WsmWorkspaceApplication getWorkspaceApplication(
       Workspace workspace, String applicationId) {
-    return applicationDao.getWorkspaceApplication(workspace.getWorkspaceId(), applicationId);
+    return getWorkspaceApplicationByWorkspaceId(workspace.getWorkspaceId(), applicationId);
+  }
+
+  public WsmWorkspaceApplication getWorkspaceApplicationByWorkspaceId(
```
cahrens, line 126: Why have 2 methods that do the same thing?

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
```diff
@@ -0,0 +1,134 @@
...
+            ControlledResourceCategory.USER_PRIVATE,
+            null);
+    var request = new CreateResourceRequestV2();
+
+    policyBuilder.addPolicies(request);
+
+    assertThat(
+        request.getPolicies().get(ControlledResourceIamRole.OWNER.toSamRole()).getMemberEmails(),
+        contains(GcpUtils.getWsmSaEmail()));
+    assertThat(
```
cahrens, line 34: So this would be the case of a Cromwell App being owned by Leo, but the user having Editor permissions?

cahrens: Though this is a USER_PRIVATE resource, so why does Leo get owner permissions on that?

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
```diff
@@ -0,0 +1,134 @@
...
+    var request = new CreateResourceRequestV2();
+
+    assertThrows(InternalLogicException.class, () -> policyBuilder.addPolicies(request));
+  }
+
+  @Test
+  void addPolicies_applicationPrivate() {
+    var policyBuilder =
+        new ControlledResourceSamPolicyBuilder(
+            ControlledResourceIamRole.WRITER,
```
cahrens, line 105: Why can a user have WRITER permissions here, but a DELETER cannot in the previous test case?

service/src/main/java/bio/terra/workspace/service/resource/controlled/model/ControlledResource.java
```diff
@@ -283,7 +283,8 @@ public ControlledResourceFields buildControlledCloneResourceCommonFields(
       @Nullable String name,
       @Nullable String description,
       String createByEmail,
-      String region) {
+      String region,
+      @Nullable String applicationId) {
```
cahrens, line 287: Why allow this to be null?

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  @Test
+  void addPolicies_applicationShared() {
+    var app =
+        new WsmWorkspaceApplication()
+            .application(new WsmApplication().serviceAccount("fake-svc@example.com"));
+    var policyBuilder =
+        new ControlledResourceSamPolicyBuilder(
+            ControlledResourceIamRole.READER,
```
cahrens, line 56: What does it mean to specify a READER role but with no user?

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
```diff
@@ -0,0 +1,134 @@
...
+            app);
+    var request = new CreateResourceRequestV2();
+
+    policyBuilder.addPolicies(request);
+
+    assertThat(
+        request.getPolicies().get(ControlledResourceIamRole.OWNER.toSamRole()).getMemberEmails(),
+        contains(GcpUtils.getWsmSaEmail()));
+    assertThat(
+        request.getPolicies().get(ControlledResourceIamRole.EDITOR.toSamRole()).getMemberEmails(),
```
cahrens, line 68: I'm not understanding why we test Editor here (vs. Reader, which was in the policy builder).

marctalbott (approved):

service/src/main/java/bio/terra/workspace/service/workspace/WsmApplicationService.java
aherbst-broad: Good catch, fixed.

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
aherbst-broad: The choice of DELETER in the previous test (`addPolicies_applicationSharedWithNoAppFails` I think) is arbitrary; the intent is to ensure that the presence of _any_ private IAM role argument triggers an exception as that is not a valid thing to specify for an application shared resource. I'll add a comment to that effect. 

See also this [doc](https://docs.google.com/document/d/1F28KHeVpTh0Zuqdux30ONabMgjj6CNmyv7B8iqcXnsE/edit?resourcekey=0-5ndsiOmBJdG7bB1gZXOy1g) that explains the resource permissions model in  more depth.

service/src/main/java/bio/terra/workspace/service/iam/SamService.java
```diff
@@ -1007,13 +1012,24 @@ public void createControlledResource(
...
+    // from the application, i.e., when a resource is cloned)
+    WsmWorkspaceApplication app = null;
+    if (resource.getApplicationId() != null
+        && (resource.getCategory().equals(ControlledResourceCategory.APPLICATION_PRIVATE)
+            || resource.getCategory().equals(ControlledResourceCategory.APPLICATION_SHARED))) {
+      app =
+          applicationService.getWorkspaceApplicationByWorkspaceId(
+              resource.getWorkspaceId(), resource.getApplicationId());
+    }
+
```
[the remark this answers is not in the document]
aherbst-broad: No; resources may be application or user managed, hence the nullable app variable here. If the applicationService is unable to find the app with a given (non-null) ID, that is likely a misconfiguration and will result in a `ApplicationNotFoundException`.

service/src/main/java/bio/terra/workspace/service/resource/controlled/model/ControlledResource.java
aherbst-broad: Same as above; controlled resources _may_ be managed by an app (but certainly not always). For application managed resources that are being cloned, we need to ensure the application ID propagates.

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
aherbst-broad: Good eye; this is a shared resource that is accessible to all user's, so the private IAM role here is actually ignored. This value can be null (or really anything) which points to this being an awkward interface. I'll add some parameter checking to ensure an exception is thrown when a private IAM role is passed but the resource is shared.

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
aherbst-broad: The "managing" application gets EDITOR, which is what this is asserting. Including READER as a parameter in the above call was misleading, so I've added a check to prevent that inconsistency.

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
aherbst-broad: Update: I've added a logger.warn() to emit on this situation. I had originally put in a stricter check that threw an exception but it caused some worrisome connected test failures that make me worry that there are callsites that are not using this API properly. Rather than refactor the world, I am going to leave that logic as is with the warning, we can reevaluate downstream if it becomes more problematic.

service/src/main/java/bio/terra/workspace/service/resource/controlled/model/ControlledResource.java
aherbst-broad: Yeah this isn't needed as a parameter; all of the callsites that thread it through are grabbing it from the source anyway. I've removed the param.

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
```diff
@@ -0,0 +1,180 @@
...
+    assertThrows(
+        InternalLogicException.class,
+        () -> policyBuilder.addPolicies(request),
+        "Specifying a private IAM role on an application shared resource is invalid");
+  }
+
+  @Test
+  void addPolicies_applicationSharedWithNoAppFails() {
+    // ensure an application shared resource cannot have a user with a private IAM role on it
+    // (tho choice of EDITOR for the private IAM role is arbitrary--any value will cause an
```
cahrens, line 118: But this test case is really about the app being null, isn't it? Perhaps this comment is on the wrong test case?

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
```diff
@@ -0,0 +1,180 @@
...
+    // exception.
+    var policyBuilder =
+        new ControlledResourceSamPolicyBuilder(
+            ControlledResourceIamRole.EDITOR,
+            null,
+            ControlledResourceCategory.APPLICATION_SHARED,
+            null);
+    var request = new CreateResourceRequestV2();
+
+    assertThrows(InternalLogicException.class, () -> policyBuilder.addPolicies(request));
```
cahrens, line 128: Since an `InternalLogicException` can be thrown for a couple of different reasons, it would be good to check the error message.

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
```diff
@@ -0,0 +1,180 @@
...
+            .application(new WsmApplication().serviceAccount("fake-svc@example.com"));
+    var policyBuilder =
+        new ControlledResourceSamPolicyBuilder(
+            ControlledResourceIamRole.EDITOR,
+            "fake@example.com",
+            ControlledResourceCategory.APPLICATION_SHARED,
+            app);
+    var request = new CreateResourceRequestV2();
+
+    assertThrows(InternalLogicException.class, () -> policyBuilder.addPolicies(request));
```
cahrens, line 144: Since an InternalLogicException can be thrown for a couple of different reasons, it would be good to check the error message.

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
```diff
@@ -0,0 +1,180 @@
...
+  void addPolicies_applicationPrivateWithNoAppFails() {
+    var policyBuilder =
+        new ControlledResourceSamPolicyBuilder(
+            ControlledResourceIamRole.EDITOR,
+            null,
+            ControlledResourceCategory.APPLICATION_PRIVATE,
+            null);
+    var request = new CreateResourceRequestV2();
+
+    assertThrows(InternalLogicException.class, () -> policyBuilder.addPolicies(request));
```
cahrens, line 178: Since an InternalLogicException can be thrown for a couple of different reasons, it would be good to check the error message.

And is it also a problem in this case that this is APPLICATION_PRIVATE, but has no email address for the EDITOR user?

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
aherbst-broad: 🤦 yup. Fixed.

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
aherbst-broad: Updated to check the error, and passed in an email, good idea -- thanks!

service/src/test/java/bio/terra/workspace/service/iam/ControlledResourceSamPolicyBuilderTest.java
aherbst-broad: Done.

cahrens (approved):

PR #1763, opened 2024-06-21
WOR-1620 WSM 404s during deployment of batch pool

blakery:
https://broadworkbench.atlassian.net/browse/WOR-1620

Switches batch account validation to use the WSM service account token to retrieve the landing zone, similar to other places.


service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/batchpool/VerifyAzureBatchPoolCanBeCreatedStepTest.java
blakery, line 1: I figured I'd also change this test to be a true unit test, while I was here, since it didn't actually need spring for anything.

marctalbott (approved):

aherbst-broad (approved):

PR #1764, opened 2024-06-21
[TOAZ-364] Support for AzureGov storage SAS tokens

aaronegrant:
Azure Gov cloud environments use an alternate domain name for storage account urls. Specifically, the domain ends with "[core.usgovcloudapi.net](http://core.usgovcloudapi.net/)", rather than "[core.windows.net](http://core.windows.net/)". See [Develop with Storage API on Azure Government - Azure Government](https://learn.microsoft.com/en-us/azure/azure-government/documentation-government-get-started-connect-to-storage)  for additional details.

WSM APIs that return Storage URLs and SAS tokens need to be updated to support this alternate domain name when running in an Azure Gov environment.


service/src/main/java/bio/terra/workspace/service/crl/CrlService.java
```diff
@@ -574,7 +577,19 @@ private AzureProfile getAzureProfile(AzureCloudContext azureCloudContext) {
     return new AzureProfile(
         azureCloudContext.getAzureTenantId(),
         azureCloudContext.getAzureSubscriptionId(),
-        AzureEnvironment.AZURE);
+        getAzureEnvironmentFromName(azureConfiguration.getAzureEnvironment()));
+  }
+
+  public AzureEnvironment getAzureEnvironmentFromName(String envName) {
```
aaronegrant, line 583: The JDK differs from the .NET SDK in that there is no 'name' attribute that we can interrogate and match on. Hard coding the 3 available AzureEnvironments in this switch statement instead

service/src/main/resources/application.yml
```diff
@@ -173,6 +174,7 @@ workspace:
     protected-data-landing-zone-defs: ["ProtectedDataResourcesFactory"]
     azure-database-util-image: "us.gcr.io/broad-dsp-gcr-public/azure-database-utils:${version.gitHash}"
     azure-database-util-logs-tail-lines: 1000
+    azure-environment: ${env.azure.environment}
```
giventocode, line 177: Is this setting required? It seems to be a duplicate from the one above.

aaronegrant: The first one pulls it from the OS environment, the second one is where the Workspace Azure config object reads it from and refers to the first one. Seems to be the pattern used. @dvoet Am I doing this config setup correctly?

dvoet (approved):

aherbst-broad (approved):

PR #1777, opened 2024-07-19
WOR-1777 Handle more instances of ManagementExceptions when deleting resources

blakery:
https://broadworkbench.atlassian.net/browse/WOR-1777

* change GetAzureDiskAttachedVmStep to extend DeleteAzureControlledResourceStep for management exception handling
  The output of GetAzureDiskAttachedVmStep is only used by `DetachAzureDiskFromVmStep`, which is already set up to handle a missing value.
* change `DetachAzureDiskFromVmStep` to extend `DeleteAzureControlledResourceStep`
* Parameterize DeleteAzureControlledResourceStep, for better/more central logging options: There was only one step that  didn't already have a parameter named `resource`, of the appropriate type, and that was trivial to add.
* Added `LandingZoneNotFoundException` to common error handling

Because this is changing the parameters of the `FlightMap`, it could cause a running flight to log a `NullPointerException` instead of the appropriate exception, if it failed for one of the covered exceptions when running `DeleteFederatedCredentialStep`.

### DIFF
--- a/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/DeleteAzureControlledResourceStep.java
+++ b/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/DeleteAzureControlledResourceStep.java
@@ -7,6 +7,7 @@ import bio.terra.stairway.StepStatus;
 import bio.terra.workspace.amalgam.landingzone.azure.LandingZoneServiceApiException;
 import bio.terra.workspace.common.exception.AzureManagementExceptionUtils;
 import bio.terra.workspace.service.resource.controlled.flight.delete.DeleteControlledResourceStep;
+import bio.terra.workspace.service.resource.controlled.model.ControlledResource;
 import com.azure.core.management.exception.ManagementException;
 import java.util.Arrays;
 import java.util.List;
@@ -16,7 +17,14 @@ import org.springframework.http.HttpStatus;
  * Common base class for deleting Azure resources. Wraps the Azure resource deletion logic with
  * handling for common exceptions.
  */
-public abstract class DeleteAzureControlledResourceStep implements DeleteControlledResourceStep {
+public abstract class DeleteAzureControlledResourceStep<R extends ControlledResource>
+    implements DeleteControlledResourceStep {
+
+  protected final R resource;
+
+  protected DeleteAzureControlledResourceStep(R resource) {
+    this.resource = resource;
+  }
 
   @Override
   public StepResult doStep(FlightContext context) throws InterruptedException {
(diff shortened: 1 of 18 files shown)


service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/GetAzureDiskAttachedVmStep.java
```diff
@@ -16,54 +15,58 @@
...
- * been detached from a virtual machine, but we need to attach it back in corresponding 'undo'
- * operation.
+ * machine to which data disk is attached and saves it into flight's working map. It is required in
+ * case a disk has been detached from a virtual machine, but we need to attach it back in
+ * corresponding 'undo' operation. This is a separate step because the data is required for the
+ * 'undo' operation at DetachAzureDiskFromVmStep, and we need to guarantee its persistence - which
+ * we can't guarantee if it is from the 'do' portion of the same step.
  */
-public class GetAzureDiskAttachedVmStep implements DeleteControlledResourceStep {
+public class GetAzureDiskAttachedVmStep
```
marctalbott, line 24: I don't have a better suggestion unfortunately, but just calling out that I find it rather confusing that this step is a `Get` step that also extends `DeleteAzureControlledResourceStep` and it has a `deleteResource` method that doesn't delete anything.

marctalbott (approved):

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/disk/GetAzureDiskAttachedVmStep.java
blakery: I know. Naming things is hard.
There is a fairly extensive explanation in the comments, for this reason.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/DeleteAzureControlledResourceStep.java
```diff
@@ -76,14 +88,32 @@ protected StepResult handleResourceDeleteException(Exception e, FlightContext co
...
+            resource.getResourceType().name(),
+            resource.getWorkspaceId());
         return StepResult.getStepResultSuccess();
       }
     }
 
+    if (e instanceof LandingZoneNotFoundException) {
+      // If the landing zone is not present, it's probably because it was removed directly
+      logger.debug(
+          "Unable to delete resource {} from workspace {}, because no landing zone was found in LZS",
```
aherbst-broad, line 112: Nit: note in the log msg that we are explicitly ignoring the error and moving on.

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/azure/DeleteAzureControlledResourceStep.java
```diff
@@ -76,14 +88,32 @@ protected StepResult handleResourceDeleteException(Exception e, FlightContext co
...
+      var code =
+          Optional.ofNullable(apiEx.getMessage())
+              .flatMap(
+                  msg ->
+                      missingResourceManagementCodes.stream()
+                          .filter(c -> msg.contains(c))
+                          .findFirst());
+      if (code.isPresent()) {
+        logger.debug(
+            "LZS encountered management exception {} when deleting resource {} from workspace {}",
```
aherbst-broad, line 101: Nit: note in the log msg that we are explicitly ignoring the exception and moving on.

aherbst-broad (approved):

PR #1782, opened 2024-08-07
[IA-5035] Deploy a GCE VM with WSM

rtitle:
https://broadworkbench.atlassian.net/browse/IA-5035

Much more detailed notes here: https://docs.google.com/document/d/1lRX_r2NwN9NlV5i8TIxYOlLKuAWAQWu0tvAsFKf-21U/edit

Fixes needed to deploy a GCE VM with the Leo requirements (init scripts, image, etc).


openapi/src/parts/controlled_gcp_gce_instance.yaml
```diff
@@ -222,6 +222,12 @@ components:
             of 64000 GB (64 TB). If not specified, this defaults to 500.
           type: integer
           format: int64
+        bootDiskSizeGb:
```
rtitle, line 225: Added optional `bootDiskSizeGb` parameter, it was previously hardcoded to 100G. The Leo image is slightly bigger.

aherbst-broad: Should this be required? What is the behavior if null is passed to the builder params?

service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/gceinstance/CreateGceInstanceStep.java
```diff
@@ -196,7 +196,7 @@ static Instance setFields(
                 .setInitializeParams(
                     new AttachedDiskInitializeParams()
                         .setSourceImage(creationParameters.getVmImage())
-                        .setDiskSizeGb(100L)),
+                        .setDiskSizeGb(creationParameters.getBootDiskSizeGb())),
```
rtitle, line 199: If null, defaults to the size of the source image (which is the GCP behavior)

openapi/src/parts/controlled_gcp_gce_instance.yaml
rtitle: If null, the GCP behavior is to default to the size of the source image.

I did discover that the `vmImage` parameter is required, though WSM doesn't treat it as such. I can do a pass to mark that required, and update the swagger documentation. (good call)

aherbst-broad (approved):

blakery (approved):

PR #1806, opened 2024-12-02
Support for deployment in Azure Government Cloud (WSM)

bennettn4:
Part of large effort to add support for services to run in Azure Government cloud.

- [wsm] Added config value azureEnvironment
- [azureDatabaseUtils] Added environment variable for AZURE_ENVIRONMENT

https://broadworkbench.atlassian.net/browse/TOAZ-372
See related pull requests here:
[workbench-libs](https://github.com/broadinstitute/workbench-libs/pull/1713) [sam](https://github.com/broadinstitute/sam/pull/1603) [bpm](https://github.com/DataBiosphere/terra-billing-profile-manager/pull/505) [wsm](https://github.com/DataBiosphere/terra-workspace-manager/pull/1806) [cromwell](https://github.com/broadinstitute/cromwell/pull/7670) [terra-helmfile](https://github.com/broadinstitute/terra-helmfile/pull/5976)

PR #1836, opened 2025-02-25
CORE-338: fix Azure unit tests, maybe fix live database utils

davidangb:
After looking into it, the azure unit tests were an isolated issue. Any problems we're seeing in production are not directly related to the test failures. However, this PR has an experiment to try to address production, too.

In this PR:
* fix a test assertion in `AzureDatabaseUtilsRunner` that was broken in #1806
* default to the correct value (`AzureCloud`, not `AZURE`) for the azure environment

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -30,7 +30,7 @@ env:
     tps: ${TPS_ADDRESS:https://tps.dsde-dev.broadinstitute.org/}
     lzs: ${LZS_ADDRESS:https://landingzone.dsde-dev.broadinstitute.org/}
   azure:
-    environment: ${AZURE_ENVIRONMENT:AZURE}
+    environment: ${AZURE_ENVIRONMENT:AzureCloud}
     customer:
       usage-attribute: ${AZURE_CUSTOMER_USAGE_ATTRIBUTE:}
 
--- a/service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/AzureDatabaseUtilsRunnerTest.java
+++ b/service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/AzureDatabaseUtilsRunnerTest.java
@@ -49,7 +49,7 @@ public class AzureDatabaseUtilsRunnerTest extends BaseMockitoStrictStubbingTest 
   @Mock private ApiAzureLandingZoneDeployedResource mockDatabase;
   @Mock private ApiAzureLandingZoneDeployedResource mockAdminIdentity;
 
-  private final String AZURE_ENV = "AzureCloud";
+  private final String AZURE_ENV = "AZURE";
 
   @BeforeEach
   public void createAzureDatabaseUtilsRunner() {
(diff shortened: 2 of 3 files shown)


service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/AzureDatabaseUtilsRunnerTest.java
```diff
@@ -171,6 +174,8 @@ private void assertResults(Map<String, String> expectedEnv) {
     var expectedEnvWithCommon = new HashMap<>(expectedEnv);
     expectedEnvWithCommon.putAll(
         Map.of(
+            "AZURE_ENVIRONMENT",
+            AZURE_ENV,
```
davidangb, line 178: #1806 added a new env var, but did not update this expectation to include the new env var. This caused tests to fail. See https://github.com/DataBiosphere/terra-workspace-manager/pull/1806/files#diff-7e2975f5c6f927eade14d4008340e3ec0d6d7df12e5669d0b3e3cd3784f0cac3

service/src/test/java/bio/terra/workspace/service/resource/controlled/cloud/azure/database/AzureDatabaseUtilsRunnerTest.java
```diff
@@ -49,8 +49,11 @@ public class AzureDatabaseUtilsRunnerTest extends BaseMockitoStrictStubbingTest
   @Mock private ApiAzureLandingZoneDeployedResource mockDatabase;
   @Mock private ApiAzureLandingZoneDeployedResource mockAdminIdentity;
 
+  private final String AZURE_ENV = "AzureCloud";
+
   @BeforeEach
   public void createAzureDatabaseUtilsRunner() {
+    when(mockAzureConfig.getAzureEnvironmentConfigString()).thenReturn(AZURE_ENV);
```
davidangb, line 56: in practice, we should have a value for the `AZURE_ENVIRONMENT` env var. This test was using an empty mock for its config. Now it has a value.

service/src/main/resources/application.yml
```diff
@@ -30,7 +30,7 @@ env:
     tps: ${TPS_ADDRESS:https://tps.dsde-dev.broadinstitute.org/}
     lzs: ${LZS_ADDRESS:https://landingzone.dsde-dev.broadinstitute.org/}
   azure:
-    environment: ${AZURE_ENVIRONMENT:AZURE}
+    environment: ${AZURE_ENVIRONMENT:AzureCloud}
```
davidangb, line 33: inspired by https://github.com/DataBiosphere/terra-landing-zone-service/pull/536. This value didn't cause more problems because there's a default case in https://github.com/DataBiosphere/terra-workspace-manager/blob/2fb17c499a2fc9cd72c785cdc0ffe2104e91f2dd/service/src/main/java/bio/terra/workspace/app/configuration/external/AzureConfiguration.java#L35-L39. However, this value is passed as-is to the database utils, where I think it _will_ cause a problem; the database utils yaml file expects `AzureCloud`, not `AZURE` in https://github.com/DataBiosphere/terra-workspace-manager/pull/1806/files#diff-48a6c4ffa6ec09eb28b0d58dee997ddf20fa5ff4c596a10e0d83a4f4b68a43bf.

marctalbott (approved):

kevinmarete (approved):

